PR #4, opened 2017-04-02
[WIP] update examples and readmes 

bannanc:
My goal with this pull request is to get the examples to run correctly, specifically I will:

- [x] update SMIRNOFF_comparison examples to import forcefield_utils correctly
- [ ] update example in README to be a code snippet that works (unless the problem there is an issue with the code somewhere)
- [ ] update README to include more information about ChemicalEnvironments. 
- [ ] Check README and examples in the examples/ChemicalEnvironments/ directory (I assume they're from smarty and are fine, but Christopher made them back in August and I haven't looked at them closely in a long time so I want to make sure they're understandable. 

I have this in a separate Fork because I didn't have permission to push branch changes on the main fork, but I'll move these to a branch when that changes.

### DIFF
--- a/examples/SMIRNOFF_comparison/compare_molecule_energies.py
+++ b/examples/SMIRNOFF_comparison/compare_molecule_energies.py
@@ -1,7 +1,7 @@
 #!/bin/env python
 
 from openforcefield.utils import *
-from openforcefield.typing.engines.smirnoff import ForceField
+from openforcefield.typing.engines.smirnoff import * 
 
 import os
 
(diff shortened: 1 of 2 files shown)


examples/SMIRNOFF_comparison/compare_molecule_energies.py
```diff
@@ -1,7 +1,7 @@
 #!/bin/env python
 
 from openforcefield.utils import *
-from openforcefield.typing.engines.smirnoff import ForceField
+from openforcefield.typing.engines.smirnoff import *
```
jchodera: We should avoid `import *` if at all possible, since this practice is [discouraged by PEP 8](https://www.python.org/dev/peps/pep-0008/#id23) (though we have a lot of non-PEP8-compliant code because we modeled this off of OpenMM, which uses C++ conventions). For now, it's fine to get things working though.

bannanc:
Closing, as I've moved changes to a branch instead of a fork, see pull request #7

PR #9, opened 2017-04-05
Replace SMIRFF with SMIRNOFF throughout

davidlmobley:
This attempts to clean up everywhere SMIRFF (the old name) appears in the repo and replace it with SMIRNOFF, except in one place within the `ForceField` class where I (at least temporarily) retained the ability for this to also process `ffxml` files with `SMIRFF` tags rather than `SMIRNOFF` tags.

### DIFF
--- a/openforcefield/tests/test_smirnoff.py
+++ b/openforcefield/tests/test_smirnoff.py
@@ -372,14 +372,14 @@ def test_create_system_boxes_features(verbose=False):
         for f in check_boxes(forcefield, description="to test Parm@frosst parameters with charge method %s" % str(chargeMethod), chargeMethod=chargeMethod, verbose=verbose):
             yield f
 
-def test_create_system_boxes_smirff99Frosst(verbose=False):
+def test_create_system_boxes_smirnoff99Frosst(verbose=False):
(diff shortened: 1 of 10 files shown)


openforcefield/tests/test_smirnoff.py
```diff
@@ -112,42 +112,42 @@
...
 
 # Set up another, super minimal FF to test that alternate aromaticity
 # models implemented properly
 ffxml_MDL_contents = u"""\
 <?xml version="1.0"?>
 
-<SMIRFF version="0.1" aromaticity_model="OEAroModel_MDL">
+<SMIRNOFF version="0.1" aromaticity_model="OEAroModel_MDL">
 
 !-- SMIRKS (SMIRKS Force Field) minimal file, not intended for use -->
```
bannanc: All "SMIRKS Force Fields" should have the native open added to it. I don't think I caught all of these in smarty or smirnoff99Frosst

davidlmobley:
@bannanc and I discussed also bringing in the latest smirnoff99Frosst into this repo, as well as prioritizing bringing in the utilities for generating it. We'll do that very soon. But, @bannanc - let's actually do that under a separate PR. We can go ahead and merge this one once it passes.

PR #10, opened 2017-04-06
Move in latest forcefield from smarty, update SMIRFF to SMIRNOFF

davidlmobley:
Together with https://github.com/open-forcefield-group/smarty/pull/247, this moves the latest force field files here, updates them to refer to SMIRNOFF rather than SMIRFF, and makes sure each item in `utilities` either occurs in `smarty` or `openforcefield` but not both.


davidlmobley:
@bannanc - do a quick review when you have a minute, then merge?

davidlmobley:
(Please note I have not yet updated the IPython notebook so that it imports from `openforcefield` rather than `smarty`, as I think there is likely some other bookkeeping to do first.)

bannanc:
I think ALL of the ipython notebooks in this repo and the examples need to be updated. There are a lot of examples that were moved over without any changes.

bannanc:
Also, as I noted in smarty, I was confused about which pull request I was on when I suggested to move the molecule files back. I think they should live here as they will be useful for many different experiments in the OFF group. 

Also, do we really want to keep the archive smirff files on github? While they are sometimes useful for use to find sources of Christopher's SMIRKS patterns, I'm not sure they should be public since they are SO CONFUSING. I don't know where a better permanent home for them would be though.

davidlmobley:
> Also, do we really want to keep the archive smirff files on github? While they are sometimes useful for use to find sources of Christopher's SMIRKS patterns, I'm not sure they should be public since they are SO CONFUSING. I don't know where a better permanent home for them would be though.

I suppose we could migrate them to a private repo on open-forcefield-group; that's probably the best thing to do, as they are of no practical value to someone who wasn't involved in the process.

Issue #13, opened 2017-04-06
Move historical force field stuff from Bayly to private repo


davidlmobley:
Under the utilities within the conversion directory, there are historical files Chris Bayly used in making smirnoff99Frosst. These are of no practical value to anyone who wasn't involved in his process, so we should move them out of this public repo to a private one.

davidlmobley:
Closed via #18 .

PR #18, opened 2017-04-14
Remove archival files which are not useful

davidlmobley:
Per #13 , removes archival files used (by hand) in constructing an initial draft of SMIRNOFF99Frosst, as these are of no practical value except to those who were involved in the process.


jchodera:
Thanks!

Issue #22, opened 2017-04-28
Loading forcefield from ffxml in openforcefield/data/forcefield fails


bannanc:
I've tried to use the following script lines:

```
from openforcefield.typing.engines.smirnoff import forcefield
ff = forcefield.ForceField('forcefield/Frosst_AlkEtOH.ffxml')
```

This gets the following error:
```
OSError: Error reading file '/Users/bannanc/anaconda3/lib/python3.6/site-packages/openforcefield/typing/engines/smirnoff/data/forcefield/Frosst_AlkEtOH.ffxml': failed to load external entity "/Users/bannanc/anaconda3/lib/python3.6/site-packages/openforcefield/typing/engines/smirnoff/data/forcefield/Frosst_AlkEtOH.ffxml"
```

There is this [#TODO](https://github.com/open-forcefield-group/openforcefield/blob/master/openforcefield/typing/engines/smirnoff/forcefield.py#L561) in the loadFiles method in forcefield.py

It seems like this would be fixed by using the get_data_filename. If that works I'll put in a pull request today.

davidlmobley:
I'm not sure this qualifies as a bug, but yes, you need `get_data_filename` if you want to specify a relative path.

davidlmobley:
Resolved by #23

jchodera:
OpenMM recently added a way in which other conda packages could register their install paths with `ForceField` so that these directories would be searched too. We should probably add this capability so that, for example, the forcefield packages we deliver can be separate packages from the openforcefield toolkit.

https://github.com/pandegroup/openmm/pull/1789

davidlmobley:
Ooh, cool.

PR #23, opened 2017-04-28
Fixing loadFile for forcefield with files in data directory

bannanc:
I'm addressing the issue #22. I'm taking advantage of the method get_data_filename from utils to get access to the data directory. 

The previous version seemed to look for a data directory at 
openforcefield/typing/engines/smirnoff/data/
instead of just
openforcefield/data/

### DIFF
--- a/openforcefield/typing/engines/smirnoff/forcefield.py
+++ b/openforcefield/typing/engines/smirnoff/forcefield.py
@@ -33,7 +33,7 @@ import lxml.etree as etree
 
 from simtk.openmm.app import element as elem
 from simtk.openmm.app import Topology
-from openforcefield.utils import generateTopologyFromOEMol
+from openforcefield.utils import generateTopologyFromOEMol, get_data_filename
 
 import os
 import math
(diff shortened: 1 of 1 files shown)


bannanc:
Also, @davidlmobley or @jchodera 
should we format the repository to require a code review before merging changes?

davidlmobley:
@bannanc - can you verify that this exception handling syntax still raises a useful exception in the scenario that you specify the name of an FFXML file which is in NEITHER location? (e.g. a file that just doesn't exist?) Sometimes I forget what happens if a new exception of the same type (in this case, `IOError`) is raised within an `except` block which is already testing for the same error.

Otherwise, this looks good.

davidlmobley:
I also just went in and protected `master` so that now status checks must pass before merging, pushes can't be made directly to master, and reviews are required except for admins. THanks.

bannanc:
@davidlmobley I just did a test run:
```
from openforcefield.typing.engines.smirnoff import forcefield
ff = forcefield.ForceField('imaginary_file.ffxml')
```
This returns a `ValueErrror` here is the full error message:
```
ValueError: Sorry! /Users/bannanc/anaconda3/lib/python3.6/site-packages/openforcefield/data/imaginary_file.ffxml does not exist. If you just added it, you'll have to re-install
```

jchodera:
Good catch on the data directory issues, @bannanc!

> should we format the repository to require a code review before merging changes?

Yes!

Issue #25, opened 2017-04-28
Implement Library Charges


davidlmobley:
**Major design decision**:

Currently, the charge method is handled on a "by system" basis; that is, as an argument to `createSystem`. So, all the components of a `System` are charged via the same mechanism, such as AM1-BCC, and we have no mechanism to flag specific components of a system for special treatment. Often, this is what we want, but *water poses challenges* (see #24 ) -- typically, for existing water models, we want to assign library charges and only charge our solute (or non-water solvent) atoms via normal charging schemes like AM1-BCC. 

How shall we achieve this? @jchodera , @bannanc , @andrizzi, thoughts? Some ideas:
- Create a new "special section" in the [SMIRNOFF format](https://github.com/open-forcefield-group/openforcefield/blob/master/The-SMIRNOFF-force-field-format.md) (kind of analogous to the `BondChargeCorrections` section) which allows SMIRKS patterns to specify use of library charges for specific groups
- Allow `chargeMethod` to be a list rather than a string, where a different charge method can be used for different components of the system

Other possibilities?

davidlmobley:
See https://github.com/open-forcefield-group/openforcefield/issues/24#issuecomment-298092869 for possible solutions to this. Probably the best is modifying the nonbonded force section to allow charge overrides, I think.

jchodera:
I think the addition of allowing an additional `charge` attribute to override the atomic charge is the simplest approach for handling special water models to enable comparisons with existing models. 

The real question here is *what are the cases outside of special water models in which we would want to override charge handling*? We already have a *very* flexible approach to molecule-based library charges (by providing `OEMol` molecules with the desired partial charges) as well as an approach to have `createSystem` compute and apply AM1-BCC-like charges to all molecules. The gaping hole in this capability is charging of polymeric systems or very large small molecules where AM1-BCC-like methods would be impractical, but are there really use cases beyond special water models where the ability to do something more than what we have now would be useful?

davidlmobley:
I can't think of that many cases offhand aside from the issues you mention.

I have done work where I had sugars in a protein where I wanted to use GLYCAM parameters for the sugars (with their own charges) and "normal" parameters for the protein and ligand and co-solvents. But in that case, to avoid having to create my own GLYCAM parameterization engine, I presumably would have just created a ParmEd system for the GLYCAM components and merged them with the remainder of the system. So I don't necessarily see this as a use case.

[...]

PR #33, opened 2017-05-01
Add tip3p SMIRNOFF parameters and fix scrambled VdW parameters in mixtures

### LINKED ISSUE #24: Add tip3p.ffxml to data/forcefield/?
Is there a file for this somewhere?

Having a separate file with the parameters for tip3p water would help setting up simulations in water solvents by simply doing
```python
smirnoff_ffxml = get_data_filename('forcefield/smirnoff99Frosst.ffxml')
tip3p_ffxml = get_data_filename('forcefield/tip3p.ffxml')
ff = ForceField(smirnoff_ffxml, tip3p_ffxml)
```
and avoid hardcoding the parameters in a Python script through `ForceField.addParameter()`. I can work on this today if you agree this is the way to go.

### LINKED ISSUE #30: Sigma and epsilon parameters scrambled with multiple molecules
I found a bug that was causing sigma and epsilon parameters to be incorrectly modified when assigning charges with multiple molecules. [This line](https://github.com/open-forcefield-group/openforcefield/blob/master/openforcefield/typing/engines/smirnoff/forcefield.py#L2023) should use `map_atom_index` instead of `atom_index`.

I'll open a PR to fix this and add a test.

andrrizzi:
Fix #24: Add `tip3p.ffxml` to `data/forcefield/`.
Fix #30: Sigma and epsilon parameters scrambled in mixtures.

- I've interpreted the `Date` and `Author` tags related to the creation of the file, but if these were intended to describe date and author of the parameters, I'll change it.
- @jchodera: I've changed the H-H constrained distance w.r.t. to what you proposed in #24 from 1.8532A to 1.5139006545247014A as those are the one that ends up being used in OpenMM. I've also changed the charges to the TIP3P ones: should I leave them there even if they're not currently supported, or do we want to add the `charge` attributes back when #25 will be implemented?
- I've added `data/monomers/tip3p_water.mol2` because `data/monomers/water.mol2` didn't have tip3p charges or a residue name recognized by OpenMM. I'm using it for tests, and I'll use this file to set up SMIRNOFF hydration free energy calculations.
- Note that the test I've added (`test_smirnoff.py: test_tip3p_solvated_molecule_energy`) [adds unnecessary bonds](https://github.com/open-forcefield-group/openforcefield/blob/tip3p-and-fix-30/openforcefield/tests/test_smirnoff.py#L675-L685) as a workaround for ParmEd/ParmEd#868 . Those bonds will have to be removed if/when that's fixed.

### DIFF
--- a/openforcefield/data/forcefield/tip3p.ffxml
+++ b/openforcefield/data/forcefield/tip3p.ffxml
@@ -1,0 +1,19 @@
+<?xml version='1.0' encoding='ASCII'?>
+<SMIRNOFF version="0.1">
+  <!-- SMIRks Native Open Force Field (SMIRNOFF) file -->
+  <Date>2017-04-29</Date>
+  <Author>J. D. Chodera, MSKCC; A. Rizzi, Weill Cornell; C. C. Bannan, UC Irvine</Author>
+  <!-- SMIRNOFF file implementing TIP3P water model. -->
+  <NonbondedForce coulomb14scale="0.833333" lj14scale="0.5" sigma_unit="nanometers" epsilon_unit="kilojoules_per_mole">
+    <!-- TIP3P water oxygen with charge override -->
+    <Atom smirks="[#1]-[#8X2H2+0:1]-[#1]" id="n1" sigma="0.31507524065751241" epsilon="0.635968" charge="-0.834"/>
+    <!-- TIP3P water hydrogen with charge override -->
+    <Atom smirks="[#1:1]-[#8X2H2+0]-[#1]" id="n2" sigma="1" epsilon="0" charge="0.417"/>
+  </NonbondedForce>
+  <Constraints distance_unit="angstroms">
+    <!-- constrain water O-H bond to equilibrium bond length (overrides earlier constraint) -->
+    <Constraint smirks="[#1:1]-[#8X2H2+0:2]-[#1]" id="c1" distance="0.9572"/>
+    <!-- constrain water H...H, calculating equilibrium length from H-O-H equilibrium angle and H-O equilibrium bond lengths -->
+    <Constraint smirks="[#1:1]-[#8X2H2+0]-[#1:2]" id="c2" distance="1.5139006545247014"/>
+  </Constraints>
+</SMIRNOFF>
--- a/openforcefield/data/forcefield/tip3p.ffxml
+++ b/openforcefield/data/forcefield/tip3p.ffxml
@@ -6,8 +6,10 @@
   <!-- SMIRNOFF file implementing TIP3P water model. -->
   <NonbondedForce coulomb14scale="0.833333" lj14scale="0.5" sigma_unit="nanometers" epsilon_unit="kilojoules_per_mole">
     <!-- TIP3P water oxygen with charge override -->
+    <!-- WARNING: charges are not supported yet (see issue openforcefield#25), and they are currently here only for documentation -->
     <Atom smirks="[#1]-[#8X2H2+0:1]-[#1]" id="n1" sigma="0.31507524065751241" epsilon="0.635968" charge="-0.834"/>
     <!-- TIP3P water hydrogen with charge override -->
+    <!-- WARNING: charges are not supported yet (see issue openforcefield#25), and they are currently here only for documentation -->
     <Atom smirks="[#1:1]-[#8X2H2+0]-[#1]" id="n2" sigma="1" epsilon="0" charge="0.417"/>
   </NonbondedForce>
   <Constraints distance_unit="angstroms">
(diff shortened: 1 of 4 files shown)


davidlmobley:
@bannanc - can you take a look as a starting point (mainly at SMIRKS, etc.), then tag me when you're done and I'll look as well?

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -2020,7 +2020,7 @@ def createForce(self, system, topology, nonbondedMethod=NoCutoff, nonbondedCutof
             for atom_mapping in atom_mappings:
                 for (atom_index, map_atom_index) in atom_mapping.items():
                     # Retrieve NB params for reference atom (charge not set yet)
-                    [charge, sigma, epsilon] = force.getParticleParameters(atom_index)
+                    charge, sigma, epsilon = force.getParticleParameters(map_atom_index)
```
bannanc: It's not clear what the difference between `atom_index` and `map_atom_index` is, perhaps we should put a comment above the for loop? I'm not particularly familiar with this portion of the code though so maybe it isn't necessary. @davidlmobley

bannanc:
@davidlmobley 
The SMIRKS patterns look fine to me (at least they are the same ones I included in the issue around this topic). You should probably take a look at the test script, but it makes sense to me. I also commented on the forcefield.py change.

openforcefield/typing/engines/smirnoff/forcefield.py
davidlmobley: This was a bugfix. But a comment wouldn't hurt.

openforcefield/data/forcefield/tip3p.ffxml
```diff
@@ -0,0 +1,19 @@
+<?xml version='1.0' encoding='ASCII'?>
+<SMIRNOFF version="0.1">
+  <!-- SMIRks Native Open Force Field (SMIRNOFF) file -->
+  <Date>2017-04-29</Date>
+  <Author>J. D. Chodera, MSKCC; A. Rizzi, Weill Cornell; C. C. Bannan, UC Irvine</Author>
+  <!-- SMIRNOFF file implementing TIP3P water model. -->
+  <NonbondedForce coulomb14scale="0.833333" lj14scale="0.5" sigma_unit="nanometers" epsilon_unit="kilojoules_per_mole">
+    <!-- TIP3P water oxygen with charge override -->
```
davidlmobley: We have charge overrides here, but the charge override isn't implemented yet, presumably. Maybe needs a warning? Or are you planning on doing that, @andrrizzi ?

openforcefield/tests/test_smirnoff.py
```diff
@@ -668,6 +664,82 @@ def test_gromacs_roundtrip():
     os.remove(grofile)
 
 
+def test_tip3p_solvated_molecule_energy():
```
davidlmobley: Nice job on this test.

openforcefield/typing/engines/smirnoff/forcefield.py
andrrizzi: I agree that we could make the variable names more explicit. `atom_mapping` comes from `topology._reference_to_topology_atom_mappings` which is documented [here](_reference_to_topology_atom_mappings).

Maybe I can rename them `reference_molecule_atom_id` and `topology_atom_id` to make the code clearer?

openforcefield/data/forcefield/tip3p.ffxml
andrrizzi: Yeah, I didn't know what to do about them (see my question in 2nd bullet point above). The charges are ignored currently, but I left them there mainly to be sure they'll be there when we'll implement #25 .

I don't have a strong opinion about them, but writing a comment warning that charges are not working until that issue is tackled sounds like a good idea!

andrrizzi:
Thanks @bannanc @davidlmobley for the review!

> running a bit of dynamics and seeing that things are properly constrained?

Sounds simple enough. I'll add a separate test for that before merging.

openforcefield/typing/engines/smirnoff/forcefield.py
davidlmobley: Sounds perfect, yes. :)

openforcefield/data/forcefield/tip3p.ffxml
davidlmobley: I like having them in the format so we don't have to adjust later, but adding a comment sounds like a good way to avoid confusion. :)

andrrizzi:
Even the second test pass! Merging.

davidlmobley:
Nice work, @andrrizzi ! Thanks!

PR #45, opened 2017-05-31
Add icu dependency, add conda-forge dependency, drop python3.4, support python 3.6

### LINKED ISSUE #44: Add explicit icu 58.* as dependency?
I run into a problem today where I had to explicitly tell conda to install `openforcefield` with `icu 58.*` (a dependency for `lxml`) or version `56.1` got installed, which would get me this error:
```
  File "create_input_files.py", line 12, in <module>
    from openforcefield.typing.engines import smirnoff
  File "/cbio/jclab/home/andrrizzi/miniconda/envs/smirnoff/lib/python3.5/site-packages/openforcefield/typing/engines/smirnoff/__init__.py", line 3, in <module>
    from openforcefield.typing.engines.smirnoff.forcefield import *
  File "/cbio/jclab/home/andrrizzi/miniconda/envs/smirnoff/lib/python3.5/site-packages/openforcefield/typing/engines/smirnoff/forcefield.py", line 32, in <module>
    import lxml.etree as etree
ImportError: libicui18n.so.58: cannot open shared object file: No such file or directory
```
The exact commands I used to solve the issue are
```
conda config --add channels omnia --add channels conda-forge
conda create --name smirnoff yank openforcefield packmol 'icu=58.*'
```
(the `icu=58.*` was not there before). This is not a problem when installing only `openforcefield`, but neither `yank` nor `packmol` requires `icu`, so this must be some sort conda weirdness.

I'd modify the conda recipe to explicitly require `58.*` version to avoid this problem.

andrrizzi:
This fixes the repo conda recipe to add explicitly the `icu` library as a dependency. I'll soon update the one in `conda-recipes` too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,9 @@ install:
   - if [ "$TRAVIS_SECURE_ENV_VARS" == false ]; then echo "OpenEye license will not be installed in forks."; fi
 
 script:
-  # Add omnia channel
+  # Add omnia and conda-forge channel
   - conda config --add channels ${ORGNAME}
+  - conda config --add channels conda-forge
   # Create and activate test environment
   - conda create --yes -n test python=$python
   - source activate test
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,8 @@ script:
 env:
   matrix:
     - python=2.7  CONDA_PY=27
-    - python=3.4  CONDA_PY=34
     - python=3.5  CONDA_PY=35
+    - python=3.6  CONDA_PY=36
 
   global:
     - ORGNAME="omnia"
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ pip install -i https://pypi.anaconda.org/OpenEye/simple OpenEye-toolkits
 ```
 Install `openforcefield` tools via conda:
 ```bash
-conda install --yes -c omnia openforcefield
+conda install --yes -c conda-forge -c omnia openforcefield
 ```
 
 # Tools
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - numpy
     - networkx
     - lxml
-    - icu ==58.*  # This is a lxml dependency but sometimes conda installs version 56
+    - icu 58*  # This is a lxml dependency but sometimes conda installs version 56
     - openmoltools >=0.7.3
     - parmed
     - matplotlib
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -14,6 +14,7 @@ requirements:
     - python
     - setuptools
     - pandas
+    - icu ==58.*  # This is a lxml dependency but sometimes conda installs version 56
 
   run:
     - python
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - numpy
     - networkx
     - lxml
-    - icu =58.*  # This is a lxml dependency but sometimes conda installs version 56
+    - icu ==58.*  # This is a lxml dependency but sometimes conda installs version 56
     - openmoltools >=0.7.3
     - parmed
     - matplotlib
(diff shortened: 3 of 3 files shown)


jchodera:
It looks like there's an error in that specification:
```
0.86s$ conda build devtools/conda-recipe
Can't build /home/travis/build/open-forcefield-group/openforcefield/devtools/conda-recipe due to environment creation error:
Invalid spec: =58.*
```

jchodera:
It looks like the `conda-forge` channel has to be added:
```
Can't build /home/travis/build/open-forcefield-group/openforcefield/devtools/conda-recipe due to unsatisfiable dependencies:
Packages/dependencies missing in current linux-64 channels: 
  - openforcefield 0.0.0 py27_0 -> icu ==58.*
  - icu ==58.*
```
@lnaden can help update the travis testing boilerplate to add the `conda-forge` channel in our "new" standard way.

andrrizzi:
I think I can figure it out from the changes @Lnaden did with YANK.

@davidlmobley if I do this now you'll have to add the `conda-forge` channel on top of `omnia` to install `openforcefield`.

davidlmobley:
@andrrizzi - is there a major downside to that? I haven't done anything with conda-forge yet, but it sounds like this is going to be the "new way" of doing things, right? I'm OK with that as long as you guys educate me. :)

jchodera:
@davidlmobley : Omnia is migrating to `conda-forge`, so this cannot be prevented, only delayed.
See here: http://www.omnia.md/blog/2017/5/18/omnias-conda-distribution-now-linked-with-conda-forge

davidlmobley:
Yes, @jchodera - have this on my radar to start trying to understand. I wasn't objecting, just asking Andrea whether there was some downside, since his phrasing suggested it was a burden somehow.

andrrizzi:
@davidlmobley no downsides I think, but you'll have to `conda config --add channels conda-forge` to make the installation work, and I think the one liner becomes `conda install -c omnia -c conda-forge openforcefield` (the order of the channels may matter).

Sorry for the misleading comment, I just wanted you and your group to be aware of this before I go ahead.

Lnaden:
Slight correction. its `conda install -c conda-forge -c omnia`, the order is important.

davidlmobley:
Oh, then let's absolutely do that, @andrrizzi . Just make sure to update the README.md to indicate this. :)

andrrizzi:
Thanks @Lnaden ! I'll do it tonight then.

andrrizzi:
Auch! Right, another thing to consider is that `conda-forge` supports only python 3.5 and python 3.6, so if somebody there is still using python 3.4, they might have to upgrade. Omnia has dropped support for 3.4 now, so the current `openforcefield` release on `anaconda` doesn't have files for it anyway.

andrrizzi:
Ok! I think I've figured it out.

davidlmobley:
> Auch! Right, another thing to consider is that conda-forge supports only python 3.5 and python 3.6, so if somebody there is still using python 3.4, they might have to upgrade. Omnia has dropped support for 3.4 now, so the current openforcefield release on anaconda doesn't have files for it anyway.

This sounds all right to me. Are you at the point where you are ready for review, @andrrizzi ?

andrrizzi:
@davidlmobley , yes, this is good to go on my side.

davidlmobley:
Though, after merging, @andrrizzi - I realized you probably want to note the relevant Python support in the README.md. Want to do another PR to update that?

jchodera:
> Though, after merging, @andrrizzi - I realized you probably want to note the relevant Python support in the README.md. Want to do another PR to update that?

Note that we aren't actually dropping support for 3.4 until we update the conda recipe in omnia-recipes.

andrrizzi:
> Want to do another PR to update that?

Sure.

> until we update the conda recipe in omnia-recipes

I think the recipe in `conda-recipes` doesn't have control on the Python versions supported. At least, the latest release of `openforcefield` on anaconda already doesn't have Python 3.4 files. It could be that it's controlled by the main scripts of the repo rather than the single recipes.

I'll soon update the recipe in `conda-recipes` too and bump build number.

Lnaden:
> I think the recipe in conda-recipes doesn't have control on the Python versions supported

The python versions are controlled by the [main build script here](https://github.com/omnia-md/conda-recipes/blob/04dfe91801249c42911408dddf2db951b24edfe6/conda-build-all#L94-L98), I dropped Python 3.4 support as part of the [change to conda-forge priority](https://github.com/omnia-md/conda-recipes/pull/778).

PR #53, opened 2017-08-03
Add monovalent ion parameters from AMBER forcefields

davidlmobley:
This PR adds monovalent ion parameters to smirnoff99Frosst's source files and to the FF; once merged I will also update the smirnoff99Frosst repo.

@bannanc - please take a quick look to make sure everything looks in order to you. To help you out, here's a copy and paste of the source material from parm99:
```
  Li+      1.1370  0.0183             Li+ Aqvist JPC 1990,94,8021. (adapted)
  Na+      1.8680  0.00277            Na+ Aqvist JPC 1990,94,8021. (adapted)
  K+       2.6580  0.000328           K+  Aqvist JPC 1990,94,8021. (adapted)
  Rb+      2.9560  0.00017            Rb+ Aqvist JPC 1990,94,8021. (adapted)
  Cs+      3.3950  0.0000806          Cs+ Aqvist JPC 1990,94,8021. (adapted)
  F-       1.75    0.061              Gough et al. JCC 13,(1992),963.
  Cl-      1.948   0.265              Fox, JPCB,102,8070,(98),flex.mdl CHCl3
  Br-      2.22    0.320              Junmei(?)
  I-       2.35    0.40               JCC,7,(1986),230;
```


jchodera:
I'm pretty sure you don't want most of those parameters---those NaCl parameters are vulnerable to crystallization, for example. Have you read through the latest AmberTools manual discussing good modern choices for ions?

jchodera:
See Section 3.6 of the [Amber 17 Manual](http://ambermd.org/doc12/Amber17.pdf).

davidlmobley:
Looks like there are three sets, @jchodera , for each water model. Assuming (for now) we are using TIP3P, which do you propose?

jchodera:
I'd suggest bundling TIP3P and the related ion parameters into a single `tip3p.ffxml` file, and TIP4P-Ew and related ion parameters into `tip4pew.ffxml` file.

davidlmobley:
AMBER manual:
> For the 12-6 LJ nonbonded model, three different parameter sets are available for each water model to meet different requirements:
> 1. 12-6 normal usage set. In which we have HFE set of the monovalent ions (could reproduce the experimental HFE),[97] the Compromise (CM) set of divalent ions (which could reproduce the experimental relative HFE and CN values),[95] and IOD set (could reproduce the experimental IOD) for the trivalent and tetravalent ions.[96] These parameters are recommended to be used in the normal MD simulations. This is because for the monovalent ions the error of the 12-6 LJ nonbonded model is pretty small (a CM set may not needed since HFE or IOD set are pretty close to each other) while for the trivalent and tetravalent metal ions the 12-6 LJ nonboned model has relative big error (a CM set could have big error for both HFE and IOD at this moment).
> 2. 12-6 HFE set to reproduce experimental HFE.[94, 96, 97] The HFE parameter set has limited error for monovalent ions, while could have remarkable error for highly charged ions. Since we use the HFE set for monovalent ions in the 12-6 normal usage set, so we don’t have a sepecific HFE set parameter file for monovalent ions.
> 3. 12-6 IOD set to reproduce experimental IOD.[94, 96, 97] Since the ion with certain parameter could reproduce similar IOD values in the three water models, so the IOD set parameters of three water models were designed identical (for the monovalent and divalent metal ions, while for the trivalent and tetravalent ions, the IOD set are estimated for each water model separately). The IOD parameter set are recommmended to be used in the structural refinement or for structural property orientated investigation.

Probably we will have to make all three sets available in separate files, though we may also want to pick a "default" set to include in the basic force field (e.g. have SMIRNOFF99Frosst include one specific set which one can override by including a different ffxml file).

davidlmobley:
@jchodera 

> I'd suggest bundling TIP3P and the related ion parameters into a single tip3p.ffxml file, and TIP4P-Ew and related ion parameters into tip4pew.ffxml file.

That doesn't solve it since there are THREE SETS PER WATER MODEL.

utilities/convert_frosst/smirnoffishFrcmod.parm99Frosst.txt
```diff
@@ -49,6 +50,17 @@ NONBON
...
+# Monovalent ions from parm99
+[#3:1]          1.1370  0.0183   parm99 
+[#11:1]         1.8680  0.00277  parm99 
+[#19:1]         2.6580  0.000328 parm99
+[#37:1]         2.9560  0.00017  parm99
+[#55:1]         3.3950  0.0000806 parm99
+[#9X0:1]        1.75    0.061    parm99
+[#17X0:1]       1.948   0.265    parm99
+[#35X0:1]       2.22    0.320    parm99
+[#53X0:1]       2.35    0.40     parm99
```
bannanc: Adding formal charges to the SMIRKS might make it easier to understand what these are. Though, they aren't necessary for actually assigning the parameters correctly.

utilities/convert_frosst/smirnoff99Frosst_AlkEthOH.ffxml
```diff
@@ -129,77 +129,77 @@
...
+    <Proper smirks="[#17:1]-[#6X4:2]-[#6X4:3]-[#17:4]" id="t8" idivf1="1" k1="0.000" periodicity1="3" phase1="0.0" idivf2="1" phase2="180.0" k2="0.450" periodicity2="1"/>
+    <Proper smirks="[#35:1]-[#6X4:2]-[#6X4:3]-[#35:4]" id="t9" idivf1="1" k1="0.000" periodicity1="3" phase1="0.0" idivf2="1" phase2="180.0" k2="0.000" periodicity2="1"/>
+    <Proper smirks="[#1:1]-[#6X4:2]-[#6X4:3]-[#8X2:4]" id="t10" idivf1="1" k1="0.000" periodicity1="3" phase1="0.0" idivf2="1" phase2="0.0" k2="0.250" periodicity2="1" periodicity3="3" k3="0.0" phase3="0.0" idivf3="1" phase4="0.0" k4="0.25" idivf4="1" periodicity4="1"/>
+    <Proper smirks="[#1:1]-[#6X4:2]-[#6X4:3]-[#9:4]" id="t11" idivf1="1" k1="0.000" periodicity1="3" phase1="0.0" idivf2="1" phase2="0.0" k2="0.190" periodicity2="1"/>
+    <Proper smirks="[#1:1]-[#6X4:2]-[#6X4:3]-[#17:4]" id="t12" idivf1="1" k1="0.000" periodicity1="3" phase1="0.0" idivf2="1" phase2="0.0" k2="0.250" periodicity2="1"/>
+    <Proper smirks="[#1:1]-[#6X4:2]-[#6X4:3]-[#35:4]" id="t13" idivf1="1" k1="0.000" periodicity1="3" phase1="0.0" idivf2="1" phase2="0.0" k2="0.550" periodicity2="1"/>
```
bannanc: Why are there so many changes in this file?

bannanc:
I still need to look more closely at the documentation, but I agree that we should pick a default to be in SMIRNOFF99Frosst.

utilities/convert_frosst/smirnoff99Frosst_AlkEthOH.ffxml
davidlmobley: Looks like it hadn't been re-generated in a long time (I don't actually use it for anything normally).

jchodera:
> Probably we will have to make all three sets available in separate files, though we may also want to pick a "default" set to include in the basic force field (e.g. have SMIRNOFF99Frosst include one specific set which one can override by including a different ffxml file).

Whoops---sorry about missing the "three per water model" issue. I agree it's good to have override choices, and set 1 (the normal usage set) should be our default.

utilities/convert_frosst/smirnoff99Frosst_AlkEthOH.ffxml
bannanc: Oh, I missed that this was in the AlkEthOH edition, my bad

utilities/convert_frosst/smirnoffishFrcmod.parm99Frosst.txt
davidlmobley: Thanks, fixed!

davidlmobley:
OK, so here's my plan:
- Use Joung/Cheatham "general purpose" TIP3P parameters by default in the force field (in this PR)
- Make separate FFXML files containing the various parameters AMBER provides for the various water models (in a separate PR) and provide these, so they can be used to override the defaults when desired

Since there are three sets of parameters per water model, I do NOT plan to include these in the water FFXML files, but just have a separate set of FFXML files that would be loaded by the user depending on what ion parameters they want.

PR #60, opened 2017-09-18
Remove extra OEAssignImplicitHydrogens which could result in protonation state changes

### LINKED ISSUE #59: Certain molecules end up with modified protonation states prior to SMIRKS matching
Shuzhe Wang found that certain molecules appear to result in the application of a parameter which is not the expected one; for example, he provided an example (below) of a molecule containing trivalent nitrogens which ends up with a torsion involving tetravalent nitrogens. It appears that this results from an extra assignment of implicit hydrogens when we are preparing to get SMIRKS matches by switching to the desired aromaticity model (i.e. in https://github.com/open-forcefield-group/openforcefield/blob/master/openforcefield/typing/engines/smirnoff/forcefield.py#L105-L110 ); the extra call to `OEAssignImplicitHydrogens` appears to sometimes result in a change in protonation state from the desired (provided) protonation state, which is obviously not what we want.

The specific case Shuzhe provided was `[H]c1c([H])c(N([H])S(=O)(=O)C([H])([H])[H])c([H])c([H])c1OC([H])([H])C([H])([H])N(C([H])([H])[H])C([H])([H])C([H])([H])c1c([H])c([H])c(N([H])S(=O)(=O)C([H])([H])[H])c([H])c1[H]` and this ends up getting a SMIRKS match to `[*:1]-[#6X4:2]-[#7X4:3]-[*:4]`.

Fix going up momentarily.

davidlmobley:
Attempts to resolve #59 

@hjuinj - can you test this out?


davidlmobley:
He notes (via Slack) that this works now, so merging.

Issue #41, opened 2017-05-25
Implement off-center charges as discussed in SMARTY issue


davidlmobley:
As discussed on SMARTY, we need to build in handling of off-center charges (https://github.com/open-forcefield-group/smarty/issues/132 ) . As I understand it, we are waiting on some changes to be made to OpenMM. @jchodera , what do we need to do to get those into OpenMM?

jchodera:
We first need to decide how to handle off-center charges for multisite water models. We can implement those now.

Separate from that, I do need to figure out what needs to get implemented in OpenMM and open an issue to request extensions to the current supported virtual sites.

jchodera:
But maybe we should have two separate issues for those tasks?

jchodera:
Here's the proposal from https://github.com/open-forcefield-group/smarty/issues/132#issuecomment-252050844 reproduced here:
```XML
<!-- sigma hole for halogens: "distance" denotes distance along the 2->1 bond vector -->
<VirtualSite type="BondCharge" smirks="[Cl:1]-[C:2]" distance="0.30"/>
<!-- monovalent lone pairs: carbonyl -->
<!-- inPlaneAngle is angle charge site (X) makes with 1 and 2, i.e. X-1-2 -->
<!-- outOfPlaneAngle is angle charge site (X) makes out of the plane of 2-1-3 measured from 1 -->
<VirtualSite type="BondCharge" smirks="[O:1]=[C:2]-[*:3]" distance="0.30" outOfPlaneAngle="0" inPlaneAngle="120"/>
<!-- divalent lone pair: pyrimidine -->
<!-- single charge site lies in plane, bisecting the angle left by taking 360 degrees minus angle(2,1,3) -->
<VirtualSite type="DivalentLonePair" smirks="[*:2]~[#7X2:1]~[*:3]" distance="0.30"/>
<!-- trivalent nitrogen lone pair -->
<!-- charge sites lie above and below the nitrogen at specified distances from the nitrogen, along the vector perpendicular to the plane of (2,3,4) that passes through the nitrogen. If the nitrogen is co-planar with the connected atom, charge sites are simply above and below the plane-->
<!-- Positive and negative values refer to above or below the nitrogen as measured relative to the plane of (2,3,4), i.e. below the nitrogen means nearer the 2,3,4 plane unless they are co-planar -->
[...]

PR #67, opened 2017-10-05
[WIP] Implement handling of virtual sites for off-center charges/etc.

davidlmobley:
This will likely take a while to finish, but I'm starting work on implementing virtual sites for off-center charges (and LJ parameters if desired). So far this just adds a draft spec of virtual sites into forcefield format document.

Done so far:
- [x] Update spec draft to explain final spec
- [x] Add diagrams of virtual sites

To do:
- [ ] Write class/generators
- [ ] Update XML parser
- [ ] Test on TIP4P, TIP5P water (energy validations should be possible)
- [ ] Write tests
...


jchodera:
cc: #41

davidlmobley:
@jchodera can you by any chance provide a pointer to docs or example code showing how to PUT something at a virtual site? I have a test notebook where I'm trying to set up TIP4P to make sure I know how to do it, and I've defined the appropriate virtual site, but I cant figure out (so far) how to put a charge there... Basic, but not obvious in the docs or code I've found.

jchodera:
Here's a gist illustrating how to do this:
https://gist.github.com/anonymous/be8693af41fe070638a27a834cfcb7e9

davidlmobley:
Awesome, that looks like it will get me going. Thanks so much!

davidlmobley:
@jchodera - from your GIST, I'm confused about part of it -- I think I'm missing something basic:
```python
# Add a virtual site to the System
# This appends one more particle to the System
virtual_site_index = system.addParticle(0.0 * unit.amu) # mass can be zero
virtual_site = openmm.OutOfPlaneSite(0, 1, 2, 1.0, 1.0, 0.0)
system.setVirtualSite(virtual_site_index, virtual_site)

# Add another site to the Lennard-Jones force for this virtual site
force.addParticle(0*charge, sigma, epsilon)
```

How does the `force.addParticle` here know that this refers to the virtual site? None of the arguments have anything to do with the virtual site. Is there an order-of-operations effect here where, if you add a particle to a force after creating a virtual site, the particle is assumed to be the virtual site? I haven't been able to find where this is documented, so I'm having trouble figuring out why THIS force is different from the ones you create earlier in your GIST which are "normal" particles. (Obviously this one is also massless, but I can't see how it is connected to the virtual site.)

What am I missing? Or is your code missing a line which connects the force/particle to the virtual site?

jchodera:
> How does the force.addParticle here know that this refers to the virtual site? None of the arguments have anything to do with the virtual site. Is there an order-of-operations effect here where, if you add a particle to a force after creating a virtual site, the particle is assumed to be the virtual site? I haven't been able to find where this is documented, so I'm having trouble figuring out why THIS force is different from the ones you create earlier in your GIST which are "normal" particles. (Obviously this one is also massless, but I can't see how it is connected to the virtual site.)

The `force.addParticle()` call adds a new charge/LJ site to the *end* of the current list of sites. `force.addParticle()` actually returns the index of the particle that was just added, which should (in this example) match the `virtual_site_index`. You will want to be careful that these match up---an `assert` to make sure they are equal couldn't hurt.
```python
# Add a virtual site to the System
virtual_site_index = system.addParticle(0.0 * unit.amu) # mass can be zero
nonbonded_site_index = force.addParticle(0*charge, sigma, epsilon)
assert virtual_site_index == nonbonded_site_index
virtual_site = openmm.OutOfPlaneSite(0, 1, 2, 1.0, 1.0, 0.0)
system.setVirtualSite(virtual_site_index, virtual_site)
```

davidlmobley:
@jchodera - now that I have figured out how to work with virtual sites, I need some help with an architectural issue. Specifically, how do we want to deal with the modified `Topology` and `coordinates` we will need for working with virtual sites? 

Right now, application of a SMIRNOFF consists of basically one step after loading the `forcefield`: We have an initial OpenMM `Topology` (and, currently, `molecules` corresponding to a list of molecules in the system), we we provide these to `forcefield.createSystem` to create and return the relevant OpenMM `system`, which one can then simulate using the original `Topology` and coordinates.

When we add virtual sites, the `system` returned will need to contain additional particles beyond what would be represented in an INITIAL OpenMM `Topology` such as one read from a PDB file. This will be useless without a corresponding `Topology`, so how should we make application of a SMIRNOFF specifying virtual sites easy for the user/how should we handle this?  Naively it seems like we have two main options:
1) Modify `createSystem` so it also returns a `Topology` corresponding to the contents of the new system if has added any virtual sites/particles, or
2) Add a pre-processing step before `createSystem` for SMIRNOFFs using virtual sites, which would provide or create a new `Topology` and coordinates which have the virtual sites added; this would then be used by `createSystem`.

I think I am leaning towards option 2, since it doesn't break our existing approach for current SMIRNOFFs and simply means application of virtual-site containing SMIRNOFFs is slightly more complex.

It might work something like this, for a SMIRNOFF containing virtual sites, assuming we already have a topology and OEMols of the components:
```python
ff = ForceField( 'vsite_forcefield.ffxml')
extended_topology, extended_positions = ff.extendSystem( topology, positions)
system = ff.createSystem( extended_topology, molecules)
```

What I'm imagining is that `extended_topology` would contain both the original (chemical) Topology as well as the extended Topology containing virtual sites (as well as information mapping atoms between the two), as the SMIRKS matching functionality in `ForceField` will need access to the original topology/molecules in order to do substructure searches, but it will also need access to the new topology information/mappings in order to know what particle numbers in the new system the relevant parameters should be assigned to. `createSystem` would raise an exception if it's applied to a regular OpenMM `Topology` rather than an extended topology containing this additional information, IF the SMIRNOFF in question specifies use of virtual sites.

`extended_positions` is a hair tricky; probably it would just be positions including the positions of the virtual sites. However, it would probably be good if we had an easy way for users to recover just the positions of their "chemical" atoms later from the full position array without too much hassle, such as if they want to convert back into OEMol or RDKit molecules from our `extended_topology`. Perhaps we would address this by providing utility functionality which takes an `extended_topology` and `extended_positions` and converts back into a standard OpenMM topology and positions.

Thoughts?

jchodera:
The second approach is how OpenMM handles this. See [`Modeller.addExtraParticles()`](http://docs.openmm.org/latest/api-python/generated/simtk.openmm.app.modeller.Modeller.html#simtk.openmm.app.modeller.Modeller.addExtraParticles), though I realize we don't actually document this in the User Guide in [Model Building and Editing](http://docs.openmm.org/latest/userguide/application.html#model-building-and-editing). Maybe raise an OpenMM issue that documentation of this feature is missing from the User Guide?

This is also likely the best approach for us, but let me think about this for a few hours. We've been aiming to make SMIRNOFF a true drop-in replacement for OpenMM's `app.ForceField`, but I wonder if that's really the best goal, and whether we might try to strive for simplicity instead even if it means evolving the API a bit.

For example instead of 
```python
ff = ForceField( 'vsite_forcefield.ffxml')
extended_topology, extended_positions = ff.extendSystem( topology, positions)
system = ff.createSystem( extended_topology, molecules)
```
we might want to just evolve the API to aim for
```python
forcefield = ForceField( 'vsite_forcefield.ffxml')
[system, new_topology] = forcefield.createSystem(topology)
```
where I think our real goals are to
* eliminate the need for `molecules` by using a `Topology` representation that also contains the necessary information about bonds or molecular identities
* automatically add new virtual sites or Drude particles that are then represented in `new_topology`
* allow automated charging, which may mean we want to add an optional or required `positions` argument to `createSystem` if there's any situation in which an initial set of positions (instead of automatically-generated conformers) might be wanted

jchodera:
This might also be a good time to create our own `Topology` object with the features you describe attached to it. We discussed some of the other features we would like at our Tuesday meeting. Maybe we should even open a new issue to discuss what features we might want, and what we can implement now vs punt to later.

davidlmobley:
I agree that this is definitely time to make our own `Topology`. I am OK with changing the API (and this might also alleviate some confusion wherein people think that, since the API is so similar, they should be able to use our FFXML files with `Modeller` and other aspects of OpenMM) though, so let me know what you think after you think about it for a few hours.

davidlmobley:
@leeping - you can check out the basic ideas for virtual sites in the spec here (though this PR will get closed and the material will be brought in separately); please let us know if it covers all the use cases you have in mind.

leeping:
@davidlmobley - Thanks for the heads-up, and I will take a look.  What should I look at specifically, so that I don't miss anything? (i.e. should I look at a file in this PR, and/or the contents of this thread)?

davidlmobley:
@leeping - check out the "virtual sites" section in this document: https://github.com/openforcefield/openforcefield/blob/510163eb0607c69f91385208f6a0f0c894525708/The-SMIRNOFF-force-field-format.md

leeping:
Thanks for sharing the spec - I think the virtual sites section could benefit from having some equations that give the site position as a function of the host atom positions.  

My experience is that the virtual sites available in Gromacs are fairly well thought-out, so I used that as a point of comparison.  The following image is from the Gromacs manual:

screenshot 2018-01-08 11 09 40

Here is the rough correspondence between SMIRNOFF and Gromacs, but I can't guarantee they are mathematically the same.  

2 -> BondCharge
3 -> Linear combination of three atomic positions; similar to DivalentLonePair with zero angle
3fd -> In the plane of three atoms with a fixed distance from a selected atom; also similar to DivalentLonePair with zero angle
3fad -> In the plane of three atoms with a fixed distance and in-plane angle from a selected atom
3out -> Out of the plane of three atoms; similar to DivalentLonePair with nonzero angle
4fdn -> Fixed distance from one atom, in the direction normal to plane formed by three other atoms; similar to TrivalentLonePair

So GROMACS has "3fad" which is not in the SMIRNOFF spec, and SMIRNOFF has "MonovalentLonePair" that is not in the GROMACS spec.  I also have a sense that a virtual site defined as the centroid or center-of-mass of an arbitrary number of atoms could be very helpful in general.  

My overall suggestion is to consider adding "3fad" as well as a "centroid / COM" virtual site.

A few side notes: 

- We should be careful about TrivalentLonePair because the GROMACS manual says they took special care to avoid some instabilities when all the atoms were in the same plane.

- As a side note I'd like the option to charge virtual sites directly so that it's not required to always use charge increments.  This is because some water models like TIP4P has all of the negative charge on the virtual site.  (Of course we could work backwards and figure out the charges / increments for TIP4P, but it still complicates implementation; also, if the oxygen charge is not exactly zero, the MD code may also not realize that there are computational savings.)  To avoid confusion, the direct charge and charge increment parameters should not both be set simultaneously.

davidlmobley:
@leeping - probably a good idea to actually have that in the spec. Here's my notes on what these would be using (though I have a note that we need to check the signs):

* `BondCharge`: `LocalCoordinatesSite([index1, index2, index3], originWeights=[0, 1, 0], xWeights=[-1,1, 0], yWeights=[-1, 0.5, 0.5], localPosition=[distance, 0, 0]` 
* `MonovalentLonePair`: `LocalCoordinatesSite([index1, index2, index3], originWeights=[1, 0, 0], xWeights=[-1,1, 0], yWeights=[0, -1, 1], localPosition=[-distance*cos(theta2)*cos(180-theta1), -distance*cos(theta2)*sin(180-theta1), -distance*sin(theta2)]` 
* `DivalentLonePair`: As in pyrimidine, something like `LocalCoordinatesSite([index1, index2, index3], originWeights=[1, 0, 0], xWeights=[-1,0.5, 0.5], yWeights=[0.5, -0.5, 0], localPosition=[-distance, 0, 0] )`; more general case such as TIP5P, `LocalCoordinatesSite([index1, index2, index3], originWeights=[1, 0, 0], xWeights=[-1,0.5, 0.5], yWeights=[0.5, -0.5, 0], localPosition=[-distance*cos(outOfPlaneAngle), 0, distance*sin(outOfPlaneAngle)] )`
* `TrivalentLonePair`: `LocalCoordinatesSite([index1, index2, index3, index4], originWeights=[1, 0, 0, 0], xWeights=[0,1, -1,0], yWeights=[-0, 11, 0, -1], localPosition=[0, 0, (+/-)distance] )`

davidlmobley:
The larger point, @leeping , is that we can trivially support anything allowed by the OpenMM virtual sites (which are quite flexible) so the main thing is to come up with what else you would want and what it should be called/how it would be defined in teh SMIRNOFF spec. i.e. if you want `3fad`, what would we call it and how would the parameters be specified?

jchodera:
I've also been trying to reason through the spec changes we discussed:

### Old way
```XML
<VirtualSites distanceUnits="angstroms" angleUnits="degrees" sigma_unit="angstroms" epsilon_unit="kilocalories_per_mole">
    <!-- sigma hole for halogens: "distance" denotes distance along the 2->1 bond vector, measured from atom 2 -->
    <!-- Specify that 0.2 charge from atom 1 and 0.1 charge units from atom 2 are to be moved to the virtual site, and a small Lennard-Jones site is to be added (sigma = 0.1*angstroms, epsilon=0.05*kcal/mol) -->
    <VirtualSite type="BondCharge" smirks="[Cl:1]-[C:2]" distance="0.30" charge1increment="+0.2" charge2increment="+0.1" sigma="0.1" epsilon="0.05" />
    <!-- Charge increments can extend out to as many atoms as are labeled, e.g. with a third atom: -->
    <VirtualSite type="BondCharge" smirks="[Cl:1]-[C:2]~[*:3]" distance="0.30" charge1increment="+0.1" charge2increment="+0.1" charge3increment="+0.05" sigma="0.1" epsilon="0.05" />
    <!-- monovalent lone pairs: carbonyl -->
    <!-- X denotes the charge site, and P denotes the projection of the charge site into the plane of 1 and 2. -->
    <!-- inPlaneAngle is angle point P makes with 1 and 2, i.e. P-1-2 -->
    <!-- outOfPlaneAngle is angle charge site (X) makes out of the plane of 2-1-3 (and P) measured from 1 -->
    <!-- Since unspecified here, sigma and epsilon for the virtual site default to zero -->
    <VirtualSite type="MonovalentLonePair" smirks="[O:1]=[C:2]-[*:3]" distance="0.30" outOfPlaneAngle="0" inPlaneAngle="120" charge1increment="+0.2" />
    <!-- divalent lone pair: pyrimidine, TIP4P, TIP5P -->
    <!-- The atoms 2-1-3 define the X-Y plane, with Z perpendicular. If outOfPlaneAngle is 0, the charge site is a specified distance along the in-plane vector which bisects the angle left by taking 360 degrees minus angle(2,1,3). If outOfPlaneAngle is nonzero, the charge sites lie out of the plane by the specified angle (at the specified distance) and their in-plane projection lines along the angle's bisector. -->
    <VirtualSite type="DivalentLonePair" smirks="[*:2]~[#7X2:1]~[*:3]" distance="0.30" OfPlaneAngle="0.0" charge1increment="+0.1" />
    <!-- trivalent nitrogen lone pair -->
    <!-- charge sites lie above and below the nitrogen at specified distances from the nitrogen, along the vector perpendicular to the plane of (2,3,4) that passes through the nitrogen. If the nitrogen is co-planar with the connected atom, charge sites are simply above and below the plane-->
    <!-- Positive and negative values refer to above or below the nitrogen as measured relative to the plane of (2,3,4), i.e. below the nitrogen means nearer the 2,3,4 plane unless they are co-planar -->
    <VirtualSite type="TrivalentLonePair" smirks="[*:2]~[#7X3:1](~[*:4])~[*:3]" distance="0.30" charge1increment="+0.1"/>
    <VirtualSite type="TrivalentLonePair" smirks="[*:2]~[#7X3:1](~[*:4])~[*:3]" distance="-0.30" charge1increment="+0.1"/>
</VirtualSites>
```
### Proposed new way
We suggested we might want to give virtual sites names that can allow us to use virtual site locations (e.g. `vsite="vs1"`) in place of SMARTS strings. But this now makes it problematic to move charges from other atoms unless they are pre-defined as part of the virtual site. Also, it's unclear how we treat electrostatic exclusions on virtual sites---should they be excluded from interacting with the atoms that define them only, or any tagged atoms, or is there some more complex way they count as "bonded" for the purposes of determining exclusions or exceptions?

See especially the NOTEs below. 
```XML
<!-- Define virtual site locations -->
<VirtualSites distance_units="angstroms" angle_units="degrees"">
    <!-- Define a virtual site for the sigma hole for halogens: "distance" denotes distance along the 2->1 bond vector, measured from atom 2 -->
    <VirtualSite vsite="vs1" type="BondCharge" smirks="[Cl:1]-[C:2]" distance="0.30" />
    <!-- monovalent lone pairs: carbonyl -->
    <!-- X denotes the charge site, and P denotes the projection of the charge site into the plane of 1 and 2. -->
    <!-- inPlaneAngle is angle point P makes with 1 and 2, i.e. P-1-2 -->
    <!-- outOfPlaneAngle is angle charge site (X) makes out of the plane of 2-1-3 (and P) measured from 1 -->
    <!-- NOTE: We must pre-specify any atoms we may want to draw charges from, weo we define both atoms bonded to the carbon here.
    <VirtualSite vsite="vs2" type="MonovalentLonePair" smirks="[O:1]=[C:2]([*:4])-[*:3]" distance="0.30" outOfPlaneAngle="0" inPlaneAngle="120"/>
</VirtualSites>
...
<!-- Define vdW sites -->
<vdWForce>
    <Site vsite="vs1" sigma="0.1" epsilon="0.05">
    <Site vsite="vs2" sigma="0.1" epsilon="0.0">
</vdWForce>
...
<!-- Define charge increments -->
<BondChargeIncrements>
    <!-- steal 0.2 charge units from tagged atom 1 and 0.1 charge units from tagged atom 2 in original VirtualSite definition -->
    <Increment vsite="vs1" charge1increment="+0.2" charge2increment="+0.1"/>
    <!-- also steal charge units from tagged atoms not involved in the actual position determination -->
    <Increment vsite="vs2" charge1increment="+0.2" charge2increment="+0.1" charge3increment="-0.05" charge4increment="-0.05"/>
</BondChargeIncrements>
...
<PointPolarizableDipoles>
    <!-- put a point-polarizable dipole on a virtual site -->
    <!-- NOTE: How do we count nearby atoms or atoms that determine the virtual site geometry in electrostatic exclusions or Thole damping?
    <PointPolarizableDipole vsite="vs1" alpha="0.05">
</PointPolarizableDipoles>
```

davidlmobley:
> But this now makes it problematic to move charges from other atoms unless they are pre-defined as part of the virtual site.

I'm not clear why this is especially problematic, @jchodera . It seems like we just make it so every atom that is part of the virtual site as a tagged atom is eligible to have charge transferred to/from it when the increment are defined, but they may or may not have charge transferred. , e.g. `charge1increment` would be charge transferred to (from?) the first labeled atom (if any), `charge2increment` the second, etc. Why is this problematic? 

> <!-- NOTE: How do we count nearby atoms or atoms that determine the virtual site geometry in electrostatic exclusions or Thole damping?

Maybe we want to just add `exclusion1`, `exclusion2`, ... `exclusionN`, ...; and `damping1`, `damping2`, ... `dampingN` where N is the numbered of labeled atoms, which would allow people to specify whatever exclusions they want? Basically just anything included in the SMIRKS pattern could be labeled, and anything labeled can have a charge increment, an exclusion value (boolean or scalar, not sure...) and a damping parameter?

jchodera:
If we're OK with pretagging every atom that may be subsequently referenced, I think we should be OK. let me rework how those files will look using this scheme.

trevorgokey:
A few notes, since I am working on this now:

1. I think 3fad is equivalent to MonovalentLonePair, if outOfPlaneAngle == 0.0

2. Doing a centroid vsite was easy enough with openmm, so I have plugged it in (but haven't thought about the spec). It could be possible to mass weight to get a COM, see below. Currently I have named it VirtualSiteCentroidType (maybe e.g. type="centroid" in the current spec)

3. I think I was reading this on the openmm github repo, but it could be useful to have both an absolute charge and a chargeincrement simultaneously. The vsite charge could be the sum of the two, and we could perhaps limit the absolute charge parameter to be integer values to maintain sanity. This will add flexibility and generality to their definition. In lieu of this, a chargescale rather than a chargeincrement could be interesting in terms of parameter fitting.

4. I found it easy enough to define a BondCharge vsite as being centered on an atom, and pointing towards the centroid of the remaining vsite atoms. Positive distances would point away from the centroid, relative to the first atom. The spec is currently equivalent when the centroid is one atom.

5. To enable mass weighting, I am assuming a `mass` boolean tag could be set per xml vsite element, which means in the internal representation, each vsite will know whether it should mass-weight (where applicable).

tagging @dmobley, @leeping

davidlmobley:
@trevorgokey that sounds fine to me, I think. What's the argument to having the absolute charge/what use case?

Please tag me in Slack if you need my eyes on a GitHub issue; when I fall behind on notifications the GitHub ones are always the first, so I'll get to your issue faster if you point me to it in Slack.

j-wags:
I'm closing this PR, since #548 is taking up the mantle. Let's make sure to have further discussions over there, so the history remains linear.

Issue #65, opened 2017-10-04
Update to be compatible with NetworkX 2.0 API


jchodera:
cc: https://github.com/open-forcefield-group/openforcefield/issues/63

Issue #63, opened 2017-10-03
tests fail when installing new master on top of old conda distribution


cbayly13:
nothing works and even tests fail after trying to install the new openforcefield master. My installation went like this:
conda create -c conda-forge -c omnia -n openff_2017sep python=3.5 openforcefield
source activate openff_2017sep
pip install jupyter
pip install OpenEye-toolkits
cd openforcefield
pip install -e ./

The traceback I get with tests all end up with the same KeyError on 'return self._nodes[n]', example:
======================================================================
ERROR: Check embedded atom parsing works
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bayly/BaylyData/projects/forcefield/openforcefield/openforcefield/tests/test_chemicalenvironment.py", line 213, in test_embedded_atoms_smirks
    env = ChemicalEnvironment(smirks)
  File "/Users/bayly/BaylyData/projects/forcefield/openforcefield/openforcefield/typing/chemistry/environment.py", line 473, in __init__
    if not self.isValid():
  File "/Users/bayly/BaylyData/projects/forcefield/openforcefield/openforcefield/typing/chemistry/environment.py", line 486, in isValid
    smirks = self._asSMIRKS()
  File "/Users/bayly/BaylyData/projects/forcefield/openforcefield/openforcefield/typing/chemistry/environment.py", line 711, in _asSMIRKS
    initialAtom = self.getAtoms()[0]
  File "/Users/bayly/miniconda/envs/openff_2017sep/lib/python3.5/site-packages/networkx/classes/reportviews.py", line 178, in __getitem__
    return self._nodes[n]
KeyError: 0

jchodera:
Something upstream may have changed. The last travis build was [14 days ago](https://travis-ci.org/open-forcefield-group/openforcefield/builds/277487572). I've enabled nightly builds so we catch issues more quickly.

jchodera:
I can reproduce this issue. Debugging...

jchodera:
I'm betting it's [networkx](https://anaconda.org/conda-forge/networkx/files), which was just updated to 2.0 on `conda-forge` 12 days ago (since our last travis test).

jchodera:
Pinning `networkx` to 1.11 fixes the issue:
```bash
conda install --yes networkx==1.11
```
I'll create a PR to change the `devtools/` recipe, but we should also pin this in the conda recipe.

jchodera:
Let me know if there are other issues, @cbayly13!

PR #71, opened 2017-10-31
[WIP] Support any networkx version

bannanc:
Following issues #63 and #65 we realized there were significant changes in networkx 2.0. I believe I've updated this code to be compatible we networkx versions 1.x and 2.0. I tested both locally on python3.6 with networkx 1.11 (currently the pinned version) and 2.0 and the tests pass.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -18,7 +18,7 @@ requirements:
   run:
     - python
     - numpy
-    - networkx ==1.11
+    - networkx
     - lxml
     - icu 58*  # This is a lxml dependency but sometimes conda installs version 56
     - openmoltools >=0.7.3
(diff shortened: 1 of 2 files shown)


jchodera:
It looks like you only use three methods of NetworkX's `Graph` API:
* `_graph.neighbors(atom)`
* `_graph.nodes()`
* `_graph.get_edge_data(atom1, atom2)`

What about factoring these three calls out into three methods that encapsulate all of the graph operations so that if things in the future change, you only need to change one or two calls instead of 24 changes distributed throughout the file?

bannanc:
@jchodera I can do that. 
forcefield.py also uses networkx though nothing used in that script seems to be affected by the changes here.

jchodera:
I think we'll want to do a few rounds of refactoring on `forcefield.py` both to isolate the `Graph` API and the OEMol/RDKit API with a level of indirection.

bannanc:
I'm refactoring to isolate networkx methods. This last commit has a couple of typos, if I fix those the tests should pass, I'll get it working in the morning.

bannanc:
Tests passed locally with those typos fixed, @jchodera or @davidlmobley could I get a review?

bannanc:
@davidlmobley Travis tests are also passing, they just hadn't finished when I commented last night. Are we go to merge here?

PR #73, opened 2017-11-11
Update outdated documentation in examples

davidlmobley:
@bannanc , can you do a quick check? Just fixes some outdated documentation.

PR #79, opened 2017-11-22
Fix outdated example documentation

davidlmobley:


davidlmobley:
Supersedes #73 which I will close.

PR #83, opened 2017-11-28
Add cyclobutane parameters to smirnoff99frosst

bannanc:
Following [smirnoff99frosst issue#65](https://github.com/openforcefield/smirnoff99Frosst/issues/65) I estimated parameters for angles in 4-membered rings.

### DIFF
--- a/utilities/convert_frosst/smirnoff99Frosst.ffxml
+++ b/utilities/convert_frosst/smirnoff99Frosst.ffxml
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='ASCII'?>
 <SMIRNOFF version="0.1" aromaticity_model="OEAroModel_MDL">
   <!-- SMIRNOFF (SMIRKS Native Open Force Field) template file -->
-  <Date>Date: Aug. 3, 2017</Date>
+  <Date>Date: Nov. 28, 2017</Date>
   <Author>C. I. Bayly, OpenEye/UC Irvine; C. C. Bannan, UC Irvine; D. L. Mobley, UC Irvine</Author>
   <!-- This file is meant for processing via openforcefield.typing.engines.smirnoff -->
   <!-- WARNING: AMBER functional forms drop the factor of 2 in the bond energy term, so cross-comparing this file with a corresponding .frcmod file, it will appear that the values here are twice as large as they should be. -->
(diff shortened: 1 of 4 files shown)


utilities/convert_frosst/smirnoff99Frosst.ffxml
```diff
@@ -141,75 +144,75 @@
     <Improper smirks="[*:1]~[#6X3:2](=[#7X2,#7X3+1:3])~[#7:4]" id="i4" k1="10.5" periodicity1="2" phase1="180."/>
     <Proper smirks="[*:1]~[*:2]~[*:3]~[*:4]" id="t1" idivf1="4" k1="3.50" periodicity1="2" phase1="180.0"/>
     <Proper smirks="[*:1]-[#6X4:2]-[#6X4:3]-[*:4]" id="t2" idivf1="1" k1="0.156" periodicity1="3" phase1="0.0"/>
-    <Proper smirks="[#6X4:1]-[#6X4:2]-[#6X4:3]-[#6X4:4]" id="t3" idivf1="1" k1="0.180" periodicity1="3" phase1="0.0" k2="0.250" idivf2="1" periodicity2="2" phase2="180.0" phase3="180.0" k3="0.200" idivf3="1" periodicity3="1"/>
+    <Proper smirks="[#6X4:1]-[#6X4:2]-[#6X4:3]-[#6X4:4]" id="t3" idivf1="1" k1="0.180" periodicity1="3" phase1="0.0" periodicity2="2" idivf2="1" k2="0.250" phase2="180.0" idivf3="1" k3="0.200" periodicity3="1" phase3="180.0"/>
```
bannanc: @davidlmobley I've seen this come up in PRs before, there isn't actually a change to these torsion parameters, but the order of things in `ffxml` entry has changed. Is there a way to prevent this from happening so that we don't have a record of so many unnecessary changes in every pull request?

davidlmobley: You're welcome to look into the XML library. I would guess it's sort of like a python dictionary -- since the order doesn't matter, very little attention is paid into whether the order is preserved.

bannanc: OK that makes sense. At this point I think we'll continue making hand tuned changes with the smirnoffishFrcmod file so it doesn't matter too much, but might be worth looking at how parameters are written out for future tracking of changes. You're probably right about it using a dictionary for the underlying structure, I'll take a look at the XML library just in case.

bannanc:
I'd like Chris to take a look at my issue first ideally. 

@davidlmobley I added a more thorough check to the convert_frcmod script that checks that the smirks string matches the section instead of just that it is parseable.

bannanc:
Chris also approved these changes so I think we're good to go, @davidlmobley if you agree can you merge it since I added another commit after your review.

davidlmobley:
Excellent. And the additional change/check looks great. Thanks, @bannanc !

davidlmobley:
@bannanc - can you do a new release here and on `openforcefield`?

davidlmobley:
(And note the updated version of this needs to get into the `openforcefield` repo.)

PR #85, opened 2017-12-04
Sphinx documentation infrastructure

jchodera:
This PR sets up basic documentation infrastructure using sphinx.

Here's a preview: http://open-forcefield-toolkit.readthedocs.io/en/latest/


jchodera:
> Just to double check, all the new files are just to support getting this documentation infrastructure?

Yes.

> I had a hard time following all of the changes in files changed, but the preview looks like a good initial setup to me.

The only relevant thing to check out at the moment is the preview: 
http://open-forcefield-toolkit.readthedocs.io/en/latest/

> I think its good to go once the conflicts for README.md and setup.py are resolved.

I'll fix that now.

bannanc:
OK, thanks for the clarification, I think it is good to go when the conflicts are fixed.

jchodera:
Conflicts resolved.

davidlmobley:
Basic layout, etc. looks great. Minor issue regarding what's present there though (http://open-forcefield-toolkit.readthedocs.io/en/latest/installation.html): 

> The simplest way to install the Open Forcefield Toolkit is via the conda package manager.

This implies there is another way to install (which of course there is) but it's never stated. 

Also, it says:
> will automatically dependencies from binary packages automatically

The word "install" is missing.

Having not worked with Sphinx before -- what do we have to do to get content into the documentation? Right now there's nothing there but install instructions. Looks like we have to write `.rst` files, which are markdown format files under `docs`? Is there a way to get basic information populated from, say, function doc strings which already contain a lot of documentation?

PR #105, opened 2018-04-06
Fix an issue with impropers where they could only keep things planar; switches impropers from six-fold to three-fold

vtlim:
For the section in the ammonia test function that sums the angles around nitrogen, I get an oechem warning: `Warning: No principle axes found during inertial alignment` but it doesn't seem to create an issue.

This is referring to the `test_smirnoff.py` file in the `test_improper_pyramidal` function at the line starting with `for atom in outmol.GetAtoms(oechem.OEIsInvertibleNitrogen())`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ script:
   # Build the recipe
   - conda build devtools/conda-recipe
   # Use oeommtools for improper test
-  - conda install -c openeye/label/Orion oeommtools -c omnia
+  - conda install --yes -c openeye/label/Orion oeommtools -c omnia
   # Install
   - conda install --yes --use-local openforcefield
   # Run tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ script:
   #- pip install $OPENEYE_CHANNEL openeye-toolkits && python -c "import openeye; print(openeye.__version__)"
   # Use beta version for partial bond orders
   - pip install --pre -i https://pypi.anaconda.org/openeye/simple openeye-toolkits && python -c "import openeye; print(openeye.__version__)"
+  # Use oeommtools for improper test
+  - conda install -c openeye/label/Orion oeommtools -c omnia
   # Build the recipe
   - conda build devtools/conda-recipe
   # Install
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,10 +23,10 @@ script:
   #- pip install $OPENEYE_CHANNEL openeye-toolkits && python -c "import openeye; print(openeye.__version__)"
   # Use beta version for partial bond orders
   - pip install --pre -i https://pypi.anaconda.org/openeye/simple openeye-toolkits && python -c "import openeye; print(openeye.__version__)"
-  # Use oeommtools for improper test
-  - conda install -c openeye/label/Orion oeommtools -c omnia
   # Build the recipe
   - conda build devtools/conda-recipe
+  # Use oeommtools for improper test
+  - conda install -c openeye/label/Orion oeommtools -c omnia
   # Install
   - conda install --yes --use-local openforcefield
   # Run tests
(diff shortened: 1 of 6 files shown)


davidlmobley:
Normally, a pull request should start off with a description of what issue it fixes. This one doesn't, so I'll briefly summarize here: 

We uncovered a profound issue (bug of sorts) in the improper torsion implementation used here. Specifically, we were applying impropers to all six paths around a trefoil, but not all six have the same handedness. This was not a problem as long as the improper was applied to keep something planar, but if impropers were applied to move a center away from being planar, it failed completely and the center would remain planar.

It is worth noting that AMBER impropers are ONLY used to keep centers planar, so this was not an issue that affected any of our existing force fields, tests, or ports from SMIRNOFF; we uncovered it in a research project exploring alternate impropers.

This pull request fixes the issue by switching to applying three-fold rather than six-fold impropers -- applying all THREE torsions in the trefoil that have the same handedness. It also adds a test to ensure impropers can give puckered geometries when the center is intended to be nonplanar.

jchodera:
@davidlmobley : Does it keep the ordering of the atoms in the impropers such that if the molecule is `i-j(k)-l`, `i-j-k-l`, `k-j-l-i`, `l-j-i-k` where the central atom is the *second* atom in each torsion? Or does it switch to my suggestion such that the torsions are `j-i-k-l`, `j-k-l-i`, `j-l-i-k` and the central atom is the *first* atom in each torsion?

jchodera:
Actually, a diagram would be of immense help!

vtlim:
image

@jchodera, I saw that specifying the central atom as first or second didn't affect the outcome as long as the outer atoms were specified in the same order (e.g., all specified clockwise or counterclockwise). So in the diagram above, the torsions are added as you said, `j-i-k-l`, `j-k-l-i`, `j-l-i-k`.

In the documentation update, I'm not sure if David meant that the central atom was second atom in the SMIRKS string or if it was called in `addTorsion` second, but I assumed the former.

jchodera:
> I saw that specifying the central atom as first or second didn't affect the outcome as long as the outer atoms were specified in the same order (e.g., all specified clockwise or counterclockwise). So in the diagram above, the torsions are added as you said, j-i-k-l, j-k-l-i, j-l-i-k.

Whether the central (`j`) atom is the first or second in the `PeriodicTorsion` makes a HUGE difference on how the value of the equilibrium torsion angle induces non-planar geometries, and even the meaning of the SIGN of the torsion. I think my suggestion---the one you implemented---is the easiest to think about conceptually, but I just want to make sure you know that it is very different from the AMBER definition.

The sensitivity to torsion deviations may also be much different, so our force constants may need to change. I haven't done a detailed comparison with AMBER, but it may be a good idea to simulate some fluctuations in AMBER and compute energies with both SMIRNOFF and AMBER parameter sets using the same spring constant, plotting the correlation to see if we need to remap force constants in a more complex way.

davidlmobley:
Oops, my browser window had not refreshed so I hit merge before seeing there was additional discussion here. 

The central atom is second in SMIRKS pattern, as per the docs.

I'm fine with adding a diagram; perhaps we can do it in a separate PR. It would also be very helpful if OpenMM added a diagram of how the angle is measured; it's really less than ideal to have to dig through the code to figure it out.

jchodera:
> It would also be very helpful if OpenMM added a diagram of how the angle is measured; it's really less than ideal to have to dig through the code to figure it out.

Do you mean for [`PeriodicTorsionForce`](http://docs.openmm.org/latest/api-python/generated/simtk.openmm.openmm.PeriodicTorsionForce.html#simtk.openmm.openmm.PeriodicTorsionForce) or for [the XML tag `<PeriodicTorsionForce>` `<Improper/>` nodes?](http://docs.openmm.org/latest/userguide/application.html#periodictorsionforce)?

davidlmobley:
For `PeriodicTorsionForce`. There's no diagram of how the angle is measured/computed exactly.

Issue #106, opened 2018-04-14
Adding new improper torsions to .offxml, SMIRKS pattern match bug (forcefield.py)


jmaat:
When trying to add the following improper torsion to my .offxml file: 
```<Improper smirks="[*:1]-[#7X3:2](-[#6a$(*1ncncc1):3])-[*:4]" id="i6" k1="1000" periodicity1="2" phase1="180."/>```

The following error occurred:
"does not specify a Improper torsion, but it is supposed to."

I commented out this part of the forcefield.py code and it got rid of the error. This piece of code most likely needs to be debugged:
https://github.com/openforcefield/openforcefield/blob/master/openforcefield/typing/engines/smirnoff/forcefield.py#L1721-L1727

jchodera:
> When trying to add the following improper torsion to my .offxml file:

It looks like the XML tag defining the improper torsion you were trying to add wasn't successfully posted. Could you try posting it again?

davidlmobley:
@jchodera - turns out we've stumbled across some SMIRKS patterns which can't yet be handled by ChemicalEnvironments, which does some checking to ensure that provided SMIRKS match the type of parameter they are being used for. Caitlin will have a look at this on Monday.

@jmaat - you'll have to paste your SMIRKS again (should be able to edit your comment above) as it didn't take properly.

davidlmobley:
Actually, I can fix it! It's there in the text but just not displaying properly, one sec.

davidlmobley:
@jmaat - GitHub uses Markdown style formatting; you can put code in single quotes (if it's a single line) or triple single quotes, if it's multi-line, to get it to be treated as and formatted as plain text so it will display properly, as was needed here.

jchodera:
> Actually, I can fix it! It's there in the text but just not displaying properly, one sec.

Thanks!

The new `Topology` PR relies entirely on `ChemicalEnvironments` to parse whether the SMARTS string is valid (as well as determine what kind of valence term a tagged SMIRKS string specifies), so it will be great to figure out what is required to fix that.

bannanc:
@jchodera
David meant he could fix Jessica's comment. I was out of town/internet
camping this weekend so I haven't had a chance to look at why these aren't
parsing correctly. If it's an easy fix I'll address it tomorrow, but if it
requires significant time then it will be longer due to SAMPL and MolSSI
deadlines that are both approaching quickly.
-- 
*****************************************
Caitlin C. Bannan
Ph.D. Candidate
[...]

Issue #100, opened 2018-02-07
Make ChemicalEnvironments more pythonic


bannanc:
Every time I look back at the `ChemicalEnvironements` code, I want to fix the way it handles the "getters" and "setters" and function naming to be more pythonic. 

As far as I know right now the only thing this is being used for is to check a SMIRKS pattern matches the section it is in using `getType`. Other than that function, @jchodera @davidlmobley is there any opposition to updating the internal structure and function naming here? 

This would be a good time to also make the `isValid` function independent of the SMIRKS parsing tool available so it can work with OEchem or RDKit.

davidlmobley:
That's OK with me.

jchodera:
This all sounds great!

I had to make some changes in `environment.py` in [the `Topology` PR](https://github.com/openforcefield/openforcefield/pull/86/files), so if you do start in on changes, you may first want to pull in the changes just to `environment.py` and then work from there.

davidlmobley:
@bannanc can this be closed?

bannanc:
No, I made some small fixes to parsing back in PR #108, but I haven't changed any of the code. I put this on the back burner thinking it didn't make sense to make large changes to this file while the topology branch wasn't merged. 

Then there was discussion in #145 and then an update in #200 about where `ChemicalEnvironments` should be stored since most of the functionality is only important in a tool like ChemPer that is trying to change SMIRKS patterns. That discussion made me more hesitant to make changes until it had a permanent home.

PR #86, opened 2017-12-04
[WIP] Drafting API for Topology

jchodera:
This is a work-in-progress PR to create the new `Topology` object discussed in #81.

Please take a look at both the proposed API and the various `TODO` items and the questions therein. I'll try to formulate some of the outstanding questions in the morning.

Once #85 is merged, we can also render the proposed API in a more readable/browsable manner.

PR #35, opened 2017-05-03
Updating jupyter notebooks to remove smarty dependencies

bannanc:
As per issue #12, many of the jupyter notebooks were moved from smarty with no changes. Also, I am now using python3 on my work desktop and found a few python2 things that I updated in the process.

PR #108, opened 2018-04-24
Update ChemicalEnvironments SMIRKS parsing

bannanc:
I have updated ChemicalEnvironments to parse more SMIRKS, per issue #106. Please note, I pulled the changes @jchodera made in to ChemicalEnvironments in #86 into this PR. I had pulled that code into my branch after writing issue #100. I haven't addressed issue #100 yet, if we just want to replace things with snake case I should be able to do that fairly easily with find and replace and could add it to this PR. When I wrote that issue I was imagining changing the `Get` and `Set` functions to also be more pythonic with `@property` and `setter` methods instead. I don't think this will take me very long, but its been REALY low on the priority list with SAMPL and MolSSI deadlines approaching. 

So currently this PR addresses only issue #106 which involved more than one bug that I found:

1. aromatic atom symbols weren't recognized when looking for atoms not in brackets. For example, `'c1ccccc1'` is technically a SMIRKS string, but before ChemicalEnvironments would have complained because it didn't know that `'c'` was an atom. Now it does. 

2. You can specify rings in recursive SMIRKS such as `"[#6X3$(*1-C=C-C1):1]"` could be used to look for a butene like group. However, the parsing didn't put the first `1` after the `*` in the right place before. 

3. While addressing 2, I discovered another bug with recursive SMIRKS. The first atom inside the recursion always refers to the first or "top" atom. In my mind, there are 3 options for what that atom could look like. 
    1.  `*` which means there is no additional information `"[#6X3$(*~[#7,#8,#9,#17,#35,#53]):1]"`. These are the most common version of this type of SMIRKS pattern
    2. a "square bracketed atom" such as `"[$(#6X3~[#7,#8,#9,#17,#35,#53])]"`. These aren't that common in human written SMIRKS, but they can appear if you had a replacement string such as `"$ewg"` where you did the replacement before making the `ChemicalEnvironment`
    3. Atomic or aromatic symbols that do not require brackets which look like: `"[$(c1ccccc1)]"`. This last type was not handled at all in the parsing, I've added coverage for this scenario. 

I've tried to add tests to cover all three of these cases. However, I would like to take this bug as an example of reasons why just "can we put it into a `ChemicalEnvironment`" is not a sufficient test. Many of these are caught because of type mismatches or because after pulling the SMIRKS into a `ChemicalEnvironment` we check that SMIRKS created by that environment is actually valid in some other toolkit. I could update the `isValid` function to check for openeye or rdkit and use which ever is available. That function is the only place `ChemicalEnvironments` are dependent on OpenEye. 

@jchodera or @davidlmobley take a look at this description and let me know if you want me to change things to snake case or address the isValid issue before we merge this.

### DIFF
--- a/openforcefield/typing/chemistry/environment.py
+++ b/openforcefield/typing/chemistry/environment.py
@@ -121,13 +121,8 @@ def _convert_embedded_SMIRKS(smirks):
             embedded = embedded[2:]
 
         else: # embedded starts with a "no bracket" atom such as 'C'
-            # covers element symbols, i.e. N,C,O,Br not followed by a number
-            element_sym = "!?[A-Z][a-z]?"
-            # covers element symbols that are aromatic:
-            aro_sym = "!?[cnops]"
-            # replacement strings
-            replace_str = "\$\w+"
-            no_bracket = r'('+'|'.join([element_sym, aro_sym, replace_str])+')'
+            # atoms by symbol don't need brackets, this covers atomic symbols and aromatic atoms
+            no_bracket = r'(!?[A-Z][a-z]?|!?[cnops])'
             embedded = embedded[1:] # remove leading '('
             match = re.match(no_bracket, embedded)
             if match is not None:
@@ -554,6 +549,7 @@ into ChemicalEnvironments." % smirks)
             smirks = self._asSMIRKS()
         if self.replacements is not None:
             smirks = oechem.OESmartsLexReplace(smirks, self.replacements)
+            print("replaced SMIRKS", smirks)
         return oechem.OEParseSmarts(qmol, smirks)
 
     def _parse_smirks(self,input_smirks):
--- a/openforcefield/typing/chemistry/environment.py
+++ b/openforcefield/typing/chemistry/environment.py
@@ -436,7 +436,6 @@ class ChemicalEnvironment(object):
         if toolkit.lower() == 'openeye':
             try:
                 from openeye import oechem
-                print('I found openeye')
                 self.toolkit = 'openeye'
             except:
                 raise Exception("Could not import openeye.oechem")
@@ -491,7 +490,7 @@ class ChemicalEnvironment(object):
             # Check that it is a valid SMIRKS
             if not self.isValid(smirks):
                 raise SMIRKSParsingError("Error Provided SMIRKS ('%s') was \
-not parseable with OpenEye tools" % smirks)
+not parseable with %s tools" % (smirks, self.toolkit))
 
             # Check for SMIRKS not supported by Chemical Environments
             if smirks.find('.') != -1:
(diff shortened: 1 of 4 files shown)


davidlmobley:
@bannanc 
> I could update the isValid function to check for openeye or rdkit and use which ever is available

Thanks for your work on this. I think this would be useful if you can get the RDKit version of it working easily, though if not it would make sense to me to do that in a separate issue/PR later. 

I'll have a quick look through your other changes, but I'm fine having the rest of the issues be dealt with separately/later.

openforcefield/typing/chemistry/environment.py
```diff
@@ -18,6 +18,9 @@
 and David Mobley, UC Irvine.
 
 """
+
+# QUESTION: should we ditch camelCase and go with PEP8 standard snake_case?
```
davidlmobley: SGTM to do eventually; ambivalent about whether it is now.

bannanc: This is a carry over from @jchodera since I took the chemenv from the Topology branch. I think the answer to this is yes eventually, but not happening right now.

bannanc:
@davidlmobley 
Otherwise I'm happy to wait on the other pythonic issues for future PRs and merge this. I'm not allowed to merge this because the tests are failing so will you merge it unless. 

Did you want me to change this question in the python file?

> # QUESTION: should we ditch camelCase and go with PEP8 standard snake_case?

davidlmobley:
Happy to merge, but what about:
> Thanks for your work on this. I think this would be useful if you can get the RDKit version of it working easily, though if not it would make sense to me to do that in a separate issue/PR later.

Also can you remove "WIP"?

davidlmobley:
Mind also disabling the py27 test at the same time, since we're going to keep having problems with that one given that py27 support is going away?

bannanc:
@davidlmobley I tried removing python 2.7 and then the other tests failed... I don't understand why, but that was [this commit](https://travis-ci.org/openforcefield/openforcefield/builds/372318575?utm_source=github_status&utm_medium=notification)

I then switched to the conda install for openeye because I thought it made sense for us to switch to that installation. (it is still failing)

bannanc:
I found a discrepancy in openeye and rdkit SMARTS parsing, I updated the test to avoid the corner case and wrote an issue in RDKit's issue tracker: https://github.com/rdkit/rdkit/issues/1842

bannanc:
@davidlmobley This is ready for you to take a look at again, I'm happy to handle the rdkit part differently, but I wanted to make sure I could test it so making it an init option seemed like the easiest way to do that for now.

davidlmobley:
Looks good to me! Thanks! Update docs in a separate PR to note that py27 may still work but is no longer supported?

PR #119, opened 2018-10-12
[WIP] Molecule, Atom, Bond, Toolkit Update

j-wags:


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jchodera: We should add `.idea` to the `.gitignore` file so that your editor doesn't check in files that don't belong in the repo.

openforcefield/tests/test_molecule.py
```diff
@@ -56,9 +58,11 @@ def assert_molecule_is_equal(molecule1, molecule2, msg):
 
 def test_cheminformatics_toolkit_is_installed():
     """Ensure that at least one supported cheminformatics toolkit is installed."""
-    if RDKIT_UNAVAILABLE and OPENEYE_UNAVAILABLE:
+    # Can't just call all_subclasses here -- AmberToolsToolkitWrapper won't suffice for this test
+    if not(RDKitToolkitWrapper.toolkit_is_available) and not(OpenEyeToolkitWrapper.toolkit_is_available):
+
         msg = 'No supported cheminformatics toolkits are installed. Please install a supported toolkit:\n'
-        msg += str(SUPPORTED_TOOLKITS)
+        msg += 'OpeneEyeToolkitWrapper or RDKitToolkitWrapper'
```
jchodera: We really need the toolkits themselves to be installed, and want to communicate to users how to install them. Perhaps we should instead print `toolkit_wrapper.installation_instructions` instead, which provides info on how to install the toolkits themselves (rather than the wrapper)?

openforcefield/topology/molecule.py
```diff
@@ -35,16 +35,15 @@
...
-from distutils.spawn import find_executable
+from collections import OrderedDict
 
 from simtk import unit
 from simtk.openmm.app import element
 
 import openforcefield
-from openforcefield.utils.toolkits import OPENEYE_INSTALLED, RDKIT_INSTALLED, TOOLKIT_PRECEDENCE, SUPPORTED_FILE_FORMATS
-from openforcefield.utils.toolkits import requires_rdkit, requires_openeye
+from openforcefield.utils.toolkits import OPENEYE_AVAILABLE, RDKIT_AVAILABLE, GLOBAL_TOOLKIT_REGISTRY
```
jchodera: Hopefully, we can go through the toolkit wrappers for all our cheminformatics needs in `molecule.py`, so we don't actually need to know which toolkit is servicing our needs (and don't need this here).

openforcefield/topology/molecule.py
```diff
@@ -537,20 +1267,48 @@ def to_dict(self):
...
-        molecule_dict['bonds'] = [ bond.to_dict() for bond in self.bonds ] # TODO: How do we make sure bonds are serializable?
+        molecule_dict['name'] = self._name
+        ## From Jeff: If we go the properties-as-dict route, then _properties should, at
+        ## the top level, be a dict. Should we go through recursively and ensure all values are dicts too?
+        molecule_dict['atoms'] = [ atom.to_dict() for atom in self._atoms ]
+        molecule_dict['virtual_sites'] = [ vsite.to_dict() for vsite in self._virtual_sites ]
+        molecule_dict['bonds'] = [ bond.to_dict() for bond in self._bonds ]
         # TODO: Charges
         # TODO: Properties
+        # From Jeff: We could have the onerous requirement that all "properties" have to_dict() functions.
```
jchodera: Another possibility would be to have something like a `Conformation` or `Positions` object that knows how to serialize/deserialize itself with `to_dict()` and `from_dict()`. That way, we can reuse the same object for both `Molecule` and `Topology`.

openforcefield/topology/molecule.py
j-wags: I agree with this. The one tricky part is [1] that our Molecule constructor wants to type-check for OEMol and RDMol-type inputs _here in the Molecule class_. My thought is that we could have the toolkits each provide a "from_object" function that will do the type checking for an input and return False it if it's not compatible, or the results of from_rdkit or from_openeye (or whatever other toolkits we add later) if it is compatible. Then the Molecule constructor could iterate through our toolkit_registry and to run from_object using each one, until we either get a Molecule out or they've all returned False.

[1] https://github.com/openforcefield/openforcefield/pull/119/files/11b71c413a5f1487e364085fdaa28f6675d3fd6f#diff-dcc45fa83688f86668e384f78c5ab22aR1223

.idea/vcs.xml
j-wags: Thanks for catching this. Resolved in https://github.com/openforcefield/openforcefield/pull/119/commits/dea744de97d776e6b6ed02618efe023d6ab3decb

openforcefield/topology/molecule.py
jchodera: This sounds like a good solution!

jchodera:
It looks like `openforcefield/utils/serialization.py` is conflicting, so you'll need to do a `git merge origin/master` after `git fetch` at some point to resolve that!

openforcefield/topology/molecule.py
j-wags: This is a really good idea -- Tracking as #138

PR #164, opened 2019-01-19
Reimplement comparison between amber and SMIRNOFF parameters

andrrizzi:
In this PR:
- Fixed unit specification of `cutoff` and `switch` in `Frosst_AlkEthOH_parmAtFrosst.offxml` according to the new specification.
- Reimplemented and renamed the function `openforcefield/tests/utils.py:compare_molecule_energies` to `compare_amber_smirnoff`.
- Fixed the syntax errors in the example `SMIRNOFF_comparison/compare_molecule_energies.py` and `SMIRNOFF_comparison/compare_set_energies.py`.

The old function `compare_molecule_energies()` used the [`SystemChecker`](https://github.com/choderalab/openmoltools/blob/master/openmoltools/system_checker.py) class in `openmoltools`, and we removed that dependency in the `topology` branch. Here, I've added a few functions to perform the same task. The implementation is hopefully a little more concise with less redundant code (there's a lot of new lines but about half of them are docstrings).

Currently `compare_molecule_energies.py` is successful while `compare_set_energies.py` still fails with the error
```
  File "compare_set_energies.py", line 39, in <module>
    results = compare_amber_smirnoff(prmtop_filepath, inpcrd_filepath, forcefield, molecule)
  File "/Users/andrea/miniconda/envs/openforcefield/lib/python3.6/site-packages/openforcefield-0.1+260.ge074487.dirty-py3.6.egg/openforcefield/tests/utils.py", line 918, in compare_amber_smirnoff
  File "/Users/andrea/miniconda/envs/openforcefield/lib/python3.6/site-packages/openforcefield-0.1+260.ge074487.dirty-py3.6.egg/openforcefield/topology/topology.py", line 1390, in from_openmm
Exception: No match found for molecule
```
I plan to look into it in a separate PR.

### DIFF
--- a/examples/SMIRNOFF_comparison/compare_molecule_energies.py
+++ b/examples/SMIRNOFF_comparison/compare_molecule_energies.py
@@ -28,3 +28,5 @@ forcefield = ForceField('forcefield/Frosst_AlkEthOH_parmAtFrosst.offxml')
 # Compare energies
 from openforcefield.tests.utils import compare_amber_smirnoff
 results = compare_amber_smirnoff(prmtop_filepath, inpcrd_filepath, forcefield, molecule)
+
+print(results)
--- a/examples/SMIRNOFF_comparison/compare_molecule_energies.py
+++ b/examples/SMIRNOFF_comparison/compare_molecule_energies.py
@@ -7,27 +7,24 @@ import os
 datapath = './AlkEthOH_inputfiles/AlkEthOH_rings_filt1'
 #molname = 'AlkEthOH_r0' #That fails, but it's complicated. Try cyclobutane
 molname = 'AlkEthOH_r51'
-mol_filename = os.path.join( datapath, molname+'.mol2')
-prmtop_filename = os.path.join( datapath, molname+'.top')
-inpcrd_filename = os.path.join( datapath, molname+'.crd')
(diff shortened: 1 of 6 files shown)


andrrizzi:
> `compare_set_energies.py` still fails

The problem is caused by an incorrect interpretation of atom types of the OpenEyeToolkitWrapper. I've created issue #165 with the details so this is ready for review.

j-wags:
Ran `antechamber` to make a SYBYL-typed mol2 for each GAFF-typed mol2 in the AlkEthOH test set for `examples/SMIRNOFF_comparison` (replaced the tarball with one containing the new files)

andrrizzi:
Thank you!

> (replaced the tarball with one containing the new files)

I've noticed that there's a lot of redundant files in `molecules` so I was thinking about cleaning that up in a separate PR. Will continue the discussion in #165 .

davidlmobley:
Thanks, guys! Great to have this shaping up. :) 

The `no match for molecule` error would suggest to me that there's a reading or perception error for one or more of the molecules. I used to get this if, for example, I read a file with GAFF atom types when I wasn't intending to, or similar.

andrrizzi:
> I read a file with GAFF atom types when I wasn't intending to

That was exactly the problem :) . Hopefully, that will be better after #168 .

Issue #117, opened 2018-10-05
Add entry point handling for plugin openforcefield group offxml search path


jchodera:
Example searching for registered files: https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py#L50-L64
Example registering files on installation: https://github.com/choderalab/openmm-forcefields/blob/master/setup.py#L43-L51

PR #170, opened 2019-01-24
Fix #163: Make ForceField look automatically in data/forcefield

### LINKED ISSUE #163: Make ForceField look automatically in data/forcefield directory instead of data/
Currently when we need to prepend `forcefield/` to all offxml files defined in the repo
```
forcefield = ForceField('forcefield/Frosst_AlkEthOH_parmAtFrosst.offxml')
```
Should we make `ForceField` look automatically in `data/forcefield/` in [this snippet](https://github.com/openforcefield/openforcefield/blob/77e66993daacdfba61d89888cca82223452b621b/openforcefield/typing/engines/smirnoff/forcefield.py#L556) instead of just `data/` and make the call a little simpler?
```
forcefield = ForceField('Frosst_AlkEthOH_parmAtFrosst.offxml')
```
Or is there a reason I don't see to avoid this?

andrrizzi:
This implements #163 .

I think I got all the `ForceField` constructors. I found a few inconsistencies every now and then.


davidlmobley:
That would be a good idea.

andrrizzi:
I can work on this if it's considered high-priority.

jchodera:
We can implement this at the same time as we update the https://github.com/openforcefield/smirnoff99frosst repo to install into the entry point.

andrrizzi:
Sounds good. I think you already captured this in #117.

Issue #138, opened 2018-11-09
Implement Conformer/Positions and PartialCharges classes that are inherently serializable and cover their own unit checking


j-wags:

j-wags:
Pydantic example of mixing default initialization with custom functionality are here

dgasmith:
Another way to do it is with an internal `data` field which is a PyDantic object similar to [this](https://github.com/MolSSI/QCFractal/blob/master/qcfractal/interface/collections/collection.py#L45). Recursive serialization is a bit harder, but do able.

j-wags:
Relevant discussion: https://github.com/openforcefield/openforcefield/pull/173#discussion_r266039941

jchodera:
@SimonBoothroyd found pydantic to be too infexible for the property estimator---it might be useful to collect those experiences here too.

j-wags:
Referencing https://github.com/openforcefield/propertyestimator/pull/18 , where Simon removed it from the propertyestimator repo

PR #173, opened 2019-02-01
Fix and test OFFXML I/O

### LINKED ISSUE #172: ImproperHandler failing on triazole
From Ian Craig:

[2019_01_25_triazole_improper_data.tar.gz](https://github.com/openforcefield/openforcefield/files/2796858/2019_01_25_triazole_improper_data.tar.gz)

> Hi. I'm running through the notebook provided during the hands-on session at the San Diego meeting earlier this month. I'm having problems parameterizing my own molecules though. In particular, the forcefield.create_openmm_system call complains that certain pairs of atoms aren't bonded according to the topology.
> I'll post the error below, but it seems that the PDB I'm providing isn't consistent with the SMILES somehow. Does the PDB have to be created in a particular way, or with a particular atom ordering, etc ..? The attached notebook and PDB should illustrate.

j-wags:
## Tasks
This PR:

- [x] Refactors `ParameterHandler`, `ParameterList`, and `ParameterType`-derived classes according to [this layout](https://docs.google.com/presentation/d/1U1ZnB-kaAA1s-lFjTQSI1hHx5kntbOw-H4HlGdV4KCo/edit#slide=id.g4ff30d61b6_0_232). This refactor isolates the data spread among the classes, such that neither `ParameterHandler` nor `ParameterIOHandler` requires a pointer to its parent `ForceField`. This refactor will enable the substitution of a Pydantic/Marshmallow/other-robust-data-model backend in the future
- [x] Implements to_{dict/list} functions for `ParameterHandler`, `ParameterList`, and `ParameterType`, allowing recursive serialization.
- [x] Establishes the interface between `ParameterHandler` and `ParameterIOHandler` to be the transfer of a hierarchical `OrderedDict` compliant with the SMIRNOFF spec.
- [x] Replaces the `lxml`-based XMLParameterIOHandler with an `xmltodict` based one.
- [x] Implements support for cosmetic tags input data
- [x] Adds tests for new features
- [x] Condenses unit/quantity serialization code into a few functions in `openforcefield/utils` so that future Pint incorporation will be smoother
- [x] Closes #172 by virtue of passing freesolv parameter comparison tests (the 4th molecule in the freesolv tests has an improper)

Besides code review, I would be interested in input on:
* The use of class-level dicts and lists in ParameterType and ParameterHandler to define class structure and behavior. Specifically:
    * Is the current implementation general enough? Will it offer enough flexibility for other developers to implement their ideas?
    * Would it be better to use different names for the class-level dicts and lists?
* Semantics for SMIRNOFF data fields (for example, a bond's `k` value). Currently I use "attrib" throughout the code, but there may be a more descriptive name like "attribute", "property", etc.
* Is there a better word than "cosmetic" for optional/extra data fields?

This PR is already too large, so I'm submitting it for review now. Subsequent PRs can address remaining issues, including
- [ ] AST-based unit {de,}serialization can't handle lists of quantities, thus breaking charge_increment serialization for virtual sites
- [ ] Reimplementing `missing_valence` checks in `ParameterHandler.create_force()` calls.
- [ ] Automatically adding `@property.setters` during ParameterHandler and ParameterType initialization, which will check for unit compatibility when setting unit-bearing attributes.
- [ ] Updating SMIRNOFF spec per #191 
- [ ] Implementing Pint-based unit backend 

## Status
- [x] Ready for review


openforcefield/tests/test_forcefield.py
```diff
@@ -251,8 +260,246 @@ def test_parameterize_ethanol_to_parmed(self):
 
         parmed_system = forcefield.create_parmed_structure(topology, positions=pdbfile.getPositions())
 
-# from_filename
-# from xml_string
+    def test_charges_from_molecule(self):
+        """Test skipping charge generation and instead getting charges from the original Molecule"""
+        # Create an ethanol molecule without using a toolkit
+        mol = Molecule()
+        mol.add_atom(6, 0, False) # C0
```
jchodera: Instead of these very long (20-line!) blocks to create a simple molecule like cyclohexane with partial charges just to test `forcefield.create_openmm_system(topology, charge_from_molecules=molecules), what about using the toolkit to do
```python
mol = OpenEyeToolkitWrapper.from_file('ethanol_with_partial_charges.mol2')
```
?

openforcefield/topology/molecule.py
```diff
@@ -1913,24 +1916,68 @@ def generate_conformers(self,
...
+
+        """
+        if isinstance(toolkit_registry, ToolkitRegistry):
+            charges = toolkit_registry.call(
+                      'compute_partial_charges_am1bcc',
+                      self
+            )
+        elif isinstance(toolkit_registry, ToolkitWrapper):
+            toolkit = toolkit_registry
+            charges = toolkit.compute_partial_charges_am1bcc(
```
jchodera: Isn't the spacing a little silly and unnecessarily verbose here? This could just be 
```python
        elif isinstance(toolkit_registry, ToolkitWrapper):
            charges = toolkit_registry.compute_partial_charges_am1bcc(self)
```

openforcefield/tests/test_forcefield.py
j-wags: Good question. To make these more of "unit" tests, I wanted to remove toolkit dependent behavior outside of the precise functionality being tested (in this case, parameter assignment and partial charge pass-through). Since reading any supported file (`mol2` or `sdf`) will go through a toolkit, that introduces another source of potential failure for this test. 

Once the 1.0.0 release is made and we have a stable serialized molecule format, we could use that as a source of toolkit-independent test molecules. 

To remove text bloat, I could also define these in the test setup at the top of the file, or in a different file altogether. What do you think?

openforcefield/topology/molecule.py
j-wags: Yeah, `yapf` made some interesting decisions about where to break lines. I'm treating these on a one-by-one basis as I find them. I'll go ahead and clean this one up in the branch.

openforcefield/tests/test_forcefield.py
jchodera: > To remove text bloat, I could also define these in the test setup at the top of the file, or in a different file altogether. What do you think?

That's a great idea. You can have a `create_test_molecule()` method that localizes all of this information into one place, and then streamline this after release.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -130,45 +247,166 @@ def __contains__(self, item):
...
+            if key in self._REQUIRE_UNITS:
+                if not val.unit.is_compatible(self._REQUIRE_UNITS[key]):
+                    msg = "{} constructor received kwarg {} with value {}, " \
+                          "which is incompatible with expected unit {}".format(self.__class__,
+                                                                               key,
+                                                                               val,
+                                                                               self._REQUIRE_UNITS[key])
+                    raise SMIRNOFFSpecError(msg)
+
+                if key in self._ATTRIBS_TO_TYPE:
```
SimonBoothroyd: Should this be indented into the `if key in self._REQUIRE_UNITS` block? In the statement just above `if key in self._ATTRIBS_TO_TYPE` is its own independent if statement.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -130,45 +247,166 @@ def __contains__(self, item):
...
+                              "which is incompatible with expected unit {}".format(self.__class__,
+                                                                                   idx_key,
+                                                                                   val,
+                                                                                   self._REQUIRE_UNITS[unidx_key])
+                        raise SMIRNOFFSpecError(msg)
+                if unidx_key in self._ATTRIBS_TO_TYPE:
+                    type_to_cast = self._ATTRIBS_TO_TYPE[unidx_key]
+                    val = type_to_cast(val)
+                getattr(self, unidx_key).append(val)
+                del kwargs[idx_key]
```
SimonBoothroyd: I'm not sure if this matters but you may want to consider the case where someone accidentally misses an index (e.g param1, param3, param4) in the kwargs dict - I think currently the code would just stop when it hit the gap?

openforcefield/typing/engines/smirnoff/parameters.py
SimonBoothroyd: You might wanter to consider refactoring this check into a _validate_kwarg method to remove the duplicate code?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -181,6 +419,69 @@ def set_smirks(self, smirks):
     # For example, if we have a parameter with units energy/distance**2, can we check to make
     # sure the dimensionality is preserved when the parameter is modified?
 
+    def to_dict(self, return_cosmetic_attributes=False):
+        """
+        Convert this ParameterType-derived object to dict. A unit-bearing attribute ('X') will be converted to two dict
+        entries, one (['X'] containing the unitless value, and another (['X_unit']) containing a string representation
```
SimonBoothroyd: I think I saw elsewhere in this PR that we can now parse quantities directly from strings using ast? Could this be swapped to use this functionality so only one entry is present?

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -575,80 +631,175 @@ def parse(self, sources, input_format=None):
...
-
-    # TODO : Move to ParameterIOHandler
-    def _parse_aromaticity_model(self, root):
-        """Parse the aromaticity model, make sure it is supported, and make sure it does not contradict previously-specified aromaticity models.
+            smirnoff_data = self.parse_smirnoff_from_source(source)
+            self.load_smirnoff_data(smirnoff_data)
+
+
+
+    def to_smirnoff_data(self):
```
SimonBoothroyd: As a very minor consistency thing, I think in some places you refer to the output of this as smirnoff_data, and others as smirnoff_dict. Perhaps naming this method `to_dict` would be more consistent with the rest of your new code?

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -575,80 +631,175 @@ def parse(self, sources, input_format=None):
...
+        # Define keys which are expected from the spec, but have no bearing on system energy
+        l1_spec_keys = ['Author','Date', 'version', 'aromaticity_model']
+
+        for parameter_name in l1_dict.keys():
+            # Skip (for now) cosmetic l1 items
+            if parameter_name in l1_spec_keys:
+                continue
+            # Handle cases where a parameter name has no info (eg. ToolkitAM1BCC)
+            if l1_dict[parameter_name] is None:
+                handler = self.get_handler(parameter_name, {})
```
andrrizzi: Just to make sure I understand, is this call here to make some sort of validation and make sure that `parameter_name` corresponds to a registered handler?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -130,45 +236,147 @@ def __contains__(self, item):
...
     Base class for SMIRNOFF parameter types.
 
     """
-
-    _VALENCE_TYPE = None  # ChemicalEnvironment valence type for checking SMIRKS is conformant
-
-    # TODO: Allow preferred units for each parameter type to be specified and remembered as well for when we are writing out
+    # This list will be used for validating input and detecting cosmetic attributes.
+    _VALENCE_TYPE = None  # ChemicalEnvironment valence type string expected by SMARTS string for this Handler
+    _SMIRNOFF_ATTRIBS = ['smirks'] # Attributes expected per the SMIRNOFF spec.
```
andrrizzi: This is definitely something that we can do in a separate PR (if we do it at all), but I found tricky to handle about the kwargs defined in class constants in my experience is inheritance. To make a trivial example, you need to re-add `smirks` to the list of `_SMIRNOFF_ATTRIBS` when you inherit from `ParameterType`, but when 2 levels of inheritance and attributes less general than `smirks` are involved, it's very easy to make a mistake that ends in a silent error.

A possible way to take advantage of the Python inheritance implementation to take care of everything could be to use properties instead. In this case, a parameter type could be defined with something like
```python
class BondType(ParameterType):
    length = SmirnoffAttribute(required_unit=unit.angstrom)
    k = SmirnoffAttribute(required_unit=unit.kilocalorie_per_mole / unit.angstrom**2, indexed=True)
    made_up_attr = OptionalAttribute()
```
With this implementation, `ParameterType` can discover required and optional attributes with something like
```python
class ParameterType:
    @classmethod
    def _get_smirnoff_attributes(cls):
        return {name: descriptor for name, descriptor in inspect.getmembers(cls)
                if isinstance(descriptor, SmirnoffAttribute)}
```
I think a design that goes in this direction could make both `ParameterType` and `ParameterHandler` easier to extend.

openforcefield/typing/engines/smirnoff/parameters.py
andrrizzi: If it helps, you could use a regex to find all the `param_name+number` strings as suggested here: https://stackoverflow.com/a/51174978 .

openforcefield/typing/engines/smirnoff/parameters.py
andrrizzi: I think @j-wags adopted this solution because (Jeff correct me if I'm wrong) it's more similar to how the force field is serialized in XML format.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -197,38 +443,115 @@ class ParameterHandler(object):
...
         """
+        Initialize a ParameterHandler, optionally with a list of parameters and other kwargs.
 
         Parameters
         ----------
-        forcefield : openforcefield.typing.engines.smirnoff.ForceField
+        permit_cosmetic_attributes : bool
+            Whether to accept non-spec kwargs
+        **kwargs : dict
+            The dict representation of the SMIRNOFF data source
```
andrrizzi: So I've been having second thoughts about this (apologies for my original suggestion!), and I'm now wondering if we should have this signature in a `from_dict(permit_cosmetic_attributes=False. **kwargs)` static constructor, and keep a classic separate construct with a signature like `__init__(parameter_types=None, permit_cosmetic_attributes=False, ..., **cosmetic_attributes)` for consistency with the rest of the API.

j-wags: I like that. So, the `from_dict` function would attach units and perform other default behaviors for preparing OFFXML data, and then pass the formatted results to `__init__`

j-wags: In the interest of conciseness, I'm going to treat this as an issue separate from this PR.

openforcefield/topology/topology.py
```diff
@@ -1754,6 +1763,41 @@ def to_openeye(self,
...
+            for top_atom in top_bond.atoms:
+                if top_atom == atomi:
+                    continue
+                if top_atom == atomj:
+                    return top_bond
+        # If atomj wasn't found in any of atomi's bonds, then they aren't bonded.
+        raise NotBondedError('No bond between atom {} and {}'.format(atomi, atomj))
+
+
     def is_bonded(self, i, j):
```
jchodera: Could this simply call `get_bond_between()`?

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -229,49 +258,73 @@ def __init__(self,
...
-
-    # TODO : This is unused and shouldn't be here -- Move to parameteriohandler
-    def _parse_version(self, root):
+        self._parameter_handler_classes = OrderedDict()  # Parameter handler classes that _can_ be initialized if needed
+        self._parameter_handlers = OrderedDict()  # ParameterHandler classes to be instantiated for each parameter type
+        self._parameter_io_handler_classes = OrderedDict()  # ParameterIOHandler classes that _can_ be initialiazed if needed
+        self._parameter_io_handlers = OrderedDict()  # ParameterIO classes to be used for each file type
+        self._parameters = ParameterList()  # ParameterHandler objects instantiated for each parameter type
+        self._aromaticity_model = None
+        # TODO: will it be possible for a ForceField to have multiple (compatible) aromaticity models?
```
jchodera: No, there should be a specific aromaticity model.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -229,49 +258,73 @@ def __init__(self,
...
+        SMIRNOFFAromaticityError if an incompatible aromaticity model is passed in.
+
+        """
+        # Implement better logic here if we ever support another aromaticity model
+        if aromaticity_model != 'OEAroModel_MDL':
+            raise SMIRNOFFAromaticityError("Read aromaticity model {}. Currently only "
+                                           "OEAroModel_MDL is supported.".format(aromaticity_model))
+
+        self._aromaticity_model = aromaticity_model
+        #self._used_aromaticity_models.add(aromaticity_model)
```
jchodera: Can this be removed?

openforcefield/topology/molecule.py
j-wags: Fixed in PR

openforcefield/tests/test_forcefield.py
j-wags: Separated code into `create_ethanol()` and `create_cyclohexane()` functions, and defined them at the top of `test_forcefield`

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -103,20 +117,123 @@ class ParameterList(list):
...
+        if type(item) is str:
+            # Try to find by SMIRKS
+            for parameter in self:
+                if parameter.smirks == item:
+                    self.remove(parameter)
+                    return
+
+        # Try numerical index access. This will grab the item to remove by index, and
+        # then call __delitem__ again on its own SMIRKS, finishing in the "if" statement above
+        item_to_remove = self[item].smirks
```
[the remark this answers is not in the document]
j-wags: I'm not positive, but I don't think I need to do this. If I run `del param_list[-1]`, the ParameterType will be removed from the list, BUT there will still be a pointer (or reference, or whatever python calls it) to it in the `_last_added_param` variable, so garbage collection should leave it alone... I'll go ahead and make a test for this to be sure.

openforcefield/tests/test_parameters.py
```diff
@@ -81,5 +96,324 @@ def test_del(self):
 
         # Test that we can delete elements by their smirks.
         del parameters['[#1:1]']
+        assert len(parameters) == 1
         assert p1 in parameters
         assert p2 not in parameters
+
+
+    def test_parameterlist_append(self):
```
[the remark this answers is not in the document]
j-wags: Fixed in 54e5f03140c615e4579ea42f0860caf26cc27195

openforcefield/typing/engines/smirnoff/parameters.py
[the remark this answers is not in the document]
j-wags: The `_last_added_param` does indeed stick around, even if the actual parameter is removed from the list. Test for this behavior added in 41000a9207cdc4430949459542138b650357f835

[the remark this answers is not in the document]
SimonBoothroyd: I think the code will definitely still work - I just thought it seems a little odd that the collection will still keep a reference to (and use the value from) an item which was removed, which may (in some odd edge cases) lead to an unexpected user experience (but is most cases probably not!). Just a minor thing anyway!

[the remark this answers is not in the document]
j-wags: True. When I initially implemented this behavior, I had thought about having it get units from the penultimately-added parameter if the last-added parameter was deleted. But then, in a worst-case scenario, we'd need a data structure keeping track of the order in which all the parameters were added, in case a user kept deleting the last-added one. 

The whole units-from-the-last-added-parameter thing is not obviously correct, but I agree with John that it is the method that minimized "surprises" for the user.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -103,20 +117,123 @@ class ParameterList(list):
...
+        super().insert(index, parameter)
+        self._last_added_param = parameter
+
+    def __delitem__(self, item):
+        """
+        Delete item by index or SMIRKS
+        """
+        if type(item) is str:
+            # Try to find by SMIRKS
+            for parameter in self:
```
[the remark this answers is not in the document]
j-wags: Agreed. I've added a `ParameterList.index()` function and use that in place of these redundant lookups. Implemented function and tests in a4bd7ba

openforcefield/typing/engines/smirnoff/io.py
```diff
@@ -283,20 +245,35 @@ def parse_file(self, source):
            * If a SMIRNOFF section that has already been read appears again, its definitions are appended to the end of the previously-read
              definitions if the sections are configured with compatible attributes; otherwise, an ``IncompatibleTagException`` is raised.
         """
+        from pyexpat import ExpatError
+        from openforcefield.typing.engines.smirnoff.forcefield import ParseError
+
+        # this handles open file-like objects (and strings)
+        try:
+            smirnoff_dict = xmltodict.parse(source, attr_prefix='')
+            print(smirnoff_dict)
```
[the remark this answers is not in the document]
j-wags: Good call. Fixed in 6f4d932

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: > Should this be indented into the if key in self._REQUIRE_UNITS block?
No, this is unfortunately spaghetti code at the moment, but it's readable. 

I like your validate idea -- Implemented it for units in d062fdc

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: > I think currently the code would just stop when it hit the gap?

So, what will happen right now if there's a gap (eg. `k1`, `k2`, and `k4` are defined) is that `k1` and `k2` will get set into a list `self.k`, with length 2. `k1` and `k2` will, in the process, be removed from `kwargs`. However, `k4` won't be removed from `kwargs`, and will get sent into the non-indexed kwarg handling code. That code won't recognize it, and will throw an exception. 

I've implemented a test for this scenario here: https://github.com/openforcefield/openforcefield/blob/ff-offxml-io/openforcefield/tests/test_parameters.py#L454-L470

> If it helps, you could use a regex to find all the param_name+number strings

Maybe it's just me, but I've had a lot of trouble with regexes in code. They're not all that readable, and can make debugging really complicated. I'm not totally against using them, but I would prefer to save them for cases where they're absolutely needed.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: @andrrizzi is correct here. Per the SMIRNOFF spec, we store units and values separately. Storing each quantity with its units will create a lot of text bloat -- for example, each bond force constant would expand from from `k=20`, with a single `k_unit=kilocalorie_per_mole / angstrom**2` in the section header,  to `k=20 kilocalorie_per_mol / angstrom **2`. Given that we'll have hundreds of these (and probably even more torsion types), this could end up being a significant amount of text.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -57,6 +56,36 @@ def __init__(self, msg):
...
+        self.msg = msg
+
+class SMIRNOFFAromaticityError(Exception):
+    """
+    Exception thrown when an incompatible SMIRNOFF aromaticity model is checked for compatibility.
+    """
+
+    def __init__(self, msg):
+        super().__init__(self, msg)
+        self.msg = msg
```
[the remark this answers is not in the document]
j-wags: Good idea. Implemented in 13dc270

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -247,31 +277,58 @@ def _initialize(self):
         )  # ParameterIO classes to be used for each file type
         self._parameters = ParameterList(
         )  # ParameterHandler objects instantiated for each parameter type
+        self._aromaticity_model = None
+        #self._used_aromaticity_models = set(
```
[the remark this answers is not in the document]
j-wags: Thanks. This was a placeholder to remind me that the aromaticity implementation is still in flux.  I've cleaned this up and recorded my thoughts on aromaticity here: #206

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -247,31 +277,58 @@ def _initialize(self):
...
+        # Use PEP-440 compliant version number comparison, if requested
+        if (not self.disable_version_check) and (
+                packaging.version.parse(str(version)) >
+                packaging.version.parse(str(self._MAX_SUPPORTED_SMIRNOFF_VERSION))):
+            raise SMIRNOFFVersionError(
+                'SMIRNOFF offxml file was written with version {}, but this version of ForceField only supports '
+                'up to version {}'.format(version, self._MAX_SUPPORTED_SMIRNOFF_VERSION))
+
+
+    def _register_aromaticity_model(self, aromaticity_model):
```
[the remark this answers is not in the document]
j-wags: Thanks for the feedback, however I don't think we want to treat aromaticity model as a property. 

The difference is that multiple OFFXMLs (or SMIRNOFF data sources) can be read into a single forcefield. The first one read should set the ForceField's aromaticity model, and each subsequent one will have its aromaticity model checked against the first's. I prefer directing all aromaticity models through this one function, but because multiple aromaticity models can be read, I think it would be confusing to handle it with a property setter.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -575,80 +631,175 @@ def parse(self, sources, input_format=None):
...
+        if not('SMIRNOFF' in smirnoff_data.keys()):
+            raise ParseError("'SMIRNOFF' must be a top-level key in the SMIRNOFF object model")
+
+        l1_dict = smirnoff_data['SMIRNOFF']
+        # Check that the aromaticity model required by this parameter set is compatible with
+        # others loaded by this ForceField
+        if 'aromaticity_model' in l1_dict.keys():
+            aromaticity_model = l1_dict['aromaticity_model']
+        else:
+            raise ParseError("'aromaticity_model' attribute must be specified in SMIRNOFF tag")
```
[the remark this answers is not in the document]
j-wags: Thanks for catching this -- cleaned up in e336b6d

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -575,80 +631,175 @@ def parse(self, sources, input_format=None):
...
+        #        "'aromaticity_model' (%s) does not match earlier read 'aromaticity_model' (%s)"
+        #        % (aromaticity_model, self._aromaticity_model))
+
+
+
+        # Check that the SMIRNOFF version of this data structure is supported by this ForceField implementation
+        if 'version' in l1_dict.keys():
+            version = l1_dict['version']
+        else:
+            raise ParseError("'version' attribute must be specified in SMIRNOFF tag")
```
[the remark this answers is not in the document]
j-wags: Fixed in 59cadf0

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: Yes, and `ForceField.get_handler(parameter_name, {})` will actually initialize the ParameterHandler if it doesn't exist. So here, just calling `get_handler` on the `ToolkitAM1BCC` tag will make this ForceField assign AM1BCC charges during system creation.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: I agree that this is bad, but won't change this right now. There are different ways we could standardize this (`__get/setstate__`, `from_dict`, `__init__`, designing for Pydantic in the future, etc) and it's big decision. Currently, there are conflicting serialization schemes throughout the toolkit (for example compare this to the `Serializable` class inheritance in `Molecule`). I think that the decision on serialization styles should be handled in a separate PR in the future.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -575,80 +631,175 @@ def parse(self, sources, input_format=None):
...
+            for parameter_tag, parameter_list in parameter_lists.items():
+                for parameter_dict in parameter_list:
+                    # Append units to parameters as needed
+                    parameter_kwargs = attach_units(parameter_dict,
+                                                     attached_units)
+                    # Add parameter definition
+                    handler.add_parameter(parameter_kwargs)
+
+
+    def parse_smirnoff_from_source(self, source):
```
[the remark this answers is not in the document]
j-wags: No action on this now, referencing #138 (issue on serialization)

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -103,20 +112,117 @@ class ParameterList(list):
...
+        for input_parameter in input_parameter_list:
+            self.append(input_parameter)
+            self._last_added_param = input_parameter_list
+
+    @property
+    def last_added_parameter(self):
+        """
+        Get a copy of the last parameter added to this ParameterList. Important
+        during serialization as the last parameter added determines the units used for
+        serializing all other parameters of this type.
```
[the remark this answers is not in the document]
j-wags: > One alternative solution, is to expose something like a default_serialization_unit property.

I think this is what we do with the `output_units` parameter for `to_dict` in our `ParameterType` and `ParameterHandler` classes.

> I wouldn't expose publicly the internals of how we make the default decision.

Not sure if I totally catch the meaning here. I agree that we shouldn't let users modify `_last_added_param`. Is this referring to something else?

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Brilliant. Totally agree. Thanks for the careful writeup. I've turned this into Issue #207 .

openforcefield/tests/test_molecule.py
```diff
@@ -504,7 +504,9 @@ def test_add_bond_charge_virtual_site(self, molecule):
                                                              distance,
                                                              sigma=0.1*unit.angstrom,
                                                              epsilon=1.0*unit.kilojoule_per_mole,
-                                                             charge_increments=unit.Quantity(np.array([0.1, 0.2, 0.3]), unit.elementary_charge)
+                                                             charge_increments=unit.Quantity(np.array([0.1, 0.2, 0.3]),
+                                                                                             unit.elementary_charge)
+                                                             #charge_increments=unit.Quantity([0.1, 0.2, 0.3], unit.elementary_charge)
```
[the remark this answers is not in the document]
j-wags: Removed in 40b4884

openforcefield/tests/test_parameters.py
```diff
@@ -15,14 +15,28 @@
...
 
 
+class TestParameterHandler:
+
+    def test_parameterhandler_with_different_units_to_dict(self):
+        """Test ParameterList.to_list() function when some parameters are in
+        different units (proper behavior is to convert all quantities to the last-
+        read unit)
+        """
+        pass
```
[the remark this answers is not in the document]
j-wags: Updated TODOs in 40b4884

openforcefield/tests/test_parameters.py
```diff
@@ -81,5 +96,324 @@ def test_del(self):
...
+                                                        phase1=30 * unit.degree,
+                                                        periodicity1=2,
+                                                        k1=5 * unit.kilocalorie_per_mole,
+                                                        phase2=31 * unit.angstrom, # This should be caught
+                                                        periodicity2=3,
+                                                        k2=6 * unit.kilocalorie_per_mole,
+                                                        )
+
+
+# test_parametertype_unit_property_getter
```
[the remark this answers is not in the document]
j-wags: Yes -- Fixed in 40b4884

openforcefield/tests/test_toolkits.py
```diff
@@ -679,14 +677,14 @@ def test_compute_partial_charges(self):
...
 
-        #with pytest.raises(Exception) as context:
-        with pytest.raises(NotImplementedError) as context:
+
+        with pytest.raises(NotImplementedError) as excinfo:
             charge_model = 'notARealChargeModel'
             molecule.compute_partial_charges(toolkit_registry=toolkit_registry)#, charge_model=charge_model)
 
         # ['cm1', 'cm2']
         for charge_model in ['gas', 'mul', 'bcc']:
```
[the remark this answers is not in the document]
j-wags: These are kinda legacy code. We haven't decided on a charge model for the SMIRNOFF spec.

For now, we support the `<ToolkitAM1BCC\>` tag, or carrying over charges from molecules.

This is part of a big-picture item: We need to restart the discussion about how charge models will be encoded in the SMIRNOFF spec. I've opened that up here: #208 

Previous discussion was here: https://openforcefieldgroup.slack.com/archives/CDR1P66Q2/p1546801608024700

openforcefield/topology/topology.py
j-wags: Yes. I have no idea why I didn't do that. Fixed in c4969c7.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: Yes -- Removed it

openforcefield/tests/test_toolkits.py
```diff
@@ -679,14 +677,14 @@ def test_compute_partial_charges(self):
...
-            with pytest.raises(NotImplementedError) as context:
+            with pytest.raises(NotImplementedError) as excinfo:
                 molecule.compute_partial_charges(toolkit_registry=toolkit_registry)#, charge_model=charge_model)
```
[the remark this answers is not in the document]
j-wags: Complicated. Until we decide on the charge model and its keywords, we can't get too far on these implementations.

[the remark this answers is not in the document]
jchodera: I was just suggesting you leave a `# TODO` note with when and how this comment should be fixed at a later time.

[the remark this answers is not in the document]
j-wags: Ah, my mistake, I should have read that closer. Didn't realize this was in a test file. TODO added in c7c4a83.

openforcefield/typing/engines/smirnoff/forcefield.py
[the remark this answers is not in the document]
andrrizzi: What I mean is that `register` implies that multiple aromaticity models can be registered. If this is true, +1 (although I'd use `add` in this case instead of `register`). If a `ForceField` instead can have only 1 aromaticity model, then `set` is probably the standard word. This is an internal-usage method anyway so not that important.

openforcefield/typing/engines/smirnoff/parameters.py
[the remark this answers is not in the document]
andrrizzi: I was referring to the public `last_added_parameter` property here, sorry! What I mean is that if in the future we want to change how we select the default units (currently we use the units of the last added parameter type), we'll have to break the public API. If we expose the information of the current default units with a `default_serialization_unit` instead of `last_added_parameter`, the transition to a different default scheme will be mostly harmless.

openforcefield/typing/engines/smirnoff/forcefield.py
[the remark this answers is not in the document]
j-wags: Ah, I agree. Changed it to `ForceField._set_aromaticity_model` in 7a61084

openforcefield/typing/engines/smirnoff/parameters.py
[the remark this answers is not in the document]
j-wags: Thanks for pushing this. I forgot that I exposed this as a `@property`, and didn't actually look at the referenced code on your comment. You're totally right, and I was lazy and didn't look closely enough at your feedback. I agree that this shouldn't be a public method. Removed in 409e2ae. 

I'm hesitant to expose `default_serialization_unit`, at this time, since we already support the `output_units` parameter during serialization. I'm going to leave it out for now but would consider adding it in a later PR.

j-wags:
@jchodera 

>I'm curious what the path to modularizing I/O might be so that future versions can support additional formats.

Right now, I'm relying on the concept that, if we're able to recursively drop all of our data structures to `dict`s of basic types (`str`, `int`, `list`, etc), then well supported serialization packages can serialize that. For example (beyond having to separate elements from attributes using `prepend_all_keys`) our [core XML serialization code is very simple](https://github.com/openforcefield/openforcefield/blob/ff-offxml-io/openforcefield/typing/engines/smirnoff/io.py#L266-L310). If we didn't care about [separating nodes from attributes](https://www.w3schools.com/XML/xml_dtd_el_vs_attr.asp), we could just run 

`xmltodict.unparse(my_forcefield.to_dict())`

and end up with a XML string containing our serialized forcefield.

The `prepend_all_keys()` code in `io.py` just separates nodes from attributes, which makes the output more readable but doesn't change the information content in a meaningful way. For example, 
```
<SMIRNOFF version="1.0" aromaticity_model="OEAroModel_MDL">

  <!-- SMIRNOFF (SMIRKS Native Open Force Field) template file -->
  <Date>2019-02-14</Date>
  <Author>C. I. Bayly, OpenEye/UC Irvine; C. C. Bannan, UC Irvine; D. L. Mobley, UC Irvine</Author>
...
```

contains the same information as 

```
<SMIRNOFF version="1.0" aromaticity_model="OEAroModel_MDL" Date="2019-02-14" Author="C. I. Bayly, OpenEye/UC Irvine; C. C. Bannan, UC Irvine; D. L. Mobley, UC Irvine">

  <!-- SMIRNOFF (SMIRKS Native Open Force Field) template file -->
...
```

PR #174, opened 2019-02-03
Travis tests for Frosst_AlkEthOH_parmAtFrosst parameter assignments

andrrizzi:
This PR automates the parameter comparison tests for the AlkEthOH set. Charges are currently ignored, but we can activate them when we'll be able to read them from file.

By default Travis tests only an (extensible) subset of 7 molecules (I found a test in `test_smirnoff.py` that was doing something similar using those molecule), but if `pytest` is called with the new `--runslow` option, all the AlkEthOH set is tested.

I've run the full set and they all pass, except for water (c1302), which was reparametrized in AlkEthOH to be TIP3P and is not covered by Frosst_AlkEthOH_parmAtFrosst.

Next step is doing the same thing with the FreeSolv files that I prepared for the SMIRNOFF 0.1 paper, although I stumbled into a couple of hiccups during my first experiments so I think we can merge this in the mean time.

Other changes:
- I’ve added a config file for pytest in `test/conftest.py` that allow us to mark tests with `pytest.mark.slow` and adds a new command line option `pytest --runslow`.
- I’ve updated `setup.cfg` with the latest cookiecutter’s version, which include the config for yapf.
- I’ve split the tests for `ParameterList` from `test_smirnoff.py` into a new `test_parameters.py` file.

### DIFF
--- a/openforcefield/tests/utils.py
+++ b/openforcefield/tests/utils.py
@@ -1095,7 +1095,7 @@ def compare_system_parameters(system1, system2, systems_labels=None,
 #=============================================================================================
 
 def compare_amber_smirnoff(prmtop_filepath, inpcrd_filepath, forcefield, molecule,
-                           **kwargs):
+                           check_parameters=True, check_energies=True, **kwargs):
     """
     Compare energies and parameters for OpenMM Systems/topologies created
     from an AMBER prmtop and crd versus from a SMIRNOFF forcefield file which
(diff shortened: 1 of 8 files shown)


openforcefield/tests/utils.py
```diff
@@ -897,42 +1047,67 @@ def compare_system_parameters(system1, system2):
...
 #=============================================================================================
 
-def compare_amber_smirnoff(prmtop_filepath, inpcrd_filepath, forcefield, molecule):
+def compare_amber_smirnoff(prmtop_filepath, inpcrd_filepath, forcefield, molecule,
+                           check_parameters=True, check_energies=True, **kwargs):
     """
     Compare energies and parameters for OpenMM Systems/topologies created
     from an AMBER prmtop and crd versus from a SMIRNOFF forcefield file which
-    should parameterize the same system with same parameters.
+    should parametrize the same system with same parameters.
```
j-wags: "parameterize", I think

j-wags: https://en.oxforddictionaries.com/definition/parameterize
https://english.stackexchange.com/questions/79847/parametrise-or-parameterise-a-curve

Looks like it's still an open question. Also it could end in "ise" in British english. Looks like, in addition to being in a scientific niche, we're also in a linguistic one. 

I have no strong opinion on which spelling we use, but for consistency we should agree on one (otherwise we'll have N+1 potential coding errors). @andrrizzi, if you have no preference, I'd suggest "parameterize". @SimonBoothroyd, will this be inconvenient for you?

andrrizzi: I think it's a good idea to stick to one convention. If @SimonBoothroyd is on board too, I'll go ahead and fix it before merging.

Issue #175, opened 2019-02-07
Parameter assignment regression test for FreeSolv fails due bug in toolkit version 0.0.4


andrrizzi:
I've attempted to test whether the current toolkit can recreate the `System`s we used for the hydration free energy calculations in the SMIRNOFF paper. Currently, the test fails for molecule `1019269` (1-butanol). Because of a different Lennard-Jones parameter that changed from version 0.0.3 (the toolkit version used for the calculations) to the current `smirnoff99Frosst.offxml`. Here are the parameters:
- 0.0.3 (id n13): https://github.com/openforcefield/openforcefield/blob/0.0.3/openforcefield/data/forcefield/smirnoff99Frosst.ffxml#L316
- 0.1 (id n12): https://github.com/openforcefield/openforcefield/blob/0.1.0/openforcefield/data/forcefield/smirnoff99Frosst.offxml#L315

Is this expected? If it is, should I just ignore the test case?

andrrizzi:
Actually, I'm not sure that's the problem because the error I get is:
```
The following particles have different parameters in the two NonbondedForce forces for the current OpenFF and SMIRNOFF 0.0.3 systems respectively:
particle 14:
    epsilon: 1.2552 kJ/mol != 0.0002204968 kJ/mol
    sigma: 9.390072489199176e-06 nm != 0.053453923088420355 nm
```
and the SMIRNOFF 0.0.3 parameters are different from the file I linked above. I'm assuming 0.03 files was used according to [this README.md](https://github.com/MobleyLab/SMIRNOFF_paper_code/blob/master/FreeSolv_repeat_subset/README.md), but I'll check whether there's no error in reading the topology or similar.

jchodera:
Neither of those could possibly be right! 

0.0.3 has zero epsilon and sigma, which we shouldn't allow:
```XML
     <Atom smirks="[#1:1]-[#8]" epsilon="0.0000" id="n13" rmin_half="0.0000"/>
```
and 0.1 has an `rmin_half` that is WAY too small for the epsilon:
```XML
    <Atom smirks="[#1:1]-[#8]" epsilon="0.3000" id="n12" rmin_half="5.27e-05"/>
```

andrrizzi:
After a little, I've noticed that
```
>>> (0.0002204968 * u.kilojoule_per_mole).value_in_unit(u.kilocalories_per_mole)
5.27e-05  # which is the same value of rmin_half in 0.1
[...]

PR #179, opened 2019-02-13
Fix human error in order of sigma/epsilon for polar hydrogens.

davidlmobley:
This fixes part of #175 -- due to human error on my part, polar hydroxyl hydrogens had their (minuscule) sigma/epsilon parameters swapped in the main toolkit version of the forcefield; the version used for our paper had the intended parameters, but the version incorporated here had the order swapped.

(This is because the release force field is built from a `frcmod` style legacy parameter file, so after fitting the parameters I had to port them into that file from the XML in which they were fit. And, because of legacy issues with the `frcmod` format, columns are not labeled, parameter order matters, etc., and I managed to input the parameters in the incorrect order when doing so. Ugh. We should be able to avoid such issues as we begin actually refitting rather than continuing to use this legacy `frcmod` file, but this was a one-off fix and I managed to make this huge error while putting it into the toolkit.)


davidlmobley:
@bannanc can you check and, after confirming/merging, do an update to the release force field as well?

PR #183, opened 2019-02-19
Add FreeSolv regression tests

### LINKED ISSUE #182: Expand error/warning message when undefined stereochemistry is detected
Currently, we [return None](https://github.com/openforcefield/openforcefield/blob/topology/openforcefield/utils/toolkits.py#L730) when OpenEye can't figure out the chirality of an atom/bond and `exception_if_undefined_stereo is False` so it is impossible to ignore the warning.

I'll fix this in the next PR.

andrrizzi:
This adds the regression test based on the FreeSolv set that we parameterized for the first paper (see https://github.com/MobleyLab/SMIRNOFF_paper_code) and address #175 .

Also:
- I've added the folder `data/forcefield/old/` with previous versions of the `smirnoff99Frosst.offxml` adapted to the new specification that we can use for the regression tests.
- I've added a tarball in `data/molecules/` with all the FreeSolv mol2 and xml files. These can be used also in other tests through the util function `openforcefield.tests.utils.get_freesolv_filepath()`.
- I've fixed #182, which made impossible parametrizing a molecule for which OpenEye could not figure out the stereochemistry.


openforcefield/utils/toolkits.py
```diff
@@ -727,7 +727,6 @@ def describe_oeatom(oeatom):
                 raise UndefinedStereochemistryError(msg)
             else:
                 print(msg)
-                return
```
j-wags: Per our discussion, I'm ok with returning the Molecule in these cases, but let's make sure to comeback and add a hefty warning for users who disable this check.

openforcefield/tests/conftest.py
```diff
@@ -24,14 +24,23 @@
...
 #=============================================================================================
 
-def untar_full_alkethoh_set():
-    """When running slow tests, we unpack the full AlkEthOH test in advance to speed things up."""
+def untar_full_alkethoh_and_freesolv_set():
+    """When running slow tests, we unpack the full AlkEthOH and FreeSolv test
+    sets in advance to speed things up.
+
+    See
+        test_forcefield.py::test_alkethoh_parameters_assignment
```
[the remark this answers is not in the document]
andrrizzi: Nope! Thanks for catching that!

openforcefield/utils/toolkits.py
andrrizzi: Sounds great!

davidlmobley:
Thanks! This was desperately needed, as you can tell by the number of issues it brought up. :)

PR #188, opened 2019-02-20
Use OrderedDict in labelMolecules.

proteneer:
Just a small quality of life improvement since XMLs are naturally ordered to start with.

### DIFF
--- a/openforcefield/typing/engines/smirnoff/forcefield.py
+++ b/openforcefield/typing/engines/smirnoff/forcefield.py
@@ -26,6 +26,7 @@ http://stackoverflow.com/questions/6949395/is-there-a-way-to-get-a-line-number-f
 # GLOBAL IMPORTS
 #=============================================================================================
 
+import collections
 import sys
 import string
 
@@ -1052,6 +1053,8 @@ To do: Update behavior of "Implied" force_type so it raises an exception if the 
             molecule. Each element is a list of the form [ ( [ atom1, ...,
             atomN], parameter_id, SMIRKS), ... ]
 
+        .. note:: this guarantees the ordering of the original xml file.
+
         """
 
         # XML modified? If so, re-parse by generators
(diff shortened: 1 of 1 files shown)


proteneer:
Looks like tests are failing due to license issues

jchodera:
> Looks like tests are failing due to license issues

Travis encrypted variables and files are only decrypted in [branches of the main repo](https://docs.travis-ci.com/user/encryption-keys/).

proteneer:
What do you recommend as a way forward?

jchodera:
For PRs on the `master` (which requires OpenEye toolkits), we might want to just get you access to creating branches in the main repo! Tagging @j-wags.

The good news is that [all dicts are ordered dicts from python 3.6 forward](https://twitter.com/raymondh/status/773978885092323328), so since we are dropping support for Python 3.5, this should be equivalent in behavior to the current toolkit with the supported Python 3.6/3.7.

proteneer:
Note that 3.6 it's still listed as an implementation detail - only CPython guarantees this. Python 3.7 enforces this strictly. I'm reluctant to move to 3.7 until tensorflow does.

andrrizzi:
Note also that `OrderedDict`s change the semantics of the equality operator so, if we decide to return as `OrderedDict` we would need to cast the labels as `dict(labels1) == dict(labels2)` when we expect the order not to matter. I've been hesitant to convert dicts into ordereddicts since I found this out.

proteneer:
I usually shy away from comparing dictionaries exactly for that reason. The ordering semantics are pretty tricky.

j-wags:
Agree with John. I've invited @proteneer as a contributor to `openforcefield`.

Hopefully we won't need to manually add each contributor in a few months (being a gatekeeper on that front seems decidedly not "open source"), as many of our tests will be able to run with an RDKit backend. We can start thinking of a policy for when we'll let PRs that can't test with OE merge into our dev branch. (I'm writing up a big thing on versioning/releasing, so I'll see about including that there)

j-wags:
I'm going to close this due to old age -- @proteneer, if you'd still like this feature, please open a PR from a branch, instead of a fork.

PR #189, opened 2019-02-21
Use AST as opposed to unsafe eval.

proteneer:
Uses an algebraic syntax tree to parse the unit bearing quantities as opposed to calling an unsafe eval call.

### DIFF
--- a/openforcefield/typing/engines/smirnoff/forcefield.py
+++ b/openforcefield/typing/engines/smirnoff/forcefield.py
@@ -25,7 +25,9 @@ http://stackoverflow.com/questions/6949395/is-there-a-way-to-get-a-line-number-f
 #=============================================================================================
 # GLOBAL IMPORTS
 #=============================================================================================
-
+import ast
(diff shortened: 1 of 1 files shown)


j-wags:
@andrrizzi, I think we are planning on making another release from `master` before we merge `topology` in. Further, we'll be supporting the 0.1.x toolkits for a few more months, even after the 1.0.0 release (exact horizon TBD, on the order of months).

Maybe let's leave this open while we implement an equivalent in the `topology` branch. I was in the process of refactoring the serialization code in the `ff-offxml-io` branch. In that branch, this new parsing code will go into the `extract_serialized_units_from_dict` function in `openforcefield.utils`. I can put it in Monday.

andrrizzi:
I've ported this changes to the `topology` branch. @j-wags, feel free to merge when ready.

j-wags:
These changes seemed to be fine in the `topology` branch. I'll merge this into `master` and watch how the travis build goes.

PR #190, opened 2019-02-22
Substitute eval with AST parser

andrrizzi:
Adapted from @proteneer's patch to the 0.1 version (see #189).

I've also added a new `test_io.py` file.

### DIFF
--- a/openforcefield/tests/test_io.py
+++ b/openforcefield/tests/test_io.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python
+
+#=============================================================================================
+# MODULE DOCSTRING
+#=============================================================================================
+
+"""
+Test classes and function in module openforcefield.typing.engines.smirnoff.io.
+
+"""
+
+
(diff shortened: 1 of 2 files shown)


andrrizzi:
Tests in `test_forcefield.py` pass locally.

openforcefield/tests/test_io.py
```diff
@@ -0,0 +1,68 @@
...
+)
+
+
+#=============================================================================================
+# QUANTITY PARSING UTILITIES
+#=============================================================================================
+
+@pytest.mark.parametrize('unit_string,expected_unit',[
+    ('kilocalories_per_mole', unit.kilocalories_per_mole),
+    ('kilocalories_per_mole/angstrom**2', unit.kilocalories_per_mole/unit.angstrom**2),
```
j-wags: Glad to see this in the test set. This is one of the trickier things we need to parse.

andrrizzi:
Great, thank you!

PR #209, opened 2019-03-18
Several fixes to Molecule stereo handling and tests

### LINKED ISSUE #196: What to do when rdkit.Chem.AssignStereochemistry does not add chiral code in Molecule.to_rdkit()?
Likely related to #146 and #156 .

With certain molecules, (e.g. DrugBank_6355, which is molecule 156 in the MiniDrugBank test set) [this call](https://github.com/openforcefield/openforcefield/blob/topology/openforcefield/utils/toolkits.py#L2112) of `rdkit.Chem.Chem.AssignStereochemistry` results in
```python
>>> rdatom.GetChiralTag()
rdkit.Chem.rdchem.ChiralType.CHI_UNSPECIFIED
```
If the code continues, [this line](https://github.com/openforcefield/openforcefield/blob/topology/openforcefield/utils/toolkits.py#L2114) results in `KeyError: '_CIPCode'`.

Differently from `from_rdkit()`, we already know what stereochemistry that atom should have so it doesn't make much sense to raise an exception.

*Is there a different way of setting the stereochemistry using S and R convention in RDKit? Alternatively, is there a way we can save the (read-only) chirality information in our `Atom` class using the CW/CCW notation alongside the S/R convention?*

### LINKED ISSUE #197: Have Molecule.from_iupac take an optional exception_if_undefined_stereo=True
The same argument is taken by `from_openeye` and `from_rdkit` and allow to manually bypass eventual errors (although the default behavior is to raise an exception) and proceed with the molecule parametrization.

### LINKED ISSUE #205: Should Molecule equality compare the exact bond order for aromatic bonds?
This snippet fails
```python
file_path = get_data_filename('molecules/MiniDrugBank_tripos.mol2')
molecules = Molecule.from_file(file_path, exception_if_undefined_stereo=False)
molecule = [mol for mol in molecules if mol.name == 'DrugBank_2210'][0]

iupac = molecule.to_iupac()
molecule_copy = Molecule.from_iupac(iupac)
assert molecule.is_isomorphic(molecule_copy)
```
because the molecule has a benzene ring for which the two methods of creating the `Molecule` end up assigning the alternate bond orders 1 and 2 differently.

I'm thinking of changing our condition for an [equal bond](https://github.com/openforcefield/openforcefield/blob/topology/openforcefield/topology/molecule.py#L1872-L1875) from
```python
(x['bond_order'] == y['bond_order']) and
(x['is_aromatic'] == y['is_aromatic']) and
(x['stereochemistry'] == y['stereochemistry'])
```
to
```python
((x['bond_order'] == y['bond_order']) or (x['is_aromatic'] == y['is_aromatic'])) and
(x['stereochemistry'] == y['stereochemistry'])
```

Are there cases in which this wouldn't do what we want?

andrrizzi:
This is a general cleanup and fixes to `test_molecules.py`. Apologies for the messy diff in `test_molecules.py` but I've reorganized and reordered the tests to figure out what was tested, what wasn't, and which tests were duplicated.

Travis tests should pass right now although there are several in `test_to_from_rdkit` that fails. I've created a new `pytest.mark.wip` marker to allow specific tests to xfail on Travis so that we can keep track of unresolved issues while still being able to test in Travis for regression tests. I've also added a `pytest --failwip` option to make them fail instead of xfail.

Other main changes in this PR:
- Fix #197: Rename `exception_if_undefined_stereo` to `allow_undefined_stereo`. Also, have `Molecule.from_iupac()` and `Molecule()` take an optional `allow_undefined_stereo` kwarg.
- Fix #196: Force `_CIPCode` property S/R in `Molecule.to_rdkit` if RDKit can't figure out the atom chirality.
- Fix #205: Do not check precise bond order equality for aromatic atoms.
- Added an optional `compare_atom_stereochemistry` to `Molecule.is_isomorphic()` to be able to compare for equality molecules that are equal up to an undefined stereochemistry.


openforcefield/topology/molecule.py
```diff
@@ -2861,6 +2884,7 @@ def from_file(filename,
 
         # Determine which toolkit to use (highest priority that's compatible with input type)
         if isinstance(toolkit_registry, ToolkitRegistry):
+            # TODO: Encapsulate this logic into ToolkitRegistry.call()?
```
j-wags: Ah, this could be a really good idea. We'd need to be careful, however, about the case where RDKit skips a molecule (because then we'd want to try to use another toolkit for it)

openforcefield/topology/molecule.py
```diff
@@ -2871,15 +2895,16 @@ def from_file(filename,
                     query_toolkit.
                     toolkit_name] = query_toolkit.toolkit_file_read_formats
             if toolkit == None:
-                raise Exception(
+                raise NotImplementedError(
```
j-wags: 👍

openforcefield/topology/molecule.py
```diff
@@ -3354,6 +3377,7 @@ def __init__(self, *args, **kwargs):
         #super(self, Molecule).__init__(*args, **kwargs)
         super(Molecule, self).__init__(*args, **kwargs)
 
+    # TODO: Change this to add_atom(Atom) to improve encapsulation and extensibility?
```
j-wags: 👍

openforcefield/utils/toolkits.py
```diff
@@ -2202,9 +2208,15 @@ def to_rdkit(molecule, aromaticity_model=DEFAULT_AROMATICITY_MODEL):
...
         # I added AssignStereochemistry which takes the directions of the bond set
-        # and assigns the stereochemistry tags on the double bonds
+        # and assigns the stereochemistry tags on the double bonds.
         Chem.AssignStereochemistry(rdmol, force=False)
 
+        # Forcefully assign stereo information on the atoms that RDKit
+        # can't figure out. This must be done last as calling AssignStereochemistry
+        # again will delete these properties (see #196).
+        for rdatom, stereochemistry in undefined_stereo_atoms.items():
+            rdatom.SetProp('_CIPCode', stereochemistry)
```
j-wags: Great.

openforcefield/topology/topology.py
```diff
@@ -861,7 +861,7 @@ def __init__(self, other=None):
         elif isinstance(other, FrozenMolecule):
             self.from_molecules([other])
         elif isinstance(other, OrderedDict):
-            self.initialize_from_dict(other)
+            self._initialize_from_dict(other)
```
j-wags: It appears that neither of these functions actually exist. Wonder how these got here in the first place. Fascinating!

It's definitely better to refer to these as private methods if they're yet-to-be-implemented.

openforcefield/topology/molecule.py
```diff
@@ -1850,29 +1856,45 @@ def from_smiles(smiles, hydrogens_are_explicit=False, toolkit_registry=GLOBAL_TO
...
                 .format(type(toolkit_registry)))
 
-    def is_isomorphic(self, other):
+    def is_isomorphic(
+            self, other,
+            compare_atom_stereochemistry=True
+    ):
         """
-        Determines whether the molecules are isomorphic by comparing their SMILESes
+        Determines whether the molecules are isomorphic by comparing their graphs.
```
j-wags: Good catch.

openforcefield/topology/molecule.py
```diff
@@ -1850,29 +1856,45 @@ def from_smiles(smiles, hydrogens_are_explicit=False, toolkit_registry=GLOBAL_TO
...
+                is_equal &= x['stereochemistry'] == y['stereochemistry']
+            return is_equal
+
+        def edge_match_func(x, y):
+            return (
+                # We don't need to check the exact bond order (which is 1 or 2)
+                # if the bond is aromatic. This way we avoid missing a match only
+                # if the alternate bond orders 1 and 2 are assigned differently.
+                (x['is_aromatic'] == y['is_aromatic'] or x['bond_order'] == y['bond_order']) and
+                (x['stereochemistry'] == y['stereochemistry'])
```
j-wags: Bond stereochemistry can also differ between OE and RDK (for example, RDK doesn't see stereochemistry for `RC=NH`, while OE does). We may want to add a parameter `compare_bond_stereochemistry` for that.

openforcefield/topology/molecule.py
```diff
@@ -1850,29 +1856,45 @@ def from_smiles(smiles, hydrogens_are_explicit=False, toolkit_registry=GLOBAL_TO
...
-                                        )
+
+        def node_match_func(x, y):
+            is_equal = (
+                (x['atomic_number'] == y['atomic_number']) and
+                (x['is_aromatic'] == y['is_aromatic']) and
+                (x['formal_charge'] == y['formal_charge'])
+            )
+            if compare_atom_stereochemistry:
+                is_equal &= x['stereochemistry'] == y['stereochemistry']
```
j-wags: Very clever! I didn't know we could do that.

openforcefield/topology/molecule.py
andrrizzi: Good idea! Will add before merging.

Issue #191, opened 2019-02-27
Changes to SMIRNOFF spec documentation


j-wags:
This issue is to keep a record of spec changes/questions to make as I find them.

* GBSA model: `radius_units` should be `radius_unit` to match up with other `_unit` fields in the spec (and because the `_unit` suffix on an attrib is used to trigger special behavior during parsing)
* It's not clear whether SA model defaults to `None` or `ACE`

j-wags:
* VDW tags' `switch=8.0*angstrom` should be changed to `switch='8.0' switch_unit='angstrom'`
* Same with VDW tags' cutoff

j-wags:
* change `switch-width` attribute of `Electrostatics` tag to `switch_width` (underscore) in compliance with other multi-word attributes
* separate out `cutoff_unit` and `switch_width_unit` attributes in `Electrostatics` tag

j-wags:
Add a note clarifying that it you should not use the same attrib name for a section header attrib and a parameter attrib (for example, I shouldn't have a section header "length" and "length_unit", and also a ParameterType "length" field, because the units might get crossed).

The spec does not currently do this. However, the parser for unit-bearing fields will have undefined behavior if someone makes a new ParameterHandler that uses the same name for a unit-bearing attrib in both the header and each parameter, and then somehow sets them to different unit systems.

j-wags:
* Resolve whether torsions get fractional bond orders

j-wags:
* The `vdW` tag goes up to a 1-5 scaling term. `Electrostatics` only goes up to 1-4. Should `Electrostatics` have a 1-5?

jchodera:
> The `vdW` tag goes up to a 1-5 scaling term. `Electrostatics` only goes up to 1-4. Should `Electrostatics` have a 1-5?

Sure, let's make them both go to 1-5.

j-wags:
Always put increment-able numbers at the end of attr names. For example `charge1increment` should become `chargeincrement1`

j-wags:
@jchodera @davidlmobley Should `ImproperTorsion`s have the same `idivf` convention as `ProperTorsion`s? I'm mostly asking with regard to the default `default_idivf='auto'` setting. 

My understanding is that this is where our "impropers-get-divided-by-a-factor-of-three" convention stems from, but I could imagine someone defining weird impropers so I don't want to leave "3" hard coded.

[...]

Issue #178, opened 2019-02-12
Make all offxml files compatible 


bannanc:
@ChayaSt attempted to run the example from the README and was running into a weird error, that initially I wasn't able to replicate, but then tried a slightly different version of the same code and we were able to isolate the issue. 

We started from a new conda environment with:
```
conda create -n chaya_test python=3.6
conda activate chaya_test
conda install --yes -c conda-forge -c omnia openforcefield
conda install -c openeye openeye-toolkits
conda install ipython jupyter
```

Then using the example we made an OEMol from a smiles and loaded the `offxml` file from the conda installation.
```
smiles = 'Cc1cccc(c1C)C'
new_mol = oechem.OEMol()
oechem.OESmilesToMol(new_mol, smiles)
oechem.OEAddExplicitHydrogens(new_mol)
FF_filename = get_data_filename('forcefield/Frosst_AlkEthOH_parmAtFrosst.offxml')
ff = forcefield.ForceField(FF_filename)
top, syst, pos  = ff_utils.create_system_from_molecule(ff, new_mol, verbose = False)
```

This gives this rather confusing error:
```
Traceback (most recent call last):
  File "test_chaya_error.py", line 15, in <module>
    top, syst, pos  = ff_utils.create_system_from_molecule(ff, new_mol, verbose = False)
  File "/Users/caitlinbannan/miniconda3/envs/chaya_test/lib/python3.6/site-packages/openforcefield/typing/engines/smirnoff/forcefield_utils.py", line 98, in create_system_from_molecule
    system = forcefield.createSystem(topology, [mol], verbose=verbose)
  File "/Users/caitlinbannan/miniconda3/envs/chaya_test/lib/python3.6/site-packages/openforcefield/typing/engines/smirnoff/forcefield.py", line 1022, in createSystem
    force.createForce(system, topology, verbose=verbose, **kwargs)
  File "/Users/caitlinbannan/miniconda3/envs/chaya_test/lib/python3.6/site-packages/openforcefield/typing/engines/smirnoff/forcefield.py", line 1499, in createForce
    _check_for_missing_valence_terms('HarmonicBondForce', topology, bonds.keys(), topology.bonds())
  File "/Users/caitlinbannan/miniconda3/envs/chaya_test/lib/python3.6/site-packages/openforcefield/typing/engines/smirnoff/forcefield.py", line 1207, in _check_for_missing_valence_terms
    raise Exception(msg)
[...]

PR #215, opened 2019-03-28
Update SMIRNOFF spec and implement Electrostatics tag handler 

j-wags:
This PR 
* Updates the SMIRNOFF spec with recent updates, closing #191 
* Corrects a long-running typo in our vdW tags (`Loentz-Berthelot` --> `Lo_r_entz-Berthelot`)
* ~Adds optional `load_box_vectors` param to `Topology.from_openmm()`, thus allowing retrieval of periodic box vectors during Topology initialization.~
* Implements `ElectrostaticsHandler`, thus adding support for the `Electrostatics` OFFXML tag
* Implements `_validate_parameters` functions for `vdWHandler` and `ElectrostaticsHandler` which ensure that their settings map to an OpenMM `NonbondedMethod` that we currently support.
* Implements checks during system creation that ensure `vdWHandler` and `ElectrostaticsHandler` are configured in a way compatible with the presence/absence of a periodic box in the `Topology` object.
* ~Implements the `ForceField.configure_vacuum()` function, which recursively calls `configure_vacuum` on all of the ForceField's ParameterHandlers. These functions have the ParameterHandlers switch to default non-periodic/vacuum phase settings.~ 
    * ~This makes it easier to "just run" a vacuum simulation. It will be useful because all of our  `smirnoff99Frosst.offxml` files have condensed-phase-specific vdW and Electrostatics parameters specified. Thankfully, we currently only support one configuration for vacuum, so switching to "vacuum settings" is currently unambiguous. We will need to refactor this later, however.~
* Exposes 1-5 scaling parameters
* Removes vdW `long_range_dispersion` attribute, and replaces it with vdw `method`, which can be set to either `cutoff` (default) or `PME`.
* Adds logic to resolve inputs to OpenMM system types specified [on slide 15 here](https://docs.google.com/presentation/d/1U1ZnB-kaAA1s-lFjTQSI1hHx5kntbOw-H4HlGdV4KCo/edit#slide=id.g554019e4c3_0_13)

### DIFF
--- a/The-SMIRNOFF-force-field-format.md
+++ b/The-SMIRNOFF-force-field-format.md
@@ -212,7 +212,7 @@ For example, to ensure water molecules are assigned partial charges for [TIP3P](
 
 In keeping with the AMBER force field philosophy, especially as implemented in small molecule force fields such as [GAFF](http://ambermd.org/antechamber/gaff.html), [GAFF2](https://mulan.swmed.edu/group/gaff.php), and [parm@Frosst](http://www.ccl.net/cca/data/parm_at_Frosst/), partial charges for small molecules are usually assigned using a quantum chemical method (usually a semiempirical method such as [AM1](https://en.wikipedia.org/wiki/Austin_Model_1)) and a [partial charge determination scheme](https://en.wikipedia.org/wiki/Partial_charge) (such as [CM2](http://doi.org/10.1021/jp972682r) or [RESP](http://doi.org/10.1021/ja00074a030)), then subsequently corrected via charge increment rules, as in the highly successful [AM1-BCC](https://dx.doi.org/10.1002/jcc.10128) approach.
 
-Here is an example:
+Here is an example: 
 ```XML
 <ChargeIncrementModel number_of_conformers="10" quantum_chemical_method="AM1" partial_charge_method="CM2" increment_unit="elementary_charge">
   <!-- A fractional charge can be moved along a single bond -->
@@ -260,7 +260,7 @@ As an example of a complete SMIRNOFF force field specification, see the [AlkEthO
 
 van der Waals force parameters, which include repulsive forces arising from Pauli exclusion and attractive forces arising from dispersion, are specified via the `<vdW>` tag with sub-tags for individual `Atom` entries, such as:
 ```XML
-<vdW potential="Lennard-Jones-12-6" combining_rules="Lorentz-Berthelot" scale12="0.0" scale13="0.0" scale14="0.5" scale15="1.0" sigma_unit="angstroms" epsilon_unit="kilocalories_per_mole" switch="8.0" switch_unit="angstrom" cutoff="9.0" cutoff_unit="angstroms" long_range_dispersion="isotropic">
+<vdW potential="Lennard-Jones-12-6" combining_rules="Lorentz-Berthelot" scale12="0.0" scale13="0.0" scale14="0.5" scale15="1.0" sigma_unit="angstroms" epsilon_unit="kilocalories_per_mole" switch_width="8.0" switch_width_unit="angstrom" cutoff="9.0" cutoff_unit="angstroms" long_range_dispersion="isotropic">
    <Atom smirks="[#1:1]" sigma="1.4870" epsilon="0.0157"/>
    <Atom smirks="[#1:1]-[#6]" sigma="1.4870" epsilon="0.0157"/>
    ...
@@ -271,8 +271,7 @@ The two are related by `r0 = 2^(1/6)*sigma` and conversion is done internally in
 Note that, if `rmin_half` is specified instead of `sigma`, `rmin_half_unit` should be specified; both can be used in the same block if desired.
 
 Attributes in the `<vdW>` tag specify the scaling terms applied to the energies of 1-2 (`scale12`, default: 0), 1-3 (`scale13`, default: 0), 1-4 (`scale14`, default: 0.5), and 1-5 (`scale15`, default: 1.0) interactions,
-as well as the distance at which a switching function is applied (`switch`, default: `"8.0"` angstroms), the cutoff (`cutoff`, default: `"9.0"` angstroms), and long-range dispersion treatment scheme (`long_range_dispersion`, default: `"isotropic"`).
-If Lennard-Jones PME is to be used, `switch` and `cutoff` are omitted and `long_range_dispersion="PME"` is used.
+as well as the distance at which a switching function is applied (`switch_width`, default: `"1.0"` angstroms), the cutoff (`cutoff`, default: `"9.0"` angstroms), and long-range dispersion treatment scheme (`long_range_dispersion`, default: `"isotropic"`).
 
 The `potential` attribute (default: `"none"`) specifies the potential energy function to use.
 Currently, only `potential="Lennard-Jones-12-6"` is supported:
@@ -284,7 +283,7 @@ Support for [other Lennard-Jones mixing schemes](https://en.wikipedia.org/wiki/C
 
 Later revisions will add support for additional potential types (e.g., `Buckingham-exp-6`), as well as the ability to support arbitrary algebraic functional forms using a scheme such as
 ```XML
-<vdW potential="4*epsilon*((sigma/r)^12-(sigma/r)^6)" scale12="0.0" scale13="0.0" scale14="0.5" scale15="1" sigma_unit="angstrom" epsilon_unit="kilocalories_per_mole" switch="8.0" switch_unit="angstrom" cutoff="9.0" cutoff_unit="angstrom" long_range_dispersion="isotropic">
+<vdW potential="4*epsilon*((sigma/r)^12-(sigma/r)^6)" scale12="0.0" scale13="0.0" scale14="0.5" scale15="1" sigma_unit="angstrom" epsilon_unit="kilocalories_per_mole" switch_width="8.0" switch_width_unit="angstrom" cutoff="9.0" cutoff_unit="angstrom" long_range_dispersion="isotropic">
    <CombiningRules>
       <CombiningRule parameter="sigma" function="(sigma1+sigma2)/2"/>
       <CombiningRule parameter="epsilon" function="sqrt(epsilon1*epsilon2)"/>
--- a/The-SMIRNOFF-force-field-format.md
+++ b/The-SMIRNOFF-force-field-format.md
@@ -216,11 +216,11 @@ Here is an example:
 ```XML
 <ChargeIncrementModel number_of_conformers="10" quantum_chemical_method="AM1" partial_charge_method="CM2" increment_unit="elementary_charge">
   <!-- A fractional charge can be moved along a single bond -->
-  <ChargeIncrement smirks="[#6X4:1]-[#6X3a:2]" charge1increment="-0.0073" charge2increment="+0.0073"/>
-  <ChargeIncrement smirks="[#6X4:1]-[#6X3a:2]-[#7]" charge1increment="+0.0943" charge2increment="-0.0943"/>
-  <ChargeIncrement smirks="[#6X4:1]-[#8:2]" charge1increment="-0.0718" charge2increment="+0.0718"/>
+  <ChargeIncrement smirks="[#6X4:1]-[#6X3a:2]" chargeincrement1="-0.0073" chargeincrement2="+0.0073"/>
+  <ChargeIncrement smirks="[#6X4:1]-[#6X3a:2]-[#7]" chargeincrement1="+0.0943" chargeincrement2="-0.0943"/>
+  <ChargeIncrement smirks="[#6X4:1]-[#8:2]" chargeincrement1="-0.0718" chargeincrement2="+0.0718"/>
   <!--- Alternatively, factional charges can be redistributed among any number of bonded atoms -->
-  <ChargeIncrement smirks="[N:1](H:2)(H:3)" charge1increment="+0.02" charge2increment="-0.01" charge3increment="-0.01"/>
+  <ChargeIncrement smirks="[N:1](H:2)(H:3)" chargeincrement1="+0.02" chargeincrement2="-0.01" chargeincrement3="-0.01"/>
 </ChargeIncrementModel>
 ```
 The sum of formal charges for the molecule or fragment will be used to determine the total charge the molecule or fragment will possess.
--- a/The-SMIRNOFF-force-field-format.md
+++ b/The-SMIRNOFF-force-field-format.md
@@ -644,7 +644,7 @@ This is a backwards-incompatible overhaul of the SMIRNOFF 0.1 draft specificatio
     * `<BondChargeCorrections>` was renamed to `<ChargeIncrementModel>` and generalized to accommodate an arbitrary number of tagged atoms
     * `<GBSAForce>` was renamed to `<GBSA>`    
 * `<PeriodicTorsionForce>` was split into `<ProperTorsions>` and `<ImproperTorsions>`
-* `<vdW>` now specifies 1-2, 1-3, 1-4, and 1-5 scaling factors via `scale12` (default: 0), `scale13` (default: 0), and `scale14` (default: 0.5) attributes. Coulomb scaling parameters have been removed from `StericsForce`.
+* `<vdW>` now specifies 1-2, 1-3, 1-4, and 1-5 scaling factors via `scale12` (default: 0), `scale13` (default: 0), `scale14` (default: 0.5), and `scale15` (default 1.0) attributes. Coulomb scaling parameters have been removed from `StericsForce`.
 * Added the `<Electrostatics>` tag to separately specify 1-2, 1-3, 1-4, and 1-5 scaling factors for electrostatics, as well as the method used to compute electrostatics (`PME`, `reaction-field`, `Coulomb`) since this has a huge effect on the energetics of the system.
 * Made it clear that `<Constraint>` entries do not have to be between bonded atoms.
 * `<VirtualSites>` has been added, and the specification of charge increments harmonized with `<ChargeIncrementModel>`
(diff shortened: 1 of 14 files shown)


openforcefield/topology/topology.py
```diff
@@ -1383,6 +1382,9 @@ def from_openmm(cls, openmm_topology, unique_molecules=None):
             OpenMM ``Topology``, these will be used in matching as well.
             If all bonds have bond orders assigned in ``mdtraj_topology``, these bond orders will be used to attempt to construct
             the list of unique Molecules if the ``unique_molecules`` argument is omitted.
+        load_box_vectors : bool. Default=False
```
jchodera: Why would this be exposed as an optional argument? The OpenMM `Topology` object only contains box vectors if the system is meant to be simulated in the condensed phase, so we should presumably pull these automatically if they are present. So I think this behavior should be automatic, and not something we need to expose to the users.

That is, the behavior should be 
* if OpenMM `Topology` has no box vectors, assume system is vacuum
* if OpenMM `Topology` has box vectors, assume system is condensed phase and copy box vectors

We can have API methods for our OFF `Topology` that allow us to modify the system to be condensed phase or vacuum, or alter the box vectors, later.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -792,6 +792,14 @@ def _resolve_parameter_handler_order(self):
     # TODO: How do we know if the system is periodic or not?
     # TODO: Should we also accept a Molecule as an alternative to a Topology?
 
+
+    def configure_vacuum(self):
```
jchodera: I'm uncertain we want to expose an API point like this. Following my conversation with @j-wags, the `ForceField` should have a recipe for how to process `Topology` into a `System`, and the `Topology` should know whether it is supposed to represent molecule or cluster in vacuum or in a condensed phase system. We can certainly add API methods to the `Topology` object that let us manipulate what system it is supposed to represent, but that should _not_ happen at the `ForceField` level, since the force field is just the recipe on how to apply parameters to a given description of a system.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1372,18 +1414,36 @@ def check_handler_compatibility(self,
...
-        force.setCutoffDistance(self._cutoff.in_units_of(unit.nanometer))
-        if 'ewaldErrorTolerance' in kwargs:
-            force.setEwaldErrorTolerance(kwargs['ewaldErrorTolerance'])
-        if 'useDispersionCorrection' in kwargs:
-            force.setUseDispersionCorrection(
-                bool(kwargs['useDispersionCorrection']))
+        # If we're using PME, then the only possible openMM Nonbonded type is LJPME
+        if self._long_range_dispersion == 'LJPME':
+            force.setNonbondedMethod(openmm.NonbondedForce.LJPME)
+            force.setCutoffDistance(self._cutoff.in_units_of(unit.nanometer))
```
jchodera: The `.in_units_of(unit.nanometer)` is totally unnecessary here (and in subsequent lines) if the quantity is a `simtk.unit.Quantity`. It will already be automatically converted to the appropriate units by OpenMM.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1372,18 +1414,36 @@ def check_handler_compatibility(self,
...
+        elif self._long_range_dispersion == 'isotropic':
+            force.setNonbondedMethod(openmm.NonbondedForce.PME)
+            force.setCutoffDistance(self._cutoff.in_units_of(unit.nanometer))
+            if hasattr(self, '_switch_width'):
```
jchodera: The logic is backwards here. This should be `if not hasattr(self, '_switch_width')`.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1372,18 +1414,36 @@ def check_handler_compatibility(self,
...
-            force.setEwaldErrorTolerance(kwargs['ewaldErrorTolerance'])
-        if 'useDispersionCorrection' in kwargs:
-            force.setUseDispersionCorrection(
-                bool(kwargs['useDispersionCorrection']))
+        # If we're using PME, then the only possible openMM Nonbonded type is LJPME
+        if self._long_range_dispersion == 'LJPME':
+            force.setNonbondedMethod(openmm.NonbondedForce.LJPME)
+            force.setCutoffDistance(self._cutoff.in_units_of(unit.nanometer))
+        elif self._long_range_dispersion == 'isotropic':
+            force.setNonbondedMethod(openmm.NonbondedForce.PME)
```
jchodera: Are we sure that the electrostatics treatment here has also selected `PME`?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1372,18 +1414,36 @@ def check_handler_compatibility(self,
...
+            force.setCutoffDistance(self._cutoff.in_units_of(unit.nanometer))
+            if hasattr(self, '_switch_width'):
+                force.setUseSwitchingFunction(False)
+            else:
+                force.setUseSwitchingFunction(True)
+                force.setSwitchingDistance((self._cutoff - self._switch_width).in_units_of(unit.nanometer))
+        elif self._long_range_dispersion == "None":
+            force.setNonbondedMethod(openmm.NonbondedForce.NoCutoff)
```
jchodera: You can use PME and no long-range dispersion. I think the logic here is confused.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: So, this is resolved using a lazy hack for now. 

The Electrostatics handler has the vdW handler as a dependency, so we know that the vdW handler (this code) will run first. The Electrostatics handler them checks the openMM system for whether LJPME is set. If it is, the electrostatics handler [ensures that its `method` is set to PME](https://github.com/openforcefield/openforcefield/blob/54682abe6af67636e6b1286e08feadc30e77725e/openforcefield/typing/engines/smirnoff/parameters.py#L1584-L1592). The ugly part is that the communication hard-coded in a way it shouldn't be. We'll be able to remove that when we separate electrostatics and vdW into separate openMM forces.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: `configure_vacuum` is my idea for fixing a major roadblock in terms of user experience. It's likely that I've been looking at this specific issue so long that I'm overlooking a better solution. But here's my reasoning:

`configure_vacuum` has several flaws, including setting things to opaque "vacuum default" values, and changing the API of every other ParameterHandler.

However, right now, our `smirnoff99Frosst` file specifies parameters specific to condensed-phase systems (for example, `PME` electrostatics). We will, presumably, put out examples for how to run vacuum phase simulations. If we don't include `configure_vacuum` I'm concerned that these vacuum simulation examples will look like:

```
ff = ForceField('smirnoff99Frosst.offxml')
ES_handler = ff.get_handler('Electrostatics')
ES_handler.method = 'Coulomb'
del ES_handler._cutoff
del ES_handler._switch
vdw_handler = ff.get_handler('vdW')
vdw_handler.long_range_dispersion = 'None'
del vdw_handler._cutoff

topology = Topology.from_openmm(pdbfile.topology, unique_molecules=molecules, load_box_vectors=False)
```

We could make the API more convenient to a degree, but the fact remains that the user would have to manually change a number of settings to get our default OFFXML files to run a vacuum simulation.

I had considered two other ideas as well:
* to automatically have the `ForceField`/`ParameterHandler`s switch to vacuum-compatible settings if a vacuum topology were passed in, but, this is as bad as `configure_vacuum`, while being even more opaque.
* to have separate `smirnoff99Frosst_condensed` and `_vacuum` OFFXMLs, but I'm hesitant to double the number of files we keep around. 

I think it's likely that there's a better way to handle this, but I wasn't able to think of it. Advice would be appreciated!

openforcefield/topology/topology.py
j-wags: > The OpenMM Topology object only contains box vectors if the system is meant to be simulated in the condensed phase, so we should presumably pull these automatically if they are present.

I'm flexible on this. My experience as a user in setting up tests was that many of our PDB files had some arbitrary values at the top, under the `CRYST1` field. These were, to my surprise, being interpreted by OpenMM as box vectors. 

Thinking more, _we're_ not responsible for the behavior of OpenMM's PDB reader. I think you're right that we shouldn't expose this, and should by default load the box vectors if they're present.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Ah, great. I wasn't sure why `in_units_of` was there in the first place, but I didn't want to break the pattern in case it was important. Cleaned up in fa632e2

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Great catch! Fixed in f842fdf

openforcefield/topology/topology.py
j-wags: Changed to your suggestion in fc031bb

j-wags:
@jchodera Thanks for the careful feedback -- This is really helpful. I'm going to bed but will pick back up on responses in the morning. Let me know if you have some time to help me map to valid nonbonded methods tomorrow.

openforcefield/typing/engines/smirnoff/forcefield.py
jchodera: The way we *should* handle this is for the forcefield OFF XML definition file to include information on how both vacuum and condensed phase systems should be treated. However, in our conversation, it became apparent there was currently only one way that vacuum systems should be treated, and that if there were more variety we might need to expose in future, we could do this by extending the spec.

My take away from our discussion was that the Topology object would know whether it represented a vacuum or condensed phase system, and that the OFF XML spec contained information on how the condensed phase representation should work (with vacuum being implicit but extensible in future). We would then exclusively have the user be concerned with what the Topology object represented for the system in condensed or vacuum phases, and the ForceField would just apply the protocol for either vacuum or condensed phases.

andrrizzi: But what about something like transforming `method` and `long_range_dispersion` into properties that automatically configure `cutoff` and `switch_width` accordingly?

That would reduce it to
```python
ff = ForceField('smirnoff99Frosst.offxml')
ff.get_handler('Electrostatics').method = 'Coulomb'
ff.get_handler('vdW').long_range_dispersion = None
```
Even better, I think, would be to not delete entirely the attributes and have, but have a property adopt the correct behavior. For example for `cutoff` we can have
```python
@property
def cutoff(self):
    if self.method == 'Coulomb':
        return None  # or alternatively raise AttributeError
    return self._cutoff
```
the effect would be the same, but if you switch the `method` back to `PME`, then the original cutoff loaded from the force field XML will still be in memory.

andrrizzi:
> wonder if it doesn't make the most sense to use `switch_width="8.0*angstrom"`

@j-wags can correct me if I'm wrong, but as far as I understand, separating value and units makes the XML easier to read by humans (it keeps the file more compact, and it forces all values to be specified in the same units so that they can be easily compared), while I think that using `value*unit` makes the XML easier to read by computers as it should simplify the serialization/deserialization of the various `ParameterType`s and `ParameterHandler`s and remove some of the code complexity.

I'm very ok with both solutions so I guess we should decide what we think it's more important between those two.

j-wags:
> wonder if it doesn't make the most sense to use switch_width="8.0*angstrom"

I agree with @andrrizzi here. It's a bit weird to specify value and unit separately in the header. However, it's much less bulky to do it this way in the parameter sections, and I think it's important to be consistent with how we handle unit-bearing quantities throughout the file.

We can listen for feedback about this during the alpha and switch over if users find it unwieldy.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: @jchodera 
> My take away from our discussion was that the Topology object would know whether it represented a vacuum or condensed phase system, and that the OFF XML spec contained information on how the condensed phase representation should work (with vacuum being implicit but extensible in future).

I see. This is a good design decision, and I hadn't fully understood it before. Three action items come from this:

1) I'll implement a switchover to vacuum settings if the Topology is provided without a periodic box.
2) I'll remove `configure_vacuum` from the public API
3) I'll open up an issue to discuss how we can encode nonperiodic and periodic defaults in the same OFFXML file (right now it happens to be easy because we only have one setting for vacuum, but it could get more complicated later)

j-wags: @andrrizzi 

> What about something like transforming method and long_range_dispersion into properties that automatically configure cutoff and switch_width accordingly?

I think that's a good idea once I have a better understanding of all of the desired logic here. Given the timeline, I'm aiming to do "space shuttle" coding, where each situation is handled more-or-less explicitly. This way we can visually ensure the logic is right, and make the API friendlier later.

That said, I'm rethinking the current data structure implementation now that I understand that a single OFFXML should contain both periodic and nonperiodic settings.

jchodera: > I'll open up an issue to discuss how we can encode nonperiodic and periodic defaults

Let's be clear about calling them "vacuum" and "condensed" since, as we discussed, we can use GBSA to simulate the condensed phase but the system would not be periodic.

> What about something like transforming method and long_range_dispersion into properties that automatically configure cutoff and switch_width accordingly?

@andrrizzi's suggestion is a great one if we want people to be able to change the nonbonded treatment via the API and have validators not cause an error because you would have to set cutoff and nonbonded treatment simultaneously. I think the idea of defaults is probably the only sensible solution.

openforcefield/topology/topology.py
```diff
@@ -2097,8 +2103,15 @@ def get_fractional_bond_order(self, iatom, jatom):
...
+        return self._box_vectors is not None
+
+    @property
+    def is_vacuum(self):
         """
-        ``True`` if the topology represents a periodic system; ``False`` otherwise
+        ``True`` if the topology represents a vacuum/nonperiodic system; ``False`` otherwise
         """
-        return self._is_periodic
+        return self._box_vectors is None
```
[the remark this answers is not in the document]
jchodera: Agreed. We need some way to represent whether the system is in vacuum (in which case we will use no-cutoff methods for now, but may add options on how to handle this differently later) or in the condensed phase (which may or may not make use of box vectors depending on whether we are using explicit or implicit solvent.

When importing from OpenMM `Topology` objects, box vectors should always mean the system is condensed phase, though we may subsequently ignore box vectors in applying GBSA parameters.

jchodera:
There could be yet another advantage to storing unit-bearing values as strings like `value*unit`: If the user writes in the file `"1.*angstrom`, would we want this to be written out preserving the original precision, or is `0.9999999999999997*angstrom` OK? These tokens start as strings in XML anyway, so future work could preserve the original strings and/or precision until they are changed so as to ensure reproducibility.

openforcefield/topology/topology.py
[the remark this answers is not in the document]
j-wags: I agree. Per our voice chat, I agree with you that we should avoid naming things in a way that could constrain our toolkit behavior in the future. I'm going to remove these properties, and just access `Topology.box_vectors` when needed.

andrrizzi:
> Probably humans for a little while, then primarily for computers later on in the project?

Sounds like a good plan to me.

j-wags:
@jchodera 

> If the user writes in the file "1.*angstrom, would we want this to be written out preserving the original precision, or is 0.9999999999999997*angstrom OK?

I'm not sure that this is related to how units values are stored. That is, even if `1.` and `angstrom` were stored separately, we'd still have this question. That said, I agree that we should make it a milestone goal to ensure that, if `1.` is read in, `1.` is written out, since that has a big effect on the human-friendliness of our files.

Let's also recall that we don't support text diff-ability in our files, because the same data can be encoded in different units, and properties of nodes can be written out in arbitrary order.

For now, I think we should stick with units and values separated. I'll open an issue to aggregate feedback on this during the alpha.

j-wags:
Opened #218 to gather feedback on the storage of unit-bearing quantities.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: @jchodera I think I've got it now -- refactored the logic in 0a55e67 if you'd like to check this out again.

The-SMIRNOFF-force-field-format.md
```diff
@@ -644,8 +644,8 @@ This is a backwards-incompatible overhaul of the SMIRNOFF 0.1 draft specificatio
     * `<BondChargeCorrections>` was renamed to `<ChargeIncrementModel>` and generalized to accommodate an arbitrary number of tagged atoms
     * `<GBSAForce>` was renamed to `<GBSA>`    
 * `<PeriodicTorsionForce>` was split into `<ProperTorsions>` and `<ImproperTorsions>`
-* `<vdW>` now specifies 1-2, 1-3, and 1-4 scaling factors via `scale12` (default: 0), `scale13` (default: 0), and `scale14` (default: 0.5) attributes. Coulomb scaling parameters have been removed from `StericsForce`.
-* Added the `<Electrostatics>` tag to separately specify 1-2, 1-3, and 1-4 scaling factors for electrostatics, as well as the method used to compute electrostatics (`PME`, `reaction-field`, `Coulomb`) since this has a huge effect on the energetics of the system.
+* `<vdW>` now specifies 1-2, 1-3, 1-4, and 1-5 scaling factors via `scale12` (default: 0), `scale13` (default: 0), and `scale14` (default: 0.5) attributes. Coulomb scaling parameters have been removed from `StericsForce`.
```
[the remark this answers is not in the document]
j-wags: I do. Good catch. Updated!

j-wags:
Thanks @davidlmobley 

Added your note to #178

Issue #216, opened 2019-03-28
Base class for TopologyMolecule/Bond/Atom and Molecule/Bond/Atom


andrrizzi:
If I understand the rationale for the current design, we currently keep distinct `topology.TopologyMolecule/Bond/Atom` and `molecule.Molecule/Bond/Atom` because we don't want to include all the information in `Molecule` in the `Topology` as well (e.g. partial charges and conformers). Nevertheless, the overlap between the two set of classes is quite large so I'm wondering if it makes sense to have a common set of base classes for both to make the interface uniform.

jchodera:
@andrrizzi : The distinction between `Molecule` and `TopologyMolecule` is
* `Molecule` stores all the information associated about a unique, real chemical molecular species
* `TopologyMolecule` just encapsulates information about the "image" of that real molecule mapped into a set of atoms in the `Topology` object

Their APIs should be very similar, with the `TopologyMolecule` providing additional information about the mapping of atoms in the `Topology` and the naming of that molecule in the topology, but many of the properties should be pass-through to `Molecule` (but immutable).

If there was a way to simplify the design by deriving from a common base class, I'm all for it!

We do want to keep in mind that the API for `Molecule` and `Topology` should be our new replacement for all the limitations of the OpenMM `Topology` molecule, so it will eventually need to be flexible enough to allow perses-style editing of molecules and topologies on the fly. Ideally, the level of abstract we pick for these objects will enable a wide variety of applications to be easily built on top of them, beyond just forcefield parameterization!

j-wags:
As we discussed earlier today, I think this would be a good idea in terms of making the API more consistent, but we'd have to be careful during implementation. Some queries, like `TopologyAtom.atomic_number` simply resolve to `self._atom.atomic_number`. However, things may get hairy when we do `TopologyAtom.bonds`, which requires mapping back to the reference atom, then collecting its bonds, then generating `TopologyBonds` for each one. It's not impossible, but it may be convoluted.

@andrrizzi also brought up a good point, which is that the current implementation tries to be as memory-efficient as possible, at the detriment of becoming more difficult to use. This memory efficiency may only be on the order of a factor of 2. The reasoning is, at the end of the day, we're going to generate an OpenMM system with N atoms in memory anyway, so how bad could it be to have _two_ data structures of N atoms?

jchodera:
We should definitely prioritize ease of use over memory efficiency!

andrrizzi:
[...]

Issue #185, opened 2019-02-20
Reactivate valence checks to raise exception when no SMARTS match is found


andrrizzi:
See also [this comment](https://github.com/openforcefield/openforcefield/issues/75#issuecomment-465211049) in #75.

andrrizzi:
Question: Do we want to raise an exception if `ForceField` has no parameters for some of the improper torsions?

davidlmobley:
@andrrizzi :
> Question: Do we want to raise an exception if ForceField has no parameters for some of the improper torsions?

Are you asking whether we should raise an exception if an improper specified in the force field is missing some parameters? My answer to that is "yes". 

But you might be asking, "Do we want to ensure that all groups which COULD be assigned an improper torsion parameter do indeed have one assigned?" If the latter is what you're asking, the answer is NO. Impropers are an exception where they are not necessarily employed in the force field. Historically they have only seen occasional use/use for only certain chemistry. We will likely be broadening use of them, but it is still not the case that everything which COULD be assigned an improper SHOULD be assigned one. Effectively impropers are "optional".

andrrizzi:
This was implemented in #217 .

PR #217, opened 2019-03-28
Reactivate atoms and bonds valence checks

andrrizzi:
This partially re-implement issue #185 . It reimplements valence checks for atoms and bonds only.

Reactivating valence checks for angles and torsions will require larger modifications, but I'd like to take a decision about #216 first in order to minimize the total time I'll spend on this, and implement it in a separate PR.

### DIFF
--- a/openforcefield/typing/engines/smirnoff/parameters.py
+++ b/openforcefield/typing/engines/smirnoff/parameters.py
@@ -870,7 +870,12 @@ class ParameterHandler(object):
         # the order of atom indices doesn't matter for comparison.
         valence_terms_dict = assigned_terms.__class__()
         for atoms in valence_terms:
-            atom_indices = (a.topology_particle_index for a in atoms)
+            try:
+                # valence_terms is a list of TopologyAtom tuples.
+                atom_indices = (a.topology_particle_index for a in atoms)
+            except TypeError:
+                # valence_terms is a list of TopologyAtom.
+                atom_indices = (atoms.topology_particle_index,)
             valence_terms_dict[atom_indices] = atoms
 
         # Check that both valence dictionaries have the same keys (i.e. terms).
@@ -1037,7 +1042,7 @@ class BondHandler(ParameterHandler):
             len(bonds) - skipped_constrained_bonds, skipped_constrained_bonds))
 
         # Check that no topological bonds are missing force parameters.
-        valence_terms = [(b.bond.atom1, b.bond.atom2) for b in topology.topology_bonds]
+        valence_terms = [list(b.atoms) for b in topology.topology_bonds]
         self._check_all_valence_terms_assigned(assigned_terms=bonds, valence_terms=valence_terms)
 
 
@@ -1455,8 +1460,9 @@ class vdWHandler(ParameterHandler):
             force.addParticle(0.0, 1.0, 0.0)
 
         # Set the particle Lennard-Jones terms.
-        for (atoms, ljtype) in atoms.items():
-            force.setParticleParameters(atoms[0], 0.0, ljtype.sigma,
+        for atom_key, ljtype in atoms.items():
+            atom_idx = atom_key[0]
+            force.setParticleParameters(atom_idx, 0.0, ljtype.sigma,
                                         ljtype.epsilon)
 
         # Check that no atoms are missing force parameters
--- a/openforcefield/typing/engines/smirnoff/parameters.py
+++ b/openforcefield/typing/engines/smirnoff/parameters.py
@@ -66,7 +66,27 @@ class IncompatibleParameterError(MessageException):
 
 
 class UnassignedValenceParameterException(Exception):
-    """Exception raised when there exist valence terms for which a ParameterHandler can't find parameters."""
+    """Exception raised when there are valence terms for which a ParameterHandler can't find parameters."""
+    pass
+
+
+class UnassignedBondParameterException(UnassignedValenceParameterException):
+    """Exception raised when there are bond terms for which a ParameterHandler can't find parameters."""
+    pass
+
+
+class UnassignedAngleParameterException(UnassignedValenceParameterException):
+    """Exception raised when there are angle terms for which a ParameterHandler can't find parameters."""
+    pass
+
+
+class UnassignedProperTorsionParameterException(UnassignedValenceParameterException):
+    """Exception raised when there are proper torsion terms for which a ParameterHandler can't find parameters."""
+    pass
+
+
+class UnassignedImroperTorsionParameterException(UnassignedValenceParameterException):
+    """Exception raised when there are improper torsions terms for which a ParameterHandler can't find parameters."""
     pass
 
 
@@ -860,7 +880,8 @@ class ParameterHandler(object):
     # -------------------------------
 
     @classmethod
-    def _check_all_valence_terms_assigned(cls, assigned_terms, valence_terms):
+    def _check_all_valence_terms_assigned(cls, assigned_terms, valence_terms,
+                                          exception_cls=UnassignedValenceParameterException):
         """Check that all valence terms have been assigned and print a user-friendly error message.
 
         Parameters
@@ -869,6 +890,9 @@ class ParameterHandler(object):
             Atom index tuples defining added valence terms.
         valence_terms : Iterable[TopologyAtom] or Iterable[Iterable[TopologyAtom]]
             Atom or atom tuples defining topological valence terms.
+        exception_cls : UnassignedValenceParameterException
+            A specific exception class to raise to allow catching only specific
+            types of errors.
 
         """
         # Convert the valence term to a valence dictionary to make sure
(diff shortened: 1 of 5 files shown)


andrrizzi:
I've also renamed `ParameterHandler.get_matches()` to `find_matches()` for consistency with `ToolkitWrapper.find_smarts_matches()` because I thought it was more representative of what it is actually doing, but I can revert that commit if you don't agree.

j-wags:
This pull request **fixes 1 alert** when merging 6fbe300b40d8187e41b245a7323c24ce8afc406e into 8ea7a09e9fdba58a0e2567da7be37e94a675aa7c - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/openforcefield/rev/pr-8e20ee96d4d296ae3e5bebb3aa1e8b907be668df)

**fixed alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -825,6 +849,58 @@ def to_dict(self, output_units=None, return_cosmetic_attributes=False):
...
+                        "- {unassigned_str}").format(parameter_handler=cls.__name__,
+                                                     unassigned_str='\n- '.join(unassigned_terms))
+        if len(not_found_terms) > 0:
+            if err_msg != "":
+                err_msg += '\n'
+            err_msg += ("{parameter_handler} assigned terms that were not found in the topology:\n"
+                        "- {not_found_str}").format(parameter_handler=cls.__name__,
+                                                    not_found_str='\n- '.join(not_found_terms))
+        if err_msg != "":
+            raise RuntimeError(err_msg)
```
jchodera: I don't think we want to raise `RuntimeError`, since these are not going to be informative about what was not matched. Instead, why don't we define our own exception hierarchy (e.g. `UnassignedParameterException > UnassignedValenceParameterException > UnassignedBondParameterException`) so that applications or tests could catch these exceptions and proceed with execution if intended. (For example, someone might be mixing a force field and NN potential outside of SMIRNOFF.)

Also, let's be sure the docstrings document exceptions we intentionally throw under different circumstances.

andrrizzi: Will do, thanks!

andrrizzi:
After conversation in #216, I've decided to implement `Topology.angles/propers/impropers` properties for now returning tuples of `Atom`s, similar to what's currently implemented in `Molecule`. This check will be important to include in the release. In the future, we may want to have `Angle/Proper/Improper` classes similar to `Bond` to make things a little more consistent.

jchodera:
Please remember to add a note to the docstring for any methods or properties where the API may not be final!

andrrizzi:
@j-wags can correct me if I'm wrong, but I'm not sure we're at that level yet, meaning that I think we're planning to generally label the API as unstable for the upcoming release.

I think it's still hard to make guarantees about what's not going to change signature in the next versions except maybe for a few methods such as `ForceField.create_openmm_system()` and `Molecule.from_anything()`, which are needed for the most common use cases of the toolkit. We're planning to take a systematic pass at the API in the short future though.

andrrizzi:
Reactivating the valence terms caused some tests to fail. Most of them complain about improper torsions that are not parametrized, but also some proper torsions as well. I'll check a few specific cases to see if this is a bug or something that we want to ignore.

j-wags:
This pull request **introduces 6 alerts** and **fixes 1** when merging 426e345e0843623a3f54082cc5696782eb99325c into 8ea7a09e9fdba58a0e2567da7be37e94a675aa7c - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/openforcefield/rev/pr-4b77780cc32ef30be2dfcf2ab0eed6686992de73)

**new alerts:**

* 6 for Property in old\-style class

**fixed alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

j-wags:
> I think we're planning to generally label the API as unstable for the upcoming release.

Right. I get the sense that we're applying "unstable API" warnings to 80% of functions right now, with no clear pathway to un-labeling them as such. If we define a process for removing the label, then it makes more sense.

Before I started typing this, I had probably agreed with @andrrizzi -- Everything will be labeled as "unstable", and then the day we do the stable release, all the labels are removed. However, maybe it would make more sense to have a meeting after the initial release, and select different chunks of the API to "stabilize" as part of different milestones. This way, we could have focused discussions about individual areas of the API and remove the "unstable" labels one at a time.

tl;dr - Let's apply "unstable API" labels to individual functions. Later, we can come through and remove them one-by-one as we "stabilize" different areas.

davidlmobley:
Impropers not parameterized is OK -- impropers are optional, essentially. But propers not parameterized is concerning.

j-wags:
This pull request **introduces 6 alerts** and **fixes 1** when merging 20ea3c02ec3c3df2ad75ddec6ac0ad4f8ad2860e into 8ea7a09e9fdba58a0e2567da7be37e94a675aa7c - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/openforcefield/rev/pr-fa87eb0d2e3fb74564a580d3011bf0cc412d5ebd)

**new alerts:**

* 6 for Property in old\-style class

**fixed alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

j-wags:
This pull request **introduces 6 alerts** and **fixes 1** when merging a21ae75248c1fd042c22faac8f3e32584d7c0498 into 8ea7a09e9fdba58a0e2567da7be37e94a675aa7c - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/openforcefield/rev/pr-a38a64951746f4589c01973fb9be96e095ec6dda)

**new alerts:**

* 6 for Property in old\-style class

**fixed alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

andrrizzi:
Thanks for the fix, @j-wags! Good to know that our tests caught that. This is ready for merging on my end.

j-wags:
This pull request **introduces 6 alerts** and **fixes 1** when merging 826716bf21f7716414834cc181726b8787b7928c into 8ea7a09e9fdba58a0e2567da7be37e94a675aa7c - [view on LGTM.com](https://lgtm.com/projects/g/openforcefield/openforcefield/rev/pr-6939c20d3c85c391f95e6f22ad5ddc4445ac46eb)

**new alerts:**

* 6 for Property in old\-style class

**fixed alerts:**

* 1 for Unused import

---

*Comment posted by [LGTM.com](https://lgtm.com)*

jchodera:
If we are not certain that most of the API will be stable at release, we really should NOT be calling this a 1.0 alpha release, since it's absurd to have to go through some crazy gymnastics of a few dozen alpha bugfix, feature, and API breaking releases and then switch to semantic versioning. Let's stick with 0.y.z releases until we feel that much of the API is stable, and not worry about whether or not the April release is "1.0 anything". It's just the first release with RDKit support and a more coherent object model for the parameter set, but it can be 0.y.z. Feature and API breaking releases can increment y, and bugfixes can increment z. If we release 1.0.0 in year one, I think we will still be in great shape.

In addition, we want to label the docstrings of classes and functions that have unstable APIs with standard warnings. We can remove those as parts of the API stabilize.

TensorFlow has been in a similar situation for a while, and we are a much smaller operation, so there is nothing to be ashamed about here. We just need to be very clear in how we communicate this to our users

j-wags:
I agree with @jchodera that the upcoming release should be 0.2.0, and that we can build toward 1.0.0. It's OK if the majority of the API is labeled as unstable for 0.2.0, but we should have a plan to collect API feedback after the initial release, and have a schedule that will have us finalizing it in manageable-sized chunks.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1387,22 +1476,21 @@ def create_force(self, system, topology, **kwargs):
...
         # Create all particles.
         for particle in topology.topology_particles:
             force.addParticle(0.0, 1.0, 0.0)
 
         # Set the particle Lennard-Jones terms.
-        for (atoms, ljtype) in atoms.items():
-            force.setParticleParameters(atoms[0], 0.0, ljtype.sigma,
+        for atom_key, ljtype in atoms.items():
+            atom_idx = atom_key[0]
+            force.setParticleParameters(atom_idx, 0.0, ljtype.sigma,
```
j-wags: Nice. This code is much clearer.

jchodera:
Sounds like a great plan, @j-wags.

We should also be sure to craft a release announcement for the News page that lays out some details of where we are, what feedback we'd like (and how), and where we're going (and how to check the Milestones).

andrrizzi:
>  we really should NOT be calling this a 1.0 alpha release

Absolutely agree.

Thank you all for the reviews!

Issue #222, opened 2019-04-02
ForceField.create_openmm_system(topology) should read the box vectors from topology


andrrizzi:
Currently, they are ignored so if `create_openmm_system(..., default_box_vectors=None)` the resulting `System` object will have arbitrary box vectors of 2nm per side.

Relatedly, now that the OFF `Topology` comes with box vectors, do we still want to keep the `default_box_vectors` argument to overwrite them? Or do we want the user to set `topology.box_vectors` before passing it to `create_openmm_system()`?

j-wags:
Oh, totally my fault. I should have removed that. The user should now set `Topology.box_vectors` before system creation.

j-wags:
Fixed in #225

PR #232, opened 2019-04-05
Cleanup SMIRNOFF_simulation example

andrrizzi:
This PR should have Travis tests passing. It might require a few tweakings to the doctests still, but the example tests should run fine.


examples/conversion_amber_gromacs/convert_to_amber_gromacs.ipynb
```diff
@@ -167,14 +194,14 @@
       "-------------------------------\n",
       "{'HarmonicAngleForce': Quantity(value=6.823125839233398, unit=kilojoule/mole),\n",
       " 'HarmonicBondForce': Quantity(value=1.4782228469848633, unit=kilojoule/mole),\n",
-      " 'NonbondedForce': Quantity(value=5127.504921744658, unit=kilojoule/mole),\n",
+      " 'NonbondedForce': Quantity(value=4.263674188305515, unit=kilojoule/mole),\n",
```
j-wags: Any thoughts as to why this changed so much?

j-wags: (It might be on my end -- Nonbonded handling wasn't implemented in a sane way until just a few days ago)

andrrizzi: Good catch! Let me investigate a little.

andrrizzi: Ah, nothing to worry about. The high energy was before the fix for #222 so probably the box vectors were too tight and the periodic distance between the two molecules was much smaller.

j-wags: Great. Thanks for checking this out!

Issue #247, opened 2019-04-08
Is "!1..." a typo in the smirnoff99Frostt


drmeister:
https://github.com/openforcefield/openforcefield/blob/master/openforcefield/data/forcefield/smirnoff99Frosst.offxml#L196

The first atom test seems a bit odd - it looks like it should read "[!#1:1]..." - otherwise it will miss other isotopes of hydrogen.

<Proper smirks="[!1:1]-[#7X4,#7X3:2]-[#6X4;r3:3]-[*:4]" id="t56" idivf1="1" k1="0.156" periodicity1="3" phase1="0.0"/>

andrrizzi:
Thanks for the report! This is probably a good question for @bannanc or @davidlmobley .

andrrizzi:
It looks like `smirnoff99Frosst.offxml` file [here](https://github.com/openforcefield/smirnoff99Frosst/blob/master/smirnoff99Frosst/smirnoff99Frosst.offxml#L196) is actually `"[#1:1]"`, not `[!1:1]` so I think it is a typo and we should correct it soon. Thanks for catching it!

This is related to #117.

andrrizzi:
Ah, no wait. For some reason the line number differs in the file we have here and in the smirnoff99Frosst repo. The parameter is the same over there as well (see [here](https://github.com/openforcefield/smirnoff99Frosst/blob/master/smirnoff99Frosst/smirnoff99Frosst.offxml#L198)), but if I'm not mistaken, the two files are supposed to be identical so this requires further investigation.

bannanc:
@andrrizzi I think the point of this issue is just that `[!1:1]` should be `[!#1:1]`. Since that parameter shouldn't be different in the file here or smirnoff99frosst the point of the issue is that we need to fix the typo...

I'm not sure what that has to do with issue #117

andrrizzi:
False alarm. That was probably a browser cache issue. Sorry for the spam.

andrrizzi:
> I'm not sure what that has to do with issue #117

I thought the two files got out of sync, and installing the smirnoff99Frosst force field through the plugin architecture would avoid issues like this.

andrrizzi:
[...]

PR #268, opened 2019-04-12
Fix `!1` typo in smirnoff99Frosst

bannanc:
As discussed in issue #247 I fixed `!1` to `!#1` in parameter `t56`. 
In doing this I followed the same procedure I usually would, I changed the relevant files in the `utilities/convert_frosst/` folder for the posterity of the forcefield and then I changed `openforcefield/data/forcefield/smirnoff99frosst.offxml`. 

In this process I noticed that the `offxml` file in the `convert_frosst` folder and the one in the smirnoff99Frosst repo have smirnoff spec version 0.1.0 in the file where the one in the data/forcefield/ folder has spec 0.2.0.

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -7,8 +7,19 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
 * ``minor`` increments add features but do not break API compatibility
 * ``micro`` increments represent bugfix releases or improvements in documentation
 
-0.2.0
------
+0.2.1 - Bugfix release
+----------------------
+
+This release features various documentation fixes, minor bugfixes, and code cleanup.
+
+Bugfixes
+""""""""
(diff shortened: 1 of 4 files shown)


jchodera:
Let me push an update to the changelog.

jchodera:
I've updated the changelog with this PR, as well as the other bugfix PRs we've already merged.

j-wags:
I'm concerned that merging this will add another chapter to our already-substantial mutiple-files-with-the-same-name-but-different-contents saga. Could we put off this merge until our versioning questions are resolved? Or is that a longer-term topic?

davidlmobley:
@j-wags seems like it depends on your idea of timescale for resolving the versioning/naming issues. This is a minor fix but should get put in place soon and won't affect anything significantly.

Also, it seems like the files Caitlin noted need to be updated ot the new spec. Should we deal with that in this same PR or do you want to deal with it in a separate PR?

jchodera:
@j-wags: Since we're migrating the smirnoff99Frosst force field to another repo, this PR can be closed and the branch deleted.

j-wags:
Agreed. Closing.

PR #269, opened 2019-04-12
Update releasehistory for 0.2.1 release

j-wags:


docs/releasehistory.rst
```diff
@@ -7,8 +7,19 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
...
-0.2.0
------
+0.2.1 - Bugfix release
+----------------------
+
+This release features various documentation fixes, minor bugfixes, and code cleanup.
+
+Bugfixes
+""""""""
+* [(PR #267)](https://github.com/openforcefield/openforcefield/pull/267): Add neglected `<ToolkitAM1BCC>` to the SMIRNOFF 0.2 spec
```
andrrizzi: This change log will be pushed in the docs, so I think we'll have to use Sphinx-style links here. Something like:
```
[PR `#267 <https://github.com/openforcefield/openforcefield/pull/267>`_]
```

j-wags: Good catch. Fixed in 3e243ff

jchodera: Whoops, my bad!

PR #271, opened 2019-04-12
Fix readthedocs build error

jchodera:
Debugging a readthedocs build error:
```
Solving environment: ...working... failed

UnsatisfiableError: The following specifications were found to be in conflict:
  - cairo==1.16
  - python==3.6
  - rdkit
Use "conda info <package>" to see the dependencies for each package.
```


jchodera:
I don't know _why_, but this does seem to fix the readthedocs issue.

PR #273, opened 2019-04-12
Fix formatting error in changelog

jchodera:
Now we're to the point of fixing the documentation errors in documenting the fixing of documentation errors...


andrrizzi:
I'll fix the conflicts before merging this.

Issue #237, opened 2019-04-07
Update travis build system to new MolSSI cookiecutter


j-wags:
## ! All tests are now being run with RDKit !

We've added `rdkit` to the `meta.yaml`, which is currently being used during travis builds. Moving forward, the package will always have RDKit+Ambertools.

This means that we need to rethink our testing matrix, which currently tries to do builds with and without RDKit and OpenEye.

This could be a good opportunity to update the build system to the new cookiecutter format, which would let us use totally different `yaml`s for testing configurations.

jchodera:
For now, a simple fix would just be to force uninstall of RDKit on test branches that shouldn't use it.

That is, change
```
  # Install RDKit
  - if [[ "$RDKIT" == true ]]; then conda install --yes -c rdkit rdkit; fi
  - if [[ "$RDKIT" == true ]]; then conda install --yes -c omnia ambertools; fi
```
to
```
  # Install RDKit
  - if [[ "$RDKIT" != true ]]; then conda remove --force --yes rdkit; fi
  - if [[ "$RDKIT" != true ]]; then conda remove --force --yes ambermini; fi
```

j-wags:
We should also move `oe_license.txt.enc` out of the top-level directory of the repo. It's confusing GitHub's license analysis bot and screwing up our `LICENSE` badge

PR #280, opened 2019-04-18
Fix #279: __init__ prints warning if OpenEye is not installed

### LINKED ISSUE #279: Update __init__ file to not raise exception
I apologize if this is a duplicate of something you already have documented @j-wags. The main [`__init__`](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/__init__.py) file in openforcefield looks like it still raises an exception if OpenEye toolkits aren't installed. It also has a few commented out lines at the bottom that seem like they should be uncommented or removed.

andrrizzi:
Fix #279.

Our tests were not catching this because we were installing OpenEye even in RDKIT-only tests. I'll wait to see the Travis tests fail before removing the exception in the main `__init__`.


andrrizzi:
Ah! We're actually catching the exception in `__init__` and just printing a warning so the tests shouldn't fail anyway.

andrrizzi:
It looks like some of our examples require OpenEye. I'll try to remove that dependency in here.

.travis.yml
```diff
@@ -81,7 +81,7 @@ install:
   - pip install codecov
   - pip install coverage==4.4
   # TODO: Go back to conda install once OpenEye is released for python 3.7 on conda-forge.
-  - pip install -i https://pypi.anaconda.org/openeye/simple openeye-toolkits
+  - if [[ "$OPENEYE" == true ]]; then pip install -i https://pypi.anaconda.org/openeye/simple openeye-toolkits; fi
```
j-wags: Oh wow. Great catch there. This is... baffling.

examples/inspect_assigned_parameters/inspect_assigned_parameters.ipynb
```diff
@@ -444,11 +444,22 @@
     "from openforcefield.typing.engines.smirnoff import ForceField\n",
     "from openforcefield.utils.structure import get_molecule_parameterIDs\n",
     "\n",
-    "# Retrieve molecules from the AlkEthOH test set\n",
-    "mols = Molecule.from_file(get_data_filename('molecules/AlkEthOH_test_filt1_tripos.mol2'))\n",
-    "# The test set is quite large, so to limit the size of the output we only \n",
-    "# analyze the first 10 molecules\n",
-    "mols = mols[:10]\n",
+    "# The set of molecules that we want to parametrize in SMILES format.\n",
```
j-wags: Oh, cool. That's a much better way to get these.

openforcefield/__init__.py
```diff
@@ -1,13 +1,3 @@
-try:
-    import openeye
-except Exception as e:
-    print(e)
-    print('Warning: Cannot import openeye toolkit; not all functionality will be available.')
```
j-wags: Agree that removing this should be fine. something in `ToolkitWrapper` creates a warning about this anyway, so two "missing OpenEye" messages were being printed before this change.

.travis.yml
```diff
@@ -70,7 +70,7 @@ install:
 
   # TODO: Do we still need the beta version or can we do simply -c openeye openeye-toolkits?
   # Use beta version for partial bond orders
-  # TODO: Reactivate conda install instead of pip install once OpenEye is released for python 3.7 on conda-forge.
+  # TODO: Reactivate conda install instead of pip install once OpenEye is released for python 3.7 on conda-forge, but I don't think we need oemmtools anymore.
```
j-wags: Yeah, it's good to keep these notes here. This will be useful when we rebuild our testing system up according to the latest cookiecutter in the near future (#237)

.travis.yml
andrrizzi: Oh, that was totally my fault. Sorry about that. I added the pip installation while trying to make tests pass for Python 3.7, for which OpenEye didn't/doesn't have a conda install. It's great to see that RDKit tests pass anyway.

andrrizzi:
Thanks for the review, @j-wags !

PR #281, opened 2019-04-18
Implementing support for partial charge I/O in SDF 

### LINKED ISSUE #250: Add support for loading partial charges from SDF tags using new standard
The OE and RDK maintainers have agreed on a spec for storing partial charges in SDF. We should implement support for I/O of this data.

### LINKED ISSUE #524: to/from_openeye can not distinguish None vs. all-zero partial charges
### Background
The specifications for each package's handling of partial charges is as follows:

* OpenFF Toolkit: `mol.partial_charges` can be `None` or a Quantity-wrapped numpy array of floats
* OpenEye: Each OEAtom has a `Get`/`SetPartialCharge` method, which accepts floats. Partial charges are initialized to `0.0` and can not be `None`.
* RDKit: Each RDAtom can additional SD data using `rdk_atom.SetDoubleProp('PartialCharge', value)`
* SDF spec we're adopting (#250): A `<atom.dprop.PartialCharges>` tag can be included, with a space-separated list of doubles (and maybe special `n/a` values to indicate that some atoms dont have charges). 

The OpenFF partial charge spec has the following equivalents in these representations:
* if `offmol.partial_charges is None`
    * OpenEye: ~**No equivalent**~ (Per @SimonBoothroyd's comment below, this can handle `float('nan')`
    * RDKit: RDAtoms don't have `PartialCharges` set
    * SDF: No `<atom.dprop.PartialCharges>` tag is included
* if `offmol.partial_charges is not None`:
    * OpenEye: OEAtoms have the appropriate `PartialCharges` set
    * RDKit: RDAtoms have the appropriate `PartialCharges` set
    * SDF: The `<atom.dprop.PartialCharges>` tag is included, with no values of `N/A`

### Problem
OFFMol roundtrips to OEMols can't preserve whether partial charges were None or all-zero.

### Current status
The above roundtrip will return all-zero `partial_charges` from OpenEye, even if the original OFFMol had `partial_charges` is `None`. 

In #281 and previous work, I've implemented tests for the above roundtrips. All of them can successfully roundtrip `None`, except OpenEye, which returns all zeroes.

(For SDF roundtrips using OpenEye, I implemented special logic to query for the presence or absence of the `<atom.dprop.PartialCharges>` tag in the SDF, and so `None` partial charges can actually survive that roundtrip.)

### Options for moving forward

We can not proceed forever with equating "all zero partial charges" and "`partial_charges is None`". `None` and "all-zeroes" have different meanings for OFFTK, and seemingly-reasonable assumptions about their equivalence/meaning have caused us to produce surprising behavior in situations like #381 and barriers in implementation in #488. Further, leaving the distinction ambiguous will cause serious headaches the day someone tries to put truly zero-partial-charge mols like H2, N2, O2, noble gas, or buckyballs through OFFTK. 

The best option I can think of quickly is a "nasty hack", such as setting a special SD tag on the OEMol that comes out of `offmol.to_openeye` that distinguishes whether `partial_charges` is `None` or really all-zero. 

**I would really like to hear if anyone can think of a less hacky solution!**

### Timeframe
This behavior has always been broken in OFFTK, so it's not critical that we come to a conclusion urgently.

j-wags:
- [x] Closes #250 by adding support for partial charge I/O in SDF. The partial charges are stored as a property in the SDF molecule block under the tag `<atom.dprop.PartialCharge>`
- [x] Closes #524
- [x] If an OFFMol's ``partial_charges`` attribute is set to ``None`` (the default value), calling ``to_openeye`` will now produce a OE molecule with partial charges set to ``nan``. This would previously produce an OE molecule with partial charges of 0.0, which was a loss of information, since it wouldn't be clear whether the original OFFMol's partial charges were all-zero or ``None``. OpenEye toolkit wrapper methods such as ``from_smiles`` and ``from_file`` now produce OFFMols with ``partial_charges = None`` when appropriate (previously these would produce OFFMols with all-zero charges).
- [x] Per the new SDF partial charge specification adopted by RDKit, ``Molecule.to_rdkit`` now sets partial charges on the RDAtom's ``PartialCharges`` property (this was previously set on the ``partial_charges`` property). If the OFFMol's partial_charges attribute is None, this property will not be defined.
- [x] Enforce the behavior during SDF I/O that a SDF may contain multiple MOLECULES, but that the OFF Toolkit will NEVER assume that it contains multiple CONFORMERS of the same molecule. This is an important distinction, since otherwise there is ambiguity around whether properties of one entry in a SDF are shared among several molecule blocks or not (More info [here](https://docs.eyesopen.com/toolkits/python/oechemtk/oemol.html#dude-where-s-my-sd-data)). If the user requests the OFF Toolkit to write a multi-conformer ``Molecule`` to SDF, only the first conformer will be written. For more fine-grained control of writing properties, conformers, and partial charges, users will need to call ``Molecule.to_rdkit`` or ``Molecule.to_openeye`` and use the flexibility offered by those packages.
- [x] Due to different constraints placed on the data types allowed by external toolkits, we make our best effort to preserve ``offmol.properties`` when converting molecules to other packages, but users should be aware that no guarantee of data integrity is made. The only data format for keys and values in the `offmol.property` dict that we will try to support through a roundtrip to another toolkit's Molecule object is ``string``.
- [x] Adds tests
- [x] Updates release notes

Status
- [x] Ready for review
- [ ] Ready for merge


openforcefield/tests/test_toolkits.py
```diff
@@ -191,8 +192,8 @@ def test_to_from_openeye_core_props_unset(self):
         assert central_carbon_stereo_specified
 
         # Do a first conversion to/from oemol
-        rdmol = molecule.to_openeye()
-        molecule2 = Molecule.from_openeye(rdmol)
+        oemol = molecule.to_openeye()
+        molecule2 = Molecule.from_openeye(oemol)
```
j-wags: This was a copy-paste typo previously

openforcefield/tests/test_toolkits.py
```diff
@@ -831,7 +987,7 @@ def test_to_from_rdkit_core_props_unset(self):
                 central_carbon_stereo_specified = True
         assert central_carbon_stereo_specified
 
-        # Do a first conversion to/from oemol
+        # Do a first conversion to/from rdmol
```
j-wags: This was a copy-paste typo

openforcefield/topology/molecule.py
```diff
@@ -2774,15 +2774,20 @@ def partial_charges(self, charges):
...
-        charges : a simtk.unit.Quantity - wrapped numpy array [1 x n_atoms]
-            The partial charges to assign to the molecule. Must be in units compatible with simtk.unit.elementary_charge
+        charges : None or a simtk.unit.Quantity - wrapped numpy array [1 x n_atoms]
+            The partial charges to assign to the molecule. If not None, must be in units compatible with simtk.unit.elementary_charge
+
         """
-        assert hasattr(charges, 'unit')
-        assert unit.elementary_charge.is_compatible(charges.unit)
-        assert charges.shape == (self.n_atoms, )
+        if charges is None:
```
j-wags: Previously, we couldn't set partial charges to `None`

openforcefield/tests/test_toolkits.py
```diff
@@ -282,15 +284,143 @@ def test_get_sdf_coordinates(self):
...
-        assert molecule._conformers[0].shape == (15,3)
+        filename = get_data_file_path('molecules/methane_multiconformer.sdf')
+        molecules = Molecule.from_file(filename, toolkit_registry=toolkit_wrapper)
+        assert len(molecules) == 2
+        assert len(molecules[0]._conformers) == 1
+        assert len(molecules[1]._conformers) == 1
+        assert molecules[0]._conformers[0].shape == (5, 3)
+
+    @pytest.mark.skipif(not OpenEyeToolkitWrapper.is_available(), reason='OpenEye Toolkit not available')
+    def test_load_multiconformer_sdf_as_separate_molecules_properties(self):
```
jthorton: Nice test coverage here

openforcefield/topology/molecule.py
```diff
@@ -3661,7 +3665,7 @@ def from_qcschema(cls, qca_record, client=None, toolkit_registry=GLOBAL_TOOLKIT_
         try:
             mapped_smiles = qca_record['attributes']['canonical_isomeric_explicit_hydrogen_mapped_smiles']
         except KeyError:
-            raise KeyError('The record must contain the hydrogen mapped smiles to be safley made from the archive.')
+            raise KeyError('The record must contain the hydrogen mapped smiles to be safely made from the archive.')
```
jthorton: Thanks!

openforcefield/utils/toolkits.py
```diff
@@ -465,20 +465,12 @@ def to_file_obj(self, molecule, file_obj, file_format):
...
         # TODO: This is inefficiently implemented. Is there any way to attach a file-like object to an oemolstream?
         with temporary_directory() as tmpdir:
             with temporary_cd(tmpdir):
                 outfile = 'temp_molecule.' + file_format
-                ofs = oechem.oemolostream(outfile)
-                openeye_format = getattr(oechem, 'OEFormat_' + file_format)
-                ofs.SetFormat(openeye_format)
-                oechem.OEWriteMolecule(ofs, oemol)
-                ofs.close()
+                self.to_file(molecule, outfile, file_format)
```
jthorton: Nice, I like that we are using one function for both files and file objects now.

openforcefield/utils/toolkits.py
```diff
@@ -501,9 +493,66 @@ def to_file(self, molecule, file_path, file_format):
         ofs = oechem.oemolostream(file_path)
         openeye_format = getattr(oechem, 'OEFormat_' + file_format)
         ofs.SetFormat(openeye_format)
+
+        # OFFTK strictly treats SDF as a single-conformer format.
+        # We need to override OETK's behavior here if the user is saving a multiconformer molecule.
```
jthorton: Slightly confused with this comment as it sounds like we can save multiconformer molecule into the sdf but we throw away all but the first conformer, I think in the spec you say we only write the first conformer to file as well. So maybe change this to something like 
# remove all but the first conformer when writing to SDF as we only support single conformer format?

openforcefield/utils/toolkits.py
```diff
@@ -535,12 +583,55 @@ def _read_oemolistream_molecules(cls, oemolistream, allow_undefined_stereo, file
...
+                # The logic below should handle either single- or multi-conformer SDFs.
+                for conf in oemol.GetConfIter():
+                    # First, we turn "conf" into an OEMCMol (OE multiconformer mol), since OTHER file formats
+                    # really are multiconformer, and we will eventually feed this into the `from_openeye` function,
+                    # which is made to ingest multiconformer mols.
+                    this_conf_oemcmol = conf.GetMCMol()
+
+                    # Then, we take any SD data pairs that were on the oemol, and copy them on to "this_conf_oemcmol".
+                    # These SD pairs will be populated if we're dealing with a single-conformer SDF.
+                    for dp in oechem.OEGetSDDataPairs(oemol):
```
jthorton: Just to make sure I follow this bit we are saying if it is a single conformer SDF the tags are in the `oemol` and will be transferred in the first loop else if its a multi conformer SDF the first loop will have no tags as they are instead in the `conf` molecule?

openforcefield/utils/toolkits.py
```diff
@@ -793,7 +880,9 @@ def describe_oeatom(oeatom):
             map_atoms[
                 oe_idx] = atom_index  # store for mapping oeatom to molecule atom indices below
             atom_mapping[atom_index] = map_id
-        molecule._properties['atom_map'] = atom_mapping
+
+        if 0 not in atom_mapping.values():
```
jthorton: Thanks for the fix to the atom mapping.

openforcefield/utils/toolkits.py
```diff
@@ -894,6 +992,7 @@ def to_openeye(molecule, aromaticity_model=DEFAULT_AROMATICITY_MODEL):
             oeatom.SetFormalCharge(atom.formal_charge)
             oeatom.SetAromatic(atom.is_aromatic)
             oeatom.SetData('name', atom.name)
+            oeatom.SetPartialCharge(float('nan'))
```
jthorton: Nice!

openforcefield/utils/toolkits.py
jthorton: Ohh I think I misunderstood what you mean, so we can write an SDF with multiple molecules in where each molecule could be a conformer of the same molecule, so in a round trip test we won't quite get the input molecule back as a molecule with N conformers but we get back N molecules that the user can then condense if they want to?

jthorton:
Overall this looks great with really good test coverage. I was slightly confused following the spec and implementation of what would happen when writing a multi conformer molecule to SDF but I think I follow now. Nothing blocking.

openforcefield/utils/toolkits.py
j-wags: Exactly. It's super confusing, but this logic is intended to handle both scenarios.

openforcefield/utils/toolkits.py
j-wags: Good call -- I've added that comment.

j-wags: Currently, we can NOT write an SDF with multiple conformers/molecules. Though, users are free to concatenate several single-molecule SDFs to get a "multi-conformer" or "multi-molecule" SDF.

openforcefield/utils/toolkits.py
```diff
@@ -501,9 +493,67 @@ def to_file(self, molecule, file_path, file_format):
...
+
+        # OFFTK strictly treats SDF as a single-conformer format.
+        # We need to override OETK's behavior here if the user is saving a multiconformer molecule.
+
+        # Remove all but the first conformer when writing to SDF as we only support single conformer format
+        if (file_format.lower() == "sdf") and oemol.NumConfs() > 1:
+            conf1 = [conf for conf in oemol.GetConfs()][0]
+            flat_coords = list()
+            for idx, coord in conf1.GetCoords().items():
+                flat_coords.extend(coord)
```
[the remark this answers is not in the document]
j-wags: Now THAT is a cool trick. Thanks!

openforcefield/utils/toolkits.py
```diff
@@ -501,9 +493,67 @@ def to_file(self, molecule, file_path, file_format):
...
+            oemol.DeleteConfs()
+            oecoords = oechem.OEFloatArray(flat_coords)
+            oemol.NewConf(oecoords)
+        # We're standardizing on putting partial charges into SDFs under the `atom.dprop.PartialCharge` property
+        if (file_format.lower() == "sdf") and (molecule.partial_charges is not None):
+            partial_charges_list = []
+            for oeatom in oemol.GetAtoms():
+                partial_charges_list.append(f'{oeatom.GetPartialCharge():f}')
+            partial_charges_str = ' '.join(partial_charges_list)
```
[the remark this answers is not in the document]
j-wags: Not particularly, and I think your solution is cleaner. Thanks!

openforcefield/tests/test_toolkits.py
```diff
@@ -337,6 +503,36 @@ def test_get_mol2_charges(self):
...
+    @pytest.mark.skipif(not OpenEyeToolkitWrapper.is_available(), reason='OpenEye Toolkit not available')
+    def test_mol2_charges_roundtrip(self):
+        """Test OpenEyeToolkitWrapper for performing a round trip of a molecule with partial charge to and from
+        a mol2 file"""
+        from openforcefield.tests.test_forcefield import create_ethanol
+        toolkit_wrapper = OpenEyeToolkitWrapper()
+        ethanol = create_ethanol()
+        # The value of ethanol.partial_charges[4] is 1e-5, but mol2 is only written to 4 digits of precision, so we
+        # multiply by ten to make this test pass within the precision of the mol2 specification
+        ethanol.partial_charges[4] *= 10
```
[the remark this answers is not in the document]
j-wags: Used your second suggestion: Added in 20e7bca

openforcefield/tests/test_toolkits.py
```diff
@@ -282,15 +320,143 @@ def test_get_sdf_coordinates(self):
...
+        sdf_text = sio.getvalue()
+        # In our current configuration, if the OFFMol doesn't have partial charges, we DO NOT want a partial charge
+        # block to be written. For reference, it's possible to indicate that a partial charge is not known by writing
+        # out "n/a" (or another placeholder) in the partial charge block atoms without charges.
+        assert '<atom.dprop.PartialCharge>' not in sdf_text
+
+    @pytest.mark.skipif(not OpenEyeToolkitWrapper.is_available(), reason='OpenEye Toolkit not available')
+    def test_sdf_properties_roundtrip(self):
+        """Test OpenEyeToolkitWrapper for performing a round trip of a molecule with partial charge to and from
+        a sdf file"""
```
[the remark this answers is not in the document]
j-wags: Good thinking. Updated in 9f6b550.

openforcefield/tests/test_toolkits.py
```diff
@@ -282,15 +320,143 @@ def test_get_sdf_coordinates(self):
...
+        # The second molecule in the SDF has the following properties and charges:
+        assert molecules[1].properties['test_property_key'] == 'test_property_value2'
+        assert molecules[1].properties['another_test_property_key'] == 'another_test_property_value'
+        np.testing.assert_allclose(molecules[1].partial_charges / unit.elementary_charge,
+                                   [0.027170, 0.027170, 0.027170, 0.027170, -0.108680])
+
+    @pytest.mark.skipif(not OpenEyeToolkitWrapper.is_available(), reason='OpenEye Toolkit not available')
+    def test_write_sdf_charges(self):
+        """Test OpenEyeToolkitWrapper for writing partial charges to a sdf file"""
+        from openforcefield.tests.test_forcefield import create_ethanol
```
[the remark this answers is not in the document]
j-wags: Good catch. Removed in b26f19e.

openforcefield/tests/test_toolkits.py
```diff
@@ -282,15 +320,143 @@ def test_get_sdf_coordinates(self):
...
-        filename = get_data_file_path('molecules/toluene.sdf')
-        molecule = Molecule.from_file(filename, toolkit_registry=toolkit_wrapper)
-        assert len(molecule._conformers) == 1
-        assert molecule._conformers[0].shape == (15,3)
+        filename = get_data_file_path('molecules/methane_multiconformer.sdf')
+        molecules = Molecule.from_file(filename, toolkit_registry=toolkit_wrapper)
+        assert len(molecules) == 2
+        assert len(molecules[0]._conformers) == 1
+        assert len(molecules[1]._conformers) == 1
+        assert molecules[0]._conformers[0].shape == (5, 3)
```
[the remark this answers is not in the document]
j-wags: I think that just got cargo-culted around from the first time I used it incorrectly. Removed throughout toolkit tests in b26f19e.

openforcefield/tests/test_toolkits.py
```diff
@@ -282,15 +320,143 @@ def test_get_sdf_coordinates(self):
...
+        # Make sure that a charge line was ever found
+        assert charge_line_found == True
+
+        # Make sure that the charges found were correct
+        assert_almost_equal(charges, [-0.4, -0.3, -0.2, -0.1, 0.00001, 0.1, 0.2, 0.3, 0.4])
+
+
+    @pytest.mark.skipif(not OpenEyeToolkitWrapper.is_available(), reason='OpenEye Toolkit not available')
+    def test_write_sdf_no_charges(self):
+        """Test OpenEyeToolkitWrapper for importing a charges from a sdf file"""
```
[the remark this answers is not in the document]
j-wags: Good call. Fixed this and the corresponding RDKit test docstring in 476cdbd

j-wags:
Thanks @mattwthompson and @jthorton for the careful reviews -- I really appreciate the attention to detail. Having some fresh pairs of eyes really helped catch things I was totally glossing over!

> Is the basic round-tripping of partial charges covered?

We cover partial charge roundtrips for SDF in the new tests in this PR, and we previously tested for round tripping with OE/RDMols in tests like [`test_to_from_openeye_core_props_filled`](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/tests/test_toolkits.py#L156) and [`test_to_from_rdkit_core_props_filled`](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/tests/test_toolkits.py#L950). "Core properties" are how I tried to distinguish "functional" attributes of a molecule, and their counterpart is "secondary" properties (anything that gets stuffed into the `offmol.properties` dict). A naive attempt at spec equivalence between toolkits can be found here: https://github.com/openforcefield/openforcefield/issues/135 (when I have time, I'll bake this into the developer docs as well)

> Is there anything in the RDKit block of tests that is fundamentally different than the OpenEye block? I took a more careful look at the latter.

I tried to make the tests as identical as possible between the two. In a future refactor, we could likely use `pytest.parametrize` to loop over the exact same tests for both.

> There is a conflict between this state of the API and what is plausibly a future state, in particular with respect to reading in multiple conformers and/or molecules from an SDF file, and how to handle the case of different conformers having different properties. For example, in this patch will write only the first conformer of a multi-conformer OpenFF Molecule, whereas in the future we that may be different. What is our plan here?

Thanks for describing this so succinctly. 

So, the status as of this posting is that:
* this branch can read multiple molecules from SDF, no problem. 
* this branch can not collapse multiple conformers of the same molecule from an SDF into a single, multi-conformer molecule
* this branch can write a _single conformer_ of a molecule to SDF
* this branch can not write multiple molecules, or multiple conformers of one molecule, to SDF
* this branch can successfully preserve the core properties of a molecule (except conformers other than the first) through an SDF round trip

Our functionality is limited with regard to "multi-conformer" SDFs because our molecule object model is fundamentally at odds with "multiconformer" SDF. In a multi-molecule SDF, different molecules/conformers can have different property values under the same property name (for example, each conformer could have different partial charges). We can not reconcile this with our Molecule object, in which _no data can be attached to individual conformers._ 

In the future, we'd like to be in a state where we can read and write multi-conformer SDFs. This will require a considerable amount of thought and design, and we may find that the shortest path to reliable and unsurprising behavior is a fundamental redesign of our Molecule class. This is not trivial -- I haven't found any such functionality in RDKit, and the only reference to it is an untouched issue from 2016: https://github.com/rdkit/rdkit/issues/1125. OpenEye has an [entire design docs section specifically discussing behavior around multiconformer SDFs](https://docs.eyesopen.com/toolkits/python/oechemtk/oemol.html#dude-where-s-my-sd-data).  

If, in the future, we do implement support for multiconformer SDFs, we could implement a reverse-compatible API as follows:
* `offmol.to_file('multiconformer.sdf', write_multiple_conformers=True, file_format='sdf')`
* `Molecule.from_file('multiconformer.sdf', collapse_conformers=True, file_format='sdf')`

By default, the `write_multiple_conformers` and `collapse_conformers` kwargs could be `False`, preserving the current behavior. 

In the short term, we can save an OFFMol to disk by pickling the results of `offmol.to_dict`. In the long term, we can work to implement the API above.

j-wags:
The failing tests are just the DDOS protection on the DOI link -- All other tests are passing. The DOI link issue can be handled separately. I'm going to merge this PR.

PR #283, opened 2019-04-21
Fix #259: Add example showing how to workaround ParmEd problems with TIP3P water.

### LINKED ISSUE #259: Improve examples illustrating the simulation of solvated protein systems
Hi all,
Thanks for your work with OFF. I noticed a call to submit questions and feedback so here I am!

I'd like to simulate a protein/ligand system with the ligand being parameterized by openforcefield, and the protein being parameterized by OpenMM using an Amber forcefield. This works really well in implicit solvent!

My sticking point is with the addition of explicit water. I've been working from this example (Im using streptavidin though), and used PDBFixer to add the solvent:
[example](https://github.com/openforcefield/openforcefield/blob/master/examples/using_smirnoff_with_amber_protein_forcefield/toluene_in_T4_lysozyme.ipynb)

That example calls `createSystem` from a parmed `Structure` object. Parmed expects `BondType` parameters, which don't exist for the water molecules. This is documented in these issues:
https://github.com/ParmEd/ParmEd/issues/930
https://github.com/ParmEd/ParmEd/issues/868

So, after all that my question is - have you used OpenMM to simulate a solvated protein system including an OFF-parameterized ligand, and how is that done?
Thanks!

Things I've tried:
Adding false BondType parameters and then using rigidWater. It appears that the false BondType parameres are honoured by OpenMM, because the water molecules have really bizarre shapes. I can provide example files but it looks like the missing `BondType` issue is already pretty well understood

andrrizzi:
Fix #259.

I've removed the cell exporting the complex `System` to OpenMM, AMBER, and GROMACS files in the T4 Lysozyme as now that there is water it's taking a lot of time and those snippets are already included in other two examples.


jchodera:
Wait, why is there water in this example? We should only be applying parameters to the small molecule and combining it with the protein.

jchodera:
Oh, I see: The goal is to also include explicit waters solvating the protein.

I think we likely need to find a more general solution that would allow solvent to be excluded from overlapping with the ligand, so we may want to take a more general approach that would work for systems beyond this single example (where the binding site is buried).

jchodera:
I think we should _start_ with a PDB file that already contains the (protein+solvent+ions), with the suggestion that our users should prepare the complex, split it into (protein+solvent+ions) as PDB and ligand as mol2, and then process it through our pipeline. We should _not_ use `Modeller.addSolvent()`, since this could lead to significant problems if the solvent overlaps with the ligand.

Later on, we can come up with more clever ways to automate preparation with SMIRNOFF + AMBER that includes a solvation step, but we will need to work out a more elaborate preparation pipeline for that to work. Instead, we can presume our users already have an existing pipeline they can use and simply want to apply a new set of SMIRNOFF + AMBER parameters.

andrrizzi:
Good point! I'll build a T4 Lysozyme + water pdb (if it's not already around) and fix the example today.

jchodera:
It might be a good idea to add some salt as well just to make sure that works.

examples/using_smirnoff_with_amber_protein_forcefield/toluene_in_T4_lysozyme.ipynb
```diff
@@ -95,29 +97,30 @@
...
    "outputs": [],
    "source": [
-    "t4_pdb_file_path = get_data_filename('proteins/T4-protein.pdb')\n",
-    "t4_pdbfile = PDBFile(t4_pdb_file_path)\n",
-    "\n",
-    "# Load the AMBER protein force field through OpenMM.\n",
+    "# Load the AMBER protein force field parameters through OpenMM.\n",
     "from simtk.openmm import app\n",
-    "omm_forcefield = app.ForceField('amber99sbildn.xml')\n",
+    "omm_forcefield = app.ForceField('amber99sbildn.xml', 'tip3p.xml')\n",
```
j-wags: @jchodera At the January meeting, [there seemed to be consensus](https://docs.google.com/presentation/d/1R2zM1bTqqDHUC534UKdI8vbdov2peNrQ3lA5-wtX_bc/edit#slide=id.g4d411744a8_0_0) that we would "pin" our small molecule forcefield to proteins parameterized using [AMBER ff14-sb](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/data/amber14/protein.ff14SB.xml) and water using [TIP3P-FB](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/data/tip3pfb.xml). Do we want to start putting that in our examples?

examples/using_smirnoff_with_amber_protein_forcefield/toluene_in_T4_lysozyme.ipynb
```diff
@@ -15,7 +15,7 @@
    "source": [
     "### Parametrize a molecule with smirnoff99Frosst\n",
     "\n",
-    "First, we parametrize the ligand (toluene) with the smirnoff99Frosst force field through the usual route"
+    "First, we parametrize the ligand (toluene) with the smirnoff99Frosst force field through the usual route to create an OpenMM `System`."
```
j-wags: One of the subtle points of this example is that the toluene coordinates are already compatible with the protein coordinates. Maybe add another copy of `Note that the coordinates of protein and ligand in this example are determined by the PDB file, and they are already consistent with the ligand being positioned in the binding pocket` here.

examples/using_smirnoff_with_amber_protein_forcefield/toluene_in_T4_lysozyme.ipynb
andrrizzi: Good idea! I'll swap the models before merging.

examples/using_smirnoff_with_amber_protein_forcefield/toluene_in_T4_lysozyme.ipynb
andrrizzi: Will do!

j-wags:
I've opened #284 to discuss and roll forward with FF pinning in our examples. In the meantime, I'll merge these changes.

davidlmobley:
I agree with John that our examples here should focus on assignment of parameters (taking already solvated system and adding parameters) rather than on building systems.

PR #291, opened 2019-04-25
API changes for ForceField serialization

### LINKED ISSUE #278: Remove {load,to}_smirnoff_dict from public ForceField API
Per the discussion in #265, we should remove `load_smirnoff_data` and `to_smirnoff_data` from the public API.

> Oh, dear. I think we might want to eliminate the public API for load_smirnoff_data we accidentally exposed in 0.2. We really don't want to have to support a public object API and and XML serialization spec and a dict based format for the same thing in perpetuity. I think we eventually just want to get down to the object model and either the dict or serialization, but not all three of these.
> 
> @j-wags: could we make the dict API private, or at least flag it with a warning that the dict structure is potentially subject to change? We can hash this out in a call next week.

### LINKED ISSUE #288: Implement ForceField.to_file and to_string
We should add FF I/O methods, so that users don't have to deal with converting FF's to dicts (which we're removing from the public API).

I propose 
`ForceField.to_file(filename, format, return_cosmetic_attributes=False)`
and
`ForceField.to_string(format, return_cosmetic_attributes=False)`

(Discussion migrated from #278 )

j-wags:
This PR:
- [x] closes #278 
- [x] closes #288
- [x] implements basic `ParameterHandler` compatibility checks
- [x] changes behavior of `ForceField`'s `register_X_handler` and `get_X_handler` functions to accept initialized objects instead of classes (details in PR comments)
- [x] enables extending parameter lists when reading multiple SMIRNOFF data sources

Status:
- [x] Ready for review

### DIFF
--- a/openforcefield/tests/test_forcefield.py
+++ b/openforcefield/tests/test_forcefield.py
@@ -153,8 +153,6 @@ xml_ff_w_cosmetic_elements = '''<?xml version='1.0' encoding='ASCII'?>
 '''
 
 
-
-
 #======================================================================
 # TEST UTILITY FUNCTIONS
 #======================================================================
@@ -423,8 +421,7 @@ class TestForceField():
         assert 'cosmetic_element="why not?"' not in string_3
         assert 'parameterize_eval="blah=blah2"' not in string_3
 
-    @pytest.mark.parametrize('filename_extension', ['xml', 'XML', '.xml', '.XML',
-                                                    'offxml', 'OFFXML', '.offxml', '.OFFXML'])
+    @pytest.mark.parametrize('filename_extension', ['xml', 'XML', 'offxml', 'OFFXML'])
     @pytest.mark.parametrize('specified_format', [None, 'xml', 'XML', '.xml', '.XML',
                                                   'offxml', 'OFFXML', '.offxml', '.OFFXML'])
     def test_xml_file_roundtrip(self, filename_extension, specified_format):
@@ -441,8 +438,7 @@ class TestForceField():
         assert open(iofile1.name).read() == open(iofile2.name).read()
 
 
-    @pytest.mark.parametrize('filename_extension', ['xml', 'XML', '.xml', '.XML',
-                                                    'offxml', 'OFFXML', '.offxml', '.OFFXML'])
+    @pytest.mark.parametrize('filename_extension', ['xml', 'XML', 'offxml', 'OFFXML'])
     @pytest.mark.parametrize('specified_format', [None, 'xml', 'XML', '.xml', '.XML',
                                                   'offxml', 'OFFXML', '.offxml', '.OFFXML'])
     def test_xml_file_roundtrip_keep_cosmetic(self, filename_extension, specified_format):
@@ -490,7 +486,7 @@ class TestForceField():
     def test_load_two_sources(self):
         """Test loading data from two SMIRNOFF data sources"""
         ff = ForceField(simple_xml_ff, xml_ff_w_cosmetic_elements, permit_cosmetic_attributes=True)
-        assert len(ff.get_handler('Bonds').parameters) == 4
+        assert len(ff.get_parameter_handler('Bonds').parameters) == 4
 
     def test_load_two_sources_incompatible_tags(self):
         """Test loading data from two SMIRNOFF data sources which have incompatible physics"""
@@ -537,12 +533,8 @@ class TestForceField():
         molecules = []
         molecules.append(Molecule.from_smiles('CCO'))
         molecules.append(Molecule.from_smiles('C1CCCCC1'))
-        # molecules = [Molecule.from_file(get_data_filename(name)) for name in ('molecules/ethanol.mol2',
-        #                                                                      'molecules/cyclohexane.mol2')]
         topology = Topology.from_openmm(pdbfile.topology, unique_molecules=molecules)
         topology.box_vectors = None
-        #forcefield.get_handler("Electrostatics", {})._method = "Coulomb"
-        #forcefield.get_handler("vdW", {})._method = "cutoff"
 
         omm_system = forcefield.create_openmm_system(topology)
 
@@ -775,8 +767,8 @@ class TestForceField():
 
         molecules = [create_ethanol()]
         forcefield = ForceField('smirnoff99Frosst.offxml')
-        forcefield.get_handler('vdW', {})._method = vdw_method
-        forcefield.get_handler('Electrostatics', {})._method = electrostatics_method
+        forcefield.get_parameter_handler('vdW', {})._method = vdw_method
+        forcefield.get_parameter_handler('Electrostatics', {})._method = electrostatics_method
 
         pdbfile = app.PDBFile(get_data_filename('systems/test_systems/1_ethanol.pdb'))
         topology = Topology.from_openmm(pdbfile.topology, unique_molecules=molecules)
--- a/openforcefield/tests/test_forcefield.py
+++ b/openforcefield/tests/test_forcefield.py
@@ -397,13 +397,11 @@ class TestForceField():
         with pytest.raises(SMIRNOFFSpecError, match="Unexpected kwarg {'parameters': 'k, length'} passed") as excinfo:
             forcefield = ForceField(xml_ff_w_cosmetic_elements)
 
-        # Create a forcefield from XML successfully
+        # Create a forcefield from XML successfully, by explicitly permitting cosmetic attributes
         forcefield_1 = ForceField(xml_ff_w_cosmetic_elements, permit_cosmetic_attributes=True)
 
         # Convert the forcefield back to XML
         string_1 = forcefield_1.to_string('XML', discard_cosmetic_attributes=False)
-
-        # Ensure that the new XML string has cosmetic attributes in it
 
         # Ensure that the new XML string has cosmetic attributes in it
         assert 'cosmetic_element="why not?"' in string_1
@@ -411,7 +409,7 @@ class TestForceField():
         with pytest.raises(SMIRNOFFSpecError, match="Unexpected kwarg {'parameters': 'k, length'} passed") as excinfo:
             forcefield = ForceField(string_1, permit_cosmetic_attributes=False)
 
-        # Complete the round trip from forcefield_1 successfully
+        # Complete the forcefield_1 --> string --> forcefield_2 roundtrip
         forcefield_2 = ForceField(string_1, permit_cosmetic_attributes=True)
 
         # Ensure that the forcefield remains the same after the roundtrip
--- a/openforcefield/tests/test_forcefield.py
+++ b/openforcefield/tests/test_forcefield.py
@@ -26,7 +26,7 @@ from openforcefield.utils.toolkits import OpenEyeToolkitWrapper, RDKitToolkitWra
 from openforcefield.utils import get_data_filename
 
 from openforcefield.topology import Molecule, Topology
-from openforcefield.typing.engines.smirnoff import ForceField, IncompatibleParameterError
+from openforcefield.typing.engines.smirnoff import ForceField, IncompatibleParameterError, SMIRNOFFSpecError
 
 
 #======================================================================
@@ -393,9 +393,11 @@ class TestForceField():
         """
         Test roundtripping a forcefield to XML with and without retaining cosmetic elements
         """
-        forcefield_1 = ForceField(xml_ff_w_cosmetic_elements, discard_cosmetic_attributes=False)
+        forcefield_1 = ForceField(xml_ff_w_cosmetic_elements, permit_cosmetic_attributes=True)
         string_1 = forcefield_1.to_string('XML', discard_cosmetic_attributes=False)
-        forcefield_2 = ForceField(string_1, discard_cosmetic_attributes=False)
+        with pytest.raises(SMIRNOFFSpecError, match="Unexpected kwarg {'parameters': 'k, length'} passed") as excinfo:
+            forcefield_2 = ForceField(string_1, permit_cosmetic_attributes=False)
+        forcefield_2 = ForceField(string_1, permit_cosmetic_attributes=True)
         string_2 = forcefield_2.to_string('XML', discard_cosmetic_attributes=False)
         assert string_1 == string_2
 
--- a/openforcefield/tests/test_forcefield.py
+++ b/openforcefield/tests/test_forcefield.py
@@ -27,6 +27,7 @@ from openforcefield.utils.toolkits import OpenEyeToolkitWrapper, RDKitToolkitWra
 from openforcefield.utils import get_data_filename
 from openforcefield.topology import Molecule, Topology
 from openforcefield.typing.engines.smirnoff import ForceField, IncompatibleParameterError, SMIRNOFFSpecError
+from openforcefield.typing.engines.smirnoff import XMLParameterIOHandler
 
 
 #======================================================================
@@ -423,7 +424,8 @@ class TestForceField():
 
     @pytest.mark.parametrize('filename_extension', ['xml', 'XML', 'offxml', 'OFFXML'])
     @pytest.mark.parametrize('specified_format', [None, 'xml', 'XML', '.xml', '.XML',
-                                                  'offxml', 'OFFXML', '.offxml', '.OFFXML'])
+                                                  'offxml', 'OFFXML', '.offxml', '.OFFXML',
+                                                  XMLParameterIOHandler()])
     def test_xml_file_roundtrip(self, filename_extension, specified_format):
         """
         Test roundtripping a ForceField to and from an XML file
@@ -432,15 +434,16 @@ class TestForceField():
         iofile1 = NamedTemporaryFile(suffix='.' + filename_extension)
         iofile2 = NamedTemporaryFile(suffix='.' + filename_extension)
         forcefield_1 = ForceField(simple_xml_ff)
-        forcefield_1.to_file(iofile1.name, format=specified_format)
+        forcefield_1.to_file(iofile1.name, io_format=specified_format)
         forcefield_2 = ForceField(iofile1.name)
-        forcefield_2.to_file(iofile2.name, format=specified_format)
+        forcefield_2.to_file(iofile2.name, io_format=specified_format)
         assert open(iofile1.name).read() == open(iofile2.name).read()
 
 
     @pytest.mark.parametrize('filename_extension', ['xml', 'XML', 'offxml', 'OFFXML'])
     @pytest.mark.parametrize('specified_format', [None, 'xml', 'XML', '.xml', '.XML',
-                                                  'offxml', 'OFFXML', '.offxml', '.OFFXML'])
+                                                  'offxml', 'OFFXML', '.offxml', '.OFFXML',
+                                                  XMLParameterIOHandler()])
     def test_xml_file_roundtrip_keep_cosmetic(self, filename_extension, specified_format):
         """
         Test roundtripping a forcefield to an XML file with and without retaining cosmetic elements
@@ -458,7 +461,7 @@ class TestForceField():
         forcefield_1 = ForceField(xml_ff_w_cosmetic_elements, permit_cosmetic_attributes=True)
 
         # Convert the forcefield back to XML, keeping cosmetic attributes
-        forcefield_1.to_file(iofile1.name, discard_cosmetic_attributes=False, format=specified_format)
+        forcefield_1.to_file(iofile1.name, discard_cosmetic_attributes=False, io_format=specified_format)
 
         # Ensure that the new XML string has cosmetic attributes in it
         assert 'cosmetic_element="why not?"' in open(iofile1.name).read()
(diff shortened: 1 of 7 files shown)


openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -652,6 +646,40 @@ def __init__(self, permit_cosmetic_attributes=False, **kwargs):
                                         "'permit_cosmetic_attributes=True'".format(key, self.__class__))
 
 
+    def _add_parameters(self, section_dict, permit_cosmetic_attributes=False):
```
j-wags: I've separated the parameter addition code out from the ParameterHandler initialization code. This way, we can go through multiple rounds of parameter addition from different SMIRNOFF data sources with a single ParameterHandler.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1069,6 +1097,33 @@ def __init__(self, **kwargs):
         # TODO: Do we want a docstring here? If not, check that docstring get inherited from ParameterHandler.
         super().__init__(**kwargs)
 
+    def check_handler_compatibility(self,
```
j-wags: I've added an explicit `check_handler_compatibility` function for each ParameterHandler initialized by SMIRNOFF99Frosst. These functions are nearly identical, and could be refactored/consolidated once we get a sense for how well they work.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -761,12 +782,74 @@ def parse_smirnoff_from_source(self, source):
...
 
+    def to_string(self, format, discard_cosmetic_attributes=True):
+        """
+        Write this Forcefield and all its associated parameters to a string in a given format which
+        complies with the SMIRNOFF spec.
+
+
+        Parameters
+        ----------
+        format : str
```
SimonBoothroyd: This looks mostly good, although I'm really not sure about having format as just some string literal... it makes it really unclear as to which values it can take (e.g 'Xml, xml, XML, .xml, .offxml, ..., json, Json, JSON... etc) and doesn't lend itself well to static linting. My personal preference would be to have an `Enum` which contains all of the available formats, and to pass this instead.

e.g: `ff_string = ff.to_string(ForceField.Format.OFFXML, True)` or `ff_string = ff.to_string(ForceField.Format.JSON, True)` if we extend support to other formats in the future.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -761,12 +782,74 @@ def parse_smirnoff_from_source(self, source):
...
+        """
+        Write this Forcefield and all its associated parameters to a string in a given format which
+        complies with the SMIRNOFF spec.
+
+
+        Parameters
+        ----------
+        filename : str
+            The filename to write to
+        format : str
```
SimonBoothroyd: Same as above ^^^^

openforcefield/typing/engines/smirnoff/parameters.py
SimonBoothroyd: Good call!

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1250,6 +1333,48 @@ def __init__(self, **kwargs):
...
+        ----------
+        other_handler : a ParameterHandler-derived object
+            The handler to compare to.
+
+        Raises
+        ------
+        IncompatibleParameterError if handler_kwargs are incompatible with existing parameters.
+        """
+        float_attrs_to_compare = []
+        string_attrs_to_compare = ['potential']
```
SimonBoothroyd: As opposed to splitting into different arrays, it may be a bit cleaner to have a single array of all attrs and then just do a type check in the below loops. If these checks (i.e `abs(this_val - other_val) > 1.e-6`) are going to be used repeatedly they could probably just be dumped into static methods on the `ParameterHandler` class.

Something like

```
@staticmethod
def compare_numbers(a, b, tolerance)
    ....
```

or something more like numpy's `isclose` method.

openforcefield/typing/engines/smirnoff/parameters.py
SimonBoothroyd: Definitely like the idea of this! And for sure when you get more of a handle on what comparisons can be made there definitely looks to be scope for condensing things, but for now is probably fine (maybe throw in some todo's in case of bus?)!

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -274,15 +274,15 @@ def __contains__(self, item):
...
         Parameters
         ----------
 
-        return_cosmetic_attributes : bool, optional. default = False
-            Whether to return non-spec attributes of each ParameterType-derived object.
+        discard_cosmetic_attributes : bool, optional. Default = True
+            Whether to discard non-spec attributes of each ParameterType-derived object.
 
         Returns
         -------
```
jchodera: Are these deep copies of `ParameterType`s, or the actual objects? Would modifying an object modify the corresponding object in `ForceField`?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -274,15 +274,15 @@ def __contains__(self, item):
         # Fall back to traditional access
         return list.__contains__(self, item)
 
-    def to_list(self, return_cosmetic_attributes=False):
+    def to_list(self, discard_cosmetic_attributes=True):
```
jchodera: Why have a `.to_list()` but not a `.from_list()`?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -447,16 +449,16 @@ def smirks(self, smirks):
             smirks, ensure_valence_type=self._VALENCE_TYPE)
         self._smirks = smirks
 
-    def to_dict(self, return_cosmetic_attributes=False):
+    def to_dict(self, discard_cosmetic_attributes=True):
```
jchodera: Why have a `.to_dict()` but not a `.from_dict()`?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -818,16 +846,16 @@ def postprocess_system(self, topology, system, **kwargs):
         """
         pass
 
-    def to_dict(self, output_units=None, return_cosmetic_attributes=False):
+    def to_dict(self, output_units=None, discard_cosmetic_attributes=True):
```
jchodera: Why have a `.to_dict()` but not a `.from_dict()`?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1166,6 +1220,35 @@ def __init__(self, **kwargs):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
+    def check_handler_compatibility(self,
+                                    other_handler):
+        """
+        Checks whether this ParameterHandler encodes the same physics as another ParameterHandler. This is
```
jchodera: I wouldn't say "the same physics", but at least "compatible physics".

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: These are serialized representations of ParameterTypes (so, not `ParameterType`s at all, but dicts).

j-wags: So modifying them has no effect on the corresponding object in `ForceField`.

openforcefield/typing/engines/smirnoff/io.py
```diff
@@ -312,8 +303,6 @@ def prepend_all_keys(d, char='@'):
                     prepend_all_keys(item)
 
         prepend_all_keys(smirnoff_data['SMIRNOFF'])
-        print(smirnoff_data)
-        print()
         return xmltodict.unparse(smirnoff_data, pretty=True)
 
     def to_xml(self, smirnoff_data):
```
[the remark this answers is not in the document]
j-wags: Like you said, it would have to be `ff_object.get_io_handler('XML').to_xml()`. And all is does is call `XMLParameterIOHandler.to_string()` anyway, which is easier to access via `ff_object.to_string('XML')`

It's so unlikely that this would ever be used, that I'd be OK to remove it.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -761,12 +782,74 @@ def parse_smirnoff_from_source(self, source):
...
-        msg = f"Source {source} does not appear to be in a known SMIRNOFF encoding.\n"
+        msg = f"Source {source} could not be read. If this is a file, ensure that the path is correct.\n"
+        msg += "If the file is present, ensure it is in a known SMIRNOFF encoding.\n"
         msg += f"Valid formats are: {valid_formats}\n"
-        msg += f"Parsing vailed with the following error:\n{exception_msg}\n"
+        msg += f"Parsing failed with the following error:\n{exception_msg}\n"
         raise IOError(msg)
 
 
+    def to_string(self, format, discard_cosmetic_attributes=True):
```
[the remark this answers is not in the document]
j-wags: Hmm, OK. I'll set the default to XML.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -530,35 +529,40 @@ def get_handler(self, tagname, handler_kwargs=None):
...
+                old_handler.check_handler_compatibility(new_handler)
+            return_handler = old_handler
+        elif tagname in self._parameter_handler_classes:
+            # Otherwise, register this handler in the forcefield
+            self.register_parameter_handler(new_handler)
+            return_handler = new_handler
+
+        return return_handler
 
     def get_io_handler(self, io_format):
```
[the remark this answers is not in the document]
j-wags: Good idea. I'll also change `get_handler` to `get_parameter_handler`.

openforcefield/tests/test_forcefield.py
```diff
@@ -378,20 +380,125 @@ def test_serialize(self):
...
+    @pytest.mark.parametrize('filename_extension', ['xml', 'XML', '.xml', '.XML',
+                                                    'offxml', 'OFFXML', '.offxml', '.OFFXML'])
+    @pytest.mark.parametrize('specified_format', [None, 'xml', 'XML', '.xml', '.XML',
+                                                  'offxml', 'OFFXML', '.offxml', '.OFFXML'])
+    def test_xml_file_roundtrip(self, filename_extension, specified_format):
+        """
+        Test roundtripping a ForceField to and from an XML file
+        """
+        # These files will be deleted once garbage collection runs (end of this function)
+        iofile1 = NamedTemporaryFile(suffix='.' + filename_extension)
```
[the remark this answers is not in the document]
j-wags: Oh, you're right. Good catch!

openforcefield/tests/test_forcefield.py
```diff
@@ -378,20 +380,125 @@ def test_serialize(self):
...
+            forcefield = ForceField(xml_ff_w_cosmetic_elements)
+
+        # Create a forcefield from XML successfully
+        forcefield_1 = ForceField(xml_ff_w_cosmetic_elements, permit_cosmetic_attributes=True)
+
+        # Convert the forcefield back to XML, keeping cosmetic attributes
+        forcefield_1.to_file(iofile1.name, discard_cosmetic_attributes=False, format=specified_format)
+
+        # Ensure that the new XML string has cosmetic attributes in it
+        assert 'cosmetic_element="why not?"' in open(iofile1.name).read()
```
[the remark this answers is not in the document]
j-wags: Hmm, I see the benefit of doing this, but I'm concerned that the system is already confusing, and adding in string formatting commands would make it even more so.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -372,68 +375,61 @@ def _register_parameter_io_handler_classes(self,
...
+        Register a new ParameterHandler for a specific tag, making it
         available for lookup in the ForceField.
 
         .. warning :: This API is experimental and subject to change.
 
         Parameters
         ----------
-        parameter_handler_class : A ParameterHandler-derived object
-            The ParameterHandler to register
+        parameter_handler : A ParameterHandler-derived object
```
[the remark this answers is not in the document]
j-wags: Yeah, now that I look at it, I feel weird about how I use the word "derived" in these docstrings. I'm removing this phrase from the files.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Good call. Added.

j-wags:
@jchodera asked several questions of the form "why is there a `to_X` function and not a `from_X` function?"

This was a really interesting question to think about. What it comes down to is that the SMIRNOFF data model makes serialization and deserialization asymmetric. That is, it would be satisfying to be able to take a SMIRNOFF dict (hierarchical dict representation of an OFFXML), and pass it to the `ForceField` initializer, and have that pass each parameter section into the appropriate `ParameterHandler` initializer, and have that pass the parameter list into the `ParameterList` initializer, and have that pass each individual parameter into the appropriate `ParameterType` initializer. But we can't.

This is because there are at least two cases where information that is scientifically important to the `ParameterType` is actually stored in the section header (and therefore becomes an attribute of the `ParameterHandler`). These are:

* units
* `Torsions`'s `default_idivf`

So, what really happens is:
* the `ParameterHandler` is initialized using the section dict as kwargs for its initializer
* the `ParameterHandler` initializes a `ParameterList` for itself (no kwargs)
* the `ParameterHandler` modifies each parameter (which is still a dict) by attaching units, 
* the `ParameterHandler` uses the unit-attached dict as kwargs to the `ParameterType` initializer, 
* the `ParameterHandler` finally appends the initialized `ParameterType` to its `ParameterList`.

Because of this, we can't directly serialize a ParameterType to a line in an OFFXML file. At a minimum, the owning `ParameterHandler` must decide which units all of its `ParameterTypes` will use, and then convert all of the serialized `ParameterType`s it owns to use those values.

I think that modifying the SMIRNOFF spec to achieve this symmetry would be more than just satisfying -- It would probably head off several errors and simplify development in the future. What do you think, @jchodera @SimonBoothroyd @andrrizzi @davidlmobley ?

jchodera:
100% supportive of modifying the spec to make the architecture simpler! This would mean each entry would just have its own units attached, right?

We can still have convenience methods that will convert all unit-bearing quantities of a force field or parameter class to a desired unit system. It could even be chained for convenience:

forcefields.standardize_units().to_file(filename, format=XML)

SimonBoothroyd:
I definitely think that would be a good way forward, even if it adds some extra bulk / repeated information into the force field file itself. At the end of the day, I think we really want people to be accessing and editing the FF objects through the python API rather than the file itself, such that we don't need to be too concerned about slightly verbose files / the previously common copy paste errors!

It seems like this change would very much help in allowing the force field to be serialized to different formats in the future, making things a bit more future proof. If I understand this correctly, making these changes would allow FF objects to be directly turned into ( / created from) a hierarchical `dict` without any extra processing, by which I mean every object which makes up the FF object would be able to turn itself into (/ be created from) a `dict`, without knowledge of any other objects which would be a big simplification! 

I guess this would also allow easier integration of libraries like `attrs` (or `pydantic`) which could allow cleaner class definitions / validation logic.

davidlmobley:
I also agree this makes sense. We don't want our thoughts of what we liked in terms of the spec to keep us from doing stuff which really matters/make software architecture hard, and most of those preferences were not strong preferences.

j-wags:
Great. Thanks for the feedback, I think this is the right way to go. I'm going to handle the spec change in a later PR so the FF versioning can move forward for now (otherwise we'll never finish 0.3.0)

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -507,20 +503,23 @@ def render_atoms(atomsets):
...
         Parameters
         ----------
-        tagame : str
+        tagname : str
             The name of the parameter to be handled.
-        handler_kwargs : dict, optional. Default=None
+        handler_kwargs : dict, optional. Default = None
             Dict to be passed to the handler for construction or checking compatibility. If None, will be assumed
             to represent handler defaults.
+        permit_cosmetic_attributes : bool, optional. Default = False
```
andrrizzi: This is outrageously nitpicky, but we used `allow_` instead of `permit_` somewhere else (at least) in `allow_undefined_stereo`, and I wonder if we should change one of the two for naming consistency.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -372,68 +375,61 @@ def _register_parameter_io_handler_classes(self,
...
-        Register a new ParameterIOHandler from a specified class, instantiating the ParameterIOHandler object and making
-        it available for lookup in the ForceField.
+        Register a new ParameterIOHandler, making it available for lookup in the ForceField.
 
         .. warning :: This API is experimental and subject to change.
 
         Parameters
         ----------
-        parameter_io_handler_class : A subclass of ParameterIOHandler
+        parameter_io_handler :  A ParameterIOHandler object
```
andrrizzi: I like it. Good call!

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -530,35 +529,40 @@ def get_handler(self, tagname, handler_kwargs=None):
...
+        new_handler = ph_class(**handler_kwargs,
+                               permit_cosmetic_attributes=permit_cosmetic_attributes)
 
-    def get_io_handler(self, io_format):
-        """Retrieve the parameter handlers associated with the provided tagname.
+        if tagname in self._parameter_handlers:
+            # If a handler of this class already exists, ensure that the two handlers encode compatible science
+            old_handler = self._parameter_handlers[tagname]
+            # If no handler kwargs were provided, skip the compatibility check
+            if handler_kwargs != {}:
```
andrrizzi: This may be harmless, but isn't `handler_kwargs = None` when handler kwargs are not provided?

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -761,12 +782,74 @@ def parse_smirnoff_from_source(self, source):
...
 
+    def to_string(self, format='XML', discard_cosmetic_attributes=True):
+        """
+        Write this Forcefield and all its associated parameters to a string in a given format which
+        complies with the SMIRNOFF spec.
+
+
+        Parameters
+        ----------
+        format : str, optional. Default='XML'
```
andrrizzi: We could think of having formats codification here, for `io_format`, and in `toolkits.py` in an `Enum` object instead of strings (probably for a later PR/API-breaking version).

We could also extend this to accept directly a `ParameterIOHandler` object so that we could easily implement translation between specification in the future. So maybe we can call this something like `parameter_io_handler`, which accept also a parameter handler key (i.e. a format) and the code could be something like
```python
if isinstance(parameter_io_handler, ParameterIOHandler):
    io_handler = parameter_io_handler
else:
    io_handler = self. get_parameter_io_handler(parameter_io_handler)
```

openforcefield/typing/engines/smirnoff/parameters.py
andrrizzi: Love it. In general, I think passing around objects (at least in the public API) rather than `dict`s will generally improve encapsulation and thus extensibility.

andrrizzi: It's probably a good idea to open an issue to remind us to look into code simplification.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: If None, it gets set to `{}` [a few lines above](https://github.com/openforcefield/openforcefield/pull/291/files/7f88b81dbd54dc76259b2f9cfbd5804b5b3e2049#diff-150e97b6404d62dc5d95291deeeae080R533)

This is because handlers like `ToolkitAM1BCCHandler` will have an empty dict as their handler kwargs (as a consequence of how `XMLParameterIOHandler` interprets elements with no attributes or children).

This function is also how people who want to access ForceField parameters will get to them (`ForceField.get_handler('Bonds')`). I vaguely remember that setting a default argument to a mutable object in dangerous in Python, so it's safer to set it to `None`, instead of `{}`.

andrrizzi: Ah! Gotcha. I missed that bit.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: Brilliant! Agree that IO format enums should come later, but will extend this to accept `ParameterIOHandlers` right now

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: That would be a good first step for streamlining. For now, I want to see what this function turns into. There may be some cases where differences in some attributes are permissible, so I want to develop these separately a bit more to see how much flexibility a single inherited solution would require.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: Good point. I agree we should standardize on one, and am basically ambivalent about which. If I had to choose, I think `allow` is slightly clearer.

j-wags:
Thanks for the feedback, everyone. I think I've either addressed everything here, or moved it to a issue that we can track. Merging this into 0.3.0.

PR #292, opened 2019-04-29
Implement Topology.to_openmm and remove toolkit_is_available

### LINKED ISSUE #214: Remove one between ToolkitWrapper.toolkit_is_available and ToolkitWrapper.is_available
I'd vote for removing `toolkit_is_available()` so that the final syntax would become something like
```python
toolkit = RDKitToolkitWrapper()
toolkit.is_available()
```
instead of `toolkit.toolkit_is_available()`.

### LINKED ISSUE #240: Implement Topology.to_openmm()
It would be useful to implement [`Topology.to_openmm()`](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/topology/topology.py#L1551-L1560) to reduce the amount of code needed to set up the protein-ligand system.

andrrizzi:
- Fix #240: Implement `Topology.to_openmm()`.
- Fix #214: I've removed the cached implementation, and renamed the static `ToolkitWrapper.toolkit_is_available()` to `ToolkitWrapper.is_available()`.

### DIFF
--- a/openforcefield/tests/test_topology.py
+++ b/openforcefield/tests/test_topology.py
@@ -402,9 +402,7 @@ class TestTopology(TestCase):
         """Test creation of an openforcefield Topology object from an OpenMM Topology and component molecules"""
         from simtk.openmm import app
         pdbfile = app.PDBFile(get_data_filename('systems/packmol_boxes/cyclohexane_ethanol_0.4_0.6.pdb'))
-        #toolkit_wrapper = RDKitToolkitWrapper()
-        #molecules = [ Molecule.from_file(get_data_filename(name)) for name in ('molecules/ethanol.mol2',
-        #                                                                       'molecules/cyclohexane.mol2')]
+
         molecules = []
         molecules.append(Molecule.from_smiles('CCO'))
         molecules.append(Molecule.from_smiles('C1CCCCC1'))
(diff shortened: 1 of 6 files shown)


openforcefield/tests/test_topology.py
```diff
@@ -413,7 +411,57 @@ def test_from_openmm(self):
...
+        ethanol = Molecule.from_smiles('CCO')
+        benzene = Molecule.from_smiles('c1ccccc1')
+        off_topology = Topology.from_molecules(molecules=[ethanol, benzene, benzene])
+
+        # Convert to OpenMM Topology.
+        omm_topology = off_topology.to_openmm()
+
+        # Check that bond orders are preserved.
+        n_double_bonds = sum([b.order == 2 for b in omm_topology.bonds()])
+        n_aromatic_bonds = sum([b.type is Aromatic for b in omm_topology.bonds()])
```
j-wags: Interesting! I didn't know that openMM does bond types like this.

openforcefield/tests/test_topology.py
```diff
@@ -413,7 +411,57 @@ def test_from_openmm(self):
...
+            assert atom.atomic_number == atom_copy.atomic_number
+
+        # Check bonds.
+        for bond_idx, bond in enumerate(off_topology.topology_bonds):
+            bond_copy = off_topology_copy.bond(bond_idx)
+            bond_atoms = [a.atomic_number for a in bond.atoms]
+            bond_atoms_copy = [a.atomic_number for a in bond_copy.atoms]
+            assert bond_atoms == bond_atoms_copy
+            assert bond.bond_order == bond_copy.bond_order
+            assert bond.bond.is_aromatic == bond_copy.bond.is_aromatic
```
j-wags: Very cool. I like how this test is done.

openforcefield/topology/topology.py
```diff
@@ -1547,20 +1547,86 @@ def from_openmm(cls, openmm_topology, unique_molecules=None):
...
-    def _to_openmm(self):
+    def to_openmm(self):
         """
         Create an OpenMM Topology object.
 
+        The OpenMM ``Topology`` object will have one residue per topology
+        molecule. Currently, the number of chains depends on how many copies
+        of the same molecule are in the ``Topology``. Molecules with more
+        than 5 copies are all assigned to a single chain, otherwise one
+        chain is created for each molecule. This behavior may change in
```
j-wags: This 5-copy cutoff is a good idea. We can discuss it more, but should put an unstable-API warning here for now.

openforcefield/topology/topology.py
```diff
@@ -1547,20 +1547,86 @@ def from_openmm(cls, openmm_topology, unique_molecules=None):
...
+
+        # Go through atoms in OpenFF to preserve the order.
+        omm_atoms = []
+        # We need to iterate over the topology molecules if we want to
+        # keep track of chains/residues as Atom.topology_molecule is
+        # instantiated every time and can't be used as a key.
+        for topology_molecule in self.topology_molecules:
+            for atom in topology_molecule.atoms:
+                reference_molecule = topology_molecule.reference_molecule
+                n_molecules = len(self._reference_molecule_to_topology_molecules[reference_molecule])
```
j-wags: I'm glad you got a chance to see the madness that is the TopologyMolecule class :-P I wished for memory efficiency, but at what cost?

j-wags:
Woah -- tests are failing because RDKit considers benzene's bond orders to all be 1, with `is_aromatic` set to `True`. I'm working on debugging this.

This doesn't seem to have affected our physics, since all of our molecule parameterization comparison tests between RDK and OETK pass. But I'll sleep better if we can get RDKit's behavior matching here.

j-wags:
I've resolved this by explicitly kekulizing the molecules after detecting aromaticity in the `from_rdkit` function. I also changed the kekulized bond orders to be explicitly imported, instead of casting the aromatic bond order value (`1.5`) to int (yielding `1`).

andrrizzi:
Thanks so much for fixing it! I thought the Kekulization would be redundant because it should already happen in `Chem.SanitizeMol()` (a couple of lines above). Maybe the fix to the code determining the aromaticity is enough. I'll check it out in another PR.

PR #298, opened 2019-05-06
Fix #276: standardize variable/method names

### LINKED ISSUE #276: Standardize nomenclature in public API not covered by PEP8
We should standardize the way we name arguments and methods that are exposed in the public API when these are not covered by PEP8.

Here are is a short proposal covering the cases I've noticed:
- I'd use `file_name` (with the underscore because of PEP8) to indicate `protein.pdb` and `file_path` for `path/to/protein.pdb`.
- Number of elements: `n_elements`.
- Words with a hyphen without an underscore (e.g., `setup` when used as noun).

In any case, I'm ok with any convention as long as we are more or less consistent when we expose it in the API.

andrrizzi:
A summary of the changes in this PR:
- I've documented better what `rdkit.SanitizeMol()` does and attempted to remove a second Kekulization. Let's see if tests still pass after #292.
- I've renamed the argument `.generate_conformers(..., num_conformers)` to `n_conformers` for consistency with other exposed attributes (e.g., `n_atoms`, `n_molecules`).
- I've removed `utils.structure.get_testdata_filename`, which was equivalent to `utils.get_data_filename`.
- I've added a couple of clarifications about naming conventions in the `docs/developing.rst` file.
- Most of this PR is about renaming things like `filename`/`filepath`/`pdbfile`/`outfile`/`mol2file` to `file_path`. Here is a more or less complete list
```
utils.get_data_filename -> get_data_file_path
utils.get_alkethoh_filepath -> get_alkethoh_file_path
utils.get_freesolv_filepath -> get_freesolv_file_path
utils.get_amber_filepath -> get_amber_file_path
utils.get_packmol_pdbfile -> get_packmol_pdb_file_path
utils.get_monomer_mol2file -> get_monomer_mol2_file_path
tests.utils.create_system_from_amber(prmtop_filepath, inpcrd_filepath -> create_system_from_amber(prmtop_file_path, inpcrd_file_path
tests.utils.compare_amber_smirnoff(prmtop_filepath, inpcrd_filepath -> compare_amber_smirnoff(prmtop_file_path, inpcrd_file_path
utils.structure.read_molecules(filename -> read_molecules(file_path
Molecule.from_file(filename, -> from_file(file_path,
Molecule.to_file(outfile, -> to_file(file_path,
ToolkitWrapper.from_file(filename, -> from_file(file_path,
ToolkitWrapper.to_file(molecule, outfile, outfile_format -> to_file(molecule, file_path, file_format
```
there is also
```
ParameterIOHandler.to_file(self, filename -> to_file(self, file_path
```
but I'll wait for #291 to be merged first since it touches the same lines of that PR.

### DIFF
--- a/openforcefield/utils/structure.py
+++ b/openforcefield/utils/structure.py
@@ -201,27 +201,6 @@ def generateTopologyFromOEMol(molecule):
 
     return topology
 
-def get_testdata_filename(relative_path):
-    """Get the full path to one of the reference files in testsystems.
-
-    In the source distribution, these files are in ``openforcefield/data/``,
-    but on installation, they're moved to somewhere in the user's python
-    site-packages directory.
-
-    Parameters
-    ----------
-    name : str
-        Name of the file to load (with respect to the repex folder).
-
-    """
-
-    from pkg_resources import resource_filename
-    fn = resource_filename('openforcefield', os.path.join('data', relative_path))
-
-    if not os.path.exists(fn):
-        raise ValueError("Sorry! %s does not exist. If you just added it, you'll have to re-install" % fn)
-
-    return fn
 
 # TODO: Do we need this?
 def normalize_molecules(molecules):
@@ -295,13 +274,13 @@ def normalize_molecules(molecules):
     return
 
 # TODO: Migrate into Molecule
-def read_molecules(filename, verbose=True):
+def read_molecules(file_path, verbose=True):
     """
     Read molecules from an OpenEye-supported file.
 
     Parameters
     ----------
-    filename : str
+    file_path : str
         Filename from which molecules are to be read (e.g. mol2, sdf)
 
     Returns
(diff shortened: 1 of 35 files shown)


andrrizzi:
I actually decided to push that last commit now so that I can use the GitHub conflict-handling GUI to fix it once #291 will be merged.

openforcefield/utils/structure.py
```diff
@@ -201,27 +201,6 @@ def generateTopologyFromOEMol(molecule):
 
     return topology
 
-def get_testdata_filename(relative_path):
```
j-wags: Glad to be removing this potentially confusing duplicate

openforcefield/utils/structure.py
```diff
@@ -311,17 +290,18 @@ def read_molecules(filename, verbose=True):
 
     """
     from openeye import oechem
+    from openforcefield.utils import get_data_file_path
```
j-wags: Huzzah! Imports inside functions.

PR #306, opened 2019-05-08
Allow Conformers with All Zero Positions for Molecules with One Atom

SimonBoothroyd:
## Description
A proposed fix for [issue 305](https://github.com/openforcefield/openforcefield/issues/305) related an exception being raised when parameterizing ions.

The change now retains conformers which have all of their positions set to zero for 'molecules' which contain only single atoms.

## Status
- [x] Ready to go


j-wags:
This is a good idea. You mentioned there are other reasons that we fail on Na+, but this will allow us to get things like Argon through. Thanks!

PR #311, opened 2019-05-10
SMIRNOFF spec 0.3 support

### LINKED ISSUE #178: Make all offxml files compatible

### LINKED ISSUE #293: Update convert_frosst
This has fallen through the cracks a few times now, so I'm making it its own issue.

https://github.com/openforcefield/openforcefield/issues/249#issuecomment-486744866

> If we catch a human error in a SMIRKS pattern or similar, this is still where we would make an update in the source files there and then re-generate the ffxml files from it, so we want to ensure these are current until we actually have done some refitting.

https://github.com/openforcefield/openforcefield/issues/178#issuecomment-478116481

> Once finalized we are also going to need to change the template files in utilities/convert_frosst to reflect new syntax, versionining, date, etc., then use this to re-generate smirnoff99Frosst in the correct format (basically it's generated from a smirnoffishFrcmod.parm99Frosst.txt file via convert_frcmod.py there which puts it into the appropriate XML format, basically for historical reasons -- in porting it, Chris didn't like working with the XML (understandably) and so dealt with an AMBER-style file resembling a frcmod file but with SMIRKS provided in the first column rather than an atom type; the convert_frcmod.py file puts this into the correct XML format).

### LINKED ISSUE #297: Refactor SMIRNOFF spec to better encapsulate data
@jchodera asked several questions of the form "why is there a `to_X` function and not a `from_X` function?"

This was a really interesting question to think about. What it comes down to is that the SMIRNOFF data model makes serialization and deserialization asymmetric. That is, it would be satisfying to be able to take a SMIRNOFF dict (hierarchical dict representation of an OFFXML), and pass it to the `ForceField` initializer, and have that pass each parameter section into the appropriate `ParameterHandler` initializer, and have that pass the parameter list into the `ParameterList` initializer, and have that pass each individual parameter into the appropriate `ParameterType` initializer. But we can't.

This is because there are at least two cases where information that is scientifically important to the `ParameterType` is actually stored in the section header (and therefore becomes an attribute of the `ParameterHandler`). These are:

* units
* `Torsions`'s `default_idivf`

So, what really happens is:
* the `ParameterHandler` is initialized using the section dict as kwargs for its initializer
* the `ParameterHandler` initializes a `ParameterList` for itself (no kwargs)
* the `ParameterHandler` modifies each parameter (which is still a dict) by attaching units, 
* the `ParameterHandler` uses the unit-attached dict as kwargs to the `ParameterType` initializer, 
* the `ParameterHandler` finally appends the initialized `ParameterType` to its `ParameterList`.

Because of this, we can't directly serialize a ParameterType to a line in an OFFXML file. At a minimum, the owning `ParameterHandler` must decide which units all of its `ParameterTypes` will use, and then convert all of the serialized `ParameterType`s it owns to use those values.

I think that modifying the SMIRNOFF spec to achieve this symmetry would be more than just satisfying -- It would probably head off several errors and simplify development in the future. What do you think, @jchodera @SimonBoothroyd @andrrizzi @davidlmobley ?

_Originally posted by @j-wags in https://github.com/openforcefield/openforcefield/pull/291#issuecomment-489141867_

j-wags:
## Summary

As you can see from the name of this branch (`update-convert-frosst`), my original goal here was to simply upgrade the convert_frosst utility to make 0.3 spec files.  However, the `convert_frcmod.py` script in `utilities/convert_frosst` originally relied on the 0.0.4 OFF toolkit's ForceField data structure and writer, and my new `convert_frosst_0.3.py` script would rely on the current version of the toolkit. So basically, it's impossible to maintain multiple versions of this script. Thankfully, it's also unnecessary. 

As pointed out by @jchodera, if we make a version of the toolkit that can read old SMIRNOFF specs, it won't be _necessary_ to make 0.3 files from frcmods. That is, `convert_frcmod.py` could continue to only make 0.1-spec files, and then a current toolkit version could read that 0.1 spec file and output an equivalent 0.3 spec version. So, `convert_frosst` needs to be decoupled from the current toolkit to remain viable. Since it's only really used for our current line of force fields, I think it should live in the `openforcefield/smirnoff99Frosst` repo, be pinned to a pre-0.2.0 version of the toolkit, and only ever make 0.1-spec files. 

Thus, I've moved `utilities/convert_frosst/` to `utilities/deprecated/convert_frosst`.

With that worked out, the goal of this branch shifted to:
- Implement support for the 0.3 SMIRNOFF spec
- Make a toolkit version that contains a pathway for "upgrading" SMIRNOFF files written in old specs
- Prove that the higher-spec files are equivalent in our parameterization tests
- Include some test forcefields in the 0.3 spec
- Add API points to add/delete cosmetic attributes on `ParameterHandler`s and `ParameterType` (this is a completely separate issue and I should have made a different PR. Sorry about that)

## This PR:
- [x] Adds functionality that can read/write 0.3 spec files (closes #297)
- [x] Adds `convert_0_2_smirnoff_to_0_3` function to `utils/utils.py` to convert 0.2 spec files to 0.3
- [x] Adds `convert_0_1_smirnoff_to_0_2` function to `utils/utils.py`  to convert 0.1 spec files to 0.2, while printing copious warnings about the assumptions being made in the conversion
    - [x] closes #178 
- [x] **Renames improperly-named toolkit forcefields in `test_forcefields/old**. (Note that these were originally derived from smirnoff99Frosst files, but were hand-modified some number of times during toolkit updates, and use a different versioning scheme)
  - [x]  `smirnoff99Frosst_0_0_2.offxml` to `test_ff_0_0_2_spec_0_2.offxml`
  - [x]  `smirnoff99Frosst_0_0_4.offxml` to `test_ff_0_0_4_spec_0_2.offxml`
  - [x]  `smirnoff99Frosst_0_0_4_fixed.offxml` to `test_ff_0_0_4_fixed_spec_0_2.offxml`
  - [x]  Creates 0.1 spec equivalents of each of the above by uncommenting the original (0.1-spec) parameter section tags that were left in the files, merging Impropers back into the PeriodicTorsions section, and removing new Electrostatics and vdW tags.
  - [x] Makes 0.3 spec versions of each of the above by loading 0.2-spec files and writing them back out using this branch's version of the toolkit. 
- [x] Tests spec converter/spec equivalence by running Freesolv tests with 0.1, 0.2, and 0.3 spec versions of files.
- [x] updates SMIRNOFF spec docs with changes for 0.3 spec
  - [x] changes `Proper/ImproperTorsions` default (and only) `potential` value from `charmm` to `k*(1+cos(periodicity*theta-phase))`
  - [x] always stores quantities as `value * units` instead of consolidating units to section headers
  - [x] Assigns specs to all parameter sections, which will be 0.3 to begin with. 
- [ ] ~Updates `convert_frosst` to produce 0.3 spec files (closes #293)~ See essay above
- [x] Makes cosmetic attribute output True by default
- [x] Adds `add` and `delete_cosmetic_attribute` functions to `ParameterHandler` and `ParameterType`
- [x] Concatenates multiple authors/dates if multiple data sources containing these fields are read
- [x] Adds `skip_version_check` kwarg to `ParameterHandler.__init__()` constructor. Since the 0.3 spec will require each SMIRNOFF section to have a version number, this a way to create a `ParameterHandler` with the default value (the maximum version that the Handler class can support).

## Status
- [x] Ready for review
- [ ] Ready to merge


j-wags:
@mkgilson 

> Is this for regular dihedrals or impropers?

Both proper and improper torsions. I believe they use the same functional form.

j-wags:
@jchodera 

> Perhaps this is where it's safest to just include the Lepton mathematical expression directly in the potential argument?

This would certainly resolve ambiguity, and it's the direction we're planning to head in the future, anyway. Looking into Lepton now... 

So this would be something like `potential="k*(1+cos(periodicity*theta-phase))"`. I'm unable to find the original PeriodicTorsionForce expression in the OpenMM [code](https://github.com/pandegroup/openmm/search?q=periodictorsionforce&unscoped_q=periodictorsionforce)/[docs](http://docs.openmm.org/latest/userguide/application.html#periodictorsionforce) but this is my best guess based on the [custom torsion potential example](http://docs.openmm.org/latest/userguide/application.html#customtorsionforce). Does this sound good? 

One downside is that this may imply that the toolkit can already handle arbitrary functional forms, but we can post some warnings in the comments and throw an exception if people try to pass in anything else.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -664,6 +678,7 @@ def _to_smirnoff_data(self, discard_cosmetic_attributes=True):
 
         smirnoff_dict = OrderedDict()
         smirnoff_dict['SMIRNOFF'] = l1_dict
+        smirnoff_dict = convert_smirnoff_data_quantitys_to_string(smirnoff_dict)
```
jchodera: "quantities" is misspelled.

j-wags: This is a style thing for me -- since the object is `Quantity`, I write the plural  as `Quantity`s in docstrings (some day we can make sure all of these resolve to links to class names). It's not necessary to do this in function names but it seemed to be consistent. Weak preference, happy to change if it's more confusing than helpful.

jchodera: What about `convert_smirnoff_data_quantity_to_string` then?

j-wags: `convert_smirnoff_data_quantity_to_string`. I've changed it to that.

j-wags:
@andrrizzi @jchodera -- This is now ready for formal review. I've updated the description to explain what this PR aims to do. Let me know your thoughts!

The-SMIRNOFF-force-field-format.md
```diff
@@ -281,18 +290,17 @@ As an example of a complete SMIRNOFF force field specification, see the prototyp
...
 The two are related by `r0 = 2^(1/6)*sigma` and conversion is done internally in `ForceField` into the `sigma` values used in OpenMM.
-Note that, if `rmin_half` is specified instead of `sigma`, `rmin_half_unit` should be specified; both can be used in the same block if desired.
 
 Attributes in the `<vdW>` tag specify the scaling terms applied to the energies of 1-2 (`scale12`, default: 0), 1-3 (`scale13`, default: 0), 1-4 (`scale14`, default: 0.5), and 1-5 (`scale15`, default: 1.0) interactions,
-as well as the distance at which a switching function is applied (`switch_width`, default: `"1.0"` angstroms), the cutoff (`cutoff`, default: `"9.0"` angstroms), and long-range dispersion treatment scheme (`long_range_dispersion`, default: `"isotropic"`).
+as well as the distance at which a switching function is applied (`switch_width`, default: `"1.0*angstrom"`), the cutoff (`cutoff`, default: `"9.0angstroms"`), and long-range dispersion treatment scheme (`long_range_dispersion`, default: `"isotropic"`).
```
andrrizzi: Change `"9.0angstroms"` to `"9.0*angstroms"`.

The-SMIRNOFF-force-field-format.md
```diff
@@ -304,31 +312,36 @@ Support for [other Lennard-Jones mixing schemes](https://en.wikipedia.org/wiki/C
...
 ```XML
-<vdW potential="Lennard-Jones-12-6" combining_rules="Lorentz-Berthelot" scale12="0.0" scale13="0.0" scale14="0.5" sigma_unit="angstroms" epsilon_unit="kilocalories_per_mole">
-   <AtomPair smirks1="[#1:1]" smirks2="[#6:2]" sigma="1.4870" epsilon="0.0157"/>   
+<vdW potential="Lennard-Jones-12-6" combining_rules="Lorentz-Berthelot" scale12="0.0" scale13="0.0" scale14="0.5" scale15="1">
+   <AtomPair smirks1="[#1:1]" smirks2="[#6:2]" sigma="1.4870*angstrom" epsilon="0.0157*kilocalories_per_mole"/>   
    ...
 </vdW>   
 ```
 
+| vdW section tag version | Required tag attributes and default values                                                                                                                                                                                        | Optional tag attributes | Required parameter attributes                 | Optional parameter attributes |
```
andrrizzi: I'm probably missing something, but why do we have default values with required attributes?

The-SMIRNOFF-force-field-format.md
```diff
@@ -409,13 +432,18 @@ If the `potential` attribute is omitted, it defaults to `harmonic`.
...
+
+
 ### `<ProperTorsions>`
 
 Proper torsions are specified via a `<ProperTorsions>...</ProperTorsions>` block, with individual `<Proper>` tags containing attributes specifying the periodicity (`periodicity#`), phase (`phase#`), and barrier height (`k#`).
 ```XML
-<ProperTorsions potential="charmm" phase_unit="degrees" k_unit="kilocalories_per_mole">
-   <Proper smirks="[a,A:1]-[#6X4:2]-[#6X4:3]-[a,A:4]" idivf1="9" periodicity1="3" phase1="0.0" k1="1.40"/>
-   <Proper smirks="[#6X4:1]-[#6X4:2]-[#8X2:3]-[#6X4:4]" idivf1="1" periodicity1="3" phase1="0.0" k1="0.383" idivf2="1" periodicity2="2" phase2="180.0" k2="0.1"/>
+<ProperTorsions version="0.3" potential="k*(1+cos(periodicity*theta-phase))">
```
andrrizzi: Loving this. So much clearer to interpret.

The-SMIRNOFF-force-field-format.md
```diff
@@ -426,27 +454,32 @@ For convenience, and optional attribute specifies a torsion multiplicity by whic
...
-Currently, only `potential="charmm"` is supported, where we utilize the functional form:
+Currently, only `potential="k*(1+cos(periodicity*theta-phase))"` is supported, where we utilize the functional form:
 ```
 U = \sum_{i=1}^N k_i * (1 + cos(periodicity_i * phi - phase_i))
 ```
 
 .. note:: **AMBER defines a modified functional form**, such that `U = \sum_{i=1}^N (k_i/2) * (1 + cos(periodicity_i * phi - phase_i))`, so that barrier heights would need to be divided by two in order to be used in the SMIRNOFF format.
 
-If the `potential` attribute is omitted, it defaults to `charmm`.
+If the `potential` attribute is omitted, it defaults to `k*(1+cos(periodicity*theta-phase))`.
```
andrrizzi: Same here. If I understand `potential` can be omitted, but it's listed among the "required" attributes.

openforcefield/typing/engines/smirnoff/io.py
```diff
@@ -263,7 +263,27 @@ def prepend_all_keys(d, char='@'):
...
 
+        # the "xmltodict" library defaults to print out all element attributes on separate lines
+        # unless they're prepended by "@"
         prepend_all_keys(smirnoff_data['SMIRNOFF'])
+
+        # Un-prepend "@" from "Author" and "Date" fields, as they SHOULD be printed on
+        # separate lines from the top-level SMIRNOFF tag
+        if '@Author' in smirnoff_data['SMIRNOFF'].keys():
+            smirnoff_data['SMIRNOFF']['Author'] = smirnoff_data['SMIRNOFF']['@Author']
+            del smirnoff_data['SMIRNOFF']['@Author']
```
andrrizzi: I think you can simplify the code by adding an `ignore_keys=['Author', 'Date']` argument in `prepend_all_keys(d, char='@', ignore_keys=frozenset())` and do something like
```python
if isinstance(d[key], list) or isinstance(d[key], dict):
    prepend_all_keys(d[key])
elif key not in ignore_keys:
    new_key = char + key
    d[new_key] = d[key]
    del d[key]
    prepend_all_keys(d[new_key])
```
This way we should be able to avoid also the re-ordering bit.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -488,9 +495,41 @@ def to_dict(self, discard_cosmetic_attributes=True):
...
+        on the object in the Open Force Field toolkit, but can be written out during output.
+
+        Parameters
+        ----------
+        attr_name : str
+            Name of the attribute to define for this ParameterType object.
+        attr_value : str
+            The value of the attribute to define for this ParameterType object.
+        """
+        setattr(self, attr_name, attr_value)
```
andrrizzi: I'm not 100% sure we want to expose the cosmetic attributes as class attributes. In XML, for example, `my-attr-name` is a perfectly legal attribute, but it's an illegal Python variable name. Maybe we just want to expose a set `ParameterHandler.cosmetic_attributes` that can host only strings that the users can modify how they want. We can create a special class of set if we want to do some validation.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -686,6 +779,37 @@ def parameters(self):
...
+        on the object in the Open Force Field toolkit, but can be written out during output.
+
+        Parameters
+        ----------
+        attr_name : str
+            Name of the attribute to define for this ParameterType object.
+        attr_value : str
+            The value of the attribute to define for this ParameterType object.
+        """
+        setattr(self, '_'+attr_name, attr_value)
```
andrrizzi: Same here.

openforcefield/utils/utils.py
```diff
@@ -274,6 +294,130 @@ def string_to_quantity(quantity_string):
     output_quantity = _ast_eval(ast.parse(quantity_string, mode='eval').body)
     return output_quantity
 
+def convert_smirnoff_data_strings_to_quantity(smirnoff_data):
```
andrrizzi: I'd simpy call this `convert_all_strings_to_quantity`. This works with any `dict` and `list` if I understand correctly, and doesn't require the data to be in a particular format connected to SMIRNOFF, if I understand this correctly.

openforcefield/utils/utils.py
```diff
@@ -274,6 +294,130 @@ def string_to_quantity(quantity_string):
...
+            obj_to_return = smirnoff_data
+        except TypeError:
+            obj_to_return = smirnoff_data
+        except SyntaxError:
+            obj_to_return = smirnoff_data
+
+    return obj_to_return
+
+
+def convert_smirnoff_data_quantity_to_string(smirnoff_data):
```
andrrizzi: Same here.

openforcefield/utils/utils.py
```diff
@@ -274,6 +294,130 @@ def string_to_quantity(quantity_string):
...
+    """
+    from simtk import unit
+    if isinstance(object, unit.Quantity):
+        return object
+    elif isinstance(object, str):
+        return string_to_quantity(object)
+    else:
+        return [object_to_quantity(sub_obj) for sub_obj in object]
+
+def assert_object_units_are_compatible(object_name, object, unit_to_check, context=None):
```
andrrizzi: I'd expect an assertion error from this function. We could consider inheriting from `IncompatibleUnitError(AssertionError)` or change the name to `check_units_are_compatible()`.

The-SMIRNOFF-force-field-format.md
```diff
@@ -131,7 +140,7 @@ This hierarchical structure means that a typical parameter file will tend to hav
 
 ### Multiple SMIRNOFF representations can be processed in sequence
 
-Multiple SMIRNOFF `.offxml` files can be loaded by the openforcefield `ForceField` in sequence.
+Multiple SMIRNOFF data sources can be loaded by the openforcefield `ForceField` in sequence.
```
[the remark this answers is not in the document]
j-wags: Good catch. Changed.

The-SMIRNOFF-force-field-format.md
j-wags: Thanks. Fixed.

openforcefield/tests/conftest.py
```diff
@@ -29,8 +29,8 @@ def untar_full_alkethoh_and_freesolv_set():
     sets in advance to speed things up.
 
     See
-        test_forcefield.py::test_alkethoh_parameters_assignment
-        test_forcefield.py::test_freesolv_parameters_assignment
+        test_forcefield_0_2.py::test_alkethoh_parameters_assignment
+        test_forcefield_0_2.py::test_freesolv_parameters_assignment
```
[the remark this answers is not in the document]
j-wags: Good catch. I had made some intermediate files and was a bit too ambitious using PyCharm's refactoring tools.

openforcefield/tests/test_smirnoff.py
```diff
@@ -32,7 +32,7 @@
 
 # TODO: Fix these imports and unskip these tests
 import pytest
-pytestmark = pytest.mark.skip('tests in test_smirnoff are outdated and should be integrated with test_forcefield.py')
+pytestmark = pytest.mark.skip('tests in test_smirnoff are outdated and should be integrated with test_forcefield_0_2.py')
```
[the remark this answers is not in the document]
j-wags: Fixed.

openforcefield/tests/test_forcefield.py
```diff
@@ -323,26 +324,26 @@ def test_create_gbsa():
     @pytest.mark.skip(reason='Needs to be updated for 0.2.0 syntax')
     def test_create_forcefield_from_url(self):
         urls = [
-            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/test_forcefields/smirnoff99Frosst.offxml',
-            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/test_forcefields/tip3p.offxml'
+            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/forcefield/smirnoff99Frosst.offxml',
+            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/forcefield/tip3p.offxml'
```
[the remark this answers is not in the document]
j-wags: No -- This wasn't interntional. I must have been sloppy when I merged master. Thanks for bringing this up.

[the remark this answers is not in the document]
j-wags: Fixed

openforcefield/tests/test_forcefield.py
```diff
@@ -323,26 +324,26 @@ def test_create_gbsa():
...
         # Test creation with smirnoff99frosst URL
         forcefield = ForceField(urls[0])
 
     @pytest.mark.skip(reason='Needs to be updated for 0.2.0 syntax')
     def test_create_forcefield_from_url_list(self):
         urls = [
-            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/test_forcefields/smirnoff99Frosst.offxml',
-            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/test_forcefields/tip3p.offxml'
+            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/forcefield/smirnoff99Frosst.offxml',
+            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/forcefield/tip3p.offxml'
```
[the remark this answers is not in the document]
j-wags: Fixed

openforcefield/tests/test_forcefield.py
```diff
@@ -323,26 +324,26 @@ def test_create_gbsa():
...
         # Test creation with multiple URLs
         forcefield = ForceField(urls)
 
     @pytest.mark.skip(reason='Needs to be updated for 0.2.0 syntax')
     def test_create_forcefield_from_url_iterator(self):
         urls = [
-            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/test_forcefields/smirnoff99Frosst.offxml',
-            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/test_forcefields/tip3p.offxml'
+            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/forcefield/smirnoff99Frosst.offxml',
+            'https://raw.githubusercontent.com/openforcefield/openforcefield/master/openforcefield/data/forcefield/tip3p.offxml'
```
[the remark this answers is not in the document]
j-wags: Fixed.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -706,46 +816,70 @@ def _load_smirnoff_data(self, smirnoff_data, allow_cosmetic_attributes=False):
...
         # Check that the SMIRNOFF version of this data structure is supported by this ForceField implementation
-        if 'version' in l1_dict:
-            version = l1_dict['version']
+        if 'version' in smirnoff_data['SMIRNOFF']:
+            version = smirnoff_data['SMIRNOFF']['version']
         else:
             raise ParseError("'version' attribute must be specified in SMIRNOFF tag")
         self._check_smirnoff_version_compatibility(str(version))
 
+        # TODO: What should we do if multiple authors or dates are read? Concatenate them?
```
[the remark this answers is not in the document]
j-wags: Cool. I don't think it's necessary to keep this TODO around and have removed it (we have too many un-resolvable TODOs hanging around anyway)

The-SMIRNOFF-force-field-format.md
j-wags: I now realize that there are no "optional tag attributes" for any parameter section tags. I'm going to remove that column from all tables, and change this to "tag attributes and default values"

openforcefield/typing/engines/smirnoff/io.py
j-wags: Good idea. I've implemented that.

openforcefield/utils/utils.py
j-wags: Agreed. That's clearer. Changed.

openforcefield/utils/utils.py
j-wags: Huh, I had never thought about exactly what was raised by assert statements. Today I learned that `AssertionError` was a thing. I've changed this to `check_units_are_compatible()`.

The-SMIRNOFF-force-field-format.md
```diff
@@ -71,6 +71,15 @@ The SMIRNOFF force field format supports versioning via the `version` attribute
 The version format is `x.y`, where `x` denotes the major version and `y` denotes the minor version.
 SMIRNOFF versions are guaranteed to be backward-compatible within the *same major version number series*, but it is possible major version increments will break backwards-compatibility.
 
+The SMIRNOFF tag versions describe the required and allowed force field-wide settings. The list of keywords is as follows:
+
+| SMIRNOFF tag version | Required attributes | Optional attributes |   |   |
+|----------------------|---------------------|---------------------|---|---|
+| 0.1                  | `aromaticity_model` | `Date`, `Author`    |   |   |
+| 0.2                  | `aromaticity_model` | `Date`, `Author`    |   |   |
+| 0.3                  | `aromaticity_model` | `Date`, `Author`    |   |   |
```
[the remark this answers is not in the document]
j-wags: I don't have a huge preference here, but I am inclined toward leaving it as-is, with three nearly-identical rows.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: > I'm not 100% sure we want to expose the cosmetic attributes as class attributes.

Maybe pedantic, but I think this implementation deals with the attributes on the instance level, and not the class level. I'd love to be corrected if I'm wrong about this.

> In XML, for example, my-attr-name is a perfectly legal attribute, but it's an illegal Python variable name.

This is a really interesting point. Thinking down this road, I wonder what kinds of trouble we might encounter when trying to read/write in other serialization formats. Can they all support the same set of attribute name strings, or is there a possibility that a name that's valid in XML won't be valid in, say, BSON? 

I suspect that Python attribute names will be the most restrictive standard -- that is, any attribute name that is valid for access as `ParameterType.attribute_name` can be serialized in any of our planned serialization formats. 

> Maybe we just want to expose a set ParameterHandler.cosmetic_attributes that can host only strings that the users can modify how they want. We can create a special class of set if we want to do some validation.

If I understand correctly, you're proposing that each `ParameterHandler`/`ParameterType` instance store a `dict` (or other mapping) called `self._cosmetic_attributes` to hold all of its key:value pairs. This would allow us to use a wider range of attributes, with names that would otherwise be invalid (like `my-attr-name`). A tradeoff we would make in this scheme is that users would need to access these attributes using `parameter.cosmetic_attributes['attrname']`, which is a bit clunkier than `parameter.attrname`

Maybe we could get the best of both worlds, by overriding the `ParameterType`/`ParameterHandler` `__getattr__` and `__setattr__` functions to first check whether actual attrs with the desired name exist, and if not, checking the `cosmetic_attribute` dict. That way, we'd provide access similar to what we have now, but also let users use otherwise-invalid attribute names by accessing the `cosmetic_attribute` dict directly.

This seems like a good direction, but I'd like to flesh it out more before we move forward. I'll turn this comment into an Issue so we can aim for it in a subsequent milestone. Until then, the current implementation should be _more_ restrictive, so we wouldn't need to break the API to transition to a more flexible scheme later.

j-wags: Opened #338 to discuss this further.

@andrrizzi, do you think it's safe to add this method to the API, or do you think it will permit a behavior that we will prohibit in the future?

andrrizzi: > I'd love to be corrected if I'm wrong about this.

You're totally right. That was poor choice of words on my part.

> Opened #338 to discuss this further.

Commented!

> do you think it's safe to add this method to the API, or do you think it will permit a behavior that we will prohibit in the future?

If we are 100% certain we'll support the dictionary syntax in the future before releasing 0.4, we might as well remove these methods to avoid breaking the API later. But if we still have doubts we can definitely keep them around until we make a decision (maybe adding a warning in the docs).

j-wags: We should continue the discussion, but I think I'll keep `add`/`delete_cosmetic_attribute` in there for the current release. When the 0.3.0 toolkit loads cosmetic attributes, they do become accessible via `ParameterType.jeffs_cosmetic_attr`, so moving to a dict would break that API.

PR #322, opened 2019-05-15
Fix #117: ForceField discovers offxml files installed through an entry point.

### LINKED ISSUE #117: Add entry point handling for plugin openforcefield group offxml search path

andrrizzi:
Fix #117.

This removes the automatic search inside the `data/forcefield` directory and implements a search through the entry point group `openff.forcefielddirs`.

Tests are not supposed to pass until we have released the `smirnoff99Frosst` package and we have converted all the references to the `smirnoff99Frosst.offxml` file to use `smirnoff99Frosst-latest.offxml` (or another version).


jchodera:
This is great!

My only request is that we use a better entry point name than `openff.forcefielddirs`, since we're going to be stuck with it for a while. It's much clearer and not at all more difficult for us to go with the more verbose `openforcefield.smirnoff_forcefield_directory`, right?

andrrizzi:
Sounds good! I'll change the entry point group name.

andrrizzi:
There's only 1 test failing, but it's a link to the new `test_forcefields` folder that doesn't exist yet in `master` so I think this should be good to go.

README.md
```diff
@@ -37,7 +37,7 @@ This repository provides tools for using the [SMIRKS Native Open Force Field (SM
 
 By convention, files containing [XML representations](https://en.wikipedia.org/wiki/XML) of SMIRNOFF force fields carry `.offxml` extensions.
 
-Example SMIRNOFF `.offxml` force field definitions can be found in [`openforcefield/data/forcefield`](https://github.com/openforcefield/openforcefield/tree/master/openforcefield/data/forcefield).
+Example SMIRNOFF `.offxml` force field definitions can be found in [`openforcefield/data/test_forcefields/`](https://github.com/openforcefield/openforcefield/tree/master/openforcefield/data/test_forcefields). These force fields are for testing only, and we neither record versions of these files, nor do we guarantee their correctness or completeness.
```
j-wags: Added a bit more on to this statement to make a clear disclaimer about toolkit FF file correctness (relevant to @jchodera )

openforcefield/tests/test_io.py
```diff
@@ -26,6 +28,12 @@ class TestXMLParameterIOHandler:
     def test_from_string(self):
         pass
 
+    def test_raise_file_not_found(self):
+        """Raise FileNotFoundError when the file doesn't exist."""
+        io_handler = XMLParameterIOHandler()
```
j-wags: Great thinking. This will be super helpful.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -55,6 +55,34 @@
...
+    """Return the list of directory paths where to search for offxml files.
+
+    This function load the information by calling all the entry points
+    registered in the "openff.forcefielddirs" group. Each entry point
+    (i.e. a function) should return a list of paths to directories
+    containing the offxml files.
+
+    Returns
+    -------
+    _installed_offxml_dir_paths : List[str]
```
j-wags: This shouldn't block the merge, just curious: What does a return value with an underscore indicate?

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -749,6 +774,25 @@ def parse_smirnoff_from_source(self, source):
...
+        if isinstance(source, str):
+            # Try first the simple path.
+            searched_dirs_paths = ['']
+            # Then try a relative file path w.r.t. an installed directory.
+            searched_dirs_paths.extend( _get_installed_offxml_dir_paths())
+            # Finally, search in openforcefield/data/.
+            # TODO: Remove this when smirnoff99Frosst 1.0.9 will be released.
+            searched_dirs_paths.append(get_data_file_path(''))
+
+            # Determine the actual path of the file.
```
j-wags: In the case that there are two files that match a given query, can we make any statement about which one will be returned? Maybe something like guaranteeing the order of entry point registration? Or can we blame any ambiguity there on user error?

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -764,18 +808,13 @@ def parse_smirnoff_from_source(self, source):
...
-                if os.path.exists(source):
-                    break
-
-            # Try parsing as a forcefield string
-            try:
-                smirnoff_data = parameter_io_handler.parse_string(source)
-                return smirnoff_data
-            except ParseError as e:
-                exception_msg = str(e)
+            except (FileNotFoundError, OSError):
```
j-wags: 👍 for clearing this part of the code up

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -909,6 +948,9 @@ def create_openmm_system(self,
         ----------
         topology : openforcefield.topology.Topology
             The ``Topology`` corresponding to the system to be parameterized
+        charge_from_molecules : List[openforcefield.molecule.Molecule], optional
```
j-wags: Thanks, this was my mistake.

openforcefield/typing/engines/smirnoff/io.py
```diff
@@ -154,74 +151,39 @@ def to_string(self, smirnoff_data):
...
     def parse_file(self, source):
         """Parse a SMIRNOFF force field definition in XML format, read from a file.
 
-        A ``ParseError`` is raised if the XML cannot be processed.
-
         Parameters
         ----------
-        source : str or file-like obj
-            File path of file-like obj specifying a SMIRNOFF force field definition in `the SMIRNOFF XML format <https://github.com/openforcefield/openforcefield/blob/master/The-SMIRNOFF-force-field-format.md>`_.
+        source : str or io.RawIOBase
```
j-wags: Cool. I'll start using `io.RawIOBase` in these situations from now on.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: This was an issue before this PR, but it could be exacerbated by allowing FFs from external sources. I've added a TODO to keep a record of this concern.

openforcefield/typing/engines/smirnoff/forcefield.py
andrrizzi: Whops! This is a typo. Thanks for catching it. I'll fix it before merging.

openforcefield/typing/engines/smirnoff/forcefield.py
andrrizzi: I'm not sure we can control it because, if it's not random, it probably depend on the order the forcefield packages will be installed. I can change the code to raise a warning or an error with a flag to load the first one to make this case a little safer.

j-wags: This is probably a complicated question, and worthy of further discussion. Like, if multiple FFs with the same name are found, but they have identical contents, maybe that's permissible? This will require some thinking about how the FF community/ecosystem will evolve. Let's open an issue to discuss, but not hold these changes up in the meantime.

PR #332, opened 2019-05-23
Fix typo in `combinePositions`

slochower:
Unclear to me whether this PR is useful, given the Slack discussion on #maintainers, but this does fix a typo that addresses a bug in `mergeStructure`.


j-wags:
Bizarrely, this is causing some Molecule tests in our Linux-with-both-cheminformatics-toolkits builds to fail. I can't think of any reason this change would have caused that. 

It's possible that this is being caused by some dependencies updating. I don't have the bandwidth to dig into this in the near future, so let's see if tonight's `master` builds have the same problem.

slochower:
Definitely odd.

j-wags:
Update: last night's `master` builds passed all tests, restarting passing builds in this PR yields more passes, restarting the two failing builds yields more failures. 

What a mystery. Let's keep this PR open.

slochower:
Shouldn't there be more information after this line?
https://travis-ci.org/openforcefield/openforcefield/jobs/536534798#L1263

j-wags:
Yup. 

Potentially related -- I may have seen silent failures like this the other week, when @MSchauperl's RESPPOL builds weren't correctly decrypting their OE license. It's unlikely that that's the case here, though.

j-wags:
This PR is making me question the fabric of reality. I'm tagging this as a "good first issue", since we don't really use this function, but I can't in good conscience close this PR.

mattwthompson:
I suppose this mysterious behavior could be hidden away with #648 👻

mattwthompson:
I'm closing this now that the module is deprecated and removed. May the mystery live on forever ...

PR #347, opened 2019-06-08
bugfix for cb6 parameterization issue

j-wags:
@slochower and @SimonBoothroyd 's work has been held up by a bug that I still totally don't understand, but was caused by code where we intended to make a copy of simtk.unit.Quantity-wrapped object, but instead wound up with a second reference to the original. Generally, I think there's a neighborhood of unexpected behavior/problems arising from my incomplete understanding of Quantity and its behavior. I hope that refactoring in Pint will make some of these relationships clearer.


slochower:
Thanks guys!

PR #349, opened 2019-06-09
Refactor ParameterType 

andrrizzi:
In this PR, I partially implemented #207 and refactored `ParameterType` to use a descriptor-based design to make it a little easier to implement new types and added a lot of examples in the docstrings that document how to do it.

I will refactor `ParameterHandler` and update the changelog in a separate PR as this is already big as it is, and it is self-contained. Plus, I want to make sure you approve this design before proceeding further.

In any case, I'll wait for 0.4.0 to be released before merging.


andrrizzi:
I realized after a bit that I basically reimplemented a subset of the features implemented in the [attrs library](https://www.attrs.org/en/stable/), which was suggested to us by Kyle Beauchamp some time ago. If we're ok adding it as a dependency and start using it, I can also refactor things to inherit from the descriptors and decorators implemented there to simplify the code.

andrrizzi:
I'm getting a lot of errors related to the `ParameterHandler`s supported versions, but I'm not sure these are related to my changes.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -293,161 +585,257 @@ def to_list(self, discard_cosmetic_attributes=True):
...
+                   f" Defined kwargs are {sorted(smirnoff_data.keys())}")
+            raise SMIRNOFFSpecError(msg)
+
+        # Finally, set attributes of this ParameterType and handle cosmetic attributes.
+        allowed_attributes = set(self._get_parameter_attributes().keys())
         for key, val in smirnoff_data.items():
-            if key in allowed_attribs:
-                # TODO: create @property.setter here if attrib requires unit
+            if key in allowed_attributes:
                 setattr(self, key, val)
```
j-wags: Ahh, so, in this case, `setattr` will send `val` through the `ParameterAttribute`s setter validator. Very cool.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -471,19 +858,20 @@ def to_dict(self, discard_cosmetic_attributes=False):
 
         """
         # Make a list of all attribs that should be included in the
-        # returned dict (call list() to make a copy)
-        attribs_to_return = list(self._REQUIRED_SPEC_ATTRIBS)
-        attribs_to_return += [opt_attrib for opt_attrib in self._OPTIONAL_SPEC_ATTRIBS if hasattr(self, opt_attrib)]
+        # returned dict (call list() to make a copy). We discard
```
j-wags: ~I think this is a behavior change from previous versions. Off the top of my head, I think it should only affect `(Im)ProperTorsion` `idivf` values (which are set to `auto` if not specified), but we should note it in the release notes.~

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2519,14 +2937,9 @@ class ChargeIncrementType(ParameterType):
...
         _ELEMENT_NAME = 'ChargeIncrement'
-        _REQUIRED_SPEC_ATTRIBS = ['smirks', 'chargeIncrement']
-        _REQUIRE_UNITS = {
-            'chargeIncrement': unit.elementary_charge
-        }
-        _INDEXED_ATTRIBS = ['chargeIncrement']
 
-        def __init__(self, node, parent):
-            super().__init__(**kwargs)
+        charge_increment = IndexedParameterAttribute(unit=unit.elementary_charge)
```
j-wags: Thanks for standardizing this -- The switch to `charge_increment` will make this consistent with fields like `aromaticity_model` and `combining_rules`. We'll need to update the SMIRNOFF spec docs with the new syntax (Oh boy, this means a `ChargeIncrement` section version update too!)

openforcefield/utils/utils.py
```diff
@@ -319,14 +327,14 @@ def convert_all_strings_to_quantity(smirnoff_data):
...
+
     else:
         try:
             obj_to_return = object_to_quantity(smirnoff_data)
-        except AttributeError:
-            obj_to_return = smirnoff_data
-        except TypeError:
-            obj_to_return = smirnoff_data
-        except SyntaxError:
+        except (AttributeError, TypeError, SyntaxError):
```
j-wags: 👍

openforcefield/utils/utils.py
```diff
@@ -365,27 +373,40 @@ def convert_all_quantities_to_string(smirnoff_data):
     return obj_to_return
 
 
+@functools.singledispatch
```
j-wags: It's cool to see advanced decorators used properly. This is the sort of feature that's so complex that I've been on the fence about using it, since it may be hard to debug, and could scare potential contributors away. I'm happy to have it in this part of the code as a trial run. 

This could be a bad philosophy for me to have, and I'm open to discussion on it.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -117,6 +111,304 @@ class NonbondedMethod(Enum):
...
+    Examples
+    -------
+
+    Create a parameter type with an optional and a required attributes.
+
+    >>> class MyParameter:
+    ...     attr_required = ParameterAttribute()
+    ...     attr_optional = ParameterAttribute(default=2)
+    ...
+    >>> my_par = MyParameter()
```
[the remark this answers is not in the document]
j-wags: Ahhh, I just finished the rest of the review, and now I see that required `ParameterAttribute`s are checked by the `ParameterType` that owns them. Since `class MyParameter` doesn't inherit from `ParameterType`, `attr_required` isn't really required. 

The docstring does become a bit confusing when we use the word "required" like this. Maybe we could change it to something like `normal_attr = ParameterAttribute()` `attr_with_default = ParameterAttribute(default=2)`?

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Actually, we will want the _opposite_ behavior when we work with `ParameterHandler` -- That is, if the input didn't specify an `Electrostatics` `cutoff` value, we absolutely want to write it out explicitly. To keep things consistent, I think we should reverse this decision, and have the toolkit write out every default value that was assumed.

openforcefield/tests/test_parameters.py
```diff
@@ -1,29 +1,195 @@
...
+
+    def test_handle_single_value(self):
+        """IndexedParameterAttribute handle also single values instead of sequences."""
+        class MyParameter:
+            attr_indexed_converter = IndexedParameterAttribute(converter=str)
+            attr_indexed_unit = IndexedParameterAttribute(unit=unit.meter)
+        my_par = MyParameter()
+
+        my_par.attr_indexed_converter = 3.0
+        assert my_par.attr_indexed_converter == '3.0'
```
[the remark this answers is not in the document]
andrrizzi: I agree. I implemented this because I believe this is the current behavior of `ParamenterType`, but I'm very ok changing it.

openforcefield/tests/test_parameters.py
```diff
@@ -300,6 +466,105 @@ def test_round_trip(self):
...
+            b = IndexedParameterAttribute()
+        with pytest.raises(TypeError, match="indexed attributes have different lengths"):
+            MyParameter(smirks='[*:1]', a1=1, a2=2, a3=3, b1=1, b2=2)
+
+    def test_error_single_value_plus_index(self):
+        """ParameterType raises an error if an indexed attribute is specified with and without index."""
+        class MyParameter(ParameterType):
+            a = IndexedParameterAttribute()
+        with pytest.raises(TypeError, match="'a' has been specified with and without index"):
+            MyParameter(smirks='[*:1]', a=1, a1=2)
```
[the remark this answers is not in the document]
andrrizzi: I've implemented it this way only because this is the current behavior and I didn't want to break the API in this PR, but I agree that passing a list is more "pythonic" than passing `a1`, `a2`, which makes sense only for serialization. In fact, I think we could even consider taking only `a` as a list in the normal constructor, and leave the `a1`, `a2`, ... syntax for a `from_dict()` constructor instead for serialization. It seems to me a more standard way of dealing with arguments in Python.

> `a` as a list, we should write a test for it as well.

Good thinking! I'll add it.

openforcefield/typing/engines/smirnoff/parameters.py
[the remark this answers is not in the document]
andrrizzi: Yeah, I don't think there's a way for descriptors to verify they've been set when the class that owns them is initialized. I'll fix the examples.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -117,6 +111,304 @@ class NonbondedMethod(Enum):
...
+    >>> my_par.length is None
+    True
+
+    Strings are parsed into Quantity objects.
+
+    >>> my_par.length = ['1 * angstrom', 0.5 * unit.nanometer]
+    >>> my_par.length[0]
+    Quantity(value=1, unit=angstrom)
+
+    Note that ``IndexedParameterAttribute`s are immutable right now. This
```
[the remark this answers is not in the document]
andrrizzi: Sounds good! We can decide to implement this here or in a separate PR so that it will be smaller to review. It's up to you. I'll want to add a bunch of tests for the new class as I think I'll have to use something similar also in the new `System`.

openforcefield/typing/engines/smirnoff/parameters.py
andrrizzi: Ah! This was kind of an accident. I can revert it for this PR and we can change it later if you prefer.

openforcefield/utils/utils.py
andrrizzi: I can revert to `if...else` if you prefer, no problem. The advantages of the dispatch design is that it's easier to extend (it can even be extended in a different module through `register`) and it's slightly faster. I take it for granted at this point, but I agree this feature is not widely used and we may not need the advantages for this function. I'll leave it up to your decision.

andrrizzi:
Thanks for the feedback! I'll make the fixes soon. Let me also know what you think of `attrs`. Ideally, I'd like us to take a decision before working on `ParameterHandler` as it'll save me some time.

openforcefield/typing/engines/smirnoff/parameters.py
andrrizzi: I missed the first case, and I agree with the second. I'll change this, thank you! Let's make sure to tell @SimonBoothroyd as this will change which attributes are exported in the forcefield tags and thus the hash that is constructing.

openforcefield/tests/test_parameters.py
```diff
@@ -1,29 +1,195 @@
...
+        """IndexedParameterAttribute handle also single values instead of sequences."""
+        class MyParameter:
+            attr_indexed_converter = IndexedParameterAttribute(converter=str)
+            attr_indexed_unit = IndexedParameterAttribute(unit=unit.meter)
+        my_par = MyParameter()
+
+        my_par.attr_indexed_converter = 3.0
+        assert my_par.attr_indexed_converter == '3.0'
+        my_par.attr_indexed_unit = '1.0*nanometer'
+        assert my_par.attr_indexed_unit == 1.0 * unit.nanometer
```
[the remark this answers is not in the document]
andrrizzi: Done in 95cacc6.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -117,6 +111,304 @@ class NonbondedMethod(Enum):
...
+        to be set, and string expressions are automatically parsed into a
+        ``Quantity``.
+    converter : callable, optional
+        An optional function that can be used to convert values before
+        setting the attribute.
+
+    Examples
+    -------
+
+    Create a parameter type with an optional and a required attributes.
```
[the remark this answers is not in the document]
andrrizzi: Thank you! Fixed.

openforcefield/typing/engines/smirnoff/parameters.py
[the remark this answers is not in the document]
andrrizzi: I've just realized that I'll have to encapsulate the initialization utilities in a separate class/decorator shared between `ParameterType` and `ParameterHandler` so I'll fix it by making `MyParameter` using that class.

openforcefield/tests/test_parameters.py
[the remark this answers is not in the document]
andrrizzi: Although, now that I think about it, it might be a little weird for `BondType.k`, which is currently an indexed attribute.

openforcefield/utils/utils.py
j-wags: Thanks for the details. Let’s keep singledispatch. :-)

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: This PR is already big enough without getting a spec change mixed in. Let’s revert this to ‘chargeincrement’ and open an issue to do this later.

openforcefield/utils/utils.py
andrrizzi: Ok!

openforcefield/typing/engines/smirnoff/parameters.py
andrrizzi: Sounds good! Done in 6028828.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Good call. I just spoke with @SimonBoothroyd about this. Basically, in his hash/equality comparisons, _False Negatives are OK_ (they may degrade performance a little, but won't trigger a new simulation to be run or bad science to happen). So he's now aware of this toolkit behavior and will design around it.

openforcefield/tests/test_parameters.py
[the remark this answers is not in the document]
andrrizzi: Although, now that I think about it, it might be a little weird for `BondType.k`, which is currently an indexed attribute.

openforcefield/typing/engines/smirnoff/parameters.py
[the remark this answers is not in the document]
j-wags: Let's include it in this PR. This would be so easy to forget (or lose in the issue tracker) that I don't want to have `master` in a temporarily API-broken state.

openforcefield/tests/test_parameters.py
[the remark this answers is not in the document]
j-wags: > In fact, I think we could even consider taking only a as a list in the normal constructor, and leave the a1, a2, ... syntax for a from_dict() constructor instead for serialization. It seems to me a more standard way of dealing with arguments in Python.

Agreed. I would be happy with an implementation which, at least for now, only accepts a list (or  iterable) here.

openforcefield/tests/test_parameters.py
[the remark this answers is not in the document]
j-wags: Ahhhh. `BondType.k` _IS_ a tough situation. I had to look into this.

The good thing is, [I had implemented it incorrectly to begin with](https://open-forcefield-toolkit.readthedocs.io/en/0.4.0/smirnoff.html#fractional-bond-orders-experimental). `BondType.k` is never indexed. If we're interpolating bond orders, we will be provided with indexed attributes `k_bondorder#` and `length_bondorder#`. We'll need to validate that we're not provided with both those and an unindexed `k` value, but at least there's not ambiguity over which attributes are indexed.

We do not need to implement support for `k_bondorder#` and `length_bondorder` right now. So let's just have `k` be a normal `ParameterAttribute`.

openforcefield/typing/engines/smirnoff/parameters.py
andrrizzi: How about this: 4ac9e06. Basically, I'm not serializing attributes that are set to a default value only if their default value is `None`. This means we are discarding attributes like `id` and `parent_id` when they are not set, but catching attributes like `cutoff`.

j-wags: I like this. So basically, the difference between a required-attribute-with-a-default and an optional attribute is whether `default=val` or `default=None`. This seems like a good decision to me.

andrrizzi: More or less. They're both optional attributes when we read the forcefield, meaning that we don't have to specify them in the constructor, only we export only non-trivial defaults. The choice is just to avoid having to export a bunch of `parent_id=null` attributes for each tag.

andrrizzi:
@j-wags, I think I've addressed everything and this is ready to be merged.

I've started to implement the callback mechanism in `openforcefield.utils.callback`, but I realized soon that the way I was going to use it would violate the observer pattern so I've implemented a `VerifiedList` class in `openforcefield.utils.collections` that solves the problem.

Eventually, I expect we'll want to make `VerifiedList` a `Callbackable` class as it will be very useful to safely expose through the API so I've left the `callback` module in, but this is something we can do separately.

openforcefield/utils/callback.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Examples
+    --------
+
+    We define a callbackable list that allows registering callback
+    functions to methods that change the list. The first four methods
+    summarize the different ways in which the ``callback_method``
+    decorator can be used to mark a method for callback registration.
+
+    >>> class CallbackableList(Callbackable, list):
```
j-wags: Super clear docstring. This was really helpful in making me understand what this does.

openforcefield/utils/callback.py
```diff
@@ -0,0 +1,198 @@
...
+
+    See Also
+    --------
+    Callbackable
+
+    """
+
+    if func is None and len(events) > 0:
+        # Decorator called with @callback_method(events=[...]).
+        return functools.partial(callback_method, events=events)
```
j-wags: I'm not sure that I understand when this is triggered, but I'm glad we have defined behavior for it.

openforcefield/utils/utils.py
```diff
@@ -49,24 +51,27 @@
...
 #=============================================================================================
 
-
 class MessageException(Exception):
     """A base class for exceptions that print out a string given in their constructor"""
     def __init__(self, msg):
         super().__init__(self, msg)
         self.msg = msg
 
+    def __str__(self):
```
j-wags: 👍

openforcefield/utils/collections.py
```diff
@@ -0,0 +1,141 @@
...
+# =====================================================================
+
+from collections import abc
+
+
+# =====================================================================
+# VALIDATED LIST
+# =====================================================================
+
+class ValidatedList(list):
```
j-wags: Aha! Ok, now I understand what you were talking about regarding the shortcomings of `Callbackable`. Yes, this is the right way to go forward for now.

openforcefield/utils/callback.py
andrrizzi: This is triggered when the decorator is used as
```python
@callback_method(events=['my_event])
def func():
    ...
```
instead of
```python
@callback_method
def func():
    ...
```

andrrizzi:
Thank you for the review!

PR #351, opened 2019-06-13
Fix #344

### LINKED ISSUE #344: Generate OFFMol molecule from OEMol is failing
Hi there, 

I was trying to generate an OFF molecule starting from an OpenEye molecule. I’m using the openforcefield v 0.3.0 installed from conda on Mac Osx. Unfortunately, I’m getting the following error:
```python
  File "stereo.py", line 9, in <module>
    molecule = Molecule(oemol, allow_undefined_stereo=True)
  File "/Users/gcalabro/local/miniconda3/envs/smirnoff_new/lib/python3.6/site-packages/openforcefield/topology/molecule.py", line 3433, in __init__
  File "/Users/gcalabro/local/miniconda3/envs/smirnoff_new/lib/python3.6/site-packages/openforcefield/topology/molecule.py", line 1517, in __init__
    result = toolkit_registry.call('from_object', other)
  File "/Users/gcalabro/local/miniconda3/envs/smirnoff_new/lib/python3.6/site-packages/openforcefield/utils/toolkits.py", line 2964, in call
    return method(*args, **kwargs)
  File "/Users/gcalabro/local/miniconda3/envs/smirnoff_new/lib/python3.6/site-packages/openforcefield/utils/toolkits.py", line 364, in from_object
    return self.from_openeye(object)
  File "/Users/gcalabro/local/miniconda3/envs/smirnoff_new/lib/python3.6/site-packages/openforcefield/utils/toolkits.py", line 749, in from_openeye
openforcefield.utils.toolkits.UndefinedStereochemistryError: (UndefinedStereochemistryError(...), 'Unable to make OFFMol from OEMol: OEMol has unspecified stereochemistry. oemol.GetTitle(): DrugBank_5354\nProblematic atoms are:\nAtom atomic num: 6, name: C16, idx: 15, aromatic: False, chiral: True with bonds:\nbond order: 1, chiral: False to atom atomic num: 6, name: C15, idx: 14, aromatic: False, chiral: False\nbond order: 1, chiral: False to atom atomic num: 15, name: P1, idx: 23, aromatic: False, chiral: False\nbond order: 1, chiral: False to atom atomic num: 16, name: S1, idx: 24, aromatic: False, chiral: False\nbond order: 1, chiral: False to atom atomic num: 1, name: H16, idx: 40, aromatic: False, chiral: False\n')
```
The script that I’m using is quite simple:

```python
from openeye import oechem

oemol = oechem.OEMol()

with oechem.oemolistream("5354.oeb") as ifs:
    oechem.OEReadMolecule(ifs, oemol)

molecule = Molecule(oemol, allow_undefined_stereo=True)
```

The flag ```allow_undefined_stereo``` does not seem to have any effect. I have attached the zipped offending molecule as well.

[5354.oeb.zip](https://github.com/openforcefield/openforcefield/files/3258917/5354.oeb.zip)

j-wags:
* Fixes #344 
* adds `allow_undefined_stereo` kwargs to `Molecule.from_smiles` and `Molecule.from_object`
* refactors test_molecule `test_to_from_openeye` and `test_to_from_rdkit` to be more explicit and test a wider range of behavior. 
* refactors test_toolkits `test_smiles_missing_stereochemistry` to test a wider range of behavior.

### DIFF
--- a/openforcefield/topology/molecule.py
+++ b/openforcefield/topology/molecule.py
@@ -1834,7 +1834,10 @@ class FrozenMolecule(Serializable):
         return smiles
 
     @staticmethod
-    def from_smiles(smiles, hydrogens_are_explicit=False, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY):
+    def from_smiles(smiles,
+                    hydrogens_are_explicit=False,
+                    toolkit_registry=GLOBAL_TOOLKIT_REGISTRY,
+                    allow_undefined_stereo=False):
         """
         Construct a Molecule from a SMILES representation
 
@@ -1846,6 +1849,10 @@ class FrozenMolecule(Serializable):
             If False, the cheminformatics toolkit will perform hydrogen addition
         toolkit_registry : openforcefield.utils.toolkits.ToolRegistry or openforcefield.utils.toolkits.ToolkitWrapper, optional, default=None
             :class:`ToolkitRegistry` or :class:`ToolkitWrapper` to use for SMILES-to-molecule conversion
+        allow_undefined_stereo : bool, default=False
+            Whether to accept SMILES with undefined stereochemistry. If False,
+            an exception will be raised if a SMILES with undefined stereochemistry
+            is passed into this function.
 
         Returns
         -------
@@ -1858,10 +1865,16 @@ class FrozenMolecule(Serializable):
 
         """
         if isinstance(toolkit_registry, ToolkitRegistry):
-            return toolkit_registry.call('from_smiles', smiles)
+            return toolkit_registry.call('from_smiles',
+                                         smiles,
+                                         hydrogens_are_explicit=hydrogens_are_explicit,
+                                         allow_undefined_stereo=allow_undefined_stereo)
         elif isinstance(toolkit_registry, ToolkitWrapper):
             toolkit = toolkit_registry
-            return toolkit.from_smiles(smiles, hydrogens_are_explicit=hydrogens_are_explicit)
+            return toolkit.from_smiles(smiles,
+                                       hydrogens_are_explicit=hydrogens_are_explicit,
+                                       allow_undefined_stereo=allow_undefined_stereo
+                                       )
         else:
             raise Exception(
                 'Invalid toolkit_registry passed to from_smiles. Expected ToolkitRegistry or ToolkitWrapper. Got  {}'
--- a/openforcefield/topology/molecule.py
+++ b/openforcefield/topology/molecule.py
@@ -1514,7 +1514,7 @@ class FrozenMolecule(Serializable):
             # Check through the toolkit registry to find a compatible wrapper for loading
             if not loaded:
                 try:
-                    result = toolkit_registry.call('from_object', other)
+                    result = toolkit_registry.call('from_object', other, allow_undefined_stereo=allow_undefined_stereo)
                 except NotImplementedError:
                     pass
                 else:
--- a/openforcefield/topology/molecule.py
+++ b/openforcefield/topology/molecule.py
@@ -1835,7 +1835,7 @@ class FrozenMolecule(Serializable):
         func_qualname = to_smiles_method.__qualname__
 
         # Check to see if a SMILES for this molecule was already cached using this method
-        if func_qualname in self._cached_smiles.keys():
+        if func_qualname in self._cached_smiles:
             return self._cached_smiles[func_qualname]
         else:
             smiles = to_smiles_method(self)
--- a/openforcefield/topology/molecule.py
+++ b/openforcefield/topology/molecule.py
@@ -1815,23 +1815,33 @@ class FrozenMolecule(Serializable):
         >>> smiles = molecule.to_smiles()
 
         """
-        smiles = self._cached_smiles
+        # Initialize cached_smiles dict for this molecule if none exists
+        if self._cached_smiles is None:
+            self._cached_smiles = {}
 
-        if smiles is not None:
-            return smiles
-
+        # Figure out which toolkit should be used to create the SMILES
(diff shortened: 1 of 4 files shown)


jchodera:
There are a bunch of failing tests in the OpenEye version like [this one](https://travis-ci.org/openforcefield/openforcefield/jobs/545052720#L1422-L1471). 

Looks like the `pip install` is now [installing the 2019.4.2. toolkit](https://travis-ci.org/openforcefield/openforcefield/jobs/545052720#L690-L705) that was released [30 Apr 2019](https://www.eyesopen.com/news/openeye-toolkits-v2019.apr), so it doesn't appear to be due to a recent toolkit update.

j-wags:
Very interesting. When I only run that test by itself, it passes. My guess is that some test that runs before must be modifying the cached test molecules. Looking into this.

j-wags:
Ahh, fascinating. #308 had us start caching SMILES to speed up hashing, but a molecule's SMILES is dependent on the toolkit used to make it. So, in our test suite, the RDKit tests run first, creating RDKit SMILES for all molecules. Then the OE tests run, but they recover the cached SMILES from RDKit and skip generating OE-format SMILES.

I think this is actually a bug that could arise from use of the pubic API. We do publicly expose the `toolkit_registry` kwarg in many functions, so I would say that returning an RDKit SMILES after the user requested an OE SMILES is pretty squarely a bug on our side.

This issues highlights the fact that, moving forward, we will need to be stricter about defining what exactly a "cached property" is. We originally intended it to mean "dependent on the molecular graph", so cache invalidations would occur in a pretty strict set of circumstances (that is, graph changes). However, an OFFMol's SMILES is not uniquely a function of `(atoms, bonds, stereochemistry)` -- it is also dependent on `(toolkit)`. 

I could see two ways forward here:
* Don't cache SMILES, which will increase system creation time by 50-100% according to some quick tests.
* Have `Molecule._cached_smiles` be a dict of the form `{ToolkitRegistry._toolkit_name: SMILES}`

I think that toolkit runtime is pretty important, so I'm partial to the latter option. I'll add this for review momentarily.

jchodera:
Some possible solutions:
* Refactor the code and tests to make it OK if two different toolkits are used---caching a SMILES representation should only help speed things up, after all. If we end up caching both RDKit and OpenEye SMILES, that's fine---we do 2x the effort, but still see a speedup.
* Switch from using SMILES (which is toolkit-dependent) for caching to something easily available but more universal, like InChI: [RDKit](https://www.rdkit.org/docs/source/rdkit.Chem.inchi.html) [InChI](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEMolToSTDInChI.html) 
* Use something else for caching, like Hill formulas

j-wags:
> Refactor the code and tests to make it OK if two different toolkits are used---caching a SMILES representation should only help speed things up, after all. If we end up caching both RDKit and OpenEye SMILES, that's fine---we do 2x the effort, but still see a speedup.

This is the safest option, and what I've implemented.

> Switch from using SMILES (which is toolkit-dependent) for caching to something easily available but more universal, like InChI: RDKit InChI

I think InChI is still "toolkit dependent", in cases where OE and RDK interpret the stereochemistry or aromaticity of the same molecule differently. This would be a _pretty good_ option, but I think the first option is better. 

> Use something else for caching, like Hill formulas

I'd thought about caching Hill formulas, but the toolkit internally uses SMILES to generate molecule hashes, so that would lead to some problematic key collisions  (isopropanol vs. propanol)

openforcefield/topology/molecule.py
```diff
@@ -1815,26 +1815,39 @@ def to_smiles(self, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY):
...
 
-        self._cached_smiles = smiles
-        return smiles
+        # Get a string representation of the function containing the toolkit name so we can check
+        # if a SMILES was already cached for this molecule. This will return, for example
+        # "RDKitToolkitWrapper.to_smiles"
+        func_qualname = to_smiles_method.__qualname__
+
+        # Check to see if a SMILES for this molecule was already cached using this method
+        if func_qualname in self._cached_smiles.keys():
```
andrrizzi: I'd remove the `keys()` with the `in` operator.

jchodera:
> I think InChI is still "toolkit dependent", in cases where OE and RDK interpret the stereochemistry or aromaticity of the same molecule differently. This would be a pretty good option, but I think the first option is better.

Our assumption is that both toolkits implement the same variant MDL aromaticity model, right?

openforcefield/topology/molecule.py
j-wags: Oh man, I totally thought I'd read some python docs that indicated that `in` checks both keys and values, and I was going to be all like "I know more than Andrea about one thing!". 

Except now I can't find it, and the Python docs unambiguously agree with you. It must have been a dream :-/

Reverting...

j-wags:
> Our assumption is that both toolkits implement the same variant MDL aromaticity model, right?

@jchodera That is our assumption, but I'm not 100% convinced (though I don't have hard data immediately available to back up any discrepancy). I'd prefer to assume "unsafe" in this case.

davidlmobley:
For the record:

>> Our assumption is that both toolkits implement the same variant MDL aromaticity model, right?

> @jchodera That is our assumption, but I'm not 100% convinced (though I don't have hard data immediately available to back up any discrepancy). I'd prefer to assume "unsafe" in this case.

I am quite confident there will be edge cases where they do not have exactly the same aromaticity model still. A lot of the back and forth about the definition of aromaticity models -- once the fundamentals are documented in place -- amounts to questions like, "Does it also consider X aromatic?" "What about Y?" We've basically gone through the alphabet once and all the "common" cases agree but may still disagree about edge cases. We will need to document any which come up and work with developers to get them corrected.

PR #357, opened 2019-06-14
Topology.from_openmm raises informative error if origin PDB lacks CONECT

### LINKED ISSUE #356: Provide more useful error message if topology.from_openmm has connectivity problems
@j-wags: I wonder if we can provide a more useful error message in cases like this. Perhaps there's an easy way to detect when the OpenMM Topology is contains single atoms that are not generally found as individual atoms, and to call this helpful Topology inspector when we fail to find a match for our molecule?

_Originally posted by @jchodera in https://github.com/openforcefield/openforcefield/issues/355#issuecomment-502008715_

j-wags:
- [x] Fixes #356 
- [x] Adds test to ensure informative error is thrown


openforcefield/topology/topology.py
```diff
@@ -1660,7 +1660,14 @@ def from_openmm(cls, openmm_topology, unique_molecules=None):
...
             if not (match_found):
                 hill_formula = Topology._networkx_to_hill_formula(omm_mol_G)
-                raise ValueError(f'No match found for molecule {hill_formula}')
+                msg = f'No match found for molecule {hill_formula}. '
+                probably_missing_conect = ['C1', 'H1', 'O1', 'N1', 'P1', 'S1', 'F1', 'Cl1', 'Br1']
+                if hill_formula in probably_missing_conect:
+                    msg += 'This would be a very unusual molecule to try and parameterize, ' \
+                           'and it is likely that the data source it was read from does not ' \
+                           'contain connectivity information. If this molecule is coming from ' \
+                           'PDB, please ensure that the file contains CONECT lines.'
```
jchodera: They're called `CONECT` "records" or "cards", I believe.
We might also cite the PDB format: https://www.wwpdb.org/documentation/file-format-content/format33/sect10.html

The important part of the PDB format is:
```
CONECT records are mandatory for HET groups (excluding water) and for other bonds not specified in the standard residue connectivity table.
```
That's the key here.

openforcefield/topology/topology.py
```diff
@@ -1660,7 +1660,14 @@ def from_openmm(cls, openmm_topology, unique_molecules=None):
                     break
             if not (match_found):
                 hill_formula = Topology._networkx_to_hill_formula(omm_mol_G)
-                raise ValueError(f'No match found for molecule {hill_formula}')
+                msg = f'No match found for molecule {hill_formula}. '
+                probably_missing_conect = ['C1', 'H1', 'O1', 'N1', 'P1', 'S1', 'F1', 'Cl1', 'Br1']
+                if hill_formula in probably_missing_conect:
+                    msg += 'This would be a very unusual molecule to try and parameterize, ' \
```
andrrizzi: In case you haven't stumble on this yet, there's an alternative suggested way of splitting strings on multiple lines with parentheses
```python
msg += ('This would be a very unusual molecule to try and parameterize, '
        'and it is likely that the data source it was read from does not ')
```

j-wags: Oh nice, converted to that format.

openforcefield/topology/topology.py
j-wags: Added link to the PDB spec. Happy to do my part in the war for the PDB file format :-)

j-wags: Also changed working to "CONECT records"

Issue #290, opened 2019-04-25
Implement GBSAParameterHandler


j-wags:
Our [GBSA implementation](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/typing/engines/smirnoff/parameters.py#L2306) is close to functional. If we're happy with the [spec section](https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#gbsa), we can go ahead and finish implementing this in a near-term feature release milestone.

proteneer:
What's missing in terms of implementation?

proteneer:
On the spec side of things, is it intended that GBSA charges will be inherited from the nonbonded charges? AFAIK OpenMM's system class doesn't actually require NonbondedForce charges to be the same as that of GBSA charges.

j-wags:
> What's missing in terms of implementation?

* Refactoring `GBSAHandler` and `GBSAType` to inherit the `ParameterHandler` base class 
* Implementing checks to ensure valid combinations of parameters are used to initialize `GBSAHandler` and `GBSAType`
* Finding a better long-term solution for assigning/passing charges between different `ParameterHandler`s
* Tests and examples

> On the spec side of things, is it intended that GBSA charges will be inherited from the nonbonded charges? AFAIK OpenMM's system class doesn't actually require NonbondedForce charges to be the same as that of GBSA charges

Good question. My own thought is that the charges should be inherited, but I'll tag @jchodera and @davidlmobley to see if that's how they see it, too.

davidlmobley:
The thought is that they would inherit the nonbonded charges AFAIK. 

Is there a use case for GBSA charges to be separate? Or something having them separate would enable that can't be accomplished by having them inherit?

I suppose I can see someone wanting to fit GBSA-specific charges, but... presumably if they were doing that they're basically buidling a GBSA-specific force field and I don't necessarily see that we need to allow them to retain their original force field in the same file. So inheritance seems good to me.

proteneer:
I'm also +1 on inheriting them. Though one corner case might be where you intentionally turn off GBSA for certain parts of the system and use vanilla electrostatics. But even then you probably want to use the same charges.

j-wags:
Whoops, this should have been closed with PR #363

PR #363, opened 2019-06-19
Implement GBSA handling

j-wags:
### Summary

- [x] #290 -- Implement `GBSAParameterHandler`
- [ ] Add example of creating system with GBSA force
  - [ ] Determine best practices for converting GBSA-utilizing systems between simulation packages
- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openforcefield/tree/master/docs)

### Status
 - [x] Ready for review

### Open questions

* ~Do we want to offer "certified" OFFXML implementations of OBC1, OBC2, and HCT? Basically equivalent to making `bondi`, `mbondi`, and `mbondi2` implementations of our own.~
    * ~These seem simple, but we should attach version numbers to them in case we mess up~
    * These are now created and have version numbers attached
* GBSAOBCForce nonbonded methods are currently restricted to `CutoffPeriodic`, `CutoffNonPeriodic`, and `NoCutoff`. What behavior is desired when this is set alongside a NonbondedForce set to use `PME`? Right now, I let them be different, but I suspect that this is dangerous, and will need to be cleaned up. The alternative is to override the `Electrostatics` and `vdW` tag settings, or throw an Exception if they want a NonbodedMethod that GBSA can't match.
* Should the SMIRNOFF spec expose `alpha`, `beta`, `gamma`? These are basically redundant in meaning with `OBC1` vs. `OBC2`.
* Should the SMIRNOFF spec expose or `salt_concentation`/`kappa`?

### Design Notes and Assumptions 
_Do not delete -- Linked to in the relevant code_

**HCT**
* Implementation: `simtk.openmm.app.internal.customgbforces.GBSAHCTForce`
* Parameter validation: Using OpenMM
* Energy validation: Using OpenMM
* SMIRNOFF parameters not exposed in object: surface_area_energy, solvent_radius
    * These values are hard-coded. We assert that the FF has matching values during system creation.
* Object parameters not in SMIRNOFF spec: kappa/salt conc
* Assumptions I've made:
    * What is the hard-coded surface area energy? Probably 5.4 cal/mol/A^2  [4][6]
    * What is the hard-coded solvent radius? (probably 1.4 A, per [1])
* WARNING: In this model `addParticle` and `setParticleParameters` expect DIFFERENT PARAMETERS. `addParticle` takes radius and screen, while `sPP` takes `radius-offset` and `scale*(radius-offset)`

**OBC1**
* Implementation: `simtk.openmm.app.internal.customgbforces.GBSAOBC1Force`
* Parameter validation: Using OpenMM
* Energy validation: Using OpenMM
* SMIRNOFF parameters not exposed in object: surface_area_energy, solvent_radius
    * These values are hard-coded. We assert that the FF has matching values during system creation.
* Object parameters not in SMIRNOFF spec: kappa/salt conc
* Assumptions I've made:
    * What is the hard-coded surface area energy? Probably 5.4 cal/mol/A^2 [4][6]
    * What is the hard-coded solvent radius? (probably 1.4 A, per [1])
* Note: This has the same functional form as `simtk.openmm.GBSAOBCForce`, there are just some internal values (`alpha`, `beta`, `gamma`) that need to be changed to turn OBC1 into OBC2 [5]. However, as far as I can tell, no model exposes alpha, beta, or gamma. Also, only classes derived from CustomAmberGBSAForce allow kappa (/salt conc) to change.
* WARNING: In this model `addParticle` and `setParticleParameters` expect DIFFERENT PARAMETERS. `addParticle` takes `radius` and `screen`, while `sPP` takes `radius-offset` and `scale*(radius-offset)`

**OBC2**
* Implementation: `simtk.openmm.GBSAOBCForce`
    * I'm choosing this over `simtk.openmm.app.internal.customgbforces.GBSAOBC2Force` because it lets us set the solvent radius. Beware, however -- This class behaves very differently from the first two.
* Parameter validation: Using OpenMM
* Energy validation: Using OpenMM
* Hard-coded vals: kappa/salt conc (=0) [3], solvent_radius=1.4 [2]
* SMIRNOFF parameters not exposed in object: solvent_radius
   * We could just figure out which values are hard-coded here, and ensure that the FF has matching values during input.

[1] https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/internal/customgbforces.py#L373
[2] http://docs.openmm.org/latest/userguide/theory.html#surface-area-term
[3] It's tempting to think that kappa could be pre-computed as an adjustment to `solute_dielectric`, but it doesn't seem like the underlying formulas permit that: https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/internal/customgbforces.py#L363-L379
[4] In the OBC paper: "The nonpolar contribution is computed via ΔGsurf = 0.005[kcal/mol] ∗︁ A[Å2]," https://onlinelibrary.wiley.com/doi/full/10.1002/prot.20033
[5] alpha/beta/gamma: OBC1: 0.8, 0, 2.91 ; OBC2: 1, 0.8, 4.85
[6] Hard coded `surface_area_energy` is probably 5.4 cal/mol/A^2 https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/internal/customgbforces.py#L373
and
http://docs.openmm.org/latest/userguide/theory.html#surface-area-term
```
>>> 28.3919551 / (4 * math.pi) # value from hard-coded eqn, removing factor of 4pi
2.259359999104074
>>> 5.4 * 4.184 # cal to joule
22.593600000000002
```
However the value used in the original OBC paper is 0.005 kcal/mol/A^2 [4]


j-wags:
@jchodera -- This is nearing completion, and I think it's ready for review, if you have time.

@andrrizzi -- You might be interested to see the FreeSolv GBSA tests. I now have all energy components of the system matching, which was, I think, a to-do item from before.

davidlmobley:
@j-wags should we enlist Yutong for help with code review on this too, since he wanted it? (Not advocating, just asking).

Congrats on getting this ready for review!!

j-wags:
That's a great idea -- @proteneer, please share any thoughts you have about this implementation or open questions remaining here.

openforcefield/tests/test_forcefield.py
```diff
@@ -1112,6 +1179,161 @@ def test_freesolv_parameters_assignment(self, freesolv_id, forcefield_version, a
...
+                                                     implicitSolvent=openmm_gbsas[gbsa_model],
+                                                     nonbondedMethod = amber_nb_method,
+                                                     nonbondedCutoff=amber_cutoff,
+                                                     gbsaModel='ACE',
+                                                     implicitSolventKappa=0.,
+                                                     )
+
+        off_gbsa_force = [force for force in off_omm_system.getForces() if
+                              (isinstance(force, openmm.GBSAOBCForce) or
+                               isinstance(force, openmm.openmm.CustomGBForce))][0]
```
andrrizzi: As a paranoid check, I'd do a `assert len(off_gbsa_forces) == 1` before retrieving the first element.

openforcefield/tests/test_forcefield.py
```diff
@@ -1112,6 +1179,161 @@ def test_freesolv_parameters_assignment(self, freesolv_id, forcefield_version, a
...
+            gb_params = openmm.app.internal.customgbforces.GBSAHCTForce.getStandardParameters(omm_top)
+        elif gbsa_model == 'OBC1':
+            gb_params = openmm.app.internal.customgbforces.GBSAOBC1Force.getStandardParameters(omm_top)
+        elif gbsa_model == 'OBC2':
+            gb_params = openmm.app.internal.customgbforces.GBSAOBC2Force.getStandardParameters(omm_top)
+
+        # Use GB params from OpenMM GBSA classes to populate parameters
+        for idx, (radius, screen) in enumerate(gb_params):
+            q, _, _2 = amber_gbsa_force.getParticleParameters(idx)
+            print(q, radius, screen, _, _2)
```
proteneer: what's the _, _2?

proteneer: i guess those are the amber radius/screens?

openforcefield/tests/test_forcefield.py
```diff
@@ -1112,6 +1179,161 @@ def test_freesolv_parameters_assignment(self, freesolv_id, forcefield_version, a
...
+        # Get context energies
+        amber_energy = get_context_potential_energy(amber_context, positions)
+        off_energy = get_context_potential_energy(off_context, positions)
+
+        # Very handy for debugging
+        # print(openmm.XmlSerializer.serialize(off_gbsa_force))
+        # print(openmm.XmlSerializer.serialize(amber_gbsa_force))
+
+        # Ensure that the GBSA energies (which we put into ForceGroup 1) are identical
+        assert amber_energy[1] == off_energy[1]
```
proteneer: I would check for the forces as well.

openforcefield/tests/test_parameters.py
```diff
@@ -908,36 +908,91 @@ def test_multi_term_proper_torsion_bad_units(self):
...
+        with pytest.raises(ValueError) as excinfo:
+            gbsa_handler.solvent_dielectric = 'string that can not be cast to float'
+
+        gbsa_handler.solute_dielectric = 2.5
+        gbsa_handler.solute_dielectric = "3.5"
+        with pytest.raises(ValueError) as excinfo:
+            gbsa_handler.solute_dielectric = 'string that can not be cast to float'
+
+        gbsa_handler.sa_model = 'ACE'
+        gbsa_handler.sa_model = None
```
proteneer: Do we ever want to allow for None? @jchodera

openforcefield/tests/test_forcefield.py
```diff
@@ -1112,6 +1179,161 @@ def test_freesolv_parameters_assignment(self, freesolv_id, forcefield_version, a
...
+
+        # Ensure that all system energies are the same
+        compare_system_energies(off_omm_system, amber_omm_system, positions, by_force_type=False)
+
```
proteneer: Do you actually test that FreeSolve free energies are equivalent?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1677,6 +1677,13 @@ def get_unitless_values(attr):
 
         for attr in identical_attrs:
             this_val, other_val = get_unitless_values(attr)
+
+            # Convert None values to str, so "==" comparison will work
+            if this_val is None:
```
proteneer: does None != None?

proteneer:
Some general questions: do you want to allow for setting non HCT/OBC1/OBC2 alpha/beta/gamma parameters?

proteneer:
Can you include a non-official smirnoff XML file that implements the standard AMBER GB parameters? Or would you guys be developing this with a custom set of SMIRKS separately?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2897,132 +2904,233 @@ class GBSAParameterHandler(ParameterHandler):
...
+            else:
+                gbsa_force.setSurfaceAreaEnergy(self.surface_area_penalty)
+
+
+        # Iterate over all defined GBSA types, allowing later matches to override earlier ones.
+        atom_matches = self.find_matches(topology)
+
+        # Create all particles.
+
+        # !!! WARNING: CustomAmberGBForceBase expects different per-particle parameters
```
proteneer: I'd want @jchodera or @peastman to also eyeball this blurb.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2897,132 +2904,233 @@ class GBSAParameterHandler(ParameterHandler):
...
+        # depending on whether you use addParticle or setParticleParameters. In
+        # setParticleParameters, we have to apply the offset and scale BEFORE setting
+        # parameters, whereas in addParticle, the offset is applied automatically, and the particle
+        # parameters are not set until an auxillary finalize() method is called. !!!
+
+        # To keep it simple, we DO NOT pre-populate the particles in the GBSA force here.
+        # We call addParticle further below instead.
+        #for topology_particle in topology.topology_particles:
```
proteneer: is this meant to be commented out?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2897,132 +2904,233 @@ class GBSAParameterHandler(ParameterHandler):
...
+        nonbonded_force = existing[0]
+
+        # No previous GBSAForce should exist, so we're safe just making one here.
+        force_map = {
+            'HCT': simtk.openmm.app.internal.customgbforces.GBSAHCTForce,
+            'OBC1': simtk.openmm.app.internal.customgbforces.GBSAOBC1Force,
+            'OBC2': simtk.openmm.GBSAOBCForce,
+            # It's tempting to do use the class below, but the customgbforce
+            # version of OBC2 doesn't provide setSolventRadius()
+            #'OBC2': simtk.openmm.app.internal.customgbforces.GBSAOBC2Force,
```
proteneer: I'm not sure how I feel about splitting this into two separate class types. Does GBSAOBCForce not support the features needed for HCT/OBC1?

openforcefield/tests/test_forcefield.py
j-wags: That's right. And the print statement is some debugging code that I'll remove before the merge.

openforcefield/tests/test_forcefield.py
j-wags: No -- Just single conformer energies.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2897,132 +2904,233 @@ class GBSAParameterHandler(ParameterHandler):
...
+        }
+        openmm_force_type = force_map[self.gb_model]
+
+        if nonbonded_force.getNonbondedMethod() == openmm.NonbondedForce.NoCutoff:
+            amber_cutoff = None
+        else:
+            amber_cutoff = nonbonded_force.getCutoffDistance().value_in_unit(unit.nanometer)
```
proteneer: Just double checking that we never want the GB force to have a different cutoff from the Nonbonded force

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2897,132 +2904,233 @@ class GBSAParameterHandler(ParameterHandler):
...
+                                           SA=self.sa_model,
+                                           cutoff=amber_cutoff,
+                                           kappa=0,
+                                           )
+            # WARNING: If using a CustomAmberGBForce, the functional form is affected by whether
+            # the cutoff kwarg is None *during initialization*. So, if you initialize it with a
+            # non-None value, and then try to change it to None, you'll basically break things.
+
+        # Set the GBSAForce to have the same cutoff as NonbondedForce
+        #gbsa_force.setCutoffDistance(nonbonded_force.getCutoffDistance())
```
proteneer: extra comment

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: GBSAOBCForce doesn't let us modify alpha/beta/gamma parameters (as far as I can tell). In some parts of the code, requesting an OBC2 model where kappa=0 yields a GBSAOBCForce [1], so its alpha/beta/gamma must be configured for OBC2. Thus I think  it's unsuitable to represent OBC1.
[1] https://github.com/pandegroup/openmm/blob/f9106ddb6a0d72c960c73b23ed0e07cd7190a80f/wrappers/python/simtk/openmm/app/internal/amber_file_parser.py#L985-L991

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Good catch. This was originally a quick fix for what I now realize is a validation failure, where we had an OFFXML with `sa_model="None"`, which got eaten as a string. I'll revisit this.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Yes. This was a huge footgun that I fired many times when I'd try to come back to this after a break. I've left copious warnings for future inhabitants. I might propose leaving a larger, more lasting warning as well: https://en.wikipedia.org/wiki/Long-time_nuclear_waste_warning_messages

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: I think that's the case because
1) The SMIRNOFF spec doesn't expose a separate cutoff or long-range method parameter for GBSA
2) I don't know enough about OpenMM to understand the technical interplay between NonbondedForce and GBSAForce. There's an implication that the GB and Nonbonded force should be parallel here, and so I extrapolated from that. 
http://docs.openmm.org/latest/api-python/generated/simtk.openmm.openmm.GBSAOBCForce.html
> When using this Force, the System should also include a NonbondedForce, and both objects must specify identical charges for all particles. Otherwise, the results will not be correct. Furthermore, if the nonbonded method is set to CutoffNonPeriodic or CutoffPeriodic, you should call setReactionFieldDielectric(1.0) on the NonbondedForce to turn off the reaction field approximation, which does not produce correct results when combined with GBSA.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Good catch! Removed in 7ae17c1

openforcefield/tests/test_forcefield.py
j-wags: 7ae17c1 removes the print statement and adds a comment+more explicit variable names

openforcefield/tests/test_forcefield.py
j-wags: Good idea! Check added in 7ae17c1

j-wags:
> Can you include a non-official smirnoff XML file that implements the standard AMBER GB parameters? Or would you guys be developing this with a custom set of SMIRKS separately?

Yes. The contents of such OFFXMLs currently exist as strings at the top of `test_forcefields.py`, and the tests are strenuous enough that I'm comfortable putting them out as individual files (with versions, so we can fix them if there's a mistake)

j-wags:
> Some general questions: do you want to allow for setting non HCT/OBC1/OBC2 alpha/beta/gamma parameters? To further expand, do we want to expose this as three distinct variants? Or do we want to fully expose the ability to set dielectrics/alpha/beta/gamma etc. and instead have HCT/OBC1/OBC2 directly implemented in the XML?

This is a good question. From what I can see, we could write our own CustomGBForce from scratch, which exposes everything the current GBSA spec does, and also allows users to set their own alpha/beta/gamma AND kappa. The selection of `OBC1`, `OBC2` or `HCT` would just lead to a lookup of specific alpha/beta/gamma values. 

I have trouble estimating the effort required for such an implementation. GBSA is a complex topic and I was off by a factor of more than 2 on the time required for this implementation. So, I'd like to ship this as the initial version of GBSA, but we can work asynchronously on the spec/validation scheme for a more flexible implementation in a separate Issue.

openforcefield/typing/engines/smirnoff/parameters.py
peastman: Yes, it's a really confusing API.  Sorry about that!  I didn't design it. :/

openforcefield/typing/engines/smirnoff/parameters.py
peastman: They should always have the same cutoff.  GBSA effectively subtracts off a fraction of the energy from the Coulomb interaction.  That requires that every particle pair either gets both interactions or neither.  If you have different cutoffs for the two, that will lead to really strange artifacts.

proteneer:
@jchodera when you get a chance can you take a look?

j-wags:
@proteneer 
> Can you include a non-official smirnoff XML file that implements the standard AMBER GB parameters?

I've added these as `test_forcefields/GBSA_{HCT,OBC1,OBC2}-1.0.offxml`. These contain a bugfix [1] that applies to OpenMM and ParmEd's per-particle parameter libraries. So, this will give different numbers for Cl-containing molecules until the next OpenMM release. 

[1] https://github.com/openmm/openmm/issues/2339

openforcefield/tests/test_parameters.py
j-wags: Yes -- Setting the `sa_model` to `None` zeroes the surface area energy term.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Reverted. @andrrizzi 's ParameterAttribute classes handle this now.

j-wags:
The failing tests are for GBSA energies... But the number of failures varies from ~20 to ~30 each time the tests run, and are only for HCT and OBC1 models. I don't see any failures on my computer, even running these tests repeatedly. It's possible that these errors are coming up if pytest on travis is parallelizing the tests, so I'll check that out tomorrow.

j-wags:
I seem to have resolved the issue.

The new tests compare GBSA energies that come from parameterizing the same molecule two different ways:

- one produced by parameterizing using the OFF toolkit, 
- the other produced by parameterizing using OpenMM's `AmberPrmtop` class 

On the OpenCL platform, there is no GBSA energy difference, and a `==` comparison passes. However, I don't think that Travis can support OpenCL, so I've manually set this to `Reference` in the tests. Somehow, the `Reference` energies begin to differ around the 8th or 9th digit, so I've added in a tolerance of 1e-5 kcal/mol to the tests, and now they seem to be passing.

openforcefield/tests/test_forcefield.py
j-wags: That's a good idea and would apply to tests beyond just this one. In the interest of cutting a release on Monday, I don't think we should include that in this PR, but I've started an Issue for it here: https://github.com/openforcefield/openforcefield/issues/398

j-wags:
Ok -- Tests are passing and this seems good to go. I'm going to merge this tomorrow around lunch unless there are any objections.

docs/releasehistory.rst
```diff
@@ -8,12 +8,19 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
...
 """"""""""""
-
+- `PR #363 <https://github.com/openforcefield/openforcefield/pull/363>`_: Implements
+  :py:class:`GBSAHandler <openforcefield.typing.engines.smirnoff.parameters.GBSAHandler>`,
+  which supports the
+  `GBSA tag in the SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#gbsa>`_.
+  Currently, only GBSAHandlers with ``gb_model="OBC2"`` support
+  setting non-default values for the ``surface_area_penalty`` term (default ``5.4*calories/mole/angstroms**2``).
+  No model currently supports setting ``solvent_radius`` to any value other than ``1.4*angstroms``.
+  Experimental SMIRNOFF-format implementations of ``HCT``, ``OBC1``, and ``OBC2`` are included for testing
```
jchodera: It should be easy to include the SA penalty for these too! What's the limiting issue?

docs/releasehistory.rst
```diff
@@ -8,12 +8,19 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
...
+  purposes (see below).
+- `PR #363 <https://github.com/openforcefield/openforcefield/pull/363>`_: Periodic box vectors are
+  now transferred in :py:meth:`Topology.to_openmm() <openforcefield.topology.Topology.to_openmm>`
```
jchodera: Now transferred from where to where?

docs/releasehistory.rst
```diff
@@ -30,6 +37,14 @@ Bugfixes
   an error about "trans COOH conformers"
 
 
+Force fields added
+""""""""""""""""""
+- `PR #363 <https://github.com/openforcefield/openforcefield/pull/363>`_: Adds
+  ``test_forcefields/GBSA_HCT-1.0.offxml``, ``test_forcefields/GBSA_OBC1-1.0.offxml``,
+  and ``test_forcefields/GBSA_OBC2-1.0.offxml``, which are experimental implementations
+  of GBSA models. These are primarily used in validation tests against OpenMM's models, and
```
jchodera: Would it be useful to mention which parameters/papers they are based on?

openforcefield/data/test_forcefields/GBSA_HCT-1.0.offxml
```diff
@@ -0,0 +1,18 @@
+<SMIRNOFF version="0.3" aromaticity_model="OEAroModel_MDL">
+    <GBSA version="0.3" gb_model="HCT" solvent_dielectric="78.5" solute_dielectric="1" sa_model="ACE" surface_area_penalty="5.4*calories/mole/angstroms**2" solvent_radius="1.4*angstroms">
```
jchodera: Can we add some provenance information for each of these experimental `.offxml` files? Where does this come from?

openforcefield/data/test_forcefields/GBSA_OBC1-1.0.offxml
```diff
@@ -0,0 +1,15 @@
+<SMIRNOFF version="0.3" aromaticity_model="OEAroModel_MDL">
```
jchodera: Can we add some provenance information for each of these experimental `.offxml` files? Where does this come from?

openforcefield/data/test_forcefields/GBSA_OBC2-1.0.offxml
```diff
@@ -0,0 +1,15 @@
+<SMIRNOFF version="0.3" aromaticity_model="OEAroModel_MDL">
```
jchodera: Can we add some provenance information for each of these experimental `.offxml` files? Where does this come from?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2905,137 +2907,243 @@ def postprocess_system(self, system, topology, **kwargs):
...
+
+        nonbonded_force = existing[0]
+
+        # No previous GBSAForce should exist, so we're safe just making one here.
+        force_map = {
+            'HCT': simtk.openmm.app.internal.customgbforces.GBSAHCTForce,
+            'OBC1': simtk.openmm.app.internal.customgbforces.GBSAOBC1Force,
+            'OBC2': simtk.openmm.GBSAOBCForce,
+            # It's tempting to do use the class below, but the customgbforce
+            # version of OBC2 doesn't provide setSolventRadius()
```
jchodera: We can implement that if needed!

openforcefield/typing/engines/smirnoff/parameters.py
jchodera: I think we should probably fork `customgbforces.py`, pull it from OpenMM into the openforcefield toolkit as an internal file, and modify it to behave as needed. `customgbforces` is essentially an internal OpenMM API and could be changed at any time, breaking our implementation.

docs/releasehistory.rst
j-wags: The current implementations of `OBC1` and `HCT` only support a hard-coded value: https://github.com/openmm/openmm/blob/master/wrappers/python/simtk/openmm/app/internal/customgbforces.py#L372-L375

docs/releasehistory.rst
j-wags: Good catch. Changed to 
```
  When using Topology.to_openmm(), periodic box vectors 
  are now transferred from the Open Force Field Toolkit Topology 
  into the newly-created OpenMM Topology.
```

docs/releasehistory.rst
jchodera: We can easily fix that once we pull that file in.

j-wags:
@jchodera @proteneer I think we generally agree that we want to create our own `OFFGBSAForce` class derived from `CustomGBForce` in the future. This will have the benefits of

- Letting us expose `solvent_radius` and `surface_area_energy` for all models
- Letting us expose `alpha`/`beta`/`gamma` (and maybe even `kappa`) for all models

I am hesitant about taking the intermediate step of forking customgbforces right now. That is, if we fork OpenMM's customgbforces file and fix a few things (eg. expose `surface_area_energy`) now, and then implement a full changeover in a few months, it will have the downsides of

- Delaying the initial release of GBSA handling, likely until September, as my hands will be full with meeting prep and release-1 support though August
- Requiring time/attention to maintain in parallel (there was a bugfix to `customgbforces` _this week_ https://github.com/openmm/openmm/issues/2339)
- Producing yet another class that people might expect ParmEd to convert to different packages

So, while I agree with the long-term plan, I'd advocate cutting a release of this GBSA right now as "version 0.3" in the spec, and then starting taking some time to think about "GBSA version 0.4" when we're not in a time crunch.

openforcefield/data/test_forcefields/GBSA_OBC1-1.0.offxml
j-wags: Added provenance info in 93f9c93

openforcefield/data/test_forcefields/GBSA_HCT-1.0.offxml
j-wags: Added provenance info in 93f9c93. The GBSA parameters in the new files are adapted from the contents of [this file](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/data/test_forcefields/Frosst_AlkEthOH_GBSA.offxml#L11) , which states that you originally wrote these terms. I subsequently changed the numbers to fit the the new spec and proper radius/screen values, so I've listed both of us as authors for the FFs.

openforcefield/data/test_forcefields/GBSA_OBC2-1.0.offxml
j-wags: Added provenance info in 93f9c93

jchodera:
@j-wags : OK, I agree with your logic. 
Let's be sure to warn people that this [can't be converted with `parmed` right now](https://github.com/ParmEd/ParmEd/blob/master/parmed/openmm/topsystem.py#L148-L150), and also to leave some open issues describing what we need to address here in the future to make this general.

PR #364, opened 2019-06-19
Add explicit hydrogens if needed in from_openeye

### LINKED ISSUE #362: Unexpected behavior: Parameter usage statistics incorrect when creating from SMILES without explicit hydrogens
We've been estimating parameter usage statistics for a set of molecules and found that parameter usage is handled unexpectedly when a molecule is provided which does not have explicit hydrogens (which is not infrequent when creating a molecule from SMILES, as the below shows). Specifically, unless explicit hydrogens are provided, not all parameters are listed. For example:

```
from openeye import oechem

# Create a simple molecule from SMILES and turn it into a topology.
smiles = "C#C"
mol = oechem.OEMol()
oechem.OESmilesToMol(mol, smiles)
molecule = Molecule.from_openeye(mol)
topology = Topology.from_molecules([molecule])

# Let's label using the smirnoff99Frosst force field
forcefield = ForceField('smirnoff99Frosst.offxml')

# Run the molecule labeling
molecule_force_list = forcefield.label_molecules(topology)

# Print out a formatted description of the parameters applied to this molecule
for mol_idx, mol_forces in enumerate(molecule_force_list):
    print(f'Forces for molecule {mol_idx}')
    for force_tag, force_dict in mol_forces.items():
        print(f"\n{force_tag}:")
        for (atom_indices, parameter) in force_dict.items():
            atomstr=''
            for idx in atom_indices:
                atomstr += '%3s' % idx
            print("atoms: %s  parameter_id: %s  smirks %s" % (atomstr, parameter.id, parameter.smirks) )
```

yields the following output:
```
Forces for molecule 0

Bonds:

Angles:

ProperTorsions:

ImproperTorsions:

vdW:
atoms:   0  parameter_id: n14  smirks [#6:1]
atoms:   1  parameter_id: n14  smirks [#6:1]

Electrostatics:

ToolkitAM1BCC:
```

Adding the single line:
`oechem.OEAddExplicitHydrogens(mol)`
after creation of the OEMol from the SMILES string results in the following (correct) output:

```
Forces for molecule 0

Bonds:
atoms:   0  1  parameter_id: b27  smirks [#6X2:1]#[#6X2:2]
atoms:   0  2  parameter_id: b85  smirks [#6X2:1]-[#1:2]
atoms:   1  3  parameter_id: b85  smirks [#6X2:1]-[#1:2]

Angles:
atoms:   0  1  3  parameter_id: a16  smirks [*:1]~[#6X2:2]~[*:3]
atoms:   1  0  2  parameter_id: a16  smirks [*:1]~[#6X2:2]~[*:3]

ProperTorsions:
atoms:   2  0  1  3  parameter_id: t156  smirks [*:1]-[*:2]#[*:3]-[*:4]

ImproperTorsions:

vdW:
atoms:   0  parameter_id: n15  smirks [#6X2:1]
atoms:   1  parameter_id: n15  smirks [#6X2:1]
atoms:   2  parameter_id: n10  smirks [#1:1]-[#6X2]
atoms:   3  parameter_id: n10  smirks [#1:1]-[#6X2]

Electrostatics:

ToolkitAM1BCC:
```

Using the internal `molecule = Molecule.from_smiles('C#C')` results in correct behavior, I believe.

This isn't actually incorrect, but also shouldn't lead to "silent failures" like at present. In my view, if a molecule is provided which is missing explicit hydrogens, the toolkit should either:
1) Add explicit hydrogens, or
2) Raise an exception that explicit hydrogens are missing

It shouldn't just print incorrect statistics. 

This has some overlap with https://github.com/openforcefield/openforcefield/issues/61 in that a molecule without all of its protons (or any other obviously incorrect molecule) shouldn't be just processed by the toolkit as if it is correct.

j-wags:
### Summary
- [x] Fixes #362 
- [x] Adds test for same

### Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/openforcefield/tests/test_toolkits.py
+++ b/openforcefield/tests/test_toolkits.py
@@ -252,11 +252,13 @@ class TestOpenEyeToolkitWrapper:
         offmol = Molecule.from_smiles(smiles_impl,
                                       toolkit_registry=toolkit_wrapper,
                                       hydrogens_are_explicit=False)
+        assert offmol.n_atoms == 4
 
         smiles_expl = "HC#CH"
         offmol = Molecule.from_smiles(smiles_expl,
                                       toolkit_registry=toolkit_wrapper,
                                       hydrogens_are_explicit=True)
+        assert offmol.n_atoms == 4
         # It's debatable whether this next function should pass. Strictly speaking, the hydrogens in this SMILES
         # _are_ explicit, so allowing "hydrogens_are_explicit=False" through here is allowing a contradiction.
         # We might rethink the name of this kwarg.
@@ -264,6 +266,7 @@ class TestOpenEyeToolkitWrapper:
         offmol = Molecule.from_smiles(smiles_expl,
                                       toolkit_registry=toolkit_wrapper,
                                       hydrogens_are_explicit=False)
+        assert offmol.n_atoms == 4
 
 
 
@@ -543,12 +546,13 @@ class TestRDKitToolkitWrapper:
         offmol = Molecule.from_smiles(smiles_impl,
                                       toolkit_registry=toolkit_wrapper,
                                       hydrogens_are_explicit=False)
+        assert offmol.n_atoms == 4
 
         smiles_expl = "[H][C]#[C][H]"
         offmol = Molecule.from_smiles(smiles_expl,
                                       toolkit_registry=toolkit_wrapper,
                                       hydrogens_are_explicit=True)
-
+        assert offmol.n_atoms == 4
         # It's debatable whether this next function should pass. Strictly speaking, the hydrogens in this SMILES
         # _are_ explicit, so allowing "hydrogens_are_explicit=False" through here is allowing a contradiction.
         # We might rethink the name of this kwarg.
@@ -556,6 +560,7 @@ class TestRDKitToolkitWrapper:
         offmol = Molecule.from_smiles(smiles_expl,
                                       toolkit_registry=toolkit_wrapper,
                                       hydrogens_are_explicit=False)
+        assert offmol.n_atoms == 4
 
 
     @pytest.mark.skipif(not RDKitToolkitWrapper.is_available(), reason='RDKit Toolkit not available')
--- a/openforcefield/tests/test_toolkits.py
+++ b/openforcefield/tests/test_toolkits.py
@@ -214,9 +214,34 @@ class TestOpenEyeToolkitWrapper:
             assert_almost_equal(pc1_ul, pc2_ul, decimal=6)
         assert molecule2.to_smiles(toolkit_registry=toolkit_wrapper) == expected_output_smiles
 
+
+    @pytest.mark.skipif(not OpenEyeToolkitWrapper.is_available(), reason='OpenEye Toolkit not available')
+    def test_from_openeye_implicit_hydrogen(self):
+        """
+        Test OpenEyeToolkitWrapper for loading a molecule with implicit
+        hydrogens (correct behavior is to add them explicitly)
+        """
+        from openeye import oechem
+
+        smiles_impl = "C#C"
+        oemol_impl = oechem.OEMol()
+        oechem.OESmilesToMol(oemol_impl, smiles_impl)
+        molecule_from_impl = Molecule.from_openeye(oemol_impl)
+
+        assert molecule_from_impl.n_atoms == 4
+
+        smiles_expl = "HC#CH"
+        oemol_expl = oechem.OEMol()
+        oechem.OESmilesToMol(oemol_expl, smiles_expl)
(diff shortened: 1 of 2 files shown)


openforcefield/utils/toolkits.py
```diff
@@ -693,6 +694,11 @@ def from_openeye(oemol, allow_undefined_stereo=False):
         from openeye import oechem
         from openforcefield.topology.molecule import Molecule
 
+
+        # Add explicit hydrogens if they're implicit
+        if oechem.OEHasImplicitHydrogens(oemol):
+            oechem.OEAddExplicitHydrogens(oemol)
```
andrrizzi: I'm not entirely sure how we should do this, actually. Does this change make `from_smiles(..., hydrogens_are_explicit=True)` (which is the default) behave unexpectedly? I think that function calls `from_openeye()`. Maybe `from_openeye()` should take an argument to control this behavior as well.

j-wags: Wow. Great thinking, I had totally missed the possible consequences there.

So, the case where this would be a problem is if the user provided a SMILES with a specific explicit protonation state, and the toolkit misinterpreted it as having implicit hydrogens (even though the user set `hydrogens_are_explicit=True`).

To prevent this from happening, I've updated both `OpenEyeToolkitWrapper` and `RDKitToolkitWrapper`'s `from_smiles` functions to assert that, if `hydrogens_are_explicit=True`, then there are zero implicit hydrogens on the molecule.

openforcefield/tests/test_toolkits.py
```diff
@@ -498,6 +527,36 @@ def test_smiles_add_H(self):
         smiles2 = molecule.to_smiles(toolkit_registry=toolkit_wrapper)
         assert smiles2 == expected_output_smiles
 
+    @pytest.mark.skipif(not OpenEyeToolkitWrapper.is_available(), reason='OpenEye Toolkit not available')
+    def test_rdkit_from_smiles_hydrogens_are_explicit(self):
```
andrrizzi: I think you intended to check for the RDKit wrapper availability here.

openforcefield/tests/test_toolkits.py
```diff
@@ -236,6 +236,35 @@ def test_from_openeye_implicit_hydrogen(self):
...
+        toolkit_wrapper = OpenEyeToolkitWrapper()
+        smiles_impl = "C#C"
+        with pytest.raises(ValueError,
+                           match="but OpenEye Toolkit interpreted SMILES 'C#C' as having implicit hydrogen") as excinfo:
+            offmol = Molecule.from_smiles(smiles_impl,
+                                          toolkit_registry=toolkit_wrapper,
+                                          hydrogens_are_explicit=True)
+        offmol = Molecule.from_smiles(smiles_impl,
+                                      toolkit_registry=toolkit_wrapper,
+                                      hydrogens_are_explicit=False)
```
andrrizzi: I'd suggest adding an assert statement here simply checking that the number of atoms is the expected one.

openforcefield/utils/toolkits.py
```diff
@@ -1065,9 +1065,16 @@ def from_smiles(self, smiles, hydrogens_are_explicit=False, allow_undefined_ster
         if not (hydrogens_are_explicit):
             result = oechem.OEAddExplicitHydrogens(oemol)
             if result == False:
-                raise Exception(
+                raise ValueError(
```
andrrizzi: 👍

openforcefield/utils/toolkits.py
```diff
@@ -1065,9 +1065,16 @@ def from_smiles(self, smiles, hydrogens_are_explicit=False, allow_undefined_ster
...
                     "Addition of explicit hydrogens failed in from_openeye")
-        # TODO: Add allow_undefined_stereo to this function, and pass to from_openeye?
+        elif hydrogens_are_explicit:
+            if oechem.OEHasImplicitHydrogens(oemol):
```
andrrizzi: I think you can substitute this nested ifs with a single `elif oechem.OEHasImplicitHydrogens(oemol):`

openforcefield/tests/test_toolkits.py
j-wags: Good catch.

openforcefield/tests/test_toolkits.py
j-wags: 👍 Done.

openforcefield/utils/toolkits.py
j-wags: I've met this suggestion half way -- I combined the nested `if` with the `elif`, just to keep the logic clear for future people who work on this file.

andrrizzi: 👍

PR #366, opened 2019-06-19
[WIP] Allow Missing Parameters

SimonBoothroyd:
A PR to add the option to bypass the missing valence term checks. The application I am interested in using this for is when computing gradients, where I want to build a system which only contains the parameter (or type of parameter) which the gradient is being determined with respect to (e.g. build a system which has only been assigned bonds, but no angle or torsion terms).


j-wags:
@SimonBoothroyd Thanks for the PR!

I think this feature would be very useful, and the implementation in this PR (threading the option through `create_openmm_system`'s `kwargs`) is exactly how it should be done. I'd approve merging this PR pending the addition of similar functionality to all `ParameterHandler`s (I only looked briefly at the file changes, but I don't think `vdWHandler` has this bypass implemented), and the addition of a test or two to ensure this kwarg is handled correctly.

I won't be able to review this until Monday, but @andrrizzi may be able to approve it sooner :-)

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -576,7 +576,7 @@ class ParameterHandler:
     _INDEXED_ATTRIBS = []  # list of parameter attribs that will have consecutive numerical suffixes starting at 1
     _REQUIRE_UNITS = {}  # dict of {header attrib : unit } for input checking
     _ATTRIBS_TO_TYPE = {} # dict of attribs that must be cast to a specific type to be interpreted correctly
-    _KWARGS = [] # Kwargs to catch when create_force is called
+    _KWARGS = ['allow_missing_parameters'] # Kwargs to catch when create_force is called
```
andrrizzi: We probably don't want `allow_missing_parameters` to be an accepted kwarg in all `ParameterHandler`s (improper torsions, for example) so we should probably add it individually to all the handlers that take it.

SimonBoothroyd: Happy to make this change, but could you clarify why this wouldn't be desired behaviour for improper torsions?

davidlmobley: @SimonBoothroyd improper torsions are used very sporadically in current FFs, so in general they will be missing far more than they are present. I assume that's what he's getting at.

SimonBoothroyd: Ah ok that makes sense. The only issue (I think this is an issue?) with not making this available to all handlers is that when my FF only contains a handler which doesn't capture this kwarg (e.g. only an ImproperHandler), a `not used by any registered force Handler` `ValueError` will be raised (I've run into a similar issue with `charged_molecules`). @andrrizzi, @j-wags would you be ok with me exposing an `ignore_unused_kwargs=False` option to `create_openmm_system` that would bypass this exception?

andrrizzi: I see. I would avoid an `ignore_unused_kwargs` argument, if possible. I think that a cleaner API would be with a `ignore_parameter_handlers` argument instead. It will probably make other parts of your code simpler too, but that will be trickier to implement when there are dependencies. For now, is it a problem to `try...catch` it?

jchodera:
Apologies for being late to this, but while this is definitely useful, there are simpler (and more reliable) ways of computing only the Force(s) in the OpenMM System you want for the purposes of finite difference gradients. These alternatives only take a few lines of Python, and since we will only use finite difference gradients until we can get an autodiff implementation working, I'd suggest using one of these two approaches inside the finite difference code instead of omitting valence parameters in the SMIRNOFF force field parameter objects:

1. Allocate Force objects to integer force groups and compute the energy for only the desired group. This takes perhaps three lines of code.
2. Use System.removeForce() to remove forces you aren't interested in. Perhaps two lines of code.

@andrrizzi can help you with either implementation.

andrrizzi:
> 1. Allocate Force objects to integer force groups and compute the energy for only the desired group. This takes perhaps three lines of code.
> 2. Use System.removeForce() to remove forces you aren't interested in. Perhaps two lines of code.

Whoops! I assumed those routes were unfeasible for the propertyestimator for some reason, but I'm very happy to write a quick snippet if you need an example of how that works.

j-wags:
_If_ we do go ahead with this (and it doesn't seem certain that we will), let's be sure to update the docs for `create_openmm_system` to include the new kwarg (we had forgotten to do this with another keyword). Previous reference: https://github.com/openforcefield/openforcefield/issues/321#issuecomment-492684533

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: I agree with @andrrizzi above. I don't think we should implement `ignore_unused_kwargs` unless more specific use cases pop up -- `create_openmm_system` is a high-traffic function and I'd like to keep it simple if possible. I actually have the same hesitation about `ignore_parameter_handlers`. I like Andrea's second suggestion, which would look like:
```
try:
    sys = ff.create_openmm_system(topology, ignore_missing_valence=True)
except ValueError:
    sys = ff.create_openmm_system(topology)
```

j-wags:
Specifically for @SimonBoothroyd, @jchodera's suggestions (1) and (2) are probably the best way to go.

In general (and my initial quick read of this PR left me with the understanding that), a user might reasonably define a FF where some angles simply don't have parameters, but others do. Thus, they would want to have an AngleHandler, but not have angles applied everywhere. 

I'd like to keep this PR moving forward for that use case.

Also, to help my understanding (and perhaps this is not exactly the same situation as here), is there some reason that we couldn't define an FF with only a `BondHandler` if we want to create a system with only bond energies?

j-wags:
With @SimonBoothroyd's blessing, I'm going to close this PR. This functionality is no longer needed.

PR #371, opened 2019-06-28
Read SMIRFF-enclosed FFs

j-wags:
### Summary

In testing the ability to read all files from a release-candidate smirnoff99Frosst package, I found that ones with an enclosing XML `SMIRFF` tag (instead of `SMIRNOFF`) were unable to be read. This PR fixes that.

### DIFF
--- a/openforcefield/typing/engines/smirnoff/forcefield.py
+++ b/openforcefield/typing/engines/smirnoff/forcefield.py
@@ -820,9 +820,9 @@ class ForceField:
 
         # Check that the SMIRNOFF version of this data structure is supported by this ForceField implementation
 
-        if "SMIRNOFF" in smirnoff_data.keys():
+        if "SMIRNOFF" in smirnoff_data:
             version = smirnoff_data['SMIRNOFF']['version']
-        elif "SMIRFF" in smirnoff_data.keys():
+        elif "SMIRFF" in smirnoff_data:
             version = smirnoff_data['SMIRFF']['version']
         else:
             raise ParseError("'version' attribute must be specified in SMIRNOFF tag")
--- a/openforcefield/typing/engines/smirnoff/forcefield.py
+++ b/openforcefield/typing/engines/smirnoff/forcefield.py
@@ -817,10 +817,32 @@ class ForceField:
             Whether to permit non-spec kwargs in smirnoff_data.
         """
         import packaging.version
+
+        # Check that the SMIRNOFF version of this data structure is supported by this ForceField implementation
+
+        if "SMIRNOFF" in smirnoff_data.keys():
+            version = smirnoff_data['SMIRNOFF']['version']
+        elif "SMIRFF" in smirnoff_data.keys():
+            version = smirnoff_data['SMIRFF']['version']
+        else:
+            raise ParseError("'version' attribute must be specified in SMIRNOFF tag")
+
+        self._check_smirnoff_version_compatibility(str(version))
+        # Convert 0.1 spec files to 0.3 SMIRNOFF data format by converting
+        # from 0.1 spec to 0.2, then 0.2 to 0.3
+        if packaging.version.parse(str(version)) == packaging.version.parse("0.1"):
+            # NOTE: This will convert the top-level "SMIRFF" tag to "SMIRNOFF"
+            smirnoff_data = convert_0_1_smirnoff_to_0_2(smirnoff_data)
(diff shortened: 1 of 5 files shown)


openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -817,11 +817,33 @@ def _load_smirnoff_data(self, smirnoff_data, allow_cosmetic_attributes=False):
             Whether to permit non-spec kwargs in smirnoff_data.
         """
         import packaging.version
+
+        # Check that the SMIRNOFF version of this data structure is supported by this ForceField implementation
+
+        if "SMIRNOFF" in smirnoff_data.keys():
+            version = smirnoff_data['SMIRNOFF']['version']
+        elif "SMIRFF" in smirnoff_data.keys():
```
andrrizzi: You can remove the `keys()` here to make use of the constant time lookup.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -817,11 +817,33 @@ def _load_smirnoff_data(self, smirnoff_data, allow_cosmetic_attributes=False):
...
+            smirnoff_data = convert_0_2_smirnoff_to_0_3(smirnoff_data)
+
+        # Convert 0.2 spec files to 0.3 SMIRNOFF data format by removing units
+        # from section headers and adding them to quantity strings at all levels.
+        elif packaging.version.parse(str(version)) == packaging.version.parse("0.2"):
+            smirnoff_data = convert_0_2_smirnoff_to_0_3(smirnoff_data)
+
         # Ensure that SMIRNOFF is a top-level key of the dict
         if not('SMIRNOFF' in smirnoff_data):
             raise ParseError("'SMIRNOFF' must be a top-level key in the SMIRNOFF object model")
```
andrrizzi: The tests pass so I might be missing something, but I would expect this exception to be raised the `SMIRFF in smirnoff_data` instead of `SMIRNOFF`.

If this is a problem, it might be a good idea to substitute the `SMIRFF` key with `SMIRNOFF` in `smirnoff_data` above to avoid using a lot of if-else statement downstream.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -817,11 +817,33 @@ def _load_smirnoff_data(self, smirnoff_data, allow_cosmetic_attributes=False):
...
 
-
         # Check that the aromaticity model required by this parameter set is compatible with
         # others loaded by this ForceField
         if 'aromaticity_model' in smirnoff_data['SMIRNOFF']:
```
andrrizzi: Possibly here too and in all the `if ... in smirnoff_data['SMIRNOFF']` clause below.

davidlmobley:
@j-wags Are you sure we don't want to just get those tags converted? Long long ago we called them "SMIRFF" forcefields rather than "SMIRNOFF", but finally decided to change the name. However, the intention wasn't to maintain backwards compatibility, just to switch.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: I'll make this clearer in the code, but the SMIRFF tag is converted to SMIRNOFF above (in `convert_0_1_smirnoff_to_0_2`)

j-wags:
@davidlmobley 

My goal here is to have every FF from the `smirnoff99Frosst` repo loadable, and all of them up to the 1.0.4(?) release are enclosed with a `SMIRFF` tag. Do you foresee any danger in loading those? I'm under the impression that the format in those files is identical to the "official" 0.1 SMIRNOFF spec.

j-wags:
@davidlmobley To clarify -- I want the toolkit to read `SMIRFF`-enclosed files because the alternatives are:
1) _modifying_ the released smirnoff99Frosst files, which we probably agree is bad, or
2) not being able to read all released `smirnoff99Frosst` files using the current toolkit. I wasn't under the impression that there was much risk in auto-upgrading the spec of these old files, but if you suspect that there is, we could remove (or do a deep dive into) this functionality.

openforcefield/typing/engines/smirnoff/forcefield.py
j-wags: If I get some free time I'm  going to make an Andrea-bot to remind me of this from now on.

PR #384, opened 2019-07-15
[WIP] Adds a (temporary) test FF with some extra improper parameters for an experimental fit

davidlmobley:
@yudongqiu this adds a temporary test force field from @jmaat (as discussed on our OpenFF slack here: https://openforcefieldgroup.slack.com/archives/C8P54J2JY/p1562883306333400 ) with additional impropers. The intent is to take a test pass through fitting and see if this would dramatically improve accuracy for some molecules.

Note the XML still contains some ForceBalance fitting code, which I think you can use, but please be sure.

### DIFF
--- a/openforcefield/data/test_forcefields/smirnoff99Frosst_testimpropers.offxml
+++ b/openforcefield/data/test_forcefields/smirnoff99Frosst_testimpropers.offxml
@@ -1,0 +1,344 @@
+<?xml version="1.0" encoding="utf-8"?>
+<SMIRNOFF version="0.3" aromaticity_model="OEAroModel_MDL">
+	<Bonds version="0.3" potential="harmonic" fractional_bondorder_method="None" fractional_bondorder_interpolation="linear">
+		<Bond smirks="[#6X4:1]-[#6X4:2]" length="1.526 * angstrom" k="620.0 * angstrom**-2 * mole**-1 * kilocalorie" id="b1"></Bond>
+		<Bond smirks="[#6X4:1]-[#6X3:2]" length="1.51 * angstrom" k="634.0 * angstrom**-2 * mole**-1 * kilocalorie" id="b2"></Bond>
+		<Bond smirks="[#6X4:1]-[#6X3:2]=[#8X1+0]" length="1.522 * angstrom" k="634.0 * angstrom**-2 * mole**-1 * kilocalorie" id="b3"></Bond>
+		<Bond smirks="[#6X3:1]-[#6X3:2]" length="1.45 * angstrom" k="820.0 * angstrom**-2 * mole**-1 * kilocalorie" id="b4"></Bond>
+		<Bond smirks="[#6X3:1]:[#6X3:2]" length="1.4 * angstrom" k="938.0 * angstrom**-2 * mole**-1 * kilocalorie" id="b5"></Bond>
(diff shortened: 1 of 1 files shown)


yudongqiu:
@davidlmobley Sure thanks. I think it's better to clarify this test a bit before I run it.

1. Previously I run the fitting for Bonds, Angles and ProperTorsions parameters, using [smirnoff99Frosst_experimental.offxml](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/data/test_forcefields/smirnoff99Frosst_experimental.offxml). For <Bonds>, `length` and `k` are included; for <Angles>, `angle` and `k` are included; For <ProperTorsions>, all `k`s like `k1`, `k2`... are included. This results in total of 504 parameters being fitted. **No Improper Torsions were fitted**

2. In this test, we have a new offxml file "smirnoff99Frosst_testimpropers.offxml", that contains additional <Improper> terms. These improper terms have ForceBalance annotations `parameterize` and `parameter_eval`.

Questions:
1. Do we want to fit improper terms in `smirnoff99Frosst_experimental.offxml`?

2. Do we want to fit improper terms in the new `smirnoff99Frosst_testimpropers.offxml`?

3. If we want to the fit improper terms in `smirnoff99Frosst_testimpropers.offxml`, wo we want to keep using the existing ForceBalance annotations `parameterize` and `parameter_eval`? 

Please be noted that we haven't tested any `parameter_eval` in SMIRNOFF fitting yet.

davidlmobley:
@yudongqiu 

What would be great if you could:

> Do we want to fit improper terms in smirnoff99Frosst_experimental.offxml?

Could you try doing this, for comparison with the below? I don't think we should be fitting these in general yet -- because the current ones just keep things flat -- but if possible I'd like you to do it once to compare with the below.

> Do we want to fit improper terms in the new smirnoff99Frosst_testimpropers.offxml?

Yes, please, let's try doing that. The `...test_impropers.offxml` should be exactly the `..._experimental.offxml` with extra improper terms.

> If we want to the fit improper terms in smirnoff99Frosst_testimpropers.offxml, wo we want to keep using the existing ForceBalance annotations parameterize and parameter_eval?

Yes, @jmaat has been using those already in fitting as I understand it. I'll ask her to confirm.

> Please be noted that we haven't tested any parameter_eval in SMIRNOFF fitting yet.

@jmaat has been using these I believe.


So to summarize I'm just proposing running an experiment where you (a) fit everything you're currently fitting but also fit the existing impropers, and (b) fit everything you're currently fitting and also fit the existing and new impropers. Then we can see how much the new ones would improve things.

jmaat:
@davidlmobley @yudongqiu I looked over `smirnoff99Frosst_testimpropers.offxml` file at the impropers section. These are the correct improper terms and parameterize tags in the .offxml file that I have been using for fitting.

yudongqiu:
@davidlmobley Thanks a lot for the explanation. I will proceed with running the two fittings you described and let you know the results. It's great to hear that @jmaat is able to run ForceBalance fitting on SMIRNOFF forcefield. I will try this and let you know if there's any problem with the recently developed codes.

davidlmobley:
@j-wags not sure we'll actually want to merge this here, which is why I didn't follow the template (and apologies for cluttering up your PRs...) This is really just a placeholder to expose a test FF with extra impropers so that Yudong can test out some science. If the science looks like a good idea, I'll clean it up for merging, update the docs and README, etc. If not, then I can close this PR and just leave it in a branch.

davidlmobley:
(This can be closed; closing.)

PR #390, opened 2019-07-19
[WIP] Add placeholder/experimental FF including a torsion for P=N double bonds

davidlmobley:
This introduces an experimental force field for fitting which would fix an issue where a torsion for P=N double bonds occurring in some molecules from our coverage set, such as the SMILES below, are not covered. The force field provided here is not intended for use, but I'm providing it so @yudongqiu can take a pass at fitting, which might yield a FF worth including here. 

(This is with reference to the [coverage set](https://github.com/openforcefield/qca-dataset-submission/tree/master/2019-06-25-smirnoff99Frost-coverage) )

Here are example molecules with this torsion:
- `[H]c1c(c(c(c(c1[H])[H])P(=N[H])(c2c(c(c(c(c2[H])[H])[H])[H])[H])c3c(c(c(c(c3[H])[H])[H])[H])[H])[H])[H]`
- `[H]C1=C(N(C(=C1[H])C(=S)N=P(N(C([H])([H])[H])C([H])([H])[H])(N(C([H])([H])[H])C([H])([H])[H])N(C([H])([H])[H])C([H])([H])[H])C([H])([H])[H])[H]`

### DIFF
--- a/openforcefield/data/test_forcefields/smirnoff99Frosst_experimental_PNdouble.offxml
+++ b/openforcefield/data/test_forcefields/smirnoff99Frosst_experimental_PNdouble.offxml
@@ -282,7 +282,7 @@
 		<Proper smirks="[*:1]~[#7X3:2]-[#15:3]~[*:4]" periodicity1="2" phase1="180.0 * degree" k1="2.5 * mole**-1 * kilocalorie" id="t152" idivf1="1.0"></Proper>
 		<Proper smirks="[*:1]-[#7:2]-[#15:3]=[*:4]" periodicity1="2" periodicity2="3" phase1="180.0 * degree" phase2="0.0 * degree" k1="3.0 * mole**-1 * kilocalorie" k2="2.3 * mole**-1 * kilocalorie" id="t154" idivf1="1.0" idivf2="1.0"></Proper>
 		<Proper smirks="[#6X3:1]-[#7:2]-[#15:3]=[*:4]" periodicity1="1" phase1="0.0 * degree" k1="2.3 * mole**-1 * kilocalorie" id="t155" idivf1="1.0"></Proper>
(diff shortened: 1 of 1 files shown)


yudongqiu:
I guess you have an @ to another person :smile:

Can you point out which is the new term added here?

jchodera:
@davidlmobley : Can you clarify if any of these molecules are included in one of the current https://github.com/openforcefield/qca-dataset-submission sets, or if it needs more dataset coverage?

jchodera:
Also, these are called [phosphazenes](https://en.wikipedia.org/wiki/Phosphazene).

davidlmobley:
That's how we found this was missing -- these parameters occur in molecules in the coverage set, but we couldn't parameterize the molecules because we were missing this torsion. :) 

It sounds like how @yudongqiu constructed the torsion drives for the coverage set means that we did NOT drive these torsions, so we may need to do supplemental torsion drives. But they made it through geometry optimization as I understand it.

davidlmobley:
@yudongqiu it's `t155b`, `[*:1]~[#7:2]=[#15:3]~[*:4]`

yudongqiu:
For a record, we need to do two things for the new term.

First is we want to get a set of initial parameters. These parameters should be reasonable, but they do not have to be very high quality. The plan for this is to run a few quick torsiondrives locally and use ForceBalance to get a roughly good fit for JUST the new term.

Second is that we want to have coverage of these parameters in our datasets, so they're also fitted in the product run when we fit all parameters together. The reason why the current "coverage Torsion Dataset" does not have this new term, is because it's generated by "selecting 5 torsions for each term exist in the `smirnoff99Frosst_experimental.offxml`. Since that version of forcefield does not have this new term, no torsions were selected to even though they show up in some of the molecules. To solve this, we will later need to improve the existing "coverage Torsion Dataset", by using the new forcefield file to select torsions and re-create the dataset.

davidlmobley:
This can be closed; closing.

PR #403, opened 2019-08-20
Make advanced_examples subfolder and add lysine_gromacs_simulation

j-wags:
### Summary

This PR is to bring in Dennis Della Corte's iPython example, with the summary:

>This example shows how to use the Open Force Field Toolkit to produce an OpenMM system for a small organic molecule (lysine), convert it to a pair of GROMACS input files (`top`/`gro`), add waters and ions, and run a GROMACS simulation. 

This is our first submitted example, and I suspect that we may receive more of these in the future. This is particularly valuable, as it explicitly shows how to go from starting structures to _running a simulation in GROMACS_. Having a mechanism/checklist for bringing in such examples will add a lot of value in our effort to ensure compatibility with the various simulation communities.  Let's use this as an opportunity to discuss a standard process for bringing in and maintaining these moving forward. 

### Open questions

* What should our policy on maintenance be?
* There appears to be [a GROMACS package available on the bioconda channel](https://anaconda.org/bioconda/gromacs/files). It posts packages for Linux and OSX. Do we want to test that out to make this example more automagic?
* I propose we keep these in a subdirectory `examples/advanced_examples/`. This will indicate that the contents may not be as actively maintained, or as automagic as the top-level. Is this generally agreeable?
* In the IPython notebook, there are some required files on DropBox -- Do we want to bring these into the repo (do we want to maintain them? Are we allowed to re-host them under the MIT license?)
* Some of the install instructions are redundant (installs the toolkit, openmm, and smirnoff99Frosst separately, but I think they all come down automatically with the current `conda` OFFTK package)


j-wags:
Note that, as we're busy preparing for the in-person meeting, it may be a few weeks before we get to this. In the meantime please feel free to post thoughts here, and we can discuss asynchronously.

PR #420, opened 2019-09-17
Convert old-style % strings to f-strings

jaimergp:
I applied [fstringify](https://github.com/jacktasia/fstringify) on the whole package. This tool works on the AST level to check that the code behavior remains the same, so this shouldn't have an effect on the tests.

`fstringify` won't convert `"%s %s" % (item1, item2)"` %-formatted tuples, so these will be converted by hand.

# Tasks
- [x] Apply `fstringify`
- [x] Manually search for %-strings not detected by fstringify

# Review task list
- All modified strings start with `f`!
- Replaced variables are in the same order as the original code
- Format syntax is always preceded by `:`


jaimergp:
Well, that should do it!

I didn't convert the notebooks because editing the raw strings is too dangerous with those escaped quotes everywhere... :fearful:

j-wags:
This looks really good. I noticed that our travis builds had started failing as of two weeks ago, so I'm going to fix that before I merge this. Thanks for pitching in, @jaimergp!

PR #422, opened 2019-09-19
[WIP] Update build system to latest cookiecutter standard

### LINKED ISSUE #237: Update travis build system to new MolSSI cookiecutter

j-wags:
Our build system has started falling behind best practices, and I suspect that some of our current build woes can be resolved by updating to [the newest cookiecutter](https://github.com/MolSSI/cookiecutter-cms). 

- Closes #237 
- Updates OE license (using Mobley lab license now)
- Band-aid for ChemicalEnvironments toolkit selector (allow simple auto-detection of available toolkits so docstring examples don't fail when RDKit isn't around)
- I deleted the unused encryption keys on Travis, since the list of secrets was 11 items long
- I changed the travis release branch regex to `maintenance/.+`, per Daniel's comment below. (We don't currently have any maintenance branches)


.travis.yml
```diff
@@ -1,102 +1,97 @@
...
-  - conda create -n test python=$PYTHON_VER pip pytest pytest-cov nbval
-  - source activate test
+# Install conda and some package management tools
+- source devtools/travis-ci/before_install.sh
+- python -V
+
+# Unpack encrypted OpenEye license file
+#- if [ "$TRAVIS_SECURE_ENV_VARS" == true ]; then openssl aes-256-cbc -K $encrypted_a4fdffb2b55a_key
+#  -iv $encrypted_a4fdffb2b55a_iv -in oe_license.txt.enc -out $OE_LICENSE -d; fi
+- if [ "$TRAVIS_SECURE_ENV_VARS" == true ]; then openssl aes-256-cbc -K $encrypted_4ea8671dd0e3_key
```
dgasmith: Great that this is optional!

devtools/conda-envs/test_env.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - openmmtools
+  - pyyaml
+  - toml
+  - bson
+  - msgpack-python
+  - xmltodict
+
+    # Pip-only installs
+  - pip:
+    - coverage==4.4
```
dgasmith: Coverage is now on conda and conda-forge.

.travis.yml
```diff
@@ -1,102 +1,97 @@
...
+# Run builds for both master and feature releases
+
+language: python
+
 branches:
   only:
   - master
   # Regex for feature release branches
-  - /.*[.]0-branch$/
+  - "/.*[.]0-branch$/"
```
dgasmith: Release branches are canonically named something like `maintenance/0.4.x` or similar.

devtools/conda-envs/test_env.yaml
j-wags: Moved to conda installs in 9c47611

.travis.yml
j-wags: Thanks for the insight. Updated in 7981ad7

PR #430, opened 2019-09-30
WBO testing

sukanyasasmal:
The purpose of this PR is to 
- [x] Bring the OE ToolkitWrapper WBO functionality to a working state
- [x] Design and implement sensible validation tests against predictable molecules with known confounding factors
- [ ] (maybe this PR, maybe another) Add support for bond order interpolation to the `BondHandler` class
- [ ] (maybe this PR, maybe another) Update the SMIRNOFF spec page to show this functionality is available and state any limitations/changes from preliminary spec
- [ ] (maybe this PR, maybe another) Add OFFXML files based on a existing bond-order-interpolating FF

Status
- [x] Ready for review
- [ ] Ready for merge


davidlmobley:
You can also put "[WIP]" in the title to make more clear this is something you're working on.

jchodera:
Could you add a bit more information about what this PR intends to accomplish?

j-wags:
@jchodera -- We've been discussing this in-person, but I should have posted more info in the PR description. I've modified the PR text above with the goals/progress regarding this work.

openforcefield/tests/test_wbo.py
```diff
@@ -0,0 +1,46 @@
...
+#mol1 = Molecule.from_smiles("C1=NC=CC(=C1)C2=CC=C(C=C2)O")
+#mol1.generate_conformers(n_conformers=1)
+#mol1.compute_wiberg_bond_orders()
+#
+#mol2 = Molecule.from_smiles("C1=NC=CC(=C1)C2=CC=C(C=C2)[O-]")
+#mol2.generate_conformers(n_conformers=1)
+#mol2.compute_wiberg_bond_orders()
+
+# Reading neutral molecule from file
+filename = get_data_file_path('molecules/molecule_neutral.sdf')
```
j-wags: Thanks for using `get_data_file_path` -- It's good that our tests use the same infrastructure.

openforcefield/tests/test_wbo.py
```diff
@@ -0,0 +1,46 @@
...
+for i in molecule2.atoms[4].bonds:
+    if i.atom1_index == 6 or i.atom2_index == 6:
+        wbo_anion = i.fractional_bond_order
+
+# Checking that only one additional bond is present ihe neutral molecule
+assert (len(molecule1.bonds)==len(molecule2.bonds)+1)
+# Wiberg bond order is higher in the anion
+assert (wbo_anion > wbo_neutral)
+
+
```
j-wags: It would be good to add sanity checks around the central bond. I don't know what's "sane" for WBOs, but maybe something like 0.75 < WBO < 1.25? 

It would also be good to check that all CC bonds are in a range near 1.5, and CH bonds are around 1 as well?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2676,7 +2676,7 @@ def create_force(self, system, topology, **kwargs):
                 #temp_mol.compute_partial_charges(quantum_chemical_method=self._quantum_chemical_method,
                 #                                 partial_charge_method=self._partial_charge_method)
                 temp_mol.compute_partial_charges_am1bcc(toolkit_registry=toolkit_registry)
-
+                temp_mol.compute_wiberg_bond_orders(toolkit_registry=toolkit_registry)
```
j-wags: I don't think we actually want this to be in the `ToolkitAM1BCCHandler`. WBOs will tie in during the assignment of bond parameter values, so this should be in `BondHandler.create_force`. We'll probably need to copy `generate_conformers` to there as well.

openforcefield/tests/test_wbo.py
```diff
@@ -0,0 +1,46 @@
...
+molecule1.compute_wiberg_bond_orders()
+# Fetching wiberg bond order of of C-C bond
+for i in molecule1.atoms[4].bonds:
+    if i.atom1_index == 6 or i.atom2_index == 6:
+        wbo_neutral = i.fractional_bond_order
+
+# Reading negative molecule from file
+filename = get_data_file_path('molecules/molecule_anion.sdf')
+molecule2 = Molecule.from_file(filename)  # not using toolkitwrapper flag
+molecule2.compute_wiberg_bond_orders()
```
j-wags: It would be good to ensure that both `charge_model='AM1'` and `PM3` keywords perform correctly in this test.

j-wags:
@sukanyasasmal I've realized that we haven't specified the extent of this PR (should this cover everything through actual bond parameter assignment?). I think, for now, we just should focus on adding this as a test and ensuring that we reproduce the _physical_ phenomenon we expect (higher bond order in the anion). Once the test is integrated, we can decide if we want this PR to go all the way through parameter assignment. 

I've left a few small suggestions as comments. But overall I think this is ready to move into `test_toolkits`!

openforcefield/tests/test_wbo.py
sukanyasasmal: Added.

openforcefield/tests/test_wbo.py
sukanyasasmal: Added sanity check for all the bonds.

openforcefield/tests/test_toolkits.py
```diff
@@ -451,6 +451,68 @@ def test_compute_wiberg_bond_orders(self):
...
+                    assert (0.85 < i.fractional_bond_order < 1.05)
+                elif (i.atom1.atomic_number == 8 or i.atom2.atomic_number == 8):
+                    # Checking C-O single bond
+                    wbo_C_O_neutral = i.fractional_bond_order
+                    assert (1.0 < wbo_C_O_neutral < 1.5)
+                else:
+                    # Should be C-C single bond
+                    assert (i.atom1_index == 4 and i.atom2_index == 6) or (i.atom1_index == 6 and i.atom2_index == 4)
+                    wbo_C_C_neutral = i.fractional_bond_order 
+                    assert (1.0 < wbo_C_C_neutral < 1.3)
```
j-wags: Nice! I really like how thorough this is.

j-wags:
@sukanyasasmal I've made two quick commits on this branch[1], so be sure to run `git pull` before you make further changes!

I think the new code in is good shape. Thanks again for adding the detailed test -- I have much more faith in our WBO implementation now, and having this test built will allow us to validate the AmberTools implementation much more easily once it's ready to be incorporated. 

For next steps, I think we should do the following:
1) Update this branch with the recent changes to `master` using the button on this page, and resolve any merge conflicts
2) Add a `Tests Added` section to the `Current Release` section in `docs/releasehistory.rst` acknowledging your contributions and citing the talk where this molecule was discussed
3) Merge the PR
4) Discuss whether you would like to work on building+testing the infrastructure to do bond order interpolation during actual parametrization.

So, let's have a Zoom call some time this week :-) 

[1] The two commits did the following
- rename the new molecules from `molecule` to `CID20742535` (the pubchem ID of the structure)
- remove the WBO-calculation call from `ToolkitAM1BCCHandler.create_force`. The molecule that exists in that function is a temporary copy that only exists during charge assignment, whereas the WBO values would be needed in `BondHandler.create_force` (which has no knowledge of `ToolkitAM1BCCHandler`'s `temp_mol`).

PR #432, opened 2019-10-02
[WIP] Fix codecov

j-wags:
Codecov has dropped by about 30% recently, as a result of changing some dependencies. This appears to be due to codecov now considering _the code within tests themselves_ to be included in the denominator for test coverage %. 

This PR is to find a way to either count the fact that those lines run _during the test itself_, or to remove them from the denominator.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,15 +91,16 @@ script:
 - if [[ "$RDKIT" == true && "$OPENEYE" == true ]];
   then
     pytest --ignore=utilities --ignore=examples/deprecated --ignore=docs
-    --ignore=devtools --doctest-modules --nbval-lax --cov=openforcefield;
+    --ignore=devtools --doctest-modules --nbval-lax --cov=openforcefield
+    --cov-config=setup.cfg;
   elif [[ "$OPENEYE" == true ]];
   then
     pytest --ignore=utilities --ignore=examples/deprecated --ignore=docs
     --ignore=devtools --ignore=examples/check_dataset_parameter_coverage
-    --nbval-lax --cov=openforcefield;
+    --nbval-lax --cov=openforcefield --cov-config=setup.cfg;
   else
     pytest --ignore=utilities --ignore=examples/deprecated
-    --nbval-lax --cov=openforcefield;
+    --nbval-lax --cov=openforcefield --cov-config=setup.cfg;
   fi
 
 notifications:
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,15 @@
+
+# Codecov configuration to make it a bit less noisy
+coverage:
+  status:
+    patch: false
+    project:
+      default:
+        threshold: 50%
+comment:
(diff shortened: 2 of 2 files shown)


j-wags:
Based on https://pytest-cov.readthedocs.io/en/latest/config.html and https://github.com/openforcefield/openforcefield/blob/master/setup.cfg#L3-L9, tests SHOULD be getting omitted...

j-wags:
I've tried adding `--cov-config=setup.cfg` to each pytest call. If this is necessary to get test lines omitted, we should push this change up to the cookiecutter as well.

.travis.yml
```diff
@@ -91,15 +91,16 @@ script:
...
     pytest --ignore=utilities --ignore=examples/deprecated --ignore=docs
-    --ignore=devtools --doctest-modules --nbval-lax --cov=openforcefield;
+    --ignore=devtools --doctest-modules --nbval-lax --cov=openforcefield
+    --cov-config=setup.cfg;
   elif [[ "$OPENEYE" == true ]];
   then
     pytest --ignore=utilities --ignore=examples/deprecated --ignore=docs
     --ignore=devtools --ignore=examples/check_dataset_parameter_coverage
-    --nbval-lax --cov=openforcefield;
+    --nbval-lax --cov=openforcefield --cov-config=setup.cfg;
```
j-wags: Pinging @dgasmith @janash @Lnaden to be aware of this issue.

Basically, I removed our old coverage version pin, and suddenly codecov dropped by 30%. This was due to the code lines _in the tests themselves_ becoming a part of the denominator for our coverage % (but oddly, not the numerator).

Per the advice [here](https://pytest-cov.readthedocs.io/en/latest/config.html), this can be avoided using the `[omit]` keyword in a config file. This is present both in the OFFTK repo and in the cookiecutter, in [`setup.cfg`](https://github.com/openforcefield/openforcefield/blob/master/setup.cfg#L3-L9). However, our `pytest` commands (and the cookiecutter's) don't specifically point to this file (`--cov-config`). 

I suspect that new versions of `coverage` or `pytest-cov` have changed the way that config files are found, such that `setup.cfg` is no longer found by default. This delta should fix that if it's also an issue on your end.

dgasmith: Thats odd, you shouldn't need to pass this in unless you are running pytest in a different folder.

j-wags: The docs are confusing on this point. In fact, they seem to say two opposite things about default behavior in adjacent sentences.

> Caveats
> =======
> 
> A unfortunate consequence of coverage.py's history is that ``.coveragerc`` is a magic name: it's the default file but it also
> means "try to also lookup coverage configuration in ``tox.ini`` or ``setup.cfg``".
> 
> In practical terms this means that if you have your coverage configuration in ``tox.ini`` or ``setup.cfg`` it is paramount
> that you also use ``--cov-config=tox.ini`` or ``--cov-config=setup.cfg``.

Issue #340, opened 2019-05-31
Simplifying handling of cosmetic attributes in SMIRNOFF files


jchodera:
A thought occurred to me about how we might be able to simplify the handling of cosmetic attributes: What if the definition of acceptable optional (cosmetic) attributes for each section was something that could be specified in the SMIRNOFF format? That is, suppose we had a standard way to describe the acceptable top-level and parameter-level optional attributes in the spec? 

This would have the benefit of eliminating the need to specify whether we want to accept, keep, or strip cosmetic attributes on reading and writing. Instead, we would have some API methods for manipulating cosmetic attributes at the object model level. Whatever attributes were present on write would simply be written as allowed cosmetic attributes.

j-wags:
This is an interesting idea. I'm trying to think of this from different perspectives:
* Who "accepts liability" for non-spec attributes? Should it be the person who writes the file, the person who reads it, both, or neither? Whenever we make someone go against the default behavior of our functions (eg. changing a kwarg from its default value), I'm taking that to mean that they're "accepting liability" for a non-default outcome. 
* If supporting workflows that require cosmetic attributes needs to add complexity _somewhere_, is it better for that complexity to be in the FF file or the script that loads the file?

My opinions are:
* The person who reads a SMIRNOFF data source should have to "accept liability" for non-spec attributes. If the cosmetic attribute is really critical to their workflow, to the point that explicitly allowing it adds a significant number of characters to their code, they should write their own ParameterHandler where the attributes aren't cosmetic.
* The person who writes a SMIRNOFF data file "acknowledges"/"accepts liability" for their addition of cosmetic attributes by including them in the first place. I think that making them list cosmetic attributes a second time in a new section has some benefit in terms of validation, but that benefit is outweighed by the added complexity.
* If additional complexity needs to be added somewhere, it would be better to add it in the FF file (which is not advertised as being human-readable) than the script that loads it (which should be human-readable). This is a somewhat weak preference, however.

The approaches we're considering are:
1) At an extreme, we could decide "no need to declare things as cosmetic, we'll just read and write every attribute by default". The upside of this is the simplicity. The downside is that, for very specific cases (like defining a ProperTorsion's `k3` without a `k2`) the user will silently get unexpected behavior. This is very bad. In this scheme the publisher has kind of "accepted liability", but no complexity is added.
[...]

Issue #310, opened 2019-05-10
Container class for parameterized systems


jchodera:
I've been discussing (with @andrrizzi @SimonBoothroyd @j-wags @proteneer) the potential for solving a number of technical debt issues we have accumulated with a new Python container class to represent a parameterized system.

To recap, we have a few major accumulated technical debt issues:
* **Use of Swig-wrapped C++ objects as containers causes problems.** Currently, [`ForceField`](https://open-forcefield-toolkit.readthedocs.io/en/latest/api/generated/openforcefield.typing.engines.smirnoff.forcefield.ForceField.html#openforcefield.typing.engines.smirnoff.forcefield.ForceField) is mainly structured around constructing an OpenMM [`System`](http://docs.openmm.org/latest/api-python/generated/simtk.openmm.openmm.System.html#simtk.openmm.openmm.System) object, which is a Swig-wrapped C++ object that violates PEP8 naming principles with an highly un-pythonic API. The overhead of these conversions can also be significant, and they are time-consuming to serialize via XML.
* **Nonstandard unit handling.** OpenMM also uses `simtk.unit`, rather than a more standard library like [`pint`](https://pint.readthedocs.io), for handling units; this was discussed [here](https://github.com/openforcefield/openforcefield/issues/151). In addition, `simtk.unit` does not easily support ensuring dimensionality is correct.
* **Difficulty decoding parameter types for conversion.** While OpenMM provides flexible `CustomForce` classes which can _encode_ many functional forms for existing parameter types or new parameter types, it becomes very difficult to _decode_ these custom forces to convert to representations suitable for other simulation packages without a lot of specialized logic to try to detect what we encoded. OpenMM also generally mixes electrostatics and vdW interactions in a single [`NonbondedForce`](http://docs.openmm.org/latest/api-python/generated/simtk.openmm.openmm.NonbondedForce.html#simtk.openmm.openmm.NonbondedForce), which creates further complexity for decoding.
* **Inextensibility.** While OpenMM `System` objects are very flexible, they are inextensible without writing C++ plugins.
* **Tracking origin parameters in parameterized systems is difficult.** The parameterized OpenMM `System` object has all parameters unrolled into individual interactions, which makes it difficult to identify how origin parameters affect multiple interactions.
[...]

PR #433, opened 2019-10-08
Library Charge Support

### LINKED ISSUE #25: Implement Library Charges

### LINKED ISSUE #434: Topology.add_molecule has undefined behavior if given an existing molecule with different atom indexing
**Describe the bug**
If the user passes a re-indexed version of a molecule which is already in a `Topology` to`Topology.add_molecule`, the molecule that is added is indexed in the same order as the original. This can lead to system creation errors down the line if the user plans to use the coordinates from the reindexed molecule. 

**To Reproduce**
```
        ff = ForceField('test_forcefields/smirnoff99Frosst.offxml')

        molecules = [Molecule.from_file(get_data_file_path('molecules/ethanol.sdf')),
                     Molecule.from_file(get_data_file_path('molecules/ethanol_reordered.sdf'))]
        top = Topology.from_molecules(molecules)
        omm_system = ff.create_openmm_system(top)
```

j-wags:
- [x] Close #25 
- [x] Fix #434 
- [x] Add common base class (`_NonbondedHandler`) for `ParameterHandlers` that deal with OpenMM's `NonbondedForce` objects
- [x] Make _NonbondedHandlers know whether a previous handler has assigned charges to a molecule (and therefore to skip it)
    - I ended up having `_NonbondedHandler`s attach this information to the input `Topology` during system creation (by creating and modifying a new dict `topology._ref_mol_to_charge_method`). This is performed using `_NonbondedHandler.mark_charges_assigned(self, ref_mol, topology)` and `NonbondedHandler.check_charges_assigned(ref_mol, topology)`. This is not an ideal solution, since the `Topology` shouldn't need to know anything about the parameters that are assigned. But until we have our own `System` class, there is no way to store this information in OpenMM's `System` class, and the `Topology` is the only other data structure we have continuous access to during system creation. 
    - Especially relevant was the addition of `test_charge_method_hierarchy` in `test_forcefield` -- Basically, a molecule should get charges from the `charge_from_molecules` kwarg, then LibraryCharges, then AM1-BCC. The new hierarchy test is fairly exhaustive in ensuring that this is the actual behavior. 

Status
- [x] Ready for review
- [ ] Ready for merge


proteneer:
Thanks for doing this!

1) Might make sense to add a test of Library Charges in conjunction with multiple overlapping BCCs. For us down the road, training to BCCs probably makes more sense than the charges directly.

2) Will there actually be a set of parameters released? (eg. for proteins)

j-wags:
@proteneer 

> Might make sense to add a test of Library Charges in conjunction with multiple overlapping BCCs. For us down the road, training to BCCs probably makes more sense than the charges directly.

That's a good point. I'd planned on doing library charges and BCCs separately, but I've had to refresh my knowledge of our charging systems during this work, so it may make sense to do them simultaneously. My goal is to get the library charge functionality+tests working and cleaned up by this afternoon. If I can do that, I'll see if I could get BCCs in early next week.

> Will there actually be a set of parameters released? (eg. for proteins)

No. During the in-person meeting, we discussed which order we should add new functionality, and time estimates for each. Missing toolkit functionality is blocking for several OFF subprojects. I pointed out the massive GBSA time overrun was largely due to my writing and validating the GBSA parameter files, not the GBSAHandler itself. So, in the interest of time, LibraryCharges and BCCs will come out without "real" parameters (except for dummy files used for tests). 

I know Caitlin had been working on porting an AMBER protein FF's charges to SMIRNOFF format, but the LibraryChargeHandler wasn't ready before she left, so that never got finished.

proteneer:
Thanks - this sounds good.

jchodera:
@j-wags : Is there an issue we can link to that describes how this functionality will work, and the order in which charge assignment schemes will operate?

j-wags:
@jchodera 

> Is there an issue we can link to that describes how this functionality will work, and the order in which charge assignment schemes will operate? Or has the spec been updated to describe this in detail?

I can add the current behavior to the spec if it's not there already. Basically, the priority will be `charge_from_molecules` > `LibraryCharges` > `ToolkitAM1BCC`. BCCs will be applied to the resulting charges afterwards, if present. 

The way that the object model works now creates some problems, like if a library charge parameter DOES match a molecule, but sets the charges for a molecule to be all zeroes, then `ToolkitAM1BCC` won't know that it matched at all, and will run AM1-BCC. This is because the `ParameterHandlers` are completely modular, and have to infer the history of the charging process solely by the information in the OpenMM `System` when their turn comes up. 

So, we'll need to do a SMIRNOFF spec change in the future, where the charge-generating handlers lose some modularity. This will likely involve making the charge-generating handlers be defined in a subsection within the `Electrostatics` tag, however this needs to be reconciled against the fact that `VirtualSites` will have _both_ charge _and_ LJ parameters. Decisions on that front also tie into the architecture of our `System` class. But this is a restatement of concerns in #310 and #340. 

Anyway, we'll be debating the spec change for several months in the best case, so since there are people funded to do this science _now_, it makes sense to put out hard-coded functionality in the short term.

jchodera:
Thanks for the detailed explanation, @j-wags!

Will the library charges have to match the entire molecule, or can they match a biopolymer residue at a time?

j-wags:
Whoops, missed the question above. Library charges will operate on a molecule as long as each atom in the molecule has _at least_ one match. The final charge that gets assigned to each atom will be determined using our normal hierarchy rules. Of course, this means that the "safest" way to use LibraryCharges is to specify a whole molecule SMIRKS. But if folks want to do partial molecule SMIRKS, that will work as long as all atoms are assigned. However if those partial molecule SMIRKS overlap with each other, it can become very complex to understand the outcome. 

Anyway, that means that we could _in theory_ support biopolymer parameterization now. However, we still don't have any concept of residues, so parameter assignment might be slow (would need to SMIRKS-match a whole protein!)

One important note to emphasize is that we _do not_ currently support using multiple charge methods for a single molecule. So, a user could not use LibraryCharges for protein residues and AM1-BCC just for capping groups and PTMs.

openforcefield/tests/test_forcefield.py
```diff
@@ -643,8 +728,7 @@ def test_parameterize_ethanol_missing_torsion(self):
 </SMIRNOFF>
 ''')
         pdbfile = app.PDBFile(get_data_file_path('systems/test_systems/1_ethanol.pdb'))
-        molecules = []
-        molecules.append(Molecule.from_smiles('CCO'))
```
j-wags: In test where atom indexing is important, I've tried to remove uses of `Molecule.from_smiles`, since that does not guarantee a specific ordering of the atoms.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2489,12 +2546,117 @@ def check_handler_compatibility(self,
                                          tolerance_attrs=float_attrs_to_compare+unit_attrs_to_compare,
                                          tolerance=self._SCALETOL)
 
+
+    def assign_charge_from_molecules(self, molecule, charge_mols):
```
j-wags: Previously, `ToolkitAM1BCCHandler` was responsible for assigning charges from molecules in the `charge_from_molecules` kwarg. However, it makes more sense to have that functionality in the `ElectrostaticsHandler`, since the addition of LibraryCharges means that we won't always a `ToolkitAM1BCCHandler` in the `ForceField`.

openforcefield/data/test_forcefields/tip3p.offxml
```diff
@@ -1,26 +1,21 @@
 <?xml version="1.0" encoding='ASCII'?>
```
j-wags: I've resuscitated the previously non-functional `tip3p.offxml`.

openforcefield/tests/test_forcefield.py
```diff
@@ -806,6 +879,70 @@ def test_parameterize_ethanol_to_parmed(self):
...
+                        nonbond_method_matched = True
+            assert nonbond_method_matched
+        else:
+            with pytest.raises(exception, match=exception_match) as excinfo:
+                # The method is validated and may raise an exception if it's not supported.
+                forcefield.get_parameter_handler('vdW', {}).method = vdw_method
+                forcefield.get_parameter_handler('Electrostatics', {}).method = electrostatics_method
+                omm_system = forcefield.create_openmm_system(topology)
+
+
```
j-wags: These are not new tests, they were just moved.

j-wags:
@jchodera @SimonBoothroyd @slochower @andrrizzi This PR is ready for review if any of you have time. It's somewhat large, so I've left some comments that should add context and speed up the review. A detailed summary of changes is in the summary of this PR and the additions to `releasehistory.rst`

Also, I'd appreciate your thoughts on the following: the previous behavior when a molecule couldn't be assigned charges (for example, if we deleted the ToolkitAM1BCCHandler) was to just let it go into the system with all zero charges. However, now that people have more freedom to mix and match charge methods, we probably want to raise an error if a molecule doesn't get charges assigned. I've gone ahead and implemented an`UnassignedChargeParameterException` that gets raised in this scenario and noted the behavior change in the release notes.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2489,12 +2551,117 @@ def check_handler_compatibility(self,
...
+        """
+        Given an input molecule, checks against a list of molecules for an isomorphic match. If found, assigns
+        partial charges from the match to the input molecule.
```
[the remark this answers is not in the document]
slochower: This looks good, but wonder if in the future, it'd be nice to do LibraryCharges on fragments.

The-SMIRNOFF-force-field-format.md
```diff
@@ -218,6 +216,13 @@ For example, to ensure water molecules are assigned partial charges for [TIP3P](
 </LibraryCharges>
 ```
 
+
+| LibraryCharges section tag version | Tag attributes and default values    | Required parameter attributes   | Optional parameter attributes |
+|------------------------------------|--------------------------------------|---------------------------------|-------------------------------|
+| 0.3                                |                                      | `smirks`, `charges` (indexed)   | `name`, `id`, `parent_id`     |
+
+
+
```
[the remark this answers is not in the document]
j-wags: Yes. Each _section_ has its own version in the SMIRNOFF 0.3 spec. 

This _seems_ excessive, but we recognized during the last spec revamp that, for example, the `Electrostatics` tag might in the future need to contain detailed PME settings, or that maybe some day we'll need to indicate that the `Angles` section can support `partial_bondorder` keywords. If we only have the top-level SMIRNOFF tag recording complete file compatibility, we'd be incrementing the number constantly. So, the top-level SMIRNOFF tag only encodes things about _general_ structure that affects (de)serialization (eg. units in headers vs units in quantity), without worrying about the specific keywords inside. This leaves it up to the specific _section_ versions to indicate changes in keywords or behavior.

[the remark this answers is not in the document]
j-wags: Confusingly, all _section_ versions begin at 0.3, since that's the first SMIRNOFF spec version that stated that sections have versions. That may have been a poor choice by me, but we did it, so that's the system now.

The-SMIRNOFF-force-field-format.md
```diff
@@ -218,6 +216,13 @@ For example, to ensure water molecules are assigned partial charges for [TIP3P](
...
 ### `<ChargeIncrementModel>`: Small molecule and fragment charges
```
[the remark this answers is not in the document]
j-wags: Covered in previous comment -- All sections got versions in the SMIRNOFF 0.3 spec. Since they implicitly "had" versions in previous SMIRNOFF specs, those were assumed to be 0.1 and 0.2. By versioning each section independently, sections can undergo rapid development/improvement, while maintaining their own sensible versioning scheme. But while sections progress rapidly, we won't need to update the _top-level_ SMIRNOFF spec version several times a year.

openforcefield/data/test_forcefields/tip3p.offxml
```diff
@@ -1,26 +1,21 @@
...
-    <!-- WARNING: charges are not supported yet (see issue openforcefield#25), and they are currently here only for documentation -->
-    <Atom smirks="[#1]-[#8X2H2+0:1]-[#1]" id="n1" sigma="0.31507524065751241" epsilon="0.635968" charge="-0.834"/>
-    <!-- TIP3P water hydrogen with charge override -->
-    <!-- WARNING: charges are not supported yet (see issue openforcefield#25), and they are currently here only for documentation -->
-    <Atom smirks="[#1:1]-[#8X2H2+0]-[#1]" id="n2" sigma="1" epsilon="0" charge="0.417"/>
+  <!-- These values are taken from OpenMM's TIP3P implementation: https://github.com/openmm/openmm/blob/f9106ddb6a0d72c960c73b23ed0e07cd7190a80f/wrappers/python/simtk/openmm/app/data/tip3p.xml -->
+  <vdW version="0.3" potential="Lennard-Jones-12-6" combining_rules="Lorentz-Berthelot" scale12="0.0" scale13="0.0" scale14="0.5" scale15="1" switch_width="1.0 * angstroms" cutoff="9.0 * angstroms" method="cutoff">
+    <Atom smirks="[#1]-[#8X2H2+0:1]-[#1]" id="n1" sigma="0.31507524065751241 * nanometers" epsilon="0.635968 * kilojoules_per_mole" />
+    <Atom smirks="[#1:1]-[#8X2H2+0]-[#1]" id="n2" sigma="1 * nanometers" epsilon="0 * kilojoules_per_mole" />
```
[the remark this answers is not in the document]
j-wags: These seem pretty safe, short of intentionally malicious use. The only further safeguard might be to explicitly state `[#1X1:1]` to ensure that none of the hydrogens have two bonds. But... I'd be impressed if that showed up in the bug tracker.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2219,12 +2224,78 @@ def create_force(self, system, topology, **kwargs):
...
+        Parameters
+        ----------
+        ref_mol : openforcefield.topology.Molecule
+            The molecule to mark as having charges assigned
+        topology : openforcefield.topology.Topology
+            The topology to record this information on.
+
+        """
+        # TODO: Change this to interface with system object instead of topology once we move away from OMM's System
+        topology._ref_mol_to_charge_method[ref_mol] = self.__class__
```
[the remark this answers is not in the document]
j-wags: The `_ref_mol_to_charge_method` dict records where each unique molecule in the topology gets its charge from. It's initialized to have them all be `None`, but those get filled in as charging occurs.

Since `_NonbondedForce` is inherited by `ElectrostaticsHandler`, `ToolkitAM1BCCHandler`, and `LibraryChargeHandler`, when `self.__class__` actually runs, it will be inside of one of those classes. So the dict will be populated with pairs of the form {`reference_molecule`: charge-assigning `_NonbondedHandler`-derived class}.

openforcefield/typing/engines/smirnoff/parameters.py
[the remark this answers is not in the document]
j-wags: This function is for handling the `charge_from_molecules` kwarg. It is not related to LibraryCharges (except that it gets higher priority in the charge method hierarchy). This code existed before the PR, it's just been moved into `ElectrostaticsHandler` because its priority was messed up when it was attached to `ToolkitAM1BCCHandler`

docs/releasehistory.rst
```diff
@@ -11,6 +11,34 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
...
+  `Issue #25 <https://github.com/openforcefield/openforcefield/issues/25>`_ by adding
+  initial support for the
+  `LibraryCharges tag in the SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#librarycharges-library-charges-for-polymeric-residues-and-special-solvent-models>`_
+  using
+  :py:class:`LibraryChargeHandler <openforcefield.typing.engines.smirnoff.parameters.LibraryChargeHandler>`.
+  For a molecule to have charges assigned using LibraryCharges, all of its atoms must be covered by
+  at least one LibraryCharge. If an atom is covered by multiple LibraryCharges, then the last
+  one read will be applied. At this time, there is no concept of "residues" during parametrization,
+  so it is not possible to parametrize some atoms in a molecule using LibraryCharges and
+  others using another method. Further, no effort is made to ensure that the net charge
```
jchodera: > Further, no effort is made to ensure that the net charge applied using LibraryCharges is the total formal charge on the molecule.

Since we have all the formal charge information for any provided molecules (from the toolkit layers), wouldn't it be trivial to add this check?

docs/releasehistory.rst
```diff
@@ -11,6 +11,34 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
...
+""""""""""""
+- `PR #433 <https://github.com/openforcefield/openforcefield/pull/433>`_: Closes
+  `Issue #25 <https://github.com/openforcefield/openforcefield/issues/25>`_ by adding
+  initial support for the
+  `LibraryCharges tag in the SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#librarycharges-library-charges-for-polymeric-residues-and-special-solvent-models>`_
+  using
+  :py:class:`LibraryChargeHandler <openforcefield.typing.engines.smirnoff.parameters.LibraryChargeHandler>`.
+  For a molecule to have charges assigned using LibraryCharges, all of its atoms must be covered by
+  at least one LibraryCharge. If an atom is covered by multiple LibraryCharges, then the last
+  one read will be applied. At this time, there is no concept of "residues" during parametrization,
```
jchodera: It's unclear if this means we can parameterize a biomolecule made up of a few "residues" or not. Can you clarify whether or not it is possible to provide a library of residue charge definitions and have the whole protein charged this way, provided all atoms have one LibraryCharge applied?

The-SMIRNOFF-force-field-format.md
```diff
@@ -218,6 +216,13 @@ For example, to ensure water molecules are assigned partial charges for [TIP3P](
 </LibraryCharges>
 ```
 
+
```
jchodera: Are the `<LibraryCharges>` examples above still accurate?

docs/releasehistory.rst
```diff
@@ -11,6 +11,34 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
...
+  so it is not possible to parametrize some atoms in a molecule using LibraryCharges and
+  others using another method. Further, no effort is made to ensure that the net charge
+  applied using LibraryCharges is the total formal charge on the molecule.
+
+Behavior changed
+""""""""""""""""
+- `PR #433 <https://github.com/openforcefield/openforcefield/pull/433>`_: If a molecule
+  can not be assigned charges by any charge-generation method, an
```
jchodera: Can you clarify whether passing in charged molecules via the API can suffice here, or if it will throw this exception if the SMIRNOFF file does not contain enough information to charge all molecules in the system (even if you pass a charged molecule in via the API).

openforcefield/tests/test_forcefield.py
jchodera: > In tests where atom indexing is important, I've tried to remove uses of Molecule.from_smiles, since that does not guarantee a specific ordering of the atoms.

We can fix this! We can support tagged SMILES, such as `[C:1][C:2][O:3]`, and ensure that we create atoms with the specified ordering. This would allow the tests to be much more compact, and could have other uses in future.

I believe `cmiles` has methods to do this---we should consult with @ChayaSt .

openforcefield/tests/test_forcefield.py
```diff
@@ -806,6 +879,70 @@ def test_parameterize_ethanol_to_parmed(self):
 
         parmed_system = forcefield.create_parmed_structure(topology, positions=pdbfile.getPositions())
 
+
+
+
+    @pytest.mark.parametrize("toolkit_registry,registry_description", toolkit_registries)
+    def test_pass_invalid_kwarg_to_create_openmm_system(self, toolkit_registry, registry_description):
```
jchodera: Do we still accept `kwargs` to `create_openmm_system()`? I thought we had deprecated that since it was so terribly dangerous!

openforcefield/tests/test_forcefield.py
j-wags: There have been requests for this functionality from several sources. I've opened an initial Issue with the spec for this: https://github.com/openforcefield/openforcefield/issues/448

docs/releasehistory.rst
j-wags: I'd thought about it, but concluded that it would add unnecessary complexity. At some point, maybe there will be a consensus that folks want a charge normalization scheme implemented, and we can incorporate a "don't-normalize-but-just-raise-exception-if-charges-don't-add-up" option. But for now, adding this functionality would make people immediately ask for a kwarg to disable it, and then in the future we'd have to break the API or have a growing tree of "if" statements to handle that kwarg's interaction with various normalization schemes.

jchodera: This seems highly dangerous to me since a mistyped library charge definition could lead to corrupted, non-integral charges. Wouldn't we want to be conservative first and then allow a way to relax things later if we find we need to?

j-wags: > This seems highly dangerous to me since a mistyped library charge definition could lead to corrupted, non-integral charges.

I see where you're coming from -- Our API should just "do the right thing", and that is the attitude I take toward users who show up with molecules and just "want to use the most recent FF". But here we're considering a different group of users -- those writing their own SMIRKS and `LibraryCharge` entries. I think it's likely that those users would find that our "helpful" guiderails actually get in their way. 

There are several use cases where raising an error would get in the way. For example, I suspect that someone will start experimenting with fragment-based charge assignment, with the intent to perform their own normalization on the charges in the resulting molecule. They'll be frustrated when we disallow this. 

I agree that it's possible for promiscuous SMIRKS to quietly ruin a system, so we may have to ensure that the FFs _published by Open Force Field_ don't contain potentially-promiscuous library charges. But, we already let people do tremendously dangerous things like introduce wildcard parameters into their `Bonds` section, while also concatenating sections from subsequently-loaded FFs (so, if a wildcard bond-containing file is read after Parsley, all bonds will match the wildcard). The spec allows people to create the FF equivalent of nuclear fuel -- It's great if someone else is handling it, but people need to understand that messing with parameter sources is dangerous.

What I'm getting at is that this isn't a _technical_ problem -- It would be easy to add the "formal charge" check and a resulting warning or exception+override. But users already complain that we have too many warnings. And an exception+override adds complexity on our end. If we go that route, I bet that `allow_nonintegral_charges=True` will just become another `allow_undefined_stereo=True`, which already seems to be cargo-culting around in scripts.

SimonBoothroyd: +1 for enforcing that the charges add up to the formal charge with the option to disable the check. I think most people (by which I mean me) would prefer to be told explicitly when the charges don't add up, thus highlighting something which is usually very wrong in the system without having to write checks for this each time a molecule is parametrized.

SimonBoothroyd: I think this is especially true when the default is to take the last charge to be read!

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2219,12 +2224,78 @@ def create_force(self, system, topology, **kwargs):
...
+    """Base class for ParameterHandlers that deal with OpenMM NonbondedForce objects."""
+    _OPENMMTYPE = openmm.NonbondedForce
+
+    def create_force(self, system, topology, **kwargs):
+        # If we aren't yet keeping track of which molecules' charges have been assigned by which charge methods,
+        # initialize a dict for that here.
+        # TODO: This should be an attribute of the _system_, not the _topology_. However, since we're still using
+        #  OpenMM's System class, I am storing this data on the OFF Topology until we make an OFF System class.
+        if not hasattr(topology, "_ref_mol_to_charge_method"):
+            topology._ref_mol_to_charge_method = {ref_mol: None for ref_mol in topology.reference_molecules}
```
SimonBoothroyd: I wonder if rather than tagging the topology directly with this attribute, it would be cleaner to carry around some additional, extensible, internal state object which gets passed between successive handlers and that can store intermediate data such as this?

docs/releasehistory.rst
jchodera: Let's check with @davidlmobley here too.
Since we already provide a mechanism for experimenting with charge models---specifying the charges on a whole molecule via an API---we don't want to allow someone to try to tile fragments in a way that leads to nonintegral charges on System creation and then try to go in and manually rescale things. They can already do whatever they want by specifying a precharged Moleculez where we should let them do whatever they like because that mechanism is intended for exploration.

 However, I know for certain one of us is going to make a mistake and accidentally mistype a SMARTS by accident and end up with nonintegral charges. We need this check to ensure we are not screwing this up. So it would seem prudent to add the check by default for molecules typed via LibraryCharge.

davidlmobley: I talked to Jeff about this in person and actually was inclined to agree with his viewpoint for all the reasons he listed, and also worried that additional developer time he'd spend trying to keep people from making mistakes would be time taken away from scientifically higher priority tasks like BCCs and WBOs. But I don't feel that strongly about it.

jchodera: Since these library charges are going to be written by hand, we would have to implement a mechanism outside of ForceField to check for this if we don't include it within ForceField. So either way, he's going to need to write this code. I'd prefer we use it to avoid being hoist by our own petard by baking it in here...

j-wags: I appreciate the thoughtful replies on this topic. This may just be an area where my perspective as a developer is different that the perspective of the users. 

I'll go ahead and implement a check for nonintegral charge assignment, which can optionally be disabled by a keyword argument (`allow_nonintegral_charges`). To future-proof this a bit, I'll have it apply to all charge methods (so, it will run in `ElectrostaticsHandler.postprocess_system`).

jchodera: That sounds like a reasonable approach!

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: This is admittedly a cheap hack. The object which this information *should* be stored on is the `System` (because then after parameterization, users will be able to look right at a particle and see what ParameterHandler the charge came from). 

I don't know how I feel about introducing an internal state object, so I'm thinking (typing) out loud here... It seems like this would partially be a "parametrization event" object, effectively like a rich log of what happened. This might be good for tracking provenance -- and actually we just talked about this at the MolSSI workshop (It would be neat if system prep tools produced an output which capture the important details of what they've done, which could be submitted as SI, relevant to @karmencj ). So I can envision it as purely an object to be written _to_. However, becomes more complex if it passes information _back_ to the FF, topology, or system during the parametrization process (which seems to be the proposal here). Unless there's some clear model of when it talks to the other objects, we could end up with some very hard-to-test behaviors. I'll keep thinking about how this could work, though.

SimonBoothroyd: > object which this information should be stored on is the System
This may be the case - although I'm not sure what the status of this object is?

> This might be good for tracking provenance
In principle it would - it's never a bad thing to keep track of what went into and came out of the different modular handlers.

> However, becomes more complex if it passes information back to the FF, topology, or system during the parametrization process [...] Unless there's some clear model of when it talks to the other objects

I don't think I understand this point - where would this complexity arise? The object wouldn't need to talk to anything, it would purely be an object which gets passed between handlers which would be a readable source of previously assigned metadata which could either be utilized if needed, or ignored if not, and then extended ready to be passed along the chain of handlers. In principle the object could be as simple as a dict of dicts, where each key is the handler name and the value the dict of provenance.

> we could end up with some very hard-to-test behaviors

I think that is inherent in having modular components which can be almost arbitrarily pieced together, but provided the required input and provided output provenance of each handler is well defined, then it shouldn't be overly bad!

Honestly I don't have particularly strong opinions about this, but given that in the future there is probably going to be cases where handlers may need access to information about what has already been done to the system (as is the case here), it might be worth thinking about how this will worker earlier rather than later!

openforcefield/tests/test_forcefield.py
j-wags: I agree these are dangerous, but I'm struggling to think of an alternative. 

Maybe the most direct way to remove `charge_from_molecules` is to create `LibraryChargeHandler.add_molecule()`, which takes `Molecule` objects as input and turns them into `LibraryCharge`s. The big missing piece here would just be the reverse of what we discussed elsewhere in this PR -- A way to go from `OFFMol` --> tagged SMILES. We could, by default, add the new whole-molecule SMARTS to the bottom of the hierarchy. 

Other kwargs, like `allow_nonintegral_charges` and `toolkit_registry`, will require further thought. But we could start phasing them out.

docs/releasehistory.rst
j-wags: Yes. I've added text to clarify that the `charge_from_molecules` kwarg counts as a recognized charge assignment method.

The-SMIRNOFF-force-field-format.md
j-wags: Wow. They actually didn't parse. Alanine was missing a `/` at the end of the LibraryCharge line, and TIP3P failed because `+0.417` doesn't parse with the `+`, so I've change it to just be `0.417`.

Fixed and added a test, with the `TODO`:
```
        # TODO: This test is practically useless while the XML strings are hard-coded at the top of this file.
        #       We should implement something like doctests for the XML snippets on the SMIRNOFF spec page.
```

docs/releasehistory.rst
j-wags: Good question. I've added text to address this directly: 

>   This functionality is thus able to apply per-residue charges similar to those in traditional
>   protein force fields. At this time, there is no concept of "residues" or "fragments" during
>   parametrization, so it is not possible to assign charges to `some` atoms in a molecule using
>   LibraryCharges, but calculate charges for other atoms in the same molecule using a different
>   method. To assign charges to a protein, LibraryCharges SMARTS must be provided for
>   the standard residues, as well as for any capping groups and post-translational modifications 
>   that are present.

docs/releasehistory.rst
j-wags: Ok, I've added a check to ensure that the partial charges on all particles in a Molecule sum up to its formal charge.  This can be overridden with the kwarg `allow_nonintegral_charge=True`.

openforcefield/typing/engines/smirnoff/parameters.py
jchodera: I'd be very careful about decorating the source `Topology` object with something that should change when the same `Topology` object is fed to a different `ForceField`. 

I do think we should seriously consider a new openforcefield `System` object as the right way to contain all of the applied parameterization data accessible via a public API.

PR #435, opened 2019-10-15
Update examples to use Parsley FF

j-wags:
This brought up two questions:
1) ~Do we want to include `notebook` as a dependency? It's already part of the single-file installer and the examples require it. This would need to be coordinated with updating the install instructions.~
2) ~Do we want to include `requests` as a dependency? The notebook test which downloads the Mobley free energy benchmark requires it.~

The above points are being handled in https://github.com/omnia-md/conda-recipes/pull/988


davidlmobley:
Perhaps if they're lightweight? If the examples require `notebook` it seems compelling that it should be included; I'd expect that if I install a package the examples will work.

j-wags:
Agreed. I've opened a PR to add them to the toolkit at omnia-md/conda-recipes#988

This PR is now ready for review.

PR #445, opened 2019-10-29
Eliminate navigation sidebar because it takes up too much room

jchodera:
Fixes https://github.com/openforcefield/openforcefield/issues/307

This PR eliminates the navigation sidebar because (1) it takes up too much space, causing readability issues, and (2) is unnecessary.

Tired:
image

Wired:
image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -139,6 +139,9 @@ html_logo = "_static/openforcefield.png"
 # Theme options are theme-specific and customize the look and feel of a
 # theme further.
 html_theme_options = {
+    # Disable the sidebar, since it takes up too much space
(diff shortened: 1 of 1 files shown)


jchodera:
Looks like we're not building RTD docs for `master`, so it will take until the next release to show up:
image

SimonBoothroyd:
Looks good - something to also consider is using [custom `autosummary` templates](https://github.com/openforcefield/propertyestimator/blob/master/docs/_templates/autosummary/class.rst) to [remove the module name from the nav bar / title](https://property-estimator.readthedocs.io/en/latest/api/generated/propertyestimator.properties.ExcessMolarVolume.html).

PR #449, opened 2019-11-01
[WIP] - added wbo dependent bond parameters to forcefield

sukanyasasmal:


j-wags:
Pinging @ChayaSt who has quite a bit of background on WBOs and torsions (in fact, she generated the data for the test case we just implemented). She will be able to answer questions about how to include partial bond orders in torsion parameters.

davidlmobley:
I think the idea is that we would make the torsional barrier scale (initially linearly) with WBO between some input values. This would only be the barrier height. So one might imagine a `k1_order1` and `k1_order2` (and similarly `kN_order1` and `kN_order2` parameters for any N) which will be user input, and then the WBO will be used to scale the barrier height linearly between those values.

That would allow the science we need, @sukanyasasmal and @j-wags .

j-wags:
I'm going to close this PR for the time being. @sukanyasasmal did great work fixing up and validating the current implementation of WBO computation, but has since moved on to a new position. @trevorgokey has begun working on bond parameter interpolation in a different branch off `master`, and so this one is no longer needed.

PR #459, opened 2019-11-20
[WIP] Developer docs

j-wags:
Drafting the content of developer docs. Keeping everything as one page for now, but will make a hierarchy later. 

May change to markdown if reStructured text adds a lot of friction. 

Status
- [ ] Ready for review
- [ ] Ready for merge


openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1572,16 +1572,17 @@ def assign_parameters(self, topology, system):
         """
         pass
 
-    def postprocess_system(self, topology, system, **kwargs):
+    def postprocess_system(self, system, topology, **kwargs):
```
trevorgokey: Seems like this belongs in another PR?

docs/developing.rst
```diff
@@ -3,14 +3,225 @@
...
+But when we do need to make tradeoffs, "assume the user is an expert" is our guiding principle.
+
+At the same time, we aim for "automagic" behavior whenever a decision will clearly go one way over another.
+System parametrization is an inherently complex topic, and the OFF toolkit would be nearly unusable if we required the user to explicitly approve every aspect of the process.
+For example, if a ``Topology`` has its ``box_vectors`` attribute defined, we assume that the resulting ``System`` should be periodic.
+
+
+
+Setting up a development environment
+""""""""""""""""""""""""""""""""""""
```
mattwthompson: Converting some notes from yesterday into partial instructions:

* Install conda
  * See https://www.anaconda.com/distribution/

* Set up conda environment
 
```
conda create --name off-dev -c conda-forge  -c omnia -c openeye openforcefield openeye-toolkits
conda remove --force openforcefield
git clone https://github.com/openforcefield/openforcefield
pip install -e .
```
* Obtain and store Open Eye license somewhere like `~/.oe_license.txt
  * Optionally store the path in environmental variable `OE_LICENSE`: `echo "export OE_LICENSE=/Users/yournamehere/.oe_license.txt" >> ~/.bashrc`

dotsdl: An alternative approach that doesn't require installation and removal of `openforcefield` itself, if desired:
```
conda create --name off-dev -c conda-forge -c omnia -c openeye openeye-toolkits
conda install --name off-dev -c conda-forge -c omnia -c openeye --only-deps openforcefield

git clone https://github.com/openforcefield/openforcefield
cd openforcefield
pip install -e .
```

j-wags: I think either one is fine. The separate `create` and `install` in @dotsdl's suggestion is more graceful, but I fear it could lead to dependency-resolving complications if our recipe gets more complex and we have to install more packages.

dotsdl:
I went ahead and commented out sections that aren't completed. I think the doc looks good as is, and will always be a living document. Merge it!

docs/developing.rst
dotsdl: Need to add:
```
conda activate off-dev
```
After conda installs and before use of `pip`.

mattwthompson:
I didn't merge this earlier because `openforcefield/typing/engines/smirnoff/parameters.py` showed up in the diff - switching the order of parameters in crucial function makes me nervous

openforcefield/typing/engines/smirnoff/parameters.py
dotsdl: Agreed. Is this intentional? Is it necessary? I think it should be removed from this PR regardless, but should it make its way in in another PR?

mattwthompson: I am also uncomfortable with this slipping in here, so I am overwriting these changes. If this was intentional it can be placed elsewhere but I suspect it was added accidentally last fall

j-wags: Oops, yes, that belongs elsewhere... I suspect that `LibraryChargeHandler` took care of this. Thanks for spotting this.

Issue #467, opened 2019-11-30
Sulfurs also have the same lone-pair stereochemistry issues nitrogens do with the OpenEye toolkit


jchodera:
**Describe the bug**
Previously, we had run into issues where the `openforcefield` toolkit using the OpenEye toolkit backend perceives trivalent nitrogens as stereocenters, causing issues for loading and parameterization of small molecules containing nitrogens. While we worked around this issue for nitrogens, it appears that sulfurs can also have the same problem.
![image](https://user-images.githubusercontent.com/3656088/69904338-87478700-1373-11ea-86e8-d392de72d185.png)
The molecule on the left has no trouble being loaded by the `openforcefield` toolkit, while the molecule on the right cannot be loaded without `allow_undefined_stereo=True`.
SDF files for both the left molecule (`CDK2-no-stereo-issue.sdf`) and right molecule (`CDK2-stereo-issue.sdf`) are provided.

**To Reproduce**
```python
molecule = Molecule.from_file('../CDK2-no-stereo-issue.sdf')
molecule = Molecule.from_file('../CDK2-stereo-issue.sdf')
```
[molecules.zip](https://github.com/openforcefield/openforcefield/files/3907418/molecules.zip)

**Output**
With the OpenEye toolkit backend:
```
---------------------------------------------------------------------------
UndefinedStereochemistryError             Traceback (most recent call last)
<ipython-input-14-354a3db15360> in <module>
----> 1 molecule = Molecule.from_file('../CDK2-stereo-issue.sdf')

~/miniconda/lib/python3.7/site-packages/openforcefield/topology/molecule.py in from_file(file_path, file_format, toolkit_registry, allow_undefined_stereo)
   2989                 file_path,
   2990                 file_format=file_format,
-> 2991                 allow_undefined_stereo=allow_undefined_stereo)
   2992         elif hasattr(file_path, 'read'):
   2993             file_obj = file_path

~/miniconda/lib/python3.7/site-packages/openforcefield/utils/toolkits.py in from_file(self, file_path, file_format, allow_undefined_stereo)
    410         from openeye import oechem
    411         ifs = oechem.oemolistream(file_path)
--> 412         return self._read_oemolistream_molecules(ifs, allow_undefined_stereo, file_path=file_path)
    413 
    414     def from_file_obj(self,

[...]

PR #469, opened 2019-11-30
Ensure atom names are unique after construction

### LINKED ISSUE #460: Create unique atom names for Molecule object on creation
Currently, when we create a `Molecule`, it's possible to construct one where the atoms lack unique names. This causes problems when we pass them to other tools that require unique atom names.

When constructing a `Molecule` object from another molecule that lacks unique atom names, we should create unique atom names. We could also check that provided atom names are unique when constructing atom-by-atom, but this is less critical.

Forcing the user to always specify a molecule as input with unique atom names seems overly burdensome.

Another alternative could be to have a `Molecule.normalize()` or `Molecule.validate()` option that checked for unique atom names and enforced them as being unique at that time, and we would only call this before using or exporting the molecule.

jchodera:
This PR adds a `Molecule._validate()` method that is called after all molecule construction events as a way to run any code required to standardize or otherwise validate the constructed `Molecule` object.

This PR also fixes #460 by ensuring that unique atom names are generated at this time if the specified atom names are not unique. Note that this may cause atom names to change if some were defined, but not all were unique.
This is done via addition of `Molecule.has_unique_atom_names` property and a `Molecule.generate_unique_atom_names()` public method.

There are some alternative strategies discussed in #460, but as https://github.com/openforcefield/openforcefield/issues/460#issuecomment-559136868 points out, if we're going to be standardizing on SDF files (which _lack_ atom names), we need to do something to ensure unique atom names are generated by the time we need to use them.

Additionally, two tests were added:
* A SMILES round-trip test, which discovered that it is impossible to do this for molecules that are perceived as lacking stereochemistry 
* A test for all new added methods

The failing test here also involves a phosphorous containing molecule (`[H]C([H])([H])P(=O)([H])[O-]`) suggesting #467 extends to phosphorous as well.


jchodera:
Correction: The minidrugbank phosphorous-containing molecule actually has a stereocenter because its substituents are different:

With the RDKit backend:
```python
>>> Molecule.from_smiles('[H]C([H])([H])P(=O)([H])[O-]')
UndefinedStereochemistryError             Traceback (most recent call last)
<ipython-input-14-9e8e428709fa> in <module>
----> 1 Molecule.from_smiles('[H]C([H])([H])P(=O)([H])[O-]')

~/miniconda/lib/python3.7/site-packages/openforcefield/topology/molecule.py in from_smiles(smiles, hydrogens_are_explicit, toolkit_registry, allow_undefined_stereo)

~/miniconda/lib/python3.7/site-packages/openforcefield/utils/toolkits.py in call(self, method_name, *args, **kwargs)

~/miniconda/lib/python3.7/site-packages/openforcefield/utils/toolkits.py in from_smiles(self, smiles, hydrogens_are_explicit, allow_undefined_stereo)

~/miniconda/lib/python3.7/site-packages/openforcefield/utils/toolkits.py in _detect_undefined_stereo(cls, rdmol, err_msg_prefix, raise_warning)

UndefinedStereochemistryError: Unable to make OFFMol from RDMol: Unable to make OFFMol from SMILES: RDMol has unspecified stereochemistry. Undefined chiral centers are:
 - Atom P (index 4)
```
image

Perhaps we shouldn't be using minidrugbank, or we should filter out the compounds that contain unspecified stereochemistry?

jchodera:
The remaining test failures appear to be due to the SMILES round-trip tests failing because we have some molecules in minidrugbank that have unspecified stereochemistry, yet we load them all in for testing. 

What approach should we take with these? Should we skip molecules with unspecified stereochemistry in our round-trip tests?

openforcefield/topology/molecule.py
```diff
@@ -1542,6 +1542,38 @@ def __init__(self,
...
+            element_counts[symbol] += 1
+            atom.name = symbol + str(element_counts[symbol])
+
+    def _validate(self):
+        """
+        Validate the molecule, ensuring it has unique atom names
+
+        """
+        if not self.has_unique_atom_names:
+            self.generate_unique_atom_names()
```
j-wags: I like the idea of having a general place to put molecule-validation tasks. I'm also in favor of providing functionality to add unique atom names. But _always_ generating unique atom names (in some cases overwriting ones that are already there), with no way to disable the behavior, may be going too far.

docs/releasehistory.rst
```diff
@@ -7,6 +7,30 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
 * ``minor`` increments add features but do not break API compatibility
 * ``micro`` increments represent bugfix releases or improvements in documentation
 
+0.6.1 - Bugfixes
+----------------
+
+New features
+""""""""""""
+- `PR #469 <https://github.com/openforcefield/openforcefield/pull/XXX>`_:
```
j-wags: While it's possible that this branch will come in on a different PR, I think we should assume this will come in, and just put the PR number right in the URL.

openforcefield/tests/test_forcefield.py
```diff
@@ -746,7 +746,7 @@ def test_parameterize_ethanol_missing_torsion(self):
         topology = Topology.from_openmm(pdbfile.topology, unique_molecules=molecules)
         with pytest.raises(UnassignedProperTorsionParameterException,
                            match='- Topology indices [(]5, 0, 1, 6[)]: '
-                                 'names and elements [(] H[)], [(] C[)], [(] C[)], [(] H[)],') \
+                                 'names and elements [(](H\d+)? H[)], [(](C\d+)? C[)], [(](C\d+)? C[)], [(](H\d+)? H[)],') \
```
j-wags: If people are going to rely on these atom numbers in larger workflows, it may be good to test for the exact value we expect instead of the `\d+` regex. So, we'd look to match `H5` `C0` etc here.

jchodera:
> I think that making the user call it manually is more in line with our philosophy

I'm worried that our philosophy is trending toward "annoy the hell out of the user by preventing them from ever doing anything easily with lots of small molecules". As it is right now, the user cannot use our tools to, say,
* Process a bunch of SMILES strings that might contain undefined stereochemistry, or even any trivalent sulfur atoms because of presumed stereochemistry atoms
* Process any SDF files because these lack atom names, even though we have standardized on SDF files
* Process a bunch of molecules from OpenEye or RDKit that may not have unique atom names

I agree that the philosophy of "raise an informative Exception subclass object for each error type that the user must correct" can be an effective way to make sure the user is informed of each and every criteria their molecules must meet to be properly processed, but I can guarantee that this is going to cause a great deal of frustration.

Is there some way we can provide both a "paranoid mode" (where we ensure that we are typing _exactly_ the molecule the user wants) and some sort of "just process these molecules mode" where molecules that don't pass our strict criteria are either transformed into a processable form or are dropped from the list? 

Even dropping may be problematic, because---as was pointed out previously---we are standardizing on SDF files, which do not _have_ atom names, which I believe makes it [impossible to transform the resulting system to AMBER/gromacs with ParmEd](https://github.com/openforcefield/openforcefield/issues/441).

If we were following the "principle of least surprise", I'd probably suggest the behavior where missing atom names are filled in silently with unique atom names that do not collide with any existing atom names, but an exception is raised if two user-specified atom names are identical. That would allow processing of molecules without issue in a manner that would not _change_ any atom names the user specified, but would raise an exception if the user-specified atom names caused a problem. Otherwise, they'd be filled in so things just work.

openforcefield/tests/test_forcefield.py
jchodera: I don't think we want users to rely on the atom names we generate following any sort of specified pattern. That would be very dangerous, since the order in which different cheminformatics tools might generate atoms from SMARTS strings is totally arbitrary.

openforcefield/topology/molecule.py
jchodera: Once we ditch ParmEd, it will no longer be necessary to generate unique atom names. But until we replace it, we're stick with the need to do so if we want our parameters to be deliverable to different molecular simulation packages.

That means that this needs to either happen to _all_ molecules we load into the toolkit, or it means we need to create more code to manage the transformation to other packages that does this silently behind the scenes. That may be an alternative that is workable here!

jchodera:
I noted this above, but it's important enough to state again here:

Once we ditch ParmEd, it will no longer be necessary to generate unique atom names. But until we replace it, we're stick with the need to do so if we want our parameters to be deliverable to different molecular simulation packages.

That means that this needs to either happen to all molecules we load into the toolkit, or it means we need to create more code to manage the transformation to other packages that does this silently behind the scenes. That may be an alternative to this approach that is workable!

j-wags:
I wrote a lot, but it boils down to this: 

> If we were following the "principle of least surprise", I'd probably suggest the behavior where missing atom names are filled in silently with unique atom names that do not collide with any existing atom names, but an exception is raised if two user-specified atom names are identical. That would allow processing of molecules without issue in a manner that would not change any atom names the user specified, but would raise an exception if the user-specified atom names caused a problem. Otherwise, they'd be filled in so things just work.

We do this, _at the time of system creation_. 

So, either we
* make a new function, `create_parmed_structure`, which runs your quoted algorithm after `create_openmm_system` to make the resulting system parmed-friendly; or
* we do this by default in `create_openmm_system`, since I don't think atom names have a functional effect in OpenMM, so no harm done.

j-wags:
To be a little less brief on my reasoning: I think we need to keep our _internal_ data structures clear of ParmEd's influence. The above discussion was driving us to choose a group of users to lose. Either:

* People who want to export to ParmEd inconveniently need to run `offmol.generate_atom_names`; or
* A person who uses OFFTK to generate conformers for alanine dipeptide puts in a molecule with two `CA`s and gets back conformers with `C2` and `C5` in their place

When really, we can just push this battle outside our borders -- Make the users identify where these parameters are going, and apply naming that works for ParmEd if that's the destination.

And maybe the second option doesn't seem too bad, but the slope toward ParmEd-ness becomes no less steep if we accept that behavior.  We just get into the "if you give a moose a muffin" situation -- If we rename the alanine atoms above, users will say "well, they're just in different residues, OFFTK should allow that". Then we'll be pushed to enforce some system of rules around atom names and residue names/numbers, in a toolkit made specifically not to care about atom types or topological residues. People will make workflows assuming the atom names we assign, and contort their inputs to satisfy the residue rules we enforce, and then we'll have this large set of expected behaviors causing friction on further developments (do virtual sites also need unique names? do we support chains?).

On a separate note:

I agree that the undefined stereochemistry thing is ugly and awkward. It may not be worth it on some fronts. I'd be open to arguments that we should just drop it for trivalent nitrogen and sulfur -- We could say stereochemistry that exchanges more frequently than once per millisecond is not stereochemistry at all for the purposes of atomistic simulation. This will be made easier by breaking out a `OFF-core` package, where we can keep all the rules that define our complicated stereochemistry definition in one consistent place.

jchodera:
@j-wags : I like your suggestions in https://github.com/openforcefield/openforcefield/pull/469#issuecomment-565669245! I'd be fine with this, but don't have the bandwidth to re-work this PR anytime soon. Should be an easy fix, though!

The only thing I'd caution about is that the wider the separation between doing the thing that triggers an error (converting a molecule with defective atom names) and the error (creating a system), the harder it is for our users _to identify where the error was actually committed_! So we have to help them along as much as possible by being maximally informative in our error messages.

> I agree that the undefined stereochemistry thing is ugly and awkward. It may not be worth it on some fronts. I'd be open to arguments that we should just drop it for trivalent nitrogen and sulfur -- We could say stereochemistry that exchanges more frequently than once per millisecond is not stereochemistry at all for the purposes of atomistic simulation. This will be made easier by breaking out a OFF-core package, where we can keep all the code that enforces our specific stereochemistry definition in one single place.

I still like the idea of an `openforcefield-core` package!

Another approach that would make our lives easier is to just say we're not going to care about stereochemistry in charge models, and that any model we ultimately adopt would have to somehow average over all possible stereochemical configurations. That would make the cheminformatics easy, but it may ultimately limit accuracy.

jchodera:
@j-wags: How should we proceed here? 

We're not going to be able to use SDF files and ParmEd to convert to other packages unless we implement something like this now, and I'd love to get this finished before the `master` branch diverges too much.

jchodera:
@j-wags: How should we proceed here?

j-wags:
> @j-wags: How should we proceed here?

I'll: 
- [ ] change `create_openmm_system` to assign atom names, using your algorithm in this PR
- [ ] disable validation for all molecules, but keep `Molecule.generate_unique_atom_names` and `Molecule.has_unique_atom_names`  and their tests. 
- [ ] cherry pick https://github.com/openforcefield/openforcefield/pull/486/commits/14c1d2f3b21b695f95acd96e85f5bb0e11ce71f9 to get the SMILES roundtrip tests working

j-wags:
Oh, OpenMM `System`s don't have atom names. Only OpenMM `Topology`s do. So a to-do item is
- [ ] add kwarg `ensure_unique_atom_names` to `Topology.to_openmm`, using your algorithm in this PR

openforcefield/tests/test_forcefield.py
j-wags: Sounds good.

docs/releasehistory.rst
j-wags: Fixed in dbc1451

PR #472, opened 2019-12-11
Create OFF Mols from QCArchive

### LINKED ISSUE #412: RDMol stereochemistry detector too strict
Trying to use:
```
openforcefield.topology.Molecule.from_rdkit( rdmol, allow_undefined_stereo=False)
```
where the molecule in question is `c1c[cH:1][c:2](cc1)[CH2:3][c:4]2ccccc2`:
image

fails due to the sp3 carbon in the center (index 3 in the smiles above). The function `RDKitToolkitWrapper::_find_undefined_stereo_atoms` in `toolkits.py` will fail anything that is `CHI_UNSPECIFIED` and possibly chiral:

```
   2304         # Find all atoms with undefined stereo.
   2305         undefined_atom_indices = []
   2306         for atom_idx, atom in enumerate(rdmol.GetAtoms()):
   2307             if (atom.GetChiralTag() == Chem.ChiralType.CHI_UNSPECIFIED and
   2308                     atom.HasProp('_ChiralityPossible')):
   2309                 undefined_atom_indices.append(atom_idx)
   2310         return undefined_atom_indices
```

However in the case here, the carbon is `CHI_UNSPECIFIED`, and `atom.HasProp('_ChiralityPossible') -> True`, causing the failure. 

I suggest being a little more relaxed, since chirality "is possible", but it is in fact not chiral. I am not super experienced here, but possibly checking for `_CIPCode` in the loop might be useful. In this case, for all atoms `atom.HasProp('_CIPCode') -> False` even after the prior call to `Chem.rdmolops.AssignStereochemistry` and any other checking function I found, so I assume RDKit is trying to tell us it is not a chiral center after all checks are made. `CHI_UNSPECIFIED` hints at the fact it's not chiral (i.e. a sufficient condition), but perhaps `_CIPCode` might be a necessary condition.

### LINKED ISSUE #448: Add functionality to create OFFMols with specific indexing
Many people have requested the ability to create OFFMols with specified atom orderings. Most commonly, this is in the context of [Canonical, isomeric, explicit hydrogen, mapped SMILES](https://github.com/openforcefield/cmiles#how-to-use-cmiles), which we attach to OFF-submitted molecules in QCArchive. However, people have also requested the ability to reorder the atoms from a molecule that already has an indexing system (like, an SDF).

This is two similar requests in one:
* `Molecule.from_object(obj, index_map=dict(current_idx:new_idx)`: Be able to read a molecule with a _defined_ indexing system (like, from SDF), but have the created OFFMol have a different atom/bond ordering.
* `Molecule.from_mapped_smiles(mapped_smiles)`: Read an explicit-hydrogen, fully mapped SMILES, and create an OFFMol with that atom ordering.
    * Needs to check+fail if H's aren't explicit. 

For both functions, we'll need to resolve what to do if the map indices don't begin at 1, of if the numbering system has gaps. For now, a reasonable behavior is probably just "fail".

### LINKED ISSUE #462: Add functionality to create OFFMol from QCAMol
Because of our project's close integration with QCArchive, we should be able to make Open Force Field Molecule objects from QCArchive molecules

Migrate the functionality from CMILES to produce OFFTK `Molecule` objects. 

We could keep the functionality in CMILES and add it as a dependency of OFFTK, but it would make more sense for it to be maintained as part of OFFTK (and to take advantage of the ToolkitWrapper system).

jthorton:
- [x] Fix #448 
- [x] Fix #462 
- [X] Fix #412 


openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3248,164 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+        new_molecule.partial_charges = new_charges * unit.elementary_charge
+
+        # remap the conformers there can be more than one
+        if self.conformers is not None:
+            for conformer in self.conformers:
+                new_conformer = np.zeros((self.n_atoms, 3))
+                for i in range(self.n_atoms):
+                    new_conformer[i] = conformer[n_to_o[i]].value_in_unit(unit.angstrom)
+                new_molecule.add_conformer(new_conformer * unit.angstrom)
+
```
j-wags: Maybe we should also copy over `self._properties` to the new molecule

j-wags: Let's check whether there are any virtualsites, and raise a `NotImplementedError` if so

openforcefield/topology/molecule.py
```diff
@@ -1890,55 +1894,119 @@ def from_smiles(smiles,
                 'Invalid toolkit_registry passed to from_smiles. Expected ToolkitRegistry or ToolkitWrapper. Got  {}'
                 .format(type(toolkit_registry)))
 
-    def is_isomorphic(
-            self, other,
-            compare_atom_stereochemistry=True,
-            compare_bond_stereochemistry=True,
+    @staticmethod
+    def are_isomorphic(
```
jchodera: Why go with a static method instead of `.is_isomorphic_with(other)`?

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3248,164 @@ def from_openeye(oemol, allow_undefined_stereo=False):
         return toolkit.from_openeye(
             oemol, allow_undefined_stereo=allow_undefined_stereo)
 
+    def to_qcshema(self):
```
jchodera: `to_qcshema` should be `to_qcschema`

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3248,164 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+        """
+
+        :return:
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def from_qcarchive(cls, qca_mol, smiles=None, allow_undefined_stereo=False):
```
jchodera: Should this be `from_qcschema` to match `to_qcschema`?

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3248,164 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+
+        Solution make a molecule from SMILES, then add the geometry as a conformer?
+        As the ordering is different we can use networkx to map the indices and rearange
+        this will also validate the info by checking that the general connection maps are the same
+
+        TODO do we get the same smiles string back?
+
+        Parameters
+        ----------
+        qca_mol : a QCArchive json representation
```
jchodera: If this is specifically the JSON representation string, we should name it something clearer, like `json` or `json_molecule`.
If we want to support alternative serializations (presuming that QCArchive may in the future), it may make more sense to accept a `dict` instead, since all serializations supported would presumably be reducible to a `dict`.

openforcefield/topology/molecule.py
j-wags: I think the staticmethod is best here, since it will give us a single function to handle `FrozenMolecule`s, `TopologyMolecule`s, and `nx.Graph`s. So, we'll still have `FrozenMolecule.is_isomorphic_with(other)`, but it can just immediately call this single, flexible method, which will sort out what `other` is and how to efficiently check it.

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3306,222 @@ def from_openeye(oemol, allow_undefined_stereo=False):
         return toolkit.from_openeye(
             oemol, allow_undefined_stereo=allow_undefined_stereo)
 
+    def to_qcshema(self, multiplicity=1, conformer=0):
```
jthorton: @j-wags

j-wags: We might consider changing this function to be part of a new QCElemental ToolkitWrapper in a later PR. That way, we could do the full molecule validation _in this function_, and return a QCElemental mol afterwards (instead of returning an un-validated dict, as we do now)

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3306,222 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+    @classmethod
+    def from_mapped_smiles(cls, mapped_smiles):
+        """
+        Create an openforcefield.topology.molecule.Molecule from a mapped SMILES made with cmiles.
+        The molecule will be in the order of the indexing in the mapped smiles string.
+        :param mapped_smiles: a cmiles style mapped smiles string with explicit hydrogens.
+        :return: openforcefield.topology.molecule.Molecule
+        """
+
+        # extract the atomic mapping
```
jthorton: look up atom mapping in the toolkits instead of parsing the strings this should be safer.

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3306,222 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+        mapping = []
+        for atom in mapped_smiles.split(':'):
+            try:
+                mapping.append(int(atom[:2]))
+            except ValueError:
+                try:
+                    mapping.append(int(atom[:1]))
+                except ValueError:
+                    pass
```
j-wags: OpenEye and RDKit should be able to read mapped SMILES and annotate the resulting Molecule objects with "map indices". I think it'd be better to let commercial-grade toolkit handle the mapping, rather than parsing the string ourselves.

openforcefield/topology/molecule.py
ChayaSt: This seems like a brittle way to extract map indices to check if all atoms have them. The ':' symbol is allowed to be used to represent aromatic bonds according to the daylight documentation (see the bonds section):
https://www.daylight.com/dayhtml/doc/theory/theory.smiles.html

The way I handle this in `cmiles` is to create an oe or rdkit mol, then check if none of the indices are zero.

openforcefield/topology/molecule.py
jthorton: Good point added this in as well.

jthorton: Added this for now, but what would need to be done to allow us to remap the sites?

openforcefield/topology/molecule.py
jthorton: Added the `.is_isomorphic_with(other)` function which now calls the static function, not sure what level of checking we should be doing here though

openforcefield/topology/molecule.py
jthorton: I think that would be a safer way to do it, shall we look at making a wrapper for this part now? or should I just try and import QCElemental and throw an error if not?

openforcefield/topology/molecule.py
jthorton: Good point I have now followed @ChayaSt method in cmiles to extract the atom maps if they are present.

openforcefield/topology/molecule.py
jthorton: thanks I have followed your method in cmiles to get the atom maps from the toolkits.

openforcefield/topology/molecule.py
j-wags: We could remap atoms in the existing implementation of VirtualSites, but the existing implementation isn't functional in molecule parametrization, and will likely undergo serious refactoring. Currently the VirtualSite implementation is largely a placeholder.

If you'd like to implement support for the placeholder implementation of VirtualSites, which probably isn't worth the time, the biggest thing will be to make a copy of each original VirtualSite in the new molecule. Importantly, the new VirtualSite will need to have remapped `vsite.atoms` in the new molecule. Note that `vsite.atoms` isn't a list of atom _indices_, rather it is a _list of actual atoms_ (I did say it will undergo refactoring).

openforcefield/topology/molecule.py
jthorton: For now, this is the `dict` representation with JSON encoding but we could also accept the actual record object as well and type check them to try and get the information out of it. Not sure what users would expect at this point though.

openforcefield/topology/molecule.py
j-wags: I'm redefining some conventions in `ToolkitWrapper` right now, so let's not make a `QCElementalToolkitWrapper` just yet. For now let's keep the code in `molecule.py`. Later, we can make the QCEl ToolkitWrapper, migrate this code to there, and change this function to just do `toolkit_registry.call('to_qcschema', mol, ...)`

openforcefield/topology/molecule.py
jthorton: Not sure about the naming here `from_qcshema` would suggest that the molecule is being made from the schema info like geometry, connectivity and elements, but this function only works at the entry-level of the archive which has the cmiles identifiers where its safe to remake the molecule. Maybe something like `from_qcentry`  ? Then we could also have a `from_qcschema` method which could try and construct the molecule using the schema info which is in every record with a warning?

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +529,109 @@ def test_name(self):
...
+        topmol = TopologyMolecule(molecule_smiles, topology)
+        assert molecule_smiles.hill_formula == Molecule.to_hill_formula(topmol)
+        # make sure the networkx matches
+        assert molecule_smiles.hill_formula == Molecule.to_hill_formula(molecule_smiles.to_networkx())
+
+    def test_isomorphic(self):
+        """Test the different levels of isomorphic matching"""
+        # check that hill formula fails are caught
+        ethanol = Molecule.from_smiles('CCO')
+        acetaldehyde = Molecule.from_smiles('CC=O')
```
j-wags: The behavior of `from_smiles` is completely delegated to the cheminformatics toolkits. It just so happens that the current implementation of their `from_smiles` functions preserves the order of the heavy atoms from the SMILES in this case, but that may not always be true. If the behavior changes in the future (or if another toolkit is added that doesn't behave this way), these tests will fail, and it'll take us a while to figure out why.

In these sorts of cases, I recommend making a test function to create a molecule using OFFTK's API. This may be paranoid, but it could save us a big headache in the future. Here's an example: https://github.com/openforcefield/openforcefield/blob/b8b7b314602a2685d5fe9441d01bb0a7f7c8badc/openforcefield/tests/test_forcefield.py#L259-L315

For now, I'd put the `create_acetaldehyde` function in test_forcefields, and just import it. It's sloppy, but that will keep them all in one place for when we consolidate them later.

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +529,109 @@ def test_name(self):
...
+        assert ethanol.is_isomorphic_with(acetaldehyde) is False
+        assert acetaldehyde.is_isomorphic_with(ethanol) is False
+        # check that different orderings work with full matching
+        ethanol_reverse = Molecule.from_smiles('OCC')
```
j-wags: Same as my comment above -- I will be merging `create_reversed_ethanol` from the branch I linked in my comment above soon. You might consider just copying the function from my branch into yours, and then whoever closes their PR last can resolve the conflict from the duplication. (I'm sure there's a "proper git" way to do this, but given my history will this sort of thing, I'd just break the repo again)

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +529,109 @@ def test_name(self):
...
+        assert ethanol.is_isomorphic_with(ethanol_reverse) is True
+        # check a reference mapping between ethanol and ethanol_reverse matches that calculated
+        ref_mapping = {0: 2, 1: 1, 2: 0, 3: 6, 4: 7, 5: 8, 6: 4, 7: 5, 8: 3}
+        assert Molecule.are_isomorphic(ethanol, ethanol_reverse, return_atom_map=True)[1] == ref_mapping
+        # check matching with nx.Graph atomic numbers and connectivity only
+        assert Molecule.are_isomorphic(ethanol, ethanol_reverse.to_networkx(), aromatic_matching=False,
+                                       formal_charge_matching=False, bond_order_matching=False,
+                                       atom_stereochemistry_matching=False,
+                                       bond_stereochemistry_matching=False)[0] is True
```
j-wags: This is a really complex function, but we should still test every permutation of these keyword arguments, for comparisons that evaluate either to `True` or `False`. Here's an example of how the toolkit tests to ensure that different permutations of nonbonded options produce the correct system: 

First it defines every permutation of inputs, and the expected output (or error message): https://github.com/openforcefield/openforcefield/blob/master/openforcefield/tests/test_forcefield.py#L318-L344

Then, it uses `pytest.parametrize` to run an array of individual tests: https://github.com/openforcefield/openforcefield/blob/master/openforcefield/tests/test_forcefield.py#L915-L930

These are all seen as separate tests by pytest, so each permutation of options either passes or fails, making debugging the separate possible causes much simpler. If the number of permutations becomes overwhelming, you can use itertools to define them all, and just define the expected behavior for the ethanol-vs-acetaldehyde comparison be `bond_order == True`

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +529,109 @@ def test_name(self):
...
+        # make sure files and smiles match
+        molecule_file = Molecule.from_file(get_data_file_path('molecules/ethanol.mol2'))
+        assert molecule_smiles.hill_formula == molecule_file.hill_formula
+        # make sure the topology molecule gives the same formula
+        from openforcefield.topology.topology import TopologyMolecule, Topology
+        topology = Topology.from_molecules(molecule_smiles)
+        topmol = TopologyMolecule(molecule_smiles, topology)
+        assert molecule_smiles.hill_formula == Molecule.to_hill_formula(topmol)
+        # make sure the networkx matches
+        assert molecule_smiles.hill_formula == Molecule.to_hill_formula(molecule_smiles.to_networkx())
```
j-wags: I really like this test. It's very well thought-out.

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +529,109 @@ def test_name(self):
...
+        mol2 = Molecule.from_smiles('Fc1ccccc1F')
+        assert mol1.is_isomorphic_with(mol2) is False
+        assert mol2.is_isomorphic_with(mol1) is False
+
+    def test_remap(self):
+        """Test the remap function which should return a new molecule in the requested ordering"""
+        # the order here is CCO
+        ethanol = Molecule.from_file(get_data_file_path('molecules/ethanol.mol2'))
+        # get ethanol in reverse order OCC
+        ethanol_reverse = Molecule.from_smiles('OCC')
```
j-wags: Same comment about expecting deterministic ordering from `from_smiles` here

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +529,109 @@ def test_name(self):
...
+        # test hill formula passes but isomorphic fails
+        mol1 = Molecule.from_smiles('Fc1ccc(F)cc1')
+        mol2 = Molecule.from_smiles('Fc1ccccc1F')
+        assert mol1.is_isomorphic_with(mol2) is False
+        assert mol2.is_isomorphic_with(mol1) is False
+
+    def test_remap(self):
+        """Test the remap function which should return a new molecule in the requested ordering"""
+        # the order here is CCO
+        ethanol = Molecule.from_file(get_data_file_path('molecules/ethanol.mol2'))
```
j-wags: Only OpenEye can read mol2, and not all test environments have it. Could you use `ethanol.sdf` and `ethanol_reordered.sdf` instead? Both OETK and RDKit can read that.

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +529,109 @@ def test_name(self):
...
+        # get ethanol in reverse order OCC
+        ethanol_reverse = Molecule.from_smiles('OCC')
+        # get the mapping between the molecules
+        mapping = Molecule.are_isomorphic(ethanol, ethanol_reverse, True)[1]
+        ethanol.add_bond_charge_virtual_site([0, 1], 0.3 * unit.angstrom)
+        # make sure that molecules with virtual sites raises an error
+        with pytest.raises(Exception):
```
j-wags: Could you be more specific, and expect `NotImplementedError`?

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +529,109 @@ def test_name(self):
...
+        assert_molecules_match_after_remap(new_ethanol, ethanol_reverse)
+
+        # catch mappings that are the wrong size
+        too_small_mapping = {0: 1}
+        with pytest.raises(Exception):
+            new_ethanol = ethanol.remap(too_small_mapping, current_to_new=True)
+
+        wrong_index_mapping = {(i + 10, new_id) for i, new_id in enumerate(mapping.values())}
+        with pytest.raises(Exception):
+            new_ethanol = ethanol.remap(wrong_index_mapping, current_to_new=True)
```
j-wags: It's awesome that you're checking this. Maybe break these checks out into a separate test?

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +529,109 @@ def test_name(self):
...
+
+        # test round trip (double remapping a molecule)
+        new_ethanol = ethanol.remap(mapping, current_to_new=True)
+        round_trip_mapping = Molecule.are_isomorphic(new_ethanol, ethanol, return_atom_map=True)[1]
```
j-wags: Maybe catch the first return value and ensure that it's `True` here (that is, ensure that the reordered molecule really _is_ isomorphic)

openforcefield/tests/test_molecule.py
jthorton: Thanks for pointing this out I have added in the functions to the test_ff file just need to remember to remove `create_reversed_ethanol()` function.

openforcefield/tests/test_molecule.py
jthorton: Good point changed all expected exceptions to catch the right errors now.

openforcefield/tests/test_molecule.py
jthorton: Thanks, I have separated them out.

openforcefield/tests/test_molecule.py
jthorton: Good idea otherwise the None type error might be hard to debug!

openforcefield/tests/test_molecule.py
jthorton: Good spot Ill change them over.

openforcefield/utils/toolkits.py
```diff
@@ -1569,14 +1572,16 @@ def from_file(self,
...
             # TODO: See if we can implement PDB+mol/smi combinations to get complete bond information.
+            #  testing to see if we can make a molecule from smiles and then use the PDB conformer as the geometry
+            #  and just reorder the molecule
             # https://github.com/openforcefield/openforcefield/issues/121
-            # rdmol = Chem.MolFromPDBFile(file_path, removeHs=False)
-            # mol = Molecule.from_rdkit(rdmol)
-            # mols.append(mol)
+            rdmol = Chem.MolFromPDBFile(file_path, removeHs=False)
+            mol = Molecule.from_rdkit(rdmol)
+            mols.append(mol)
```
j-wags: Let's make sure to roll this change back -- Users should run a new function`molecule.from_pdb_and_smiles` (which requires a SMILES) to load from PDB. Inside that function, it could run code similar to here [1], specifically using `AssignBondOrdersFromTemplate`. Once the RDMol is prepared using that, we could just run `Molecule.from_rdkit` to make the OFFMol.

As you said, basically the only thing the PDB is used for in this process will be to get a conformer of the molecule (but guaranteeing the correct indexing for that geometry)

[1] https://practicalcheminformatics.blogspot.com/2018/09/assigning-bond-orders-to-pdb-ligands.html

jthorton: I am a little worried that this function might not transfer the stereochemistry which is assigned when making the molecule from SMILES. It might be better to approach it the other way around make a molecule from SMILES and reindex it to match the PDB order then just transfer the conformation?

openforcefield/topology/molecule.py
```diff
@@ -1788,7 +1786,9 @@ def __eq__(self, other):
            No effort is made to ensure that the atoms are in the same order or that any annotated properties are preserved.
 
         """
-        return self.is_isomorphic(other)
+        # updated to use the new isomorphic checking method, with full matching
+        # TODO the doc string did not match the previous function what matching should this method do?
+        return Molecule.are_isomorphic(self, other, return_atom_map=False)[0]
```
jthorton: @j-wags I have changed this method to use the new full isomorphic matching method, but is this what we want? The docstring suggests we are checking the smiles and I think that's whats the old `is_isomorphic` function used to do, do we want to go back to that or keep this more strict matching checking?

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3341,319 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+                except InvalidConformerError:
+                    print('Invalid conformer for this molecule, the geometry could not be attached.')
+            # attach a dict that has the initial molecule ids and the number of the conformer it is stored in
+            offmol._properties['initial_molecules'] = initial_ids
+
+        return offmol
+
+    @classmethod
+    @RDKitToolkitWrapper.requires_toolkit()
+    def from_pdb_and_smiles(cls, file_path, smiles, allow_undefined_stereo=False):
```
jthorton: @j-wags this is the implementation of the from PDB and smiles classmethod, I think this might be the better option as this way make sure we have the bond orders and stereochemistry assigned when making a molecule from smiles, then we just reorder the atoms and transfer the coordinates from the PDB, we could be even safer and instead reorder the coordinates in the PDB and attach them without interfering with the molecule from smiles.
They only way I could see to do this was to move the code out of a toolkit wrapper however, is this okay?

openforcefield/topology/molecule.py
j-wags: Ah, I like what you've done here.

_Just in case_ the default kwargs for `are_isomorphic` change in the future, and to make this function more readable, could you explicitly set the kwargs to `are_isomorphic` here (so, like, everything is `True`, except checking for identical atom ordering)?

openforcefield/utils/toolkits.py
j-wags: Ahhh, so RDKit's `AssignBondOrdersFromTemplate` doesn't pull in stereochemistry. Darn.

It looks like there's a PR to add stereo to that function, but it's been parked for 3 years. https://github.com/rdkit/rdkit/pull/886

Ok, so thinking of alternate solutions...

### Option 1
Most directly, we could just use `AssignBondOrdersFromTemplate`, and then run `Chem.AssignStereoFrom3D` on the resulting rdmol, similar to [1]. In this case, I think we'd ignore stereo from the SMILES and just trust whatever's in the geometry. It's... not perfect, but it'd work.

### Option 2 
> reindex it to match the PDB order then just transfer the conformation?

I see, but this would require an initial source of connectivity, no? Or we could do like using `simtk.openmm.app.PDBFile` to load the PDB, then convert it to a networkX graph, and then do the matching from there. That could be nice, since then we'd guarantee the user the stereochemistry from the SMILES.

I'll let you pick which one. 2) seems a bit more difficult, but marginally better in the long run.

[1] https://github.com/openforcefield/openforcefield/blob/master/openforcefield/utils/toolkits.py#L1551-L1553

openforcefield/topology/molecule.py
j-wags: Accepting both would be ideal, imho.

openforcefield/topology/molecule.py
```diff
@@ -1888,64 +1887,162 @@ def from_smiles(smiles,
...
+        # else:
         #    other_fm = other
-        #self_smiles = self.to_smiles(toolkit_registry=toolkit_registry)
-        #other_smiles = other_fm.to_smiles(toolkit_registry=toolkit_registry)
-        #return self_smiles == other_smiles
+        # self_smiles = self.to_smiles(toolkit_registry=toolkit_registry)
+        # other_smiles = other_fm.to_smiles(toolkit_registry=toolkit_registry)
+        # return self_smiles == other_smiles
+
+    def is_isomorphic_with(self, other):
```
j-wags: I'd recommend adding **kwargs here, to pass on to the underlying staticmethod.

openforcefield/topology/molecule.py
j-wags: > not sure what level of checking we should be doing here though

We can add `**kwargs` to the `is_isomorphic_with` function signature, and just pass them all on to `are_isomorphic`, though we will want to update the docstring for that funciton to reflect the change. I've added another comment about this at the appropriate line.

openforcefield/topology/molecule.py
j-wags: And also copying the docstring for all the additional keyword arguments that can be accepted by `are_isomorphic` into this function's docstring as well.

jthorton: Good idea I have added this in now.

openforcefield/utils/toolkits.py
jthorton: So long as the PDB has the connect terms at the bottom I can make a molecule from this and turn it into a networkx graph then pull in the conformation from the molecule and the mapping by doing an isomorphic check see here https://github.com/openforcefield/openforcefield/pull/472#discussion_r365340467

openforcefield/topology/molecule.py
jthorton: okay, for now, I think it is best to just leave out the sites until the implementation is final.

jthorton:
At the QCFractal call we talked about deduplication of data in the QCArchive, and for this to happen entries must be exactly the same this includes the ordering of the atoms and the conformer coordinates down to a nano Bohr. The current `to_qcschema` method just makes a schema representation of the current molecule ordering, so the plan is to expose another function like `mol.to_canonical_order(h_last=True)`  which will act like that implemented in cmiles (https://github.com/openforcefield/cmiles/blob/b41a8d8f188a988918cd2ee34c4dbdc44a18b0f5/cmiles/_cmiles_rd.py#L64) and use the toolkits to reorder the molecule.

openforcefield/topology/molecule.py
```diff
@@ -1888,64 +1887,185 @@ def from_smiles(smiles,
...
+            mol1, mol2, return_atom_map=False,
+            aromatic_matching=True,
+            formal_charge_matching=True,
+            bond_order_matching=True,
+            atom_stereochemistry_matching=True,
+            bond_stereochemistry_matching=True,
     ):
         """
-        Determines whether the molecules are isomorphic by comparing their graphs.
+        Determines whether the two molecules are isomorphic by comparing their graphs.
```
trevorgokey: We should clarify to users what exactly is checked. For example, does this look at partial charges? It would be helpful to know all of the datapoints used to determine isomorphism.

openforcefield/topology/molecule.py
```diff
@@ -1888,64 +1887,185 @@ def from_smiles(smiles,
...
+
+                return is_equal
+        else:
+            edge_match_func = None
+
+        # Here we should work out what data type we have, also deal with lists?
+        def find_data_type(data):
+            """Find the data type and return the networkx graph"""
+
+            try:
```
trevorgokey: This should probably explicitly check for types rather than catch potentially generic `AttributeError`

```suggestion
            if isinstance(data, TopologyMolecule):
```

openforcefield/topology/molecule.py
```diff
@@ -1888,64 +1887,185 @@ def from_smiles(smiles,
...
+                return data.to_networkx()
+            except AttributeError:
+                try:
+                    # TopologyMolecule class instance
+                    return data.reference_molecule.to_networkx()
+                except AttributeError:
+                    if isinstance(data, nx.Graph):
+                        return data
+
+        mol1_netx = find_data_type(mol1)
```
trevorgokey: It would be helpful if the name of this function is more descriptive, e.g. `to_networkx`.

openforcefield/topology/molecule.py
```diff
@@ -1888,64 +1887,185 @@ def from_smiles(smiles,
...
+            bond_stereochemistry_matching=True,
     ):
         """
-        Determines whether the molecules are isomorphic by comparing their graphs.
+        Determines whether the two molecules are isomorphic by comparing their graphs.
 
         Parameters
         ----------
-        other : an openforcefield.topology.molecule.FrozenMolecule
+        mol1 : an openforcefield.topology.molecule.FrozenMolecule or TopologyMolecule or nx.Graph()
```
trevorgokey: If nx.Graphs are given, it would be helpful to know here what custom attributes must be present in the obj for the function to work correctly.

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3364,319 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+        connectivity = [(bond.atom1_index, bond.atom2_index, bond.bond_order) for bond in self.bonds]
+        symbols = [Element.getByAtomicNumber(atom.atomic_number).symbol for atom in self.atoms]
+
+        schema_dict = {'symbols': symbols, 'geometry': geometry, 'connectivity': connectivity,
+                       'molecular_charge': charge, 'molecular_multiplicity': multiplicity}
+
+        return qcel.models.Molecule.from_data(schema_dict, validate=True)
+
+    @classmethod
+    def from_mapped_smiles(cls, mapped_smiles, allow_undefined_stereo=False):
```
trevorgokey: Does this function (intend to) handle cases where partial maps exist, e.g. torsiondrives where only the 4 driven atoms have an index?

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3364,319 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+
+            # the pdb conformer is in the correct order so just attach it here
+            new_mol.add_conformer(pdbmol.conformers[0])
+
+            return new_mol
+
+        else:
+            raise InvalidConformerError('The PDB and SMILES structures do not match.')
+
+    def remap(self, mapping_dict, current_to_new=True):
```
trevorgokey: This function seems like a re-implementation of `_copy_initializer`, except that things are reordered here. I am concerned that it will be difficult to maintain these versions separately.

j-wags: This is a really good big-picture suggestion, and it's something that I'd totally passed over. I don't think we need to tackle it in this PR, but I'll create an Issue out of it.

openforcefield/topology/molecule.py
jthorton: No only fully mapped smiles would work here due to the reordering requiring a new index for each atom, I could change the name to further state this? `from_fully_mapped_smiles` ?

openforcefield/topology/molecule.py
jthorton: Good point! ill add it to the docstring.

openforcefield/topology/molecule.py
jthorton: Good idea I think `to_networkx` makes sense here.

openforcefield/topology/molecule.py
jthorton: This was my original idea but I think I was hitting cyclic import errors, where the `Molecule` class is imported for isomorphic checking in the topology file.

openforcefield/topology/molecule.py
jthorton: I think the new full docstring along with the kwargs should explain all node/edge attributes which can be checked. Let me know if this is clearer?

openforcefield/topology/molecule.py
jthorton: Good point attribute changes in future will mean that both functions need to be updated for now the testing of `remap` where the `dict()` of the atoms and bonds is compared after transferring attributes should catch where we introduce a new attribute and warn that the function is to be updated to support this.

jthorton: I have also changed this function to use the dict representation of atoms and bonds in line with `copy_initializer` to help with transferring all attributes.

openforcefield/topology/molecule.py
trevorgokey: I don't see why partial ordering wouldn't work, but at the same time as a user I could take my partial ordering, and then just do a sequential map on the other atoms. Not a big deal, but something this function could also probably do.

For now I would keep the original name as it is more intuitive to search `from_map` in the API, and then just throw an exception if anything is mapped to 0, indicating a partial map.

openforcefield/topology/molecule.py
trevorgokey: Yes, the function is good as it stands, but I was just concerned as this function seems like a core piece of functionality that is really needed. I am interpreting this function as follows. Isomorphism is defined as whether it has the same atoms (nodes) and connectivity (edges). This satisfies the definition of isomorphism. Then, it goes and establishes that several auxiliary properties satisfy an equality test, and the function returns `False` if any of these are not equal, `True` otherwise.

Because of the way I see it working, I was concerned with what should be considered. Sitting and thinking more, I understand the spirit of the function and think it's nice for a subset of users. However, the way I'd probably use it, is set them all to `False`, then cherry pick the attributes I want to check manually, so I can figure out what is different. If the two mols are ordered the same, and this function returns `True` with all options as `False`, then I can just go down the list with an equality operator and check these individually.

openforcefield/topology/molecule.py
jthorton: Yeah, I think we could support that in future and probably even offer a way of generating the partial maps using something like the isomorphic checking function.

openforcefield/topology/molecule.py
jthorton: Yes, you are right the minimum requirement for the isomorphic check is the atomic numbers on the nodes and the connectivity (edges with no attributes) all the others are optional extras which can also be compared if users want to. I think your right that the majority of users will only use the minimum checking with all extras set to `False`, it might be worth at this point changing the default behaviour to `False` for easier default use, @j-wags would you agree?

openforcefield/topology/molecule.py
j-wags: Ah, when we hit cyclic import errors, the trick is to move the errors from the module level into the methods themselves. We should probably do this throughout the codebase, but for now I do it to fix import problems when they come up.

openforcefield/topology/molecule.py
j-wags: I'm mostly with Josh here -- Arbitrarily ordering non-indexed atoms would be a dangerous behavior, where we could quickly implement it in some arbitrary way, but then some user would make workflows reliant on that arbitrary algorithm, and we'd either get angry emails when we change the ordering later, or we'd have to maintain this arbitrary behavior forever.

openforcefield/topology/molecule.py
j-wags: Did you add this? Just double checking that I didn't miss it.

jthorton:
> At the QCFractal call we talked about deduplication of data in the QCArchive, and for this to happen entries must be exactly the same this includes the ordering of the atoms and the conformer coordinates down to a nano Bohr. The current `to_qcschema` method just makes a schema representation of the current molecule ordering, so the plan is to expose another function like `mol.to_canonical_order(h_last=True)` which will act like that implemented in cmiles (https://github.com/openforcefield/cmiles/blob/b41a8d8f188a988918cd2ee34c4dbdc44a18b0f5/cmiles/_cmiles_rd.py#L64) and use the toolkits to reorder the molecule.

After looking into this I have found that RDKit and Openeye will order the atoms differently for submission to QCArchive, by looking at provenance on the first entry of every dataset I can see they have all been prepared with openeye so far. Thus to ensure deduplication of records by QCArchive we should continue to use this method. However, future users could also use RDKit or no order to submit to the public/local archive (where they will see some duplication) we could fix this by post-processing submissions to the public archive with one toolkit into consistent ordering. For local instance users, we could expose both canonicalisation methods of the toolkits in the hope that people are consistent with there use of toolkit and accept a small amount of duplication if not.

Alternatively, we could introduce a totally unique ordering system which is built into the OFFTK and will be used automatically when submitting jobs to the archive using the toolkit. This will, however, mean that the future submissions are ordered differently to those already submitted and there may be a small number of duplicate records that slip through. We should also be careful with the design of this ordering as changing it in future versions will mean that we fall back into this trap again. 

Does anyone have a preference for one method? 

I think exposing both canonicalisation methods of the toolkits aligns with the current aims of the OFFTK.

openforcefield/topology/molecule.py
jthorton: Yes sorry I forgot to push it should show now.

jthorton:
The tests are failing due to the notebook example which I have added showing how to create OFFMols from the QCArchive, some of these are expected as I have some API examples which are not implemented at the bottom which I would like some advice on, manly concerning how we add final molecules from optimisation records. 

The notebook also highlighted an unexpected failure whereby an OFFMol could not be made from an RDKit molecule which was created using the `canonical_isomeric_explicit_hydrogen_mapped_smiles` cmiles tags due to unspecified stereochemistry, however, making the molecule with the openeye backend does not raise such an error.

openforcefield/topology/molecule.py
j-wags: > I think your right that the majority of users will only use the minimum checking with all extras set to False,

@jthorton I don't think this is true. In the context of OFFTK, when most users want to know if two molecules are "identical", they mean "one molecule would have the same parameters assigned as another". My understanding of how this sort of equality could be checked is by 
1) comparing canonical SMILES, or 
2) having all the keyword arguments to `is_isomorphic` set to `True` except `atom_index` (and maybe `bond_order`, since the "same" molecule could have two valid kekule resonance structures, so we'd just check `is_aromatic`). 

> However, the way I'd probably use it, is set them all to False, then cherry pick the attributes I want to check manually, so I can figure out what is different

@trevorgokey This seems like an unusual use case -- When do we have a situation when a person has two molecules that differ only by (for example) one atom's element, and are looking for a function that tells them that? I'm imagining a case where a user compares two nearly-identical molecules -- one with a tetravalent N+ and the other with a tetravalent C in the same place -- and wants the OFFTK to tell them that these differ by element and formal charge. Maybe I'm not thinking of a more common use case, but that seems pretty far removed from OFFTK's core mission (parameterize molecules). People looking to do that sort of cheminformatics analysis should probably use RDKit. 

So, to summarize: 
* I think that uncolored graph comparison of molecules is an unusual thing to do outside of fairly involved cheminformatics
* OFFTK users are not looking to do fairly-involved cheminformatics (they can instead use RDKit)
* a good number of users will be unpleasantly surprised if our API used a different meaning of "isomorphism" than they had in mind, and people would shoot themselves in the foot by making the wrong assumption about the API.

> it might be worth at this point changing the default behaviour to False for easier default use, @j-wags would you agree?

@trevorgokey and I talked about this for some time yesterday, I do agree that "isomorphism" most directly implies "two graphs containing uncolored nodes connected by uncolored edges are superimposable" so we are technically using the word incorrectly. What we are really performing in the existing `Molecule.is_isomorphic' function is a _colored_ isomorphism check (where colors are element, bond order, etc). So, we are currently _technically_ using the word wrong, but to use it correctly would cause our users to shoot themselves in the foot

This is going to be an API breaking change either way (`is_isomorphic` is being deleted, and its replacement `is_isomorphic_with` uses different kwargs) , so we have some flexibility to start fresh here.

So this is a tricky call. I propose the following:

* We rename the current `are`/`is_isomorphic_with` to something more descriptive like `Molecule.map_molecule_graphs` or `compare_molecule_graphs`, which communicates that it will be more general
* Given that OFFTK is a parameterization engine foremost, we ensure that the `FrozenMolecule.__eq__` operator (used for `offmol1 == offmol2`) means "offmol1 would be assigned the same parameters as offmol2". In my mind this means "same canonicalized SMILES', but it may also correspond to a combination of graph comparison kwargs (though resonance structures/tautomers are tricky wrt. atom formal charge and bond orders) 
* We provide any number of "convenience methods" for molecule comparison, which basically are just pre-wrapped versions of `compare_molecular_graphs` with different keywords set.

@dmobley @jchodera This seems like a part of the API that's at high risk for foot-shooting, so your thoughts would be appreciated.

openforcefield/topology/molecule.py
j-wags: This is a nifty implementation. It's great that it conforms to the PDB ordering in the end.

> They only way I could see to do this was to move the code out of a toolkit wrapper however, is this okay?

It could be OK if there's no alternative, but could you expand on why this couldn't be in RDKitToolkitWrapper. Is it an import error?

openforcefield/topology/molecule.py
j-wags: Ok, I've (finally) split this off into #496. Again, this is a good point, but I think this PR is creeping larger and larger, so I'd like to consolidate our gains before it gets too wide.

openforcefield/topology/molecule.py
jthorton: Thanks and I think that is the most important part here as people will expect the ordering to be the same.

I also think this can be moved to the RDKitToolkitWrapper and have tried this in the latest commit, I think my original plan was letting people read PDB files with the RDKit `from_file` function then trying to catch this and create the molecule but I was worried some molecules directly from the PDB may slip through.

This also opens up the possibility of extending the `add_conformer` API to accept files with the molecules in any order as we can now graph match the structures provided connectivity and reorder the coordinates in the file and attach them to the molecule. This may be best placed in another PR though?

openforcefield/topology/molecule.py
jthorton: Good point that was probably an oversight part and exact matches will be useful and important when it comes to deduplication of fragments in the bespoke workflow for example. 

I understand where you guys are coming from with the naming but this is still a type of isomorphic checking and in my mind, in this area, I think it would be rare for someone to only want to compare uncoloured graphs unless they were setting up some relative free energy calculation and needed a mapping between two topologies but this is beyond the scope of the toolkit. 

Following on from that  the return result (`True` or `False`) is expected with the naming `Molecule.is_isomorphic_with(other)` and `Molecule.are_isomorphic(mol1, mol2)` however, I don't feel it is clear with `Molecule.compare_molecule_graphs()`. Something like `Molecule.atoms_and_bonds_match(other)` is more descriptive and has an expected return type, but I would prefer the current naming.

openforcefield/topology/molecule.py
jthorton: Thanks for that I have managed to get around this now by moving the imports around.

j-wags:
> The notebook also highlighted an unexpected failure whereby an OFFMol could not be made from an RDKit molecule which was created using the canonical_isomeric_explicit_hydrogen_mapped_smiles cmiles tags due to unspecified stereochemistry, however, making the molecule with the openeye backend does not raise such an error.

I bet this is the same issue as #412, where RDKit effectively includes map index in the CIP ranking algorithm, and so molecules that really _aren't_ stereogenic can be perceived as having stereochemistry if map indices are present.

j-wags:
> Alternatively, we could introduce a totally unique ordering system which is built into the OFFTK and will be used automatically when submitting jobs to the archive using the toolkit. This will, however, mean that the future submissions are ordered differently to those already submitted and there may be a small number of duplicate records that slip through. We should also be careful with the design of this ordering as changing it in future versions will mean that we fall back into this trap again.

I think this is basically inevitable -- We won't be using OpenEye forever (since we eventually want to offer a fully open-source stack), and so OETK canonicalization will eventually become outdated. In some sense, though, we never escape toolkit dependence -- If we implement our own canonicalization scheme based solely on (element, coordinates), then we still have the problem of _where does the conformer come from_? So maybe RDKit canonical SMILES ordering is the lesser of all evils? Or a clever toolkit built-in ordering scheme? Or just a way to search over SMILES in the QCArchive and detect if a torsiondrive has already been performed with acceptably-similar settings to the scan we intend to do?

This is a hard question, and I don't think we should try to address it in this PR. Can we start collecting our thoughts on this in a separate issue for discussion?

openforcefield/topology/molecule.py
j-wags: Posting a Slack DM discussion with some interesting API points: 

> Trevor Gokey  23:55
> Hey Jeff, I was playing around with the isomorph thing after reading your new comments. What do you think of this pseudocode? It envisions how I see using isomorphs, at least in my cases with dealing with the wild wild QCArchive. I see things this way because I am often trying to debug/categorize molecules that I know nothing about, and need to tease them apart. For me this works way better than a single function that just says yes/no isomorphic if I am e.g. trying to figure out why parameters are not the same between molecules.
```
def describe(mol1, mol2):
    exact = True
    #### check to see if heavy atom structure is the same to catch e.g. tautomers
    #### check formal charges ignoring hydrogen, this skips tautomer checks
    map = isomorphic(mol1, mol2, colors=None, ignore_H=True)
    if not map:
        return False
    print("Same heavy atom structure")
    equality_ops = [ eq.symbols, eq.aromaticity, eq.formal_charges ]
    for eq in equality_ops:
        if eq(mol1, mol2, map):
            print( eq, "True")
        else:
            exact = False
    ##### check for complete-atom isomorphs, with hydrogen
    ##### if different hydrogen and different formal charge, probably tautomer
    ##### if different hydrogen and same formal_charge, then different total charge (partial_charge)
    ##### need to establish same hydrogen before asking about stereochemistry, so check only if the molecule is completely isomorphic
    map = isomorphic(mol1, mol2, colors=None, ignore_H=False)
    if not map:
        print("Not same hydrogens: Tautomers?")
        return False
    equality_ops = [ eq.stereocenters, eq.parameter_labels, eq.formal_charges, eq.partial_charges ]
    for eq in equality_ops:
        if eq(mol1, mol2, map):
            print( eq, "True")
        else:
            exact = False
    if exact:
        print("Same exact molecule")
    return exact
```

> So you can see it is straightforward to check any other parameters or attributes in the molecules, as long as isomorphism is established. It is obviously very rough and maybe missing a few key points, but wanted to know your input before putting on the github discussion. I can see using colors as a shortcut to just check internally, making it easier for the user, e.g. isomorphic(mol1, mol2, colors=[eq.symbols, eq.aromaticity]) . Of course, here in my case by default colors=None and ignore_H=False .


> Jeffrey Wagner  12:44
> I have a lot of thoughts on this:
> * That's a beautiful API example, and I like the idea of using enums instead of a bunch of kwargs. 
> * Your use case is still weird, but I understand it better now. In some sense, I'd call what you're doing "dangerous", but I'm coming around to the idea that some fraction of users will also do that sort of analysis, no matter what we think. So if we can provide them clear, well-designed isomorphism tools, we'll enable more cool, reproducible science. And if we don't have to go too far out of the way to include that, it's a worthwhile goal.
> * I'm thinking that, if we three stakeholders have this many opinions on this API/functionality, then other toolkit users probably have a bunch of other opinions that I can't conceive of as well. The new isomorphism functionality isn't in the master branch yet, so we haven't broken the public API. We could keep all of the new code in the PR, but hide the parts under discussion as private functions (so are_isomorphic --> _are_isomorphic). This would start letting power users (like you and me) start banging on the underlying algorithms, while we continue to discuss the API. 


> Trevor Gokey  12:58
> As a small implementation detail, the enum you see is actually a list of functions, at least in the way I see it. Of course it could be implemented in various ways. I really think some nice public API to establish equality of our data is important. For example, FrozenMolecule.__eq__ could just be applying all of those `eq` functions.
> As for the dev versions: Yes, as we bang on this a bit, I definitely agree that we should have some "dev names" for these functions. Maybe not prepending "_" as it technically marks them private. Maybe a new namespace, e.g. Molecule.dev.are_isomorphic.
> In any case, a low level and high level API are pretty typical, so having both is a Good Thing(tm)

davidlmobley:
Agree this is a hard issue and that we may have to canonicalize order somehow. Probably a separate issue is the place to discuss.

openforcefield/topology/molecule.py
jthorton: I think I have a bit of a better understanding of this use case though I would still say its a rare one. I think a function like this constructed by a user could be achieved using the proposed `are_isomorphic` function with one additional kwarg of `include_hs=True` which should help you catch tautomers we could even expose a function to try and compare molecules and identify tautomers such as this `Molecule.are_tautomers(mol1, mol2)`. We would also need to change all `colour` matching options to `False` then you should be able to iterate over the matching where the enums represent kwarg dicts for example

```python
matching_options = [{'formal_charge_matching': True}, {'bond_order_matching': True}]
for kwargs in matching_options:
    if Molecule.are_isomorphic(mol1, mol2, **kwargs):
        # do somthing with the molecules
```

This would also mean that the default matching is basic graph matching (colourless edges and nodes with `atomic_number` attribute matching).

jthorton:
> I bet this is the same issue as #412, where RDKit effectively includes map index in the CIP ranking algorithm, and so molecules that really _aren't_ stereogenic can be perceived as having stereochemistry if map indices are present.

I think you are right so we are back to this point again. I think we should aim to fix this if we can, so do we want to just strip the atom map from the molecule before running `chem.AssignStereochemistry`, I am not sure what openeye does in this case but I don't get the error with that toolkit so I assume it is just ignoring the atom map? So to make the behaviour the same between the toolkit backends we should always strip and ignore the atom mapping when doing sterochem (note that the atom mapping will still be preserved and returned to the user in the `properties` dict of the `Molecule` under the key `atom_map`).

jthorton:
> Agree this is a hard issue and that we may have to canonicalize order somehow. Probably a separate issue is the place to discuss.

I think in the next part of the development of the bespoke workflow we will also want to look at standardising and automating QCArchive submission and picking this issue back up (https://github.com/openforcefield/qca-dataset-submission/issues/53#) we can roll this into that discussion. Should I just make the `to_qcschema` method private till then?

j-wags:
> Should I just make the to_qcschema method private till then?

Yes! Let's punt on that since we'll need to resolve it soon anyway.

j-wags:
@trevorgokey 

> I definitely agree that we should have some "dev names" for these functions. Maybe not prepending "_" as it technically marks them private. Maybe a new namespace, e.g. Molecule.dev.are_isomorphic.
> In any case, a low level and high level API are pretty typical, so having both is a Good Thing(tm)

I thought about it this weekend, and the `dev` namespace is a great idea. I don't think we should do it in this PR (since this is becoming way too broad), but having a dev namespace would really make it much easier to develop for the toolkit. API development is a chicken-and-egg problem, and being able to take a half-step while we get to try use cases will let us move much faster.

openforcefield/topology/molecule.py
```diff
@@ -2952,31 +3148,35 @@ def from_file(file_path,
...
-                        f"molecule and use openforcefield.topology.Molecule.from_rdkit, but we do not recommend this."
+                           f"have Corina atom types, and this is not common in the simulation community. For this " \
+                           f"reason, the Open Force Field Toolkit does not use " \
+                           f"RDKit to read .mol2. Consider reading from SDF instead. If you would like to attempt " \
+                           f"to use RDKit to read mol2 anyway, you can load the molecule of interest into an RDKit " \
+                           f"molecule and use openforcefield.topology.Molecule.from_rdkit, but we do not recommend this."
+                elif file_format == 'PDB' and RDKitToolkitWrapper.is_available():
+                    msg += "RDKit can not safely read PDBs on their own. Information about bond order and aromaticity " \
+                           "is likely to be lost. PDBs can be used along with a valid smiles string with RDKit using " \
+                           "the constructor Molecule.from_pdb_and_smiles(file_path, smiles)"
```
j-wags: 👍

openforcefield/topology/molecule.py
```diff
@@ -2952,31 +3148,35 @@ def from_file(file_path,
...
-                raise NotImplementedError(
-                    "Toolkit {} can not read file {} (format {}). Supported formats for this toolkit "
-                    "are {}".format(toolkit.toolkit_name, file_path,
-                                    file_format,
-                                    toolkit.toolkit_file_read_formats))
+                msg = f"Toolkit {toolkit.toolkit_name} can not read file {file_path} (format {file_format}). Supported " \
+                      f"formats for this toolkit are {toolkit.toolkit_file_read_formats}."
+                if toolkit.toolkit_name == 'The RDKit' and file_format == 'PDB':
+                    msg += "RDKit can however read PDBs with a valid smiles string using the " \
+                           "Molecule.from_pdb(file_path, smiles) constructor"
```
j-wags:
```suggestion
                           "Molecule.from_pdb_and_smiles(file_path, smiles) constructor"
```

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3376,343 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+        """
+        Generate the qschema input format used to submit jobs to archive
+        or run qcengine calculations locally, the molecule is placed in canonical order first.
+        spec can be found here <https://molssi-qc-schema.readthedocs.io/en/latest/index.html>
+
+        .. warning :: This API is experimental and subject to change.
+
+        Parameters
+        ----------
+        multiplicity : int, default=1, the multiplicity of the molecule required for qcschema
```
j-wags:
```suggestion
        multiplicity : int, default=1, 
            The multiplicity of the molecule required for qcschema
```

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3376,343 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+
+        conformer : int, default=0, the index of the conformer that should be used for qcschema
```
j-wags:
```suggestion
        conformer : int, default=0
            The index of the conformer that should be used for qcschema
```

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3376,343 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+
+        Returns
+        ---------
+        qcelemental.models.Molecule : a validated qcschema
```
j-wags:
```suggestion
        qcelemental.models.Molecule 
            A validated qcschema
```

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3376,343 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+    def from_mapped_smiles(cls, mapped_smiles, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY, allow_undefined_stereo=False):
+        """
+        Create an openforcefield.topology.molecule.Molecule from a mapped SMILES made with cmiles.
+        The molecule will be in the order of the indexing in the mapped smiles string.
+
+        .. warning :: This API is experimental and subject to change.
+
+        Parameters
+        ----------
+        mapped_smiles: str, a cmiles style mapped smiles string with explicit hydrogens.
```
j-wags:
```suggestion
        mapped_smiles: str
            A CMILES-style mapped smiles string with explicit hydrogens.
```

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3376,343 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+
+        toolkit_registry : openforcefield.utils.toolkits.ToolRegistry or openforcefield.utils.toolkits.ToolkitWrapper, optional, default=None
```
j-wags:
```suggestion
        toolkit_registry : openforcefield.utils.toolkits.ToolkitRegistry or openforcefield.utils.toolkits.ToolkitWrapper, optional, default=None
```

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3376,343 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+        connectivity = [(bond.atom1_index, bond.atom2_index, bond.bond_order) for bond in canonical_mol.bonds]
+        symbols = [Element.getByAtomicNumber(atom.atomic_number).symbol for atom in canonical_mol.atoms]
+
+        schema_dict = {'symbols': symbols, 'geometry': geometry, 'connectivity': connectivity,
+                       'molecular_charge': charge, 'molecular_multiplicity': multiplicity}
+
+        return qcel.models.Molecule.from_data(schema_dict, validate=True)
+
+    @classmethod
+    def from_mapped_smiles(cls, mapped_smiles, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY, allow_undefined_stereo=False):
```
j-wags: (Blocking) Let's make this private (`def _from_mapped_smiles`) until we can discuss this more, for two reasons:
1) #412 will make this break for a lot of molecules
2) I think that supporting `offmol._properties['atom_map']` leads down a dangerous road, and would like to chat in person before we commit to implementing that.

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3376,343 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+        """
+        Create a Molecule from  a QCArchive entry based on the cmiles information.
+
+        If we also have a client instance/address we can go and attach the starting geometry.
+
+        Parameters
+        ----------
+        qca_record : dict, a QCArchive dict with json encoding or record instance
+
+        client : optional, default=None, a qcportal.FractalClient instance so we can pull the initial molecule geometry.
```
j-wags:
```suggestion
        client : optional, default=None
            A qcportal.FractalClient instance so we can pull the initial molecule geometry.
```

openforcefield/topology/molecule.py
```diff
@@ -3176,6 +3376,343 @@ def from_openeye(oemol, allow_undefined_stereo=False):
...
+    @classmethod
+    def from_qcschema(cls, qca_record, client=None, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY, allow_undefined_stereo=False):
+        """
+        Create a Molecule from  a QCArchive entry based on the cmiles information.
+
+        If we also have a client instance/address we can go and attach the starting geometry.
+
+        Parameters
+        ----------
+        qca_record : dict, a QCArchive dict with json encoding or record instance
```
j-wags:
```suggestion
        qca_record : dict
            A QCArchive dict with json encoding or record instance
```

openforcefield/topology/molecule.py
j-wags: > This may be best placed in another PR though?

Agreed. This would be extremely nifty, but we can save it for a later day.

openforcefield/utils/toolkits.py
```diff
@@ -1036,6 +1039,48 @@ def to_smiles(molecule):
...
+
+        Parameters
+        ----------
+        molecule: openforcefield.topology.Molecule
+            The input molecule
+
+         Returns
+        -------
+        molecule : openforcefield.topology.Molecule
+            An openforcefield-style molecule.
```
j-wags:
```suggestion
            The input molecule, with canonically-indexed atoms and bonds.
```

openforcefield/utils/toolkits.py
```diff
@@ -1701,6 +1810,44 @@ def to_file(self, molecule, file_path, file_format):
...
+
+        Parameters
+        ----------
+        molecule: openforcefield.topology.Molecule
+            The input molecule
+
+         Returns
+        -------
+        molecule : openforcefield.topology.Molecule
+            An openforcefield-style molecule with atoms in canonical order.
```
j-wags:
```suggestion
            The input molecule, with canonically-indexed atoms and bonds.
```

openforcefield/tests/test_forcefield.py
```diff
@@ -269,6 +269,93 @@ def create_ethanol():
...
+    acetaldehyde.add_atom(1, 0, False)  # H6
+    acetaldehyde.add_bond(0, 1, 1, False)  # C0 - C1
+    acetaldehyde.add_bond(1, 2, 2, False)  # C1 = O2
+    acetaldehyde.add_bond(0, 3, 1, False)  # C0 - H3
+    acetaldehyde.add_bond(0, 4, 1, False)  # C0 - H4
+    acetaldehyde.add_bond(0, 5, 1, False)  # C0 - H5
+    acetaldehyde.add_bond(1, 6, 1, False)  # C1 - H6
+    charges = unit.Quantity(np.array([0, 0, 0, 0, 0, 0, 0]), unit.elementary_charge)
+    acetaldehyde.partial_charges = charges
+    return acetaldehyde
```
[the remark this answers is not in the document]
jthorton: Can never have too many!

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +530,326 @@ def test_name(self):
...
+        # get benzene with all aromatic atoms/bonds labeled
+        benzene = Molecule.from_smiles('c1ccccc1')
+        # get benzene with no aromatic labels
+        benzene_no_aromatic = create_benzene_no_aromatic()
+        # now test all of the variations
+        assert Molecule.are_isomorphic(benzene, benzene_no_aromatic, aromatic_matching=inputs['aromatic_matching'],
+                                       formal_charge_matching=inputs['formal_charge_matching'],
+                                       bond_order_matching=inputs['bond_order_matching'],
+                                       atom_stereochemistry_matching=inputs['atom_stereochemistry_matching'],
+                                       bond_stereochemistry_matching=inputs['bond_stereochemistry_matching'])[0] is inputs['result']
```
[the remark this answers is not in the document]
jthorton: Thanks, first time doing testing like this wasn't bad :+1:

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +530,326 @@ def test_name(self):
...
+                del bond_dict['atom1']
+                del bond_dict['atom2']
+                remapped_bond_dict = remaped_bonds[key].to_dict()
+                del remapped_bond_dict['atom1']
+                del remapped_bond_dict['atom2']
+            assert mol1.n_bonds == mol2.n_bonds
+            assert mol1.n_angles == mol2.n_angles
+            assert mol1.n_propers == mol2.n_propers
+            assert mol1.n_impropers == mol2.n_impropers
+            assert mol1.total_charge == mol2.total_charge
```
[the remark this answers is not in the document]
jthorton: I have added this to the check as well but changed the construction of the normal ethanol to use the toolkit independent function instead of the sdf file to make sure the charges are assigned.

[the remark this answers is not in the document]
jthorton: We might want to update this test when we can read charges from SDF files?

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +530,326 @@ def test_name(self):
...
+        # get ethanol in canonical order
+        ethanol = create_ethanol()
+        # get reversed non canonical ethanol
+        reversed_ethanol = create_reversed_ethanol()
+        # get the canonical ordering
+        canonical_ethanol = reversed_ethanol.canonical_order_atoms(rdkit)
+        # make sure the mapping between the ethanol and the rdkit ref canonical form is the same
+        assert (True, {0: 2, 1: 0, 2: 1, 3: 8, 4: 3, 5: 4, 6: 5, 7: 6, 8: 7}) == Molecule.are_isomorphic(canonical_ethanol,
+                                                                                                         ethanol,
+                                                                                                         True)
```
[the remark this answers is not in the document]
jthorton: I have changed them to make them look the same apart from the final ordering.

openforcefield/tests/test_molecule.py
```diff
@@ -529,6 +530,326 @@ def test_name(self):
...
+        # As the method can take a record instance or a dict with JSON encoding test both
+        # test incomplete dict
+        example_dict = {'name': 'CH4'}
+        with pytest.raises(KeyError):
+            mol = Molecule.from_qcschema(example_dict)
+
+        # test an object that is not a record
+        wrong_object = 'CH4'
+        with pytest.raises(AttributeError):
+            mol = Molecule.from_qcschema(wrong_object)
```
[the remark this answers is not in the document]
jthorton: Your right it looks a tad unfinished, the method should also work without a client call in the function I have updated the test now to use a record loaded from a file.

openforcefield/topology/molecule.py
jthorton: Yeah, it might be better to follow the toolkits and have an atom map index attribute on each atom which can be set? This could be a safer way of storing the atom map and reordering molecules using the indexes? This would also be helpful for the fragmenter move I think.

I think #412 is fixed using the atom map stashing that I put into the `from_smiles` method in toolkit but I should possibly test it on more molecules?

jthorton:
The large notebook energy was due to an incorrect unit conversion in QCEngine which will be fixed with this [PR](https://github.com/MolSSI/QCEngine/pull/226), for now, I use RDKit and explain what units the energy is returned in. One of the LGTM errors about the Non-iterable is not something I have changed so not sure what is going on there. Just waiting on tests now.

jthorton:
Any idea why LGTM JavaScript is failing? @j-wags

j-wags:
No clue. We don't have Javascript AFAIK. This is a LGTM problem. Feel free to merge, this is not blocking.

PR #497, opened 2020-01-27
Test switching ambermini to ambertools in travis

j-wags:
- [x] Verify that builds pass when we switch dependencies from `omnia:ambermini` to `omnia:ambertools`


j-wags:
Looks good. Merging.

PR #499, opened 2020-01-29
Add note about maintaining py36 to travis matrix

j-wags:
This PR puts a comment in our `.travis.yml` file to not remove the py36 builds until at least Feb 2022. 

@proteneer, a downstream user of OFFTK, has requested that we maintain py36 compatibility for at least the next two years. I don't see a problem with this, especially since there are probably several more silent downstream users that would benefit from this guarantee. I don't know of any enhancements in python>37 that would greatly improve our workflow, so I think a 2-year guarantee is fine. Are there any objections?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,10 @@ branches:
 
 matrix:
   include:
+  # The maintainer of a downstream project has requested that 
+  # we maintain py36 compatibility until at least Feb 2022.
+  # Contact Jeff Wagner or Yutong Zhao if this is an issue.
+
   # Test osx with RDKit
   - os: osx
     language: generic


dgasmith:
3.7 was pretty minor for synchronous code (async was pretty hefty). Most of the upgrades are useful for debugging (breakpoints, new fstrings). Type annotations are pretty crucial for those going deep into that however though you can probably get around this with putting the 3.7 features in `if TYPE_CHECKING:` blocks.

Main item is conda-forge will not keep up this promise. Python 3.6 is scheduled for deprecation fairly soon, 3.8 builds are nearing done.

@proteneer Out of curiosity, could you share why this needed?

j-wags:
> Most of the upgrades are useful for debugging (breakpoints, new fstrings). Type annotations are pretty crucial for those going deep into that however though you can probably get around this with putting the 3.7 features in if TYPE_CHECKING: blocks.

Hm, I didn't know about the breakpoints. OFFTK hasn't been making heavy use of type annotation (though @SimonBoothroyd has been doing something like that, so let's loop him in). 

FYI -- Python 3.6's end of life is Dec 2021 [1], so guaranteeing until Feb 2022 seems fitting on that calendar. Though I agree that conda-forge's support schedule seems to be a bit more aggressive. 

[1] https://devguide.python.org/#status-of-python-branches

SimonBoothroyd:
> Hm, I didn't know about the breakpoints. OFFTK hasn't been making heavy use of type annotation (though @SimonBoothroyd has been doing something like that, so let's loop him in).

I think 3.6 support would only be provided for the OFFTK right? If so I don't think this will affect me. 

In general though as Daniel says, 3.7 and 3.8 do add some nice features for typing, such as `TypedDict` (although this is available as a backwards compatible module), performance improvements for typing (which I read somewhere were large in some cases but can't find the article now), and some QoL things such as `Union` objects no longer collapsing explicit subtypes.

jthorton:
I think the toolkit could probably support this, however if we are to refactor the molecule model within the next two years could this be an issue? I assume we will be doing extensive type checking with the new model.

j-wags:
> however if we are to refactor the molecule model within the next two years could this be an issue? I assume we will be doing extensive type checking with the new model.

Hmm, I assume that if we're aiming to make a highly interoperable molecule class, we'll lose a lot of friends by excluding python<3.7, so there's a good chance that won't hold us back. 

> I think 3.6 support would only be provided for the OFFTK right? If so I don't think this will affect me.

Correct. Though, a good policy might be to keep compatibility with the versions supported by conda-forge, since we'll head there soon. 

> In general though as Daniel says, 3.7 and 3.8 do add some nice features for typing, such as TypedDict (although this is available as a backwards compatible module), performance improvements for typing (which I read somewhere were large in some cases but can't find the article now), and some QoL things such as Union objects no longer collapsing explicit subtypes.

Is that what you do in your free time? It's scary how much you know.

j-wags:
Ok, since nobody has raised hard objections, I'm going to merge this for the OFF toolkit.

PR #506, opened 2020-02-06
[WIP] Find rotatable bonds

### LINKED ISSUE #500: Include find rotatable bonds in OETKW and RDKTKW
This is needed for the bespoke workflow.

### LINKED ISSUE #513: OFFMol.to_rdkit loses aromaticity info
**Describe the bug**
`Molecule.to_rdkit` loses track of atom aromaticity in the example below. This happens in the line https://github.com/openforcefield/openforcefield/blob/master/openforcefield/utils/toolkits.py#L2282
```
Chem.SanitizeMol(rdmol, Chem.SANITIZE_ALL ^ Chem.SANITIZE_ADJUSTHS ^ Chem.SANITIZE_SETAROMATICITY)
```
which messes with aromaticity even though we tell it not to.

**To Reproduce**
```
smiles = "[H]c1c(c(c(c(c1OC2=C(C(=C(N3C2=C(C(=C3[H])C#N)[H])[H])F)[H])OC([H])([H])C([H])([H])N4C(=C(C(=O)N(C4=O)[H])[H])[H])[H])F)[H]"
mol = Molecule.from_smiles(smiles)

for atom in mol.atoms:
    print('OFF: ', atom.is_aromatic)
rdmol = mol.to_rdkit()
from rdkit import Chem

for rdatom in rdmol.GetAtoms():
    print('RDK: ', rdatom.GetIsAromatic())

```

> OFF:  False
> OFF:  True
> OFF:  True
> OFF:  True
> OFF:  True
> OFF:  True
> OFF:  True
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> OFF:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False
> RDK:  False

**Proposed Solution**
Add 
```
        if aromaticity_model == 'OEAroModel_MDL':
            Chem.SetAromaticity(rdmol, Chem.AromaticityModel.AROMATICITY_MDL)
        else:
            raise ValueError(f"Aromaticity model {aromaticity_model} not recognized")
``` 
to here: https://github.com/openforcefield/openforcefield/blob/master/openforcefield/utils/toolkits.py#L2282

**Steps to close Issue**
- [x] Code change above
- [x] Add this as a test case in `test_toolkits` for `RDKitToolkitWrapper`

jthorton:
- [X] Fixes #500 
- [X] Fixes #513

### DIFF
--- a/openforcefield/tests/test_molecule.py
+++ b/openforcefield/tests/test_molecule.py
@@ -1059,57 +1059,6 @@ class TestMolecule:
     # Test Molecule methods.
     # ----------------------
 
-    def test_find_rotatable_bonds(self):
-        """Test finding rotatable bonds while ignoring some groups"""
-
-        # test a simple molecule
-        ethanol = create_ethanol()
-        bonds = ethanol.find_rotatable_bonds()
-        assert len(bonds) == 2
-        for bond in bonds:
-            assert ethanol.atoms[bond.atom1_index].atomic_number != 1
-            assert ethanol.atoms[bond.atom2_index].atomic_number != 1
-
-        # now ignore the C-O bond
-        bonds = ethanol.find_rotatable_bonds('[#6:1]-[#8:2]')
-        assert len(bonds) == 1
-        assert ethanol.atoms[bonds[0].atom1_index].atomic_number == 6
-        assert ethanol.atoms[bonds[0].atom2_index].atomic_number == 6
-
-        # now ignore the C-C bond
-        bonds = ethanol.find_rotatable_bonds('[#6:1]-[#6:2]')
-        assert len(bonds) == 1
-        assert ethanol.atoms[bonds[0].atom1_index].atomic_number == 6
-        assert ethanol.atoms[bonds[0].atom2_index].atomic_number == 8
-
-        # ignore a list of searches
-        bonds = ethanol.find_rotatable_bonds(['[#6:1]-[#8:2]', '[#6:1]-[#6:2]'])
-        assert bonds == []
-
-        # test  molecules that should have no rotatable bonds
-        cyclohexane = create_cyclohexane()
-        bonds = cyclohexane.find_rotatable_bonds()
-        assert bonds == []
-
-        methane = Molecule.from_smiles('C')
-        bonds = methane.find_rotatable_bonds()
-        assert bonds == []
-
-        ethene = Molecule.from_smiles('C=C')
-        bonds = ethene.find_rotatable_bonds()
-        assert bonds == []
-
-        terminal = '[*]~[*:1]-[X2H1,X3H2,X4H3:2]-[#1]'
-        # test removing terminal rotors
-        toluene = Molecule.from_file(get_data_file_path('molecules/toluene.sdf'))
-        bonds = toluene.find_rotatable_bonds()
-        assert len(bonds) == 1
-        assert toluene.atoms[bonds[0].atom1_index].atomic_number == 6
-        assert toluene.atoms[bonds[0].atom2_index].atomic_number == 6
-
-        bonds = toluene.find_rotatable_bonds(terminal)
-        assert bonds == []
-
     def test_add_conformers(self):
         """Test addition of conformers to a molecule"""
         import numpy as np
--- a/openforcefield/tests/test_molecule.py
+++ b/openforcefield/tests/test_molecule.py
@@ -35,7 +35,7 @@ from openforcefield.topology.molecule import Molecule, Atom
 from openforcefield.utils import get_data_file_path
 # TODO: Will the ToolkitWrapper allow us to pare that down?
 from openforcefield.utils.toolkits import OpenEyeToolkitWrapper, RDKitToolkitWrapper, AmberToolsToolkitWrapper, ToolkitRegistry
-
+from openforcefield.tests.test_forcefield import create_ethanol, create_cyclohexane
 
 #=============================================================================================
 # TEST UTILITIES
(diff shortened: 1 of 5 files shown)


jthorton:
This is the my initial go at adding the ability to find rotatable bonds, which seems to work quite well. I have not added a new bond attribute `is_rotatable` yet as it was not clear if it was needed and how it would behave. For example would the attribute update after each query (similar to how the WBO is stored in the `fractional_bond_order` attribute) meaning the function would do something like

```python
def find_rotatable_bonds(self, ignore_functional_groups=None, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY):
    rotatable_bond_smarts = '[!$(*#*)&!D1:1]-&!@[!$(*#*)&!D1:2]'
    # get all of the general matches
    general_matches = self.chemical_environment_matches(query=rotatable_bond_smarts,
                                                        toolkit_registry=toolkit_registry)
    ignore_functional_groups = [list of SMARTS to ignore]
    ignore_bonds = toolkit_registryfind_smarts_matches to find matches for the items in 
    ignore_functional_groups
    Remove ignore_bonds from rotatable_bonds
    
    # now reset the is_rotatable attribute on each bond to match the search
    for bond in rotatbale_bonds:
        bond.is_rotatable = True

    return rotatable_bonds
 ```

File: openforcefield/topology/molecule.py [hunk 1]
```diff
@@ -1344,6 +1344,12 @@ def molecule_bond_index(self):
             raise ValueError('This Atom does not belong to a Molecule object')
         return self._molecule.bonds.index(self)
 
+    def __repr__(self):
+        return f"Bond(atom1 index={self.atom1_index}, atom2 index={self.atom2_index})"
+
+    def __str__(self):
+        return f"<Bond atom1 index='{self.atom1_index}', atom2 index='{self.atom2_index}'>"
+
```
j-wags: Nice. Thanks for adding these.

openforcefield/topology/molecule.py
```diff
@@ -2194,11 +2200,71 @@ def to_networkx(self):
...
+                                                            toolkit_registry=toolkit_registry)
+
+        # this will give all forwards and backwards matches, so condense them down with this function
+        def condense_matches(matches):
+            condensed_matches = set()
+            for m in matches:
+                condensed_matches.add(tuple(sorted(m)))
+            return condensed_matches
+
+        general_bonds = condense_matches(general_matches)
```
j-wags: Good thinking!

openforcefield/tests/test_molecule.py
```diff
@@ -688,6 +688,57 @@ def test_equality(self):
...
+        assert len(bonds) == 2
+        for bond in bonds:
+            assert ethanol.atoms[bond.atom1_index].atomic_number != 1
+            assert ethanol.atoms[bond.atom2_index].atomic_number != 1
+
+        # now ignore the C-O bond
+        bonds = ethanol.find_rotatable_bonds('[#6:1]-[#8:2]')
+        assert len(bonds) == 1
+        assert ethanol.atoms[bonds[0].atom1_index].atomic_number == 6
+        assert ethanol.atoms[bonds[0].atom2_index].atomic_number == 6
```
j-wags: Could you copy this block and also run the check when the SMIRKS indices are reversed? The `sorted` line in the new function should handle it, but it's always good to be sure.

openforcefield/topology/molecule.py
```diff
@@ -2194,11 +2200,71 @@ def to_networkx(self):
...
+
+        # now refine the list using the ignore groups
+        if ignore_functional_groups is not None:
+            matches_to_ignore = set()
+            if not isinstance(ignore_functional_groups, list):
+                ignore_functional_groups = [ignore_functional_groups]
```
j-wags: Since there are iterable objects other than lists, we try to follow a design pattern like this: 

https://github.com/openforcefield/openforcefield/blob/ee0f715e094e6f1749bf60f3a3b48174a9e6a9b7/openforcefield/topology/topology.py#L1030-L1035)

openforcefield/tests/test_molecule.py
```diff
@@ -688,6 +688,57 @@ def test_equality(self):
     # Test Molecule methods.
     # ----------------------
 
+    def test_find_rotatable_bonds(self):
+        """Test finding rotatable bonds while ignoring some groups"""
```
j-wags: Could you move this to `test_toolkits`, and run once with OETK and another time with RDKit? I know the different setups are _technically_ covered in our Travis testing matrix, but weird things can happen when both toolkits are installed that hide sneaky problems.

openforcefield/topology/molecule.py
jthorton: Thanks for catching that I have made that change.

openforcefield/tests/test_molecule.py
jthorton: No problem its best to be safe here, I have moved the tests.

openforcefield/tests/test_molecule.py
jthorton: Good point, I have added that as well.

PR #508, opened 2020-02-07
Implement AmberTools WBO

### LINKED ISSUE #507: Implement and test `compute_wiberg_bond_orders` in AmberToolsToolkitWrapper
Wiberg bond order calculation was added to AmberTools19. They can be calculated at this moment from the `conda-forge:ambertools:19.11` package, the `ambermd:ambertools:19.0` or from AmberTools19 compiled from source. 

Make sure that `sqm.in` includes the `printbondorders=1` keyword. For example:
```
 &(ambertool) jwagner@MBP-S$ cat sqm.in 
Find Bond Orders
 &qmmm
  scfconv=1.0000E-8,                                                          
  verbosity=1,                                                                 
  qmmask=':1',                                                                 
  diag_routine=1,                                                              
  printcharges=1,
  qm_theory='AM1',                                                             
  printcharges=1,
  printbondorders=1,
 /
   6     C        2.5780       -0.0520        0.0000 
   6    C1        1.8410       -1.2370       -0.0000 
   6    C2        1.9230        1.1800        0.0000 
   6    C3        0.4480       -1.1950       -0.0000 
   6    C4        0.5310        1.2320       -0.0000 
   6    C5       -0.2090        0.0410       -0.0000 
   6    C6       -1.6730        0.1250       -0.0000 
   8     O       -2.3890        1.1330       -0.0000 
   8    O1       -2.3280       -1.0750        0.0000 
   1     H        3.6770       -0.0890        0.0000 
   1    H1        2.3590       -2.2080       -0.0010 
   1    H2        2.5050        2.1140        0.0000 
   1    H3       -0.1420       -2.1260        0.0000 
   1    H4        0.0010        2.1990       -0.0000 
   1    H5       -3.2860       -0.9150        0.0010 
```

After running `sqm -i sqm.in`, the bond orders are printed out in `mdout` as follows:
```
Bond Orders
 
  QMMM:    NUM1 ELEM1 NUM2 ELEM2      BOND_ORDER
  QMMM:       2   C      1   C        1.41107532
  QMMM:       3   C      1   C        1.41047804
  QMMM:       3   C      2   C        0.00549987
  QMMM:       4   C      1   C        0.00732955
  QMMM:       4   C      2   C        1.42657761
  QMMM:       4   C      3   C        0.10781945
  QMMM:       5   C      1   C        0.00741156
  QMMM:       5   C      2   C        0.10784428
  QMMM:       5   C      3   C        1.42745189
  QMMM:       5   C      4   C        0.00751630
  QMMM:       6   C      1   C        0.10899281
  QMMM:       6   C      2   C        0.00591824
  QMMM:       6   C      3   C        0.00595150
  QMMM:       6   C      4   C        1.37732350
  QMMM:       6   C      5   C        1.37408715
  QMMM:       7   C      1   C        0.00255253
...
```

* Ensure that the correct AmberTools package (version>=19) comes in the standard conda install 
* Implement `AmberToolsToolkitWrapper.compute_wiberg_bond_orders`
* Filter bond orders to only pull values for actual bonds (they're currently reported for all atom pairs)
* Decide whether another function should be made to report the complete NxN WBO matrix
* Implement the same tests as `OpenEyeToolkitWrapper.compute_wiberg_bond_orders` uses

j-wags:
- [x] Closes #507 by implementing `AmberToolsToolkitWrapper.assign_fractional_bond_orders`
- [x] Change `Molecule.compute_wiberg_bond_orders` to `Molecule.assign_fractional_bond_orders` (This goes along with another planned change, from `Molecule.compute_partial_charge` to `Molecule.assign_partial_charges`)
- [x] Change `ToolkitWrapper.compute_wiberg_bond_orders` to `ToolkitWrapper.assign_fractional_bond_orders`
- [x] Add `use_conformers` kwarg to `assign_fractional_bond_orders`, which defaults to None. If None, the function generates its own conformer using an available ToolkitWrapper. Otherwise, if the user provides an iterable of conformers, the FBO calculation will use those.
- [x] Change keyword name for `assign_fractional_bond_orders` from `charge_model` to `bond_order_model`. The previous use of `charge_model` as a keyword name is a copy-paste mistake, since we copied so much of that function from `Molecule.compute_partial_charges`.
- [x] Change accepted keyword values for `assign_fractional_bond_orders` from `am1` and `pm3` to `am1-wiberg` and `pm3-wiberg`. This is because we may introduce bond order population analyses other than Wiberg in the future.
- [x] Changes the expected Wiberg bond orders for the existing test molecule. Due to small differences between the values returned by OpenEye and AmberTools I've relaxed the allowed aromatic bond order range. 


j-wags:
Ok, so for our "detailed" test molecule (protonated and deprotonated versions of [1]), AmberTools narrowly failed some tests that we had calibrated against OETK. Basically, we tested to ensure that every aromatic bond fell in the range of 1.15 to 1.60 in both the neutral molecule and anion. AmberTools returns values in the range 1.06 to 1.61, which caused these tests to fail. Here's a full summary:

| Charge state | AmberTools WBO range | OE WBO range |
|--|--|--|
| neutral | 1.34 - 1.44 | 1.35 - 1.44 |
| anion | 1.06 - 1.61 | 1.12 - 1.55 |

So basically AmberTools returns slightly more extreme values, but everything is around the same magnitude. Because these differences are small, so I'm going to relax the aromatic tolerances for these tests to accept between 1.05 and 1.65. 

Huge thanks to @sukanyasasmal and @Chayast and for making this thoughtful test. It was very quick to plug in AmberTools and get results!

[1] 
Screen Shot 2020-02-08 at 10 20 06 PM

[2] Printout of aromatic bond orders in these tests (molecule is loaded from SDF so indexing is the same):
```
OpenEye neutral aromatic order: 1.4249782910821187
OpenEye neutral aromatic order: 1.4247535696661562
OpenEye neutral aromatic order: 1.3993880829721475
OpenEye neutral aromatic order: 1.386557474305936
OpenEye neutral aromatic order: 1.3864636016050376
OpenEye neutral aromatic order: 1.3721272356508605
OpenEye neutral aromatic order: 1.3798238226278605
OpenEye neutral aromatic order: 1.4308855238777132
OpenEye neutral aromatic order: 1.3485824537022644
OpenEye neutral aromatic order: 1.3484389250490467
OpenEye neutral aromatic order: 1.4361705759121075
OpenEye anion aromatic order: 1.4214715268113562
OpenEye anion aromatic order: 1.4082261746981803
OpenEye anion aromatic order: 1.4083384738085785
OpenEye anion aromatic order: 1.421413561882213
OpenEye anion aromatic order: 1.3418572306137813
OpenEye anion aromatic order: 1.3418629033811214
OpenEye anion aromatic order: 1.282103697809885
OpenEye anion aromatic order: 1.2819662464523884
OpenEye anion aromatic order: 1.5524099209448596
OpenEye anion aromatic order: 1.1228566259220674
OpenEye anion aromatic order: 1.1228857329592095
OpenEye anion aromatic order: 1.5523016417471136
```
```
AmberTools neutral aromatic order: 1.42847694
AmberTools neutral aromatic order: 1.42838216
AmberTools neutral aromatic order: 1.38847364
AmberTools neutral aromatic order: 1.38967052
AmberTools neutral aromatic order: 1.38955476
AmberTools neutral aromatic order: 1.36395535
AmberTools neutral aromatic order: 1.38584981
AmberTools neutral aromatic order: 1.42623811
AmberTools neutral aromatic order: 1.3548221
AmberTools neutral aromatic order: 1.3413636
AmberTools neutral aromatic order: 1.44752967
AmberTools anion aromatic order: 1.42697576
AmberTools anion aromatic order: 1.40317476
AmberTools anion aromatic order: 1.40317794
AmberTools anion aromatic order: 1.42697246
AmberTools anion aromatic order: 1.31119999
AmberTools anion aromatic order: 1.31119746
AmberTools anion aromatic order: 1.23414108
AmberTools anion aromatic order: 1.23414195
AmberTools anion aromatic order: 1.61292386
AmberTools anion aromatic order: 1.06733526
AmberTools anion aromatic order: 1.06733486
AmberTools anion aromatic order: 1.61292474
```

jchodera:
It's likely the case that OpenEye `quacpac` uses some "quick and dirty" tolerances and convergence criteria to give a semiquantitative result faster, so this is a perfectly reasonable approach!

docs/releasehistory.rst
```diff
@@ -33,18 +47,18 @@ Behavior changed
 """"""""""""""""
 - `PR #469 <https://github.com/openforcefield/openforcefield/pull/469>`_:
   When running :py:meth:`Topology.to_openmm <openforcefield.topology.Topology.to_openmm>`, unique atom names
-  are generated (overriding any existing atom names) if the provided atom names are not unique. This
+  are generated if the provided atom names are not unique (overriding any existing atom names). This
```
j-wags: Changes to `releasehistory` below here are small grammar/formatting/syntax fixes and are not related to this PR.

openforcefield/utils/toolkits.py
```diff
@@ -2948,6 +2949,161 @@ def compute_partial_charges_am1bcc(self, molecule):
...
+        with temporary_directory() as tmpdir:
+            with temporary_cd(tmpdir):
+                net_charge = molecule.total_charge
+                # Write out molecule in SDF format
+                self._rdkit_toolkit_wrapper.to_file(
+                    molecule, 'molecule.sdf', file_format='sdf')
+                # Prepare sqm.in file as if we were going to run charge calc
+                # TODO: Add error handling if antechamber chokes
+                subprocess.check_output([
+                    "antechamber", "-i", "molecule.sdf", "-fi", "sdf", "-o",
```
jthorton: I haven't used `subprocess.check_output` before so I am not sure if you can get the same effect using `subprocess.run` but this is recommended after python 3.5 https://docs.python.org/3/library/subprocess.html.

openforcefield/utils/toolkits.py
```diff
@@ -2948,6 +2949,161 @@ def compute_partial_charges_am1bcc(self, molecule):
...
+         """
+
+        # Find the path to antechamber
+        # TODO: How should we implement find_executable?
+        ANTECHAMBER_PATH = find_executable("antechamber")
+        if ANTECHAMBER_PATH is None:
+            raise (IOError("Antechamber not found, cannot run "
+                           "AmberToolsToolkitWrapper.assign_fractional_bond_orders()"))
+
+        if len(molecule._conformers) == 0:
```
jthorton: I understand this point but this could be a pain for users, is it not better to assume we need to generate a conformer if one is not supplied and just do this for them?

openforcefield/utils/toolkits.py
```diff
@@ -2948,6 +2949,161 @@ def compute_partial_charges_am1bcc(self, molecule):
...
+            raise ValueError(
+                "No conformers present in molecule submitted for fractional bond order calculation. Consider "
+                "loading the molecule from a file with geometry already present or running "
+                "molecule.generate_conformers() before calling molecule.assign_fractional_bond_orders"
+            )
+        if len(molecule._conformers) > 1:
+            logger.warning(f"Warning: In AmberToolsToolkitWrapper.assign_fractional_bond_orders: "
```
jthorton: how slow would it be to generate the bond order for all of the supplied conformers and average over them in some way? Could this even be done in parallel?

openforcefield/utils/toolkits.py
```diff
@@ -2948,6 +2949,161 @@ def compute_partial_charges_am1bcc(self, molecule):
...
+
+
+        # Compute bond orders
+        bond_order_model_to_antechamber_keyword = {'am1-wiberg': 'mul'}
+        supported_bond_order_models = list(bond_order_model_to_antechamber_keyword.keys())
+        bond_order_model = bond_order_model.lower()
+        if bond_order_model is None:
+            bond_order_model = 'am1-wiberg'
+        if bond_order_model not in supported_bond_order_models:
+            raise ValueError(f"Bond order model '{bond_order_model}' is not supported by AmberToolsToolkitWrapper. "
```
jthorton: Nice, this is a very clear error message!

openforcefield/utils/toolkits.py
```diff
@@ -2948,6 +2949,161 @@ def compute_partial_charges_am1bcc(self, molecule):
...
+
+                # Run sqm to get bond orders
+                subprocess.check_output(["sqm", "-i", "sqm.in", "-o", "sqm.out", "-O"])
+
+                bond_orders = self._get_fractional_bond_orders_from_sqm_out('sqm.out')
+
+        # Note that sqm calculate WBOs for ALL PAIRS of atoms, not just those that have
+        # bonds defined in the original molecule. So here we iterate over the bonds in
+        # the original molecule and only nab the WBOs for those.
+        for bond in molecule.bonds:
```
jthorton: Is it possible this could go wrong and we pull out a bond order for the wrong pair of atoms due to file reordering somewhere in the workflow? Should we compare atom symbols as well to make sure the bond is correct?

jthorton:
Overall looks really good, with great test coverage! One thing to check is subprocess I think could get the same result using run with `check=True`.

openforcefield/utils/toolkits.py
j-wags: @dgasmith recently recommended using `check_output`, but that was in the context of replacing several `os.system` calls. 

The big difference between `run` and `check_output` [1] according to the docs is that `check_output` operates like `run`, except the keyword argument `check=True` is always set. This is good for our use case, because if some external dependency throws an error on the command line, we should raise an exception for the user [2]. I'm happy to standardize on either solution, but right now it will save us a few characters to use `check_output` everywhere.

[1] https://docs.python.org/3/library/subprocess.html#subprocess.check_output
[2] Except I've heard tales that tleap is shady about this -- apparently it can raise a non-zero exit value, but finish successfully

openforcefield/utils/toolkits.py
j-wags: Good call, and actually this parallels an API change that we'll make with `compute_partial_charges`. I'll make this change.

openforcefield/utils/toolkits.py
j-wags: I think conformer averaging would be a nice thing to do. But, like ELF10, it's not clear that there's a single common-sense way to do it. So, I'd like to leave conformer averaging algorithms as something for the community to experiment with. Once someone wants to try it, they can invent a new keyword for `bond_order_model`, and implement support for it in a ToolkitWrapper.

openforcefield/utils/toolkits.py
j-wags: > Is it possible this could go wrong and we pull out a bond order for the wrong pair of atoms due to file reordering somewhere in the workflow

Yes. 

Right now, `AmberToolsToolkitWrapper.assign_fractional_bond_orders` and `compute_partial_charges` assume that the atom indexing is not modified by antechamber. I don't know if this is guaranteed everywhere, but nobody has reported a bug so far. The toolkit makes several assumptions like this, and beyond our "throw a lot of molecules through it, occasionally reversing the atom ordering, and see if it's alright" tests, I doubt even the AMBER manual says anything about this. 

So, this is something that occasionally keeps me up at night, but other than sifting through a bunch of fortran, I'm not sure that we can really prove this assumption. 

Long story short, "added in 43377fe "

openforcefield/utils/toolkits.py
j-wags: Addresses in 8bd7fec

j-wags:
Thanks for the quick review, @jthorton. I've addressed your comments and will merge when tests pass

openforcefield/utils/toolkits.py
davidlmobley: Definitely seems like a thing to do separately/unresolved science.

PR #509, opened 2020-02-09
Fix 491

### LINKED ISSUE #491: Maximum Smarts Matches for Larger Molecules
**Describe the bug**

Currently both the RDKit and OpenEye toolkit backends cap the maximum number of matches which can be found for a given smarts pattern to [1000](https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.html#rdkit.Chem.rdchem.Mol.GetSubstructMatches) and [1024](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OESubSearch.html#OEChem::OESubSearch::GetMaxMatches) respectively (relevant to the toolkit `_find_smarts_matches` methods).

When trying to assign parameters to larger molecules such as (bio)polymers this leads to many `UnassignedXParameterException` exceptions being raised for even simple things like an aromatic torsion, but only for valence terms between atoms with higher indices.

In the case of OpenEye a hard to spot `Warning: default number of matches limit reached in substructure search` warning is given during parameter assignment, while RDKit gives no warning.

**Possible solutions**

Both RDKit and OpenEye offer ways to change this cap. In the [OpenEye toolkit](https://github.com/openforcefield/openforcefield/blob/97d3602e6306d546c3a5d35283ae3141e6e6de81/openforcefield/utils/toolkits.py#L1415) the cap can be raised using the OESubSearch.SetMaxMatches method, and in the [RDKit toolkit](https://github.com/openforcefield/openforcefield/blob/97d3602e6306d546c3a5d35283ae3141e6e6de81/openforcefield/utils/toolkits.py#L2248) by passing a different maximum to `GetSubstructMatches`.

It would be good to expose an option for setting this value on `ToolkitWrapper` classes which implement a `find_smarts_matches` method, and possibly increase the default value for the cap.

trevorgokey:

This turns off the max search limit in the toolkits so that parameters are assigned to the entire molecule rather than terminating early. This PR tests the ability to
- Parse a large molecule, in this case a C-C bond in an n-alkane where n=3000
- Create an OpenMM system of T4-protein.mol2, which is in the data/proteins folder

An issue that this brings (#510) is that after loading the protein, the formal charge is believed to be -55, although the partial charges in the mol2 sum to 8. This check is skipped in OpenMM, and is not addressed in this PR.

Closes #491


jchodera:
@trevorgokey : Thanks for tackling this, and for adding the test! 

Why would you need to use `allow_nonintegral_charges` if the charge is integral but nonzero? I imagine it's the case that whatever charge sum error is causing issues is resulting in a net charge that is not close to the integer -55, which is the real issue in requiring `allow_nonintegral_charges=True`.

Is there any danger in removing the bound on the maximum number of matches for all cases?

Finally, out of curiosity, how slow is it in applying parameters to an entire protein? I think we'll need to look into how badly this scales with biopolymer size.

trevorgokey:
@jchodera Good to hear that it seems ok; I tried to write tests that match the others as closely as possible.

The `allow_nonintegral_charges` issue had to be addressed because OpenMM was raising an exception, and causing the tests to fail. The error is not really that the partial charges don't sum to an integer, but that they don't match the formal charge. I think the error is misleading in this case. You can see the output of this error in #510.

And as for the `allow_undefined_stereo` issue, I just bypassed it since it was not really part of the tests. With the updates described below, this can be set to `False` again since it checks out now.

I used both my own old amber14 installation, and ambermini from my conda env (16.16.0) and I get the same output. The bond orders are all 1. That said, plugging this into OE works and gets me 'am' and 'ar' bond orders. I inserted the charges I got from ff14SB and reran the tests with sanity checks turned back on, and they passed. 

I want to say `label_parameters` takes about ~2 minutes, and `create_openmm_system` about ~5 minutes for this protein with 2634 atoms.

Thanks for the catch, I was weary using/saving the mol2 file, and glad you had the eyes to quickly catch what was wrong.

jchodera:
Thanks for the timing info! It's clear we will need to start another issue to talk about how to make biopolymer parameterizarion fast enough to be usable with openff-1.0.0. If you have a chance to do some timings with the RDKit and OpenEye backends and open a new issue about addressing the speed problems, that would be awesome!

openforcefield/tests/test_forcefield.py
```diff
@@ -1940,6 +1940,37 @@ def test_molecule_energy_gb_no_sa(self, zero_charges, gbsa_model):
...
+        assert len(labels["Angles"]) ==           4789
+        assert len(labels["ProperTorsions"]) ==   6973
+        assert len(labels["ImproperTorsions"]) == 528
+
+        # Need to set allow_nonintegral_charges=True in create_openmm_system
+        # since it believes the formal charge is -55
+        # and the sum of partial charges is ~8.0
+        # Happens because all bond orders in the mol2 file were 1 (made from tleap), 
+        # causing the errors in formal charge calculation (#512)
+        # Fixed using OE to make the mol2, and inserting ff14SB charges
```
j-wags: The first part of this comment is no longer applicable -- Could you reword it?

davidlmobley:
@trevorgokey looks like you are almost done here, just needs to have a couple changes and resolve conflicts.

PR #520, opened 2020-02-15
Mark protein param test as slow and shorten C3000 test

j-wags:
Following #491 

The new protein parameterization test is slow (like, >5 minutes), so I'm marking it as `slow`.  The C3000 also takes a while, so I'm changing it down to C600, which should still ensure that the max-SMARTS-matches increase is working.

### DIFF
--- a/openforcefield/tests/test_forcefield.py
+++ b/openforcefield/tests/test_forcefield.py
@@ -1973,6 +1973,7 @@ class TestForceFieldParameterAssignment:
         # Ensure that all system energies are the same
         compare_system_energies(off_omm_system, amber_omm_system, positions, by_force_type=False)
 
+    @pytest.mark.slow
     @pytest.mark.skipif(not OpenEyeToolkitWrapper.is_available(),
                         reason='Test requires OE toolkit to read mol2 files')
     @pytest.mark.parametrize("toolkit_registry,registry_description", toolkit_registries)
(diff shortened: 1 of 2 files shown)


jchodera:
> The new protein parameterization test is slow (like, >5 minutes)

Do we have any ideas for performance improvements?

j-wags:
> Do we have any ideas for performance improvements?

We may see a speedup by treating polymers like `Topology`s and residues like `TopologyMolecule`s -- Basically, we could provide a `delimiter_smarts` (eg a peptide bond SMARTS) and then "fragment" the protein into a few unique residues. We may hit some complications at the residue junctions, which may apply some restrictions on the SMARTS grammar we could use. But if could just parametrize the unique residues, we could save time on the SMARTS searching and "unroll" the parameters over every instance of the residue rapidly.

jchodera:
> We may see a speedup by treating polymers like Topologys and residues like TopologyMolecules -- Basically, we could provide a delimiter_smarts (eg a peptide bond SMARTS) and then "fragment" the protein into a few unique residues. We may hit some complications at the residue junctions, which may apply some restrictions on the SMARTS grammar we could use. But if could just parametrize the unique residues, we could save time on the SMARTS searching and "unroll" the parameters over every instance of the residue rapidly.

Let's find a time to talk about this concept.

PR #521, opened 2020-02-18
From to InChI

### LINKED ISSUE #474: Add InChI Support
It would be useful for the toolkit to support creating `Molecule` objects from [InChI](https://iupac.org/who-we-are/divisions/division-details/inchi/) strings in addition to SMILES strings. InChI keys appear to be cheminformatic-toolkit independant due to their fixed spec.

Add a new `Molecule.from_inchi(inchi: str) -> Molecule` class method which creates a `Molecule` from an InChI string.

jthorton:
- [X] Fix #474 

This PR adds support to build OFFMols from InChI strings and to get an InChI and InChI key representation of the molecule using both the standard and fixed hydrogen format, these identifiers can then be used in the generation of QCArchive submissions.


jchodera:
This is awesome!

One thing I've noticed is that this code pattern appears many times in toolkit-wrapped functions
```python
@staticmethod
    def from_inchi(inchi, allow_undefined_stereo=False, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY):
        """
        Construct a Molecule from a InChI representation
        Parameters
        ----------
        inchi : str
            The InChI representation of the molecule.
        allow_undefined_stereo : bool, default=False
            Whether to accept InChI with undefined stereochemistry. If False,
            an exception will be raised if a InChI with undefined stereochemistry
            is passed into this function.
        toolkit_registry : openforcefield.utils.toolkits.ToolRegistry or openforcefield.utils.toolkits.ToolkitWrapper, optional, default=None
            :class:`ToolkitRegistry` or :class:`ToolkitWrapper` to use for SMILES-to-molecule conversion
        Returns
        -------
        molecule : openforcefield.topology.Molecule
        Examples
        --------
        make cis-1,2-Dichloroethene
        >>> molecule = Molecule.from_inchi('InChI=1S/C2H2Cl2/c3-1-2-4/h1-2H/b2-1-')
        """

        if isinstance(toolkit_registry, ToolkitRegistry):
            molecule = toolkit_registry.call('from_inchi',
                                             inchi,
                                             allow_undefined_stereo=allow_undefined_stereo)
        elif isinstance(toolkit_registry, ToolkitWrapper):
            toolkit = toolkit_registry
            molecule = toolkit.from_inchi(inchi,
                                          allow_undefined_stereo=allow_undefined_stereo)
        else:
            raise Exception(
                'Invalid toolkit_registry passed to from_inchi. Expected ToolkitRegistry or ToolkitWrapper. Got  {}'

        return molecule

    def to_inchi(self, fixed_hydrogens=False, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY):
        """
        Create an InChI string for the molecule using the requested toolkit backend.
        InChI is a standardised representation that does not capture tautomers unless specified using the fixed hydrogen
        layer.
        For information on InChi see here https://iupac.org/who-we-are/divisions/division-details/inchi/
        Parameters
        ----------
        fixed_hydrogens: bool, default=False
            If a fixed hydrogen layer should be added to the InChI, if `True` this will produce a non standard specific
            InChI string of the molecule.
        toolkit_registry : openforcefield.utils.toolkits.ToolRegistry or openforcefield.utils.toolkits.ToolkitWrapper, optional, default=None
            :class:`ToolkitRegistry` or :class:`ToolkitWrapper` to use for SMILES-to-molecule conversion
        Returns
        --------
        inchi: str
            The InChI string of the molecule.
        Raises
        -------
        InvalidToolkitError
             If an invalid object is passed as the toolkit_registry parameter
        """

        if isinstance(toolkit_registry, ToolkitRegistry):
            inchi = toolkit_registry.call('to_inchi',
                                          self,
                                          fixed_hydrogens=fixed_hydrogens)
        elif isinstance(toolkit_registry, ToolkitWrapper):
            toolkit = toolkit_registry
            inchi = toolkit.to_inchi(self,
                                     fixed_hydrogens=fixed_hydrogens)
        else:
            raise InvalidToolkitError(
                'Invalid toolkit_registry passed to to_inchi. Expected ToolkitRegistry or ToolkitWrapper. Got  {}'

        return inchi
```
I wonder if we could instead create a decorator or other metaprogramming scheme that would allow us to specify which methods should automatically have `to_X()`, `from_X()` methods filled in. For example, if we could just add a `TOOLKIT_DISPATCH_METHODS = ['inchi', 'smiles', ...]` and have the rest filled in automatically.

jthorton:
Thanks! 

Yeah I agree there is a lot of repeated code in these toolkit functions and I like this idea of a decorator here would be my take on something like this 

```python
# test wrapper function
def handle_via_toolkit(orig_func):
    """Check if the function can be performed by the toolkit or not"""

    @wraps(orig_func)
    def wrapper(*args, **kwargs):
        toolkit_functions = ['to_smiles', 'to_inchi']
        if orig_func.__name__ in toolkit_functions:
            toolkit_registry = kwargs.get('toolkit_registry', GLOBAL_TOOLKIT_REGISTRY)

            if isinstance(toolkit_registry, ToolkitRegistry):
                data = toolkit_registry.call(orig_func.__name__, *args, **kwargs)

            elif isinstance(toolkit_registry, ToolkitWrapper):
                function = getattr(toolkit_registry, orig_func.__name__)
                data = function(*args, **kwargs)
            else:
                raise Exception(
                     f'Invalid toolkit_registry passed to {orig_func.__name__}. Expected ToolkitRegistry or '
                     f'ToolkitWrapper. Got  {type(toolkit_registry)}')

            return data

        else:
            return orig_func(*args, **kwargs)

    return wrapper
``` 
This decorator would then be placed on all functions that are handled via the toolkit back ends and this would allow us to keep the functions in the `Molecule` class and the docstrings but the function could just pass as the logic will be handled by the decorator. We could also have the `toolkit_functions` list automatically populate so we don't have to remember to update the list every time we add a function as well. The nice part of doing it this way is I still get autocomplete to work with the functions in pycharm, the only downside is we lose hints on the function arguments which I am not sure how to get around.  

Also, it looks like the test is failing because of an example I have included which causes openeye atom indexing to skip one when made from InChI. So this line causes an error (https://github.com/openforcefield/openforcefield/blob/5baae2983a181582c0f3715376f4bc0a6bcbd079/openforcefield/utils/toolkits.py#L841) this is due to the `map_atoms` dict being indexed with `oeatom.GetIdx()` as keys and the corresponding `atom_index` from the OFFMol as values. However, when we later access the mapping we enumerate the atoms instead of using the `oeatom.GetIdx()` key that the dict was built with. I have changed this to be consistent just waiting on tests now.

openforcefield/tests/test_molecule.py
```diff
@@ -372,6 +372,59 @@ def test_to_networkx(self, molecule):
...
+        inchi = molecule.to_inchi()
+        non_standard = molecule.to_inchi(True)
+
+    @pytest.mark.parametrize('molecule', mini_drug_bank())
+    def test_to_inchikey(self, molecule):
+        """Test the conversion to standard and non-standard InChIKey"""
+        inchikey = molecule.to_inchikey()
+        non_standard_key = molecule.to_inchikey(True)
+
+    def test_from_bad_inchi(self):
```
j-wags: Could you move this test to `test_toolkits` and parametrize it to run separately on RDKit and Openeye?

openforcefield/tests/test_molecule.py
```diff
@@ -372,6 +372,59 @@ def test_to_networkx(self, molecule):
...
+                  {'molecule': create_reversed_ethanol(), 'standard_inchi': 'InChI=1S/C2H6O/c1-2-3/h3H,2H2,1H3',
+                   'fixed_hydrogen_inchi': 'InChI=1/C2H6O/c1-2-3/h3H,2H2,1H3'},
+                  {'molecule': create_acetaldehyde(), 'standard_inchi': 'InChI=1S/C2H4O/c1-2-3/h2H,1H3',
+                   'fixed_hydrogen_inchi': 'InChI=1/C2H4O/c1-2-3/h2H,1H3'},
+                  {'molecule': create_cyclohexane(), 'standard_inchi': 'InChI=1S/C6H12/c1-2-4-6-5-3-1/h1-6H2',
+                   'fixed_hydrogen_inchi': 'InChI=1/C6H12/c1-2-4-6-5-3-1/h1-6H2'}
+                  ]
+
+    @pytest.mark.parametrize('data', inchi_data)
+    def test_from_inchi(self, data):
```
j-wags: Could you move this test to `test_toolkits` and parametrize it to run separately on RDKit and Openeye?

openforcefield/utils/toolkits.py
```diff
@@ -1040,6 +1041,80 @@ def to_smiles(molecule):
...
+            InChI string of the molecule.
+
+        Returns
+        --------
+        inchi: str
+            The InChI string of the molecule.
+        """
+
+        from openeye import oechem
+        oemol = OpenEyeToolkitWrapper.to_openeye(molecule)
```
j-wags: Per the discussion on "what should be a class method vs. instance method" last week, I think having to write something like this means that this should be an instance method and that this line should be `oemol = self.to_openeye(molecule)`.

I know it's used inconsistently across the toolkits, but starting the transformation to instancemethods now can let us make CachingToolkitWrappers a drop-in replacement later.

openforcefield/utils/toolkits.py
```diff
@@ -2381,6 +2533,70 @@ def to_rdkit(cls, molecule, aromaticity_model=DEFAULT_AROMATICITY_MODEL):
...
+            InChI string of the molecule.
+
+        Returns
+        --------
+        inchi_key: str
+            The InChIKey representation of the molecule.
+        """
+
+        from rdkit import Chem
+        rdmol = cls.to_rdkit(molecule)
```
j-wags: This, too, I think could become an instancemethod, with `self.to_rdkit` in this spot.

j-wags:
I like your idea above, will reply with more substance soon. I won't have time to fully review this today, but I left some requests for changes.

openforcefield/tests/test_molecule.py
```diff
@@ -372,6 +372,59 @@ def test_to_networkx(self, molecule):
...
+        """Test conversion to standard and non-standard InChI"""
+        inchi = molecule.to_inchi()
+        non_standard = molecule.to_inchi(True)
+
+    @pytest.mark.parametrize('molecule', mini_drug_bank())
+    def test_to_inchikey(self, molecule):
+        """Test the conversion to standard and non-standard InChIKey"""
+        inchikey = molecule.to_inchikey()
+        non_standard_key = molecule.to_inchikey(True)
+
```
j-wags: (Blocking) Would we expect a OFFMol to survive an InChI roundtrip? The interaction of our chosen aromaticity model/kekulization scheme and InChI's information content is complicated, and we'd need something like a mathematical proof to know whether were safe here. But, I'm not that smart, so if we DO expect an OFFMol to survive an InChI roundtip, **we should make roundtrip tests using MiniDrugBank for both toolkits here**. Our standard for "survive a roundtrip" could just be `is_isomorphic_with`.

openforcefield/tests/test_molecule.py
jthorton: Good point ill make sure these are all moved to cover RDKit as well.

openforcefield/utils/toolkits.py
jthorton: Yeah I think you're right this is quite awkward to write this way, Ill change these over.

openforcefield/tests/test_molecule.py
jthorton: Good point I think from reading the spec of standard InChI we would expect some information to be lost. This has been extensively tested by Openeye [here](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEMolToInChI.html#OEChem::OEMolToInChI) and in the round trip testing, I have used non-standard InChi with a fixed hydrogen layer which seems to give the best results. We probably want to keep an eye on the test as some current issues could be fixed in future if requirements around nitrogen stereochemistry were changed.

jthorton:
The current functionality seems to work well however I would check that I am using the RDKit API correctly by getting an answer to this [issue](https://github.com/rdkit/rdkit/issues/2969) before merging this branch. If anyone has any other InChI tests they would like added please let me know.

jthorton:
Okay, it seems like the RDKit implementation is correct. This is ready for a review.

openforcefield/tests/test_toolkits.py
```diff
@@ -18,14 +18,66 @@
...
+rdkit_inchi_stereochemistry_lost = ['DrugBank_5414', 'DrugBank_2955', 'DrugBank_5737', 'DrugBank_3332', 'DrugBank_3461',
+                                    'DrugBank_6026', 'DrugBank_3622', 'DrugBank_3726', 'DrugBank_6222', 'DrugBank_3844',
+                                    'DrugBank_6304', 'DrugBank_6305', 'DrugBank_6329', 'DrugBank_6509', 'DrugBank_6647',
+                                    'DrugBank_1897', 'DrugBank_4778', 'DrugBank_2148', 'DrugBank_2178', 'DrugBank_2538',
+                                    'DrugBank_2592', 'DrugBank_4249', 'DrugBank_5076', 'DrugBank_5418', 'DrugBank_3930',
+                                    'DrugBank_1634', 'DrugBank_1962', 'DrugBank_5043', 'DrugBank_2519']
+
+rdkit_inchi_isomorphic_fails = ['DrugBank_178', 'DrugBank_246', 'DrugBank_5847', 'DrugBank_700', 'DrugBank_1564',
+                                'DrugBank_1700', 'DrugBank_4662', 'DrugBank_2052', 'DrugBank_2077', 'DrugBank_2082',
+                                'DrugBank_2210', 'DrugBank_2642']
```
j-wags: Nice! Thanks for going through these, I know it's a pain.

j-wags:
Also, be sure to update `releasehistory.rst`! This InChI stuff is tricky, and we need to have a gigantic warning with flashing lights that roundtrips frequently won't work and that loading from InChI is dangerous, and a big statement about how we support (and maybe even recommend) "nonstandard" inchis. 

Reasonably skilled comp chemists (you and I) didn't know this all a few weeks ago, and so the majority of our userbase won't either.

openforcefield/tests/test_toolkits.py
```diff
@@ -18,14 +18,66 @@
...
+def get_mini_drug_bank(toolkit):
+    """Read the mini drug bank sdf file with the toolkit and return the molecules"""
+
+    # This is a work around a weird error where even though the test is skipted due to a missing toolkit
+    #  we still try and read the file with the toolkit
+    try:
+        molecules = Molecule.from_file(get_data_file_path('molecules/MiniDrugBank.sdf'), 'sdf', toolkit_registry=toolkit,
+                                       allow_undefined_stereo=True)
+    except ModuleNotFoundError:
+        molecules = []
```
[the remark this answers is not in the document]
jthorton: I think we should only fall into this if we try to use a toolkit that is not installed but, I may be able to get around this using the `is_available` method instead.

jthorton:
Thanks for the review Jeff, I have managed to remove the `try/except` block from the new sdf drug bank loader and moved the test to the `test_molecule` section. I totally agree as well that the tests should be clearly divided between the two files, so following with this idea we will want to have every function in the `Molecule` class to be tested at least once in the `test_molecule` file with different args if possible, and then more performance testing (like to from file) in the `test_toolkits` files for the same functions.  

I think the move of the tests from molecule to toolkits/ testing refactor should be a different issue as I expect that we may open up a can of worms here, I'll open an issue so we can plan out exactly how we will divide the testing before we starting changing things.

PR #527, opened 2020-02-21
[WIP] Multiframe xyz

### LINKED ISSUE #523: Writing to MOL error
**Describe the bug**
When trying to write molecules to `MOL` files with RDKit I get an error despite this being one of the supported file formats listed by the wrapper.
**To Reproduce**
```python
from openforcefield.utils.toolkits import RDKitToolkitWrapper

mol.generate_conformers(n_conformers=1)
# try and write to mol file with rdkit
mol.to_file('test.mol', 'mol', RDKitToolkitWrapper())
```
> KeyError: 'MOL'

I think the writer dictionaries [here](https://github.com/openforcefield/openforcefield/blob/5baae2983a181582c0f3715376f4bc0a6bcbd079/openforcefield/utils/toolkits.py#L1789) and [here](https://github.com/openforcefield/openforcefield/blob/5baae2983a181582c0f3715376f4bc0a6bcbd079/openforcefield/utils/toolkits.py#L1820) just need to be updated. It may also be better to store the reference to these objects in one variable to save repetition of code and make updating writers easier for example

```python
from rdkit import Chem
_toolkit_file_write_formats = {'SDF':Chem.SDWriter , 'MOL':Chem.SDWriter , 'SMI':Chem.SmilesWriter , 'PDB':Chem.PDBWriter, 'TDT': Chem.TDTWriter }
```  

the function to write the files could then be something like

```python
def to_file(self, molecule, file_path, file_format):
        """
        Writes an OpenFF Molecule to a file-like object
        Parameters
        ----------
        molecule : an OpenFF Molecule
            The molecule to write
        file_path
            The file path to write to
        file_format
            The format for writing the molecule data
        Returns
        ------
        """
        from rdkit import Chem
        with open(file_path, 'w') as file_obj:
            rdmol = self.to_rdkit(molecule)
            try:
                writer = self._toolkit_file_write_formats[file_format](file_obj)
                writer.write(rdmol)
                writer.close()
            except KeyError:
                rasie InvalidFileTypeError(f'The file type {file_format} is not currently supported to be wrote to by RDKit.')
```

the readers could also be stored this way

```python
from rdkit import Chem
_toolkit_file_read_formats = {'SDF': Chem.SupplierFromFilename, 'MOL': Chem.SupplierFromFilename, 'SMI': Chem.SmilesMolSupplier, 'TDT':Chem.TDTMolSupplier}
```
Not sure how useful this second option is however as depending on the input different actions are performed this would only be simpler if the functions could be unified in some way. 

 - Operating system linux

jthorton:
Here I aim to add support for wrtting multiframe and single frame XYZ files and fixing the MOL file type writer and adding more tests for each of the file output types.
- [ ] Fix #523 


j-wags:
Could you specify a few things about the behavior of OFFTK with respect to XYZ? Namely:
* Will we only ever write conformers of the **same** molecule to a single XYZ file, or could multiple distinct molecules be written to an XYZ?
* Do we expect a molecule's connectivity and/or formal charges to survive a round trip to XYZ?
* Do we expect a molecule's `name` to survive a round trip to XYZ?
* Do we expect a molecule's `partial_charges` to survive a round trip to XYZ?
* Do we expect a molecule's `properties` dict to survive a round trip to XYZ?

Issue #353, opened 2019-06-13
Topology creation is very slow for large system


j-wags:
## Description

System is a box of 1000 hexadecanes.

```
from openforcefield.typing.engines.smirnoff import ForceField
from openforcefield.topology import Molecule, Topology
from simtk.openmm import app
import time

hexadecane = Molecule.from_smiles('CCCCCCCCCCCCCCCC')
pdbfile = app.PDBFile('output.pdb')
start_time = time.time()
top = Topology.from_openmm(pdbfile.topology, unique_molecules=[hexadecane])
print("--- Topology creation finished in %s seconds ---" % (time.time() - start_time))
ff = ForceField('test_forcefields/smirnoff99Frosst.offxml')
```
> --- Topology creation finished in 1585.9495000839233 seconds ---

```
start_time = time.time()

system = ff.create_openmm_system(top)
print("--- Ssytem creation finished in %s seconds ---" % (time.time() - start_time))
```
> --- Ssytem creation finished in 57.06095290184021 seconds ---

[output.pdb.gz](https://github.com/openforcefield/openforcefield/files/3288425/output.pdb.gz)

## Thoughts

I suspect that this slowdown occurs in the graph matching stage of topology creation. Hexadecane has a huge number of self-symmetries. While the function only takes the first molecular topology match that's found, it may be generating all of them unnecessarily.

Relevant code is here: https://github.com/openforcefield/openforcefield/blob/master/openforcefield/topology/topology.py#L1635-L1664

We should poke around NetworkX to see if there's a faster algorithm that this code could use.

jaimergp:
I had to do something similar (identify copies of small molecules in a system) and found that "serializing" all connected molecules to canonical SMILES (as provided by `rdkit`) and comparing those was enough for identification purposes. Would that be enough in this case or am I missing something?

jchodera:
It looks like you also do the expensive graph matching for all molecules, then try to correct for having different mappings afterwards. We really need to do some fast hashing / caching to avoid this expense. 

What if we used an algorithm like this:
* For each Topology molecule, construct a hashable immutable object (like a tuple) that encodes the sequence of elements and bond indices within the molecule:
```python
( (element_1, element_2, ..., element_N), ((bond1_atomindex1, bond1_atomindex2), ..., (bondN_atomindex1, bond1_atomindex2))
```
[...]

PR #529, opened 2020-02-26
Mol fix and xyz

### LINKED ISSUE #523: Writing to MOL error

jthorton:
Here I aim to add support for writing multiframe and single frame XYZ files and fixing the MOL file type writer and adding more tests for each of the file output types. We also have another speed improvement for #353.
- [X] Fix #523 

This is a new branch for the fix as the old one was not from master.


jthorton:
As I did the branch wrong I have copied the comments across from the old PR

> Could you specify a few things about the intended behavior of OFFTK with respect to XYZ? 
 Namely:
> * Will we only ever write conformers of the **same** molecule to a single XYZ file, or could multiple distinct molecules be written to a single XYZ?

My intention was that only one molecule would be written to an XYZ file but multipule conformers could be, however there is a way around this for users by doing something like this:

```python
methane = Molecule.from_smiles('C')
ethane = Molecule.from_smiles('CC')

xyz = open('molecules.xyz', 'w')
methane.to_file(xyz, 'xyz')
xyz = open('molecules.xyz', 'a')
ethane.to_file(xyz, 'xyz')
```
would give :
> 5
CH4
C        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000
8
C2H6
C        0.0000000000    0.0000000000    0.0000000000
C        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000
H        0.0000000000    0.0000000000    0.0000000000

As we will not support the reading of this file type this is not an issue for us but could be for users when interfacing with other programs so am not sure if we should stop this behaviour or not. 

> * Is atom indexing guaranteed to be preserved when writing XYZ?

**Yes** the file is generated based on the current molecule atom ordering, as this is done by the toolkit directly not some back end toolkit the order should never change. 

> * What is the behaviour if `offmol.conformers` is None?

If there are no conformers then we set all of the positions of the atoms to `[0,0,0]` and write out the file copying the behaviour of writing a PDB with no conformers.

 >* Do we expect XYZ to be loadable back into OFFTK in any form? If so...

**No** there is even less information in here than in a PDB file so we should never load from a XYZ file. 

>     * Do we expect a molecule's connectivity and/or formal charges to survive a round trip to XYZ?
>     * Do we expect a molecule's `name` to survive a round trip to XYZ?
>     * Do we expect a molecule's `partial_charges` to survive a round trip to XYZ?
>     * Do we expect a molecule's `properties` dict to survive a round trip to XYZ?

jthorton:
Tests are failing due to the Mol writer fix I have put in, due to the list of file writers being a class level object and we require the import of RDKit to link the file sufix to the file writer in this part:

```python
class RDKitToolkitWrapper(ToolkitWrapper):
    """
    RDKit toolkit wrapper

    .. warning :: This API is experimental and subject to change.
    """

    from rdkit import Chem

    _toolkit_name = 'The RDKit'
    _toolkit_installation_instructions = 'A conda-installable version of the free and open source RDKit cheminformatics ' \
                                         'toolkit can be found at: https://anaconda.org/rdkit/rdkit'
    _toolkit_file_read_formats = ['SDF', 'MOL', 'SMI']  #TODO: Add TDT support
    _toolkit_file_write_formats = {'SDF': Chem.SDWriter, 'MOL': Chem.SDWriter,
                                   'SMI': Chem.SmilesWriter, 'PDB': Chem.PDBWriter, 'TDT': Chem.TDTWriter}

```

any environment that does not have RDKit installed will fail at this point as it is always run. This could be fixed by possibly  having this dict be filled during an `__init__` call  which could import rdkit something like:

```python
class RDKitToolkitWrapper(ToolkitWrapper):

    _toolkit_file_write_formats = {'SDF': None, 'MOL': None,
                                   'SMI': None, 'PDB': None, 'TDT': None}

    def __init__(self):
        try:
            from rdkit import Chem
            self._toolkit_file_write_formats = {'SDF': Chem.SDWriter, 'MOL': Chem.SDWriter,
                                                'SMI': Chem.SmilesWriter, 'PDB': Chem.PDBWriter, 'TDT': Chem.TDTWriter}
        except ImportError:
            pass
```

We then just have to make sure we don't get an `AttributeError` if the dict still has None in it..

openforcefield/utils/toolkits.py
```diff
@@ -1523,11 +1526,24 @@ class RDKitToolkitWrapper(ToolkitWrapper):
...
     """
+
     _toolkit_name = 'The RDKit'
     _toolkit_installation_instructions = 'A conda-installable version of the free and open source RDKit cheminformatics ' \
                                          'toolkit can be found at: https://anaconda.org/rdkit/rdkit'
     _toolkit_file_read_formats = ['SDF', 'MOL', 'SMI']  #TODO: Add TDT support
-    _toolkit_file_write_formats = ['SDF', 'MOL', 'SMI', 'PDB']
+
+    # Note any new file write formats should be added here with a string of class/method used
+    _toolkit_file_write_formats = {'SDF': 'Chem.SDWriter', 'MOL': 'Chem.SDWriter', 'SMI': 'Chem.SmilesWriter',
```
jthorton: This list now only has string representations of the functions which can later be called using eval, so when someone updates the writers to add more functionality in future they only have to update this dictionary for RDKit.

openforcefield/utils/toolkits.py
```diff
@@ -1784,17 +1799,16 @@ def to_file_obj(self, molecule, file_obj, file_format):
...
-            'SDF': Chem.SDWriter,
-            'PDB': Chem.PDBWriter,
-            'SMI': Chem.SmilesWriter,
-            'TDT': Chem.TDTWriter
-        }
-        writer = rdkit_writers[file_format](file_obj)
-        writer.write(rdmol)
-        writer.close()
+        try:
+            writer = eval(self._toolkit_file_write_formats[file_format])(file_obj)
```
jthorton: Here we call `eval` on the string representation of the file writer class/function.

j-wags:
Excellent.

> so am not sure if we should stop this behaviour or not.

I don't think we need to police the one-molecule-per-XYZ behavior if a user opens a file in append mode. If that messes up their workflow, it's their own fault :-) 

I agree with all of your answers on the behavior and think this will be a great addition. Thanks, Josh!

j-wags:
> Tests are failing due to the Mol writer fix I have put in, due to the list of file writers being a class level object

I wonder if these should be instance-level? I've (occasionally) debugged problems by making two RDKitToolkitWrappers, and overridden the behavior of one of them by redefining its attributes. But it gets annoying when that attribute is class-level, since then it affects the other one as well. 

Is there a good reason to keep the read and write formats at the class level? If not, I'd be fine with defining them on an instance-level (in `__init__`) and avoiding all this `eval` business.

jthorton:
Yeah, that should be fine as we can do the import in the init, would we want to change this behaviour for both of the toolkit wrappers to make them consistent?

jthorton:
Just need one final check that I have RDKit file writers set up correctly so that Mol and SDF write the same file format.  Also, do we need more tests for each of the file writing formats similar to the XYZ testing I have put in?

docs/releasehistory.rst
```diff
@@ -54,6 +54,9 @@ New features
          tests will fail when using this format due to a loss of information. We have also added support for fixed
          hydrogen layer nonstandard InChI which can help in the case of tautomers, but overall creating molecules from InChI should be
          avoided.
+- `PR #529 <https://github.com/openforcefield/openforcefield/pull/529>`_: Adds the ability to write out to XYZ files via
+  :py:meth: `Molecule.to_file <openforcefield.topology.Molecule.to_file>` both single frame and multiframe files are supported.
```
j-wags:
```suggestion
  :py:meth: `Molecule.to_file <openforcefield.topology.Molecule.to_file>`. Both single frame and multiframe XYZ files are supported.
```

openforcefield/tests/test_toolkits.py
```diff
@@ -38,6 +38,7 @@ def get_mini_drug_bank(toolkit):
     # This is a work around a weird error where even though the test is skipted due to a missing toolkit
     #  we still try and read the file with the toolkit
     if toolkit.is_available():
+        toolkit = toolkit()
```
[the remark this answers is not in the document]
jthorton: Yeah no problem!

openforcefield/topology/molecule.py
```diff
@@ -3454,6 +3448,51 @@ def from_file(file_path,
 
         return mols
 
+    def _to_xyz_file(self, file_path):
+        """
+        Write the current molecule and its conformers to a multiframe xyz file, if the molecule
+        has no current coordinates all atoms will be set to 0,0,0 in keeping with the behaviour of the
+        backend toolkits.
+
```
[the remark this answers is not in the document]
jthorton: Good idea, with there being no official spec as far as I know this is really helpful.

jthorton:
Thanks for the feedback Jeff I have made the changes you suggested and will merge once tests pass.

PR #535, opened 2020-03-04
Cmiles options

### LINKED ISSUE #534: Expand to_smiles API
In order to produce more [`Cmiles`](https://github.com/openforcefield/cmiles/blob/b41a8d8f188a988918cd2ee34c4dbdc44a18b0f5/cmiles/generator.py#L16)  identifiers, we need to expand the `to_smiles` API to allow for extra arguments which define the flavour of the smiles returned. Such as `isomeric`, `explicit_hydrogen` and `mapped`.

These options would allow the creation of the following variants 
```python
canonical_smiles = mol.to_smiles(isomeric=False, explicit_hydrogen=False, mapped=False)

canonical_isomeric_smiles = mol.to_smiles(isomeric=True, explicit_hydrogen=False, mapped=False)

canonical_explicit_hydrogen_smiles = mol.to_smiles(isomeric=False, explicit_hydrogen=True, mapped=False)

# default 
canonical_isomeric_explicit_hydrogen_smiles = mol.to_smiles(isomeric=True, explicit_hydrogen=True, mapped=False)

canonical_isomeric_explicit_hydrogen_mapped_smiles  = mol.to_smiles(isomeric=True, explicit_hydrogen=True, mapped=True)
```
are there any other options which we should include at this point?

jthorton:
- [X] Fix #534 

This PR will expand the `to_smiles` API options to allow the creation of cmiles identifiers.


jthorton:
In order to be able to generate the unique index for a molecule who will be submitted for torsion driving, I also need to add a way to generate a mapped smile for just the atoms involved in the torsion as the current method will only generate the mapped smiles for the whole molecule.

jthorton:
In testing with the automated qcarchive submission pipeline, I found that the canonical ordering would often fail on molecules with undefined stereo even after users have allowed this so I have also fixed this issue. 

I have also removed the automatic ordering of a molecule when using the `to_qcschema` method as this makes it difficult to generate the cmiles information in the correct order. Now the method just gives a schema representation of the exact molecule it is called on, to get the molecule in an canonical order for schema the canonical order method must be called first like in the follwoing example which will be part of the automated submission pipeline:

```python
ethanol = Molecule.from_smiles('CO')
ethanol.generate_conformers()

# canonical order the atoms
ordered_ethanol = ethanol.canonical_order_atoms()

# now make the schema and the cmiles tags for this ordering
schema = ordered_ethanol.to_qcschema()
mapped_hydrogen_smiles = ordered_ethanol.to_smiles(mapped=True)
```

openforcefield/tests/test_molecule.py
```diff
@@ -303,16 +303,126 @@ def test_to_from_smiles(self, molecule, toolkit):
...
+            toolkit = toolkit_class()
+            # this uses no toolkit back end so no smiles should be saved
+            mol = create_ethanol()
+
+            # now lets populate the cache with a test result
+            # first we need to make the cache key for the default input
+            isomeric, explicit_hydrogens, mapped = True, True, False
+            cache_key = toolkit.to_smiles.__qualname__ + str(isomeric) + str(explicit_hydrogens) + str(mapped)
+            cache_key += str(mol._properties.get('atom_map', None))
+            mol._cached_smiles = {cache_key: None}
```
j-wags: I don't _love_ that we're reaching down into private attributes here, but I can't think of any other way to run this test, and I think it's good that you thought to test it!

openforcefield/topology/molecule.py
```diff
@@ -1826,14 +1826,21 @@ def __eq__(self, other):
...
 
         Parameters
         ----------
+        isomeric: bool optional, default= True
+            return an isometric smiles, requires all stereochemistry to be defined
+        explicit_hydrogens: bool optional, default=True
+            return a smiles string containing all hydrogens explicitly
+        mapped: bool optional, default=False
+            return a explicit hydrogen mapped smiles, the atoms to be mapped can be controlled by supplying an
+            atom map into the properties dictionary
```
j-wags: Same comment as above about providing more info about how `offmol.properties['atom_map']` is handled -- The same docstring can just be copied into here.

openforcefield/topology/molecule.py
```diff
@@ -1868,14 +1875,14 @@ def to_smiles(self, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY):
         # Get a string representation of the function containing the toolkit name so we can check
         # if a SMILES was already cached for this molecule. This will return, for example
         # "RDKitToolkitWrapper.to_smiles"
-        func_qualname = to_smiles_method.__qualname__
-
+        smiles_hash = to_smiles_method.__qualname__ + str(isomeric) + str(explicit_hydrogens) + str(mapped)
```
j-wags: Clever!

openforcefield/utils/toolkits.py
```diff
@@ -1045,11 +1050,51 @@ def to_smiles(molecule):
...
-        oemol = OpenEyeToolkitWrapper.to_openeye(molecule)
-        smiles = oechem.OECreateSmiString(
-            oemol, oechem.OESMILESFlag_DEFAULT | oechem.OESMILESFlag_Hydrogens
-            | oechem.OESMILESFlag_Isotopes | oechem.OESMILESFlag_BondStereo
-            | oechem.OESMILESFlag_AtomStereo)
+        oemol = self.to_openeye(molecule)
+
+        # this sets up the default settings following the old DEFAULT flag
+        # more information on flags can be found here
+        # <https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemConstants/OESMILESFlag.html#OEChem::OESMILESFlag>
```
j-wags: Huzzah! Thanks for providing a docs link

openforcefield/utils/toolkits.py
```diff
@@ -1992,24 +2037,60 @@ def canonical_order_atoms(self, molecule):
...
         Parameters
         ----------
         molecule : An openforcefield.topology.Molecule
             The molecule to convert into a SMILES.
+        isomeric: bool optional, default= True
+            return an isometric smiles, requires all stereochemistry to be defined
+        explicit_hydrogens: bool optional, default=True
+            return a smiles string containing all hydrogens explicitly
+        mapped: bool optional, default=False
+            return a explicit hydrogen mapped smiles
```
j-wags: Copy extended docstring for map handling here.

openforcefield/tests/test_molecule.py
```diff
@@ -303,16 +303,126 @@ def test_to_from_smiles(self, molecule, toolkit):
         undefined_stereo = molecule.name in undefined_stereo_mols
 
         smiles1 = molecule.to_smiles(toolkit_registry=toolkit_wrapper)
-        if undefined_stereo:
-            molecule2 = Molecule.from_smiles(smiles1,
-                                             allow_undefined_stereo=True,
-                                             toolkit_registry=toolkit_wrapper)
-        else:
-            molecule2 = Molecule.from_smiles(smiles1,
-                                             toolkit_registry=toolkit_wrapper)
```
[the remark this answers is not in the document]
jthorton: No problem changed back.

openforcefield/tests/test_molecule.py
```diff
@@ -303,16 +303,126 @@ def test_to_from_smiles(self, molecule, toolkit):
...
+                    ]
+
+    @pytest.mark.parametrize('toolkit_class', [OpenEyeToolkitWrapper, RDKitToolkitWrapper])
+    @pytest.mark.parametrize('data', smiles_types)
+    def test_smiles_types(self, data, toolkit_class):
+        """Test that the toolkit is passing the correct args to the toolkit backends across different combinations."""
+
+        if toolkit_class.is_available():
+            toolkit = toolkit_class()
+            mol = Molecule.from_smiles(data['molecule_input'], toolkit_registry=toolkit)
```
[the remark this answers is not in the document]
jthorton: I understand what you mean but I am not sure that would fix the problem as if the canonicalisation method changes all tests will fail regardless of input ordering and the expected results will have to be updated., I think this is always going to be a problem when explicitly testing the output smiles strings. I think with a little more work I can be cleaver about the testing and get  around this Ill draft it up and see what you think.

openforcefield/utils/toolkits.py
```diff
@@ -1045,11 +1050,51 @@ def to_smiles(molecule):
...
+
+            if atom_map is None:
+                # now we need to add the atom map to the atoms
+                for oeatom in oemol.GetAtoms():
+                    oeatom.SetMapIdx(oeatom.GetIdx() + 1)
+            else:
+                for atom in oemol.GetAtoms():
+                    try:
+                        # try to set the atom map
+                        map_idx = atom_map[atom.GetIdx()] + 1
```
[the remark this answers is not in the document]
jthorton: Yeah this is a strange one here I am assuming our maps start from a 0 index which is not valid in the backend toolkits so I have to increment them, I can add a simple check to see if the user has started from 1 or 0 I think.

openforcefield/tests/test_molecule.py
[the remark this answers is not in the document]
j-wags: I like your new solution. API-produced molecules are ugly to implement, but will save us a ton of time in the long run.

jthorton:
Thanks Jeff Ill get it merged in and yeah I know what you mean that is an awkward point, for now, they should be able to start from 0 or 1 but if we finish the spec and say 1 only we can make that change. My other thought is that to be consistent with the atom map returned from `Molecule.are_isomorphic` which starts from 0 so if we do change we should change all of these functions to make sure they take the same types of atom maps.

PR #543, opened 2020-03-12
Enumerating states (tautomers, stereo)

### LINKED ISSUE #526: Enumerating states.
During the QCArchive submission framework, we plan on enumerating several states and it would be nice to have API points that can easily allow users to expand states such as tautomers, stereoisomers and protonation states. 

I am not too sure on what would be best here either one general state enumerator that takes an argument corresponding to the sate or a separate API point for each state we require. 
so 
```python
# options would be stereoisomers, tautomers, protonation?
molecules = molecule.enumerate_state(state='stereoisomers')
```

or 

```python
molecules = molecule.enumerate_stereoisomers()
```

After a quick look into this, I have found that RDKit can do 
- [stereoisomers](https://www.rdkit.org/docs/source/rdkit.Chem.EnumerateStereoisomers.html) 
- [tautomers](https://rdkit.blogspot.com/2020/01/trying-out-new-tautomer.html) are coming

I did not find any info for protonation though someone else may know?

Openeye I believe can already do this and tautomer and stereoisomer expansion has already been done in fragmenter 
- [steroisomers and tautomers](https://github.com/openforcefield/fragmenter/blob/master/fragmenter/states.py) fragmenter link
- [protonation](https://docs.eyesopen.com/toolkits/python/quacpactk/OEProtonFunctions/OEGetReasonableProtomers.html?highlight=oegetreasonabletautomers) openeye docs

tagging @jchodera and @ChayaSt

jthorton:
- [X] Partial-Fix #526 

This is my initial go at adding support for stereo and tautomer enumeration, so far only openeye can do both and rdkit can do stereo as we are waiting for the next rdkit release to add tautomers.

The performance between the two toolkits is different as expected.


jthorton:
This should now allow users to enumerate stereo centres and bonds as well as tautomers using the toolkit. Protomers are not yet supported until available in RDKit.

jthorton:
Not sure I understand how any of these RDKit builds are passing yet as they should all fail while pulling from conda-forge until the new version is available which should be soon see here https://github.com/conda-forge/rdkit-feedstock/pull/44.

jthorton:
I think the new version of RDKit is now failing a molecule that used to pass the round trip test.

jthorton:
Fixing the smiles round trip test the causes this test to fail `test_to_from_rdkit` as the undefined stereo chem error is not raised 👎

openforcefield/tests/test_molecule.py
```diff
@@ -1053,6 +1053,173 @@ def test_wrong_index_mapping(self):
...
+
+        # now make sure we can generate them all
+        protomers = mol.enumerate_formalcharges(max_states=10)
+
+        assert mol not in protomers
+        assert len(protomers) == 3
+
+        # make sure the molecules are different
+        for protomer in protomers:
+            assert mol != protomer
```
mattwthompson, line 1137: Isn't this handled by `assert mol not in protomers` above?

openforcefield/tests/test_molecule.py
```diff
@@ -1053,6 +1053,173 @@ def test_wrong_index_mapping(self):
...
+            assert len(isomers) == 2
+
+            assert mol not in isomers
+            # make sure the input molecule is only different by bond stereo
+            for ismol in isomers:
+                assert Molecule.are_isomorphic(mol, ismol, return_atom_map=False, bond_stereochemistry_matching=False)[0] is True
+                assert mol.is_isomorphic_with(ismol) is False
+
+            # make sure the isomers are different
+            assert Molecule.are_isomorphic(isomers[0], isomers[1])[0] is False
```
mattwthompson, line 1158: Just a little cleaner
```suggestion
            assert isomers[0].is_isomorphic_with(isomers[1]) is False
```

openforcefield/tests/test_molecule.py
```diff
@@ -1053,6 +1053,173 @@ def test_wrong_index_mapping(self):
...
+            assert len(isomers) == 2
+            # make sure the mol is not in the isomers and that they only differ by stereo chem
+            assert mol not in isomers
+            for ismol in isomers:
+                assert ismol.n_conformers != 0
+                assert Molecule.are_isomorphic(mol, ismol, return_atom_map=False, atom_stereochemistry_matching=False)[0] is True
+                assert mol.is_isomorphic_with(ismol) is False
+
+            # make sure the two isomers are different
+            assert Molecule.are_isomorphic(isomers[0], isomers[1])[0] is False
```
mattwthompson, line 1182: Same as above

openforcefield/topology/molecule.py
```diff
@@ -3580,6 +3580,111 @@ def to_file(self,
...
+
+        else:
+            raise ValueError(
+                "'toolkit_registry' must be either a ToolkitRegistry or a ToolkitWrapper"
+            )
+
+        return molecules
+
+    @OpenEyeToolkitWrapper.requires_toolkit()
+    def enumerate_formalcharges(self, max_states=10):
```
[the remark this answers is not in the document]
jthorton: Yeah, I wasn't sure what to go with in this case, but promoters are probably better.

openforcefield/tests/test_molecule.py
jthorton: Ahh good catch I think I was supposed to be checking that each protomer in the list was unique, I have updated the test now.

jthorton:
Thanks for the review @mattwthompson  and same I am not as confident on protomers so if anyone else wants to check that would be great.

jthorton:
no new reviews so going to merge.

PR #548, opened 2020-03-24
VirtualSite support

### LINKED ISSUE #651: Orphaned TopologyParticle?
There is a `TopologyMolecule.particle` method that attempts to return instances of `TopologyParticle`, which is never defined in the current codebase. ([Not tested], and that's probably for the better)

```
(openff-dev) [openforcefield] grep -R "class Topology" .                                                                                                       ci-lint  ✭ ✱
./openforcefield/topology/topology.py:class TopologyAtom(Serializable):
./openforcefield/topology/topology.py:class TopologyBond(Serializable):
./openforcefield/topology/topology.py:class TopologyVirtualSite(Serializable):
./openforcefield/topology/topology.py:class TopologyMolecule:
./openforcefield/topology/topology.py:class Topology(Serializable):
(openff-dev) [openforcefield] grep -R "class TopologyParticle" .                                                                                               ci-lint  ✭ ✱
```

My other guess, that it's something @trevorgokey is working on in https://github.com/openforcefield/openforcefield/pull/548, didn't bear any fruit. 

There's some history (#233, #235) so I figure it may have just slipped through the cracks? I will remove this if there are not immediate plans for it.

trevorgokey:
### Checklists

- [x] Update [the SMIRNOFF specification](https://github.com/openforcefield/openforcefield/blob/master/The-SMIRNOFF-force-field-format.md)

- [x] Implement/parse the VirtualSite offxml specification
- [x] Implement create_force

- Fixes #651

~~This PR is very experimental and can change without warning.~~
Edit: This PR is near completion, so functionality will likely not be changed when merged.

- [x] Ready to go!

### Old/Outdated notes:

<details>
<summary> Limitations noticed:</summary>

It seems to not be possible to apply two virtual sites using the same virtual site type and smirks combination. For example, my cursory testing shows that, for a triplet of atoms (e.g. a water molecule), it is only possible to apply exactly one DivalentLonePair virtual site. This limits the ability to add multiple virtual sites on a single pattern.
</details>

<details>
<summary>Internal representation:</summary>

The current implementation is represented internally, and to the user, as the following. This is applying random VirtualSites to a water molecule, and below is what what is returned by `label_molecules` for VirtualSites only:
```
*** VirtualSites
{'store': OrderedDict([((0, 1),
                        <VirtualSiteBondChargeType with smirks: [O:1]-[H:2]  epsilon: 0.05 kcal/mol  distance: 0.3 A  chargeincrement1: 0.2 e  chargeincrement2: 0.1 e  type: BondCharge  sigma: 0.1 A  >),
                       ((0, 1, 2),
                        <VirtualSiteDivalentLonePairType with smirks: [H:1]-[O:2]-[H:3]  epsilon: 0.05 kcal/mol  distance: 0.3 A  chargeincrement1: 0.2 e  chargeincrement2: 0.1 e  type: DivalentLonePair  outOfPlaneAngle: 1.0 deg  inPlaneAngle: 0.0 deg  sigma: 0.1 A  >),
                       ((1, 2),
                        <VirtualSiteBondChargeType with smirks: [O:1]-[H:2]  epsilon: 0.05 kcal/mol  distance: 0.3 A  chargeincrement1: 0.2 e  chargeincrement2: 0.1 e  type: BondCharge  sigma: 0.1 A  >)])}
```
The spec. shows elements as simply `VirtualSite`, but I discovered that I needed to specialize the elements based on the `type` relatively early for validation purposes. This differs somewhat what the specification/xml represents, so I am not sure if the above representation is desirable.
</details>


j-wags:
Awesome progress. Recapping our meeting today: 

* **We should allow multiple matches for VirtualSites, and multiple of the exact same SMIRKS**
* When initializing a ForceField, should we deduplicate VirtualSites with same SMIRKS, type, angles, distances, chargeincrements?
    * **We should not deduplicate VirtualSites during VirtualSiteType or VirtualSiteHandler initialization**
* Does OpenMM have a problem if two virtualsites are really close to each other (eg. do they cause an energy explosion?) Under what conditions does OpenMM automatically deduplicate or apply nonbonded exceptions between virtualsites?
    * Unresolved -- We should check
* Should we throw an exception if two virtualsites are defined on top of each other (same reference atoms, angles, distances) during system creation?
    * **Yes, but add kwarg to create_openmm_system for `allow_overlapping_virtual_sites`**
* We should change the tagname of the “VirtualSite” elements to be distinct -- that is, the XML element should be titled “BondTypeVirtualSite” etc. We should change the SMIRNOFF spec docs accordingly.

I had some thoughts on implementation after our meeting:

* If at all possible, the contents of the VirtualSiteType subclasses should contain anything essential about the functional form of the VirtualSite. That way, if a user wants to experiment by making a new VirtualSite subclass, they shouldn’t need to change OpenFF toolkit code, but rather they should be able to “register” a new VirtualSite subclass, and then the Toolkit should automatically load it when initializing a ForceField and recognize the new unique tagname.
    * I don’t know exactly what the OpenMM API for defining VirtualSites looks like, but it would be ideal if each VirtualSiteType subclass contained the necessary OpenMM keywords/classes/methods to add the proper VirtualSite to an OpenMM System. Right now, we kinda do this with [the ParameterHandler _OPENMMTYPE class attribute](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/typing/engines/smirnoff/parameters.py#L2050), where we just hold on to a “type” object for each OpenMM `Force`, and repeatedly throw keywords at its constructor during system creation for each SMIRKS match we find.
    * We may need a more sophisticated method for doing this depending on how the OpenMM VirtualSite API looks. If it’s really hairy, we might want VirtualSiteType to be the first ParameterTYPE with its own `create_force` method 

Very roughly, an API example could look like this:
```

class MyNewVirtualSiteHandler(VirtualSiteHandler):
    _TAGNAME = ”VirtualSites”
    class MyNewVirtualSiteType(VirtualSiteType):
        _ELEMENT_NAME = ”MyNewVirtualSite”
        def _create_force(self, openmm_system, particle_indices):
            (Here the OpenMM System is modified to contain the desired virtual site)

my_topology = Molecule.from_smiles(‘HOH’).to_topology()
my_ff = ForceField(‘my_ff_with_my_new_virtualsite_type.offxml’, parameter_handler_classes=[BondHandler, AngleHandler… MyNewVirtualSiteHandler])
system = my_ff.create_openmm_system(my_topology)
```

* To enable a plugin-like architecture for new VirtualSite types, I think that populating a dictionary/registry with available VirtualSite subclasses would work. We do something like this when initializing ParameterHandlers. This mapping could have keys of VirtualSite tagnames and values of VirtualSiteType subclasses. Here are some existing code snippets that do something similar:
    * [“Registering” all parameterhandlers](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/typing/engines/smirnoff/forcefield.py#L277-L278)
    * [Lookup for parameterhandlers](https://github.com/openforcefield/openforcefield/blobmaster/openforcefield/typing/engines/smirnoff/forcefield.py#L1239-L1248) 
    * [Finding all subclasses](https://github.com/openforcefield/openforcefield/blob/20712c1a4bf09cd25da8f13305904633ea135ee6/openforcefield/utils/utils.py#L86-L90) (linking just for reference, we could just call this function)

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2874,7 +2903,295 @@ def create_force(self, system, topology, **kwargs):
...
+
+        def get_openmm_virtual_site(self, atoms, mass=None):
+            assert len(atoms) == 3
+            originwt = np.zeros_like(atoms)
+            originwt[0] = 1.0 # 
+
+            xdir = [-1.0, 1.0, 0.0]
+            ydir = [-1.0, 0.0, 1.0]
+
+            theta = self.inPlaneAngle.value_in_unit( self.inPlaneAngle.unit)
```
j-wags: We should be explicit about which unit we convert to here -- Does numpy expect radians?

trevorgokey: Yes! Good catch

j-wags:
Notes from meeting today:
* OpenMM Topology doesn't contain VSites, so we don't need to worry about them in `OFFTop.from_openmm`
* If isomoprhism/deduplication gets complicated, we could remove VSites from the public OFFMol API, and only let them be added during parameterization
* Should we provide a function to "label" an OFFTopology with virtualsites? This seems like a good idea, and it'd be most direct to repurpose `create_openmm_system` for this. `create_openmm_system` could return BOTH the resulting OMMSystem AND a modified copy of the input OFFTopology. This resulting OFFTopology would have COPIES of reference molecules and vsites with final charges. I'm not sure what would happen with constraints. Returning the resulting OFFTopology should be triggered by a new kwarg, `return_topology=False`, which will run `return system, topology` at the end of `create_openmm_system` is `True`. Should be trivial to add, since we already make a deep copy of the topology at the beginning of `create_openmm_system`.
* When do we want to deduplicate SMARTS matches (by permuting tuples of match indices)?  Is it absolutely necessary to add a unique  option to virtualsites? 
    * For BondCharge sites, it's clear that we do NOT want to attempt deduplication -- N2 should have symmetrically-applied vsites.
    * For DivalentLonePairs, deduplication would SOLVE unavoidable duplication issue with TIP4P, but would make TIP5P require two separate vsite SMIRKS (+N and -N degrees). So it seems that DivalentLonePairs should be deduplicated.
    * For TrivalentLonePairs, deduplication is definitely wanted, since it will otherwise unavoidably add 6 copies of the virtual site for NH3's lone pair. This deduplication might be performed by taking the match index tuple (a,b,c,d), and deduplicating permutations of b, c, and d. This... SHOULD be safe, though I wonder if there are pathological cases that would cause trouble here.
    * I'm not sure about deduplicating matches between DIFFERENT parameters. SMIRNOFF's "hierarchy" philosophy would make me think that we could add, for example, a "generic" trivalent nitrogen lone pair using `N([*])([*])([*])`, and then define a more specific lone pair for secondary nitrogen centers using  `N([C])([C])([*])` further down in the list, expecting the second to override the first when it applies. However, there are also other situations where I'd expect virtualsites to be additive. So, for now, the best way forward is probably to NOT attempt deduplication between different parameters, and just caution users to craft exclusive SMARTS if they want them to apply exclusively.

j-wags:
Two useful resources on OpenMM's behavior with respect to virtualsites:
* https://github.com/openmm/openmm/issues/2045
* https://github.com/michellab/Sire/issues/299

trevorgokey:
Apologies for the commit spam. Tests were working on my local machine and not working on GH. After digging I realized I hadn't updated the OE license and many tests were being skipped.

mattwthompson:
Oh, also, please merge master back into here and apply linting changes. It should make some of the diffs smaller.

mattwthompson:
I bit the bullet and linted and merged master back in, albeit with some errors. Hopefully they should be feasible to catch, there are probably just a few arguments that I dropped by accident. The changes that the master branch has undergone in the past couple weeks have been substantial, so it was never going to be straightforward.

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -942,25 +943,62 @@ def _load_smirnoff_data(self, smirnoff_data, allow_cosmetic_attributes=False):
             # Get the parameter types serialization that is not passed to the ParameterHandler constructor.
             ph_class = self._get_parameter_handler_class(parameter_name)
             try:
-                parameter_list_tagname = ph_class._INFOTYPE._ELEMENT_NAME
+                infotype = ph_class._INFOTYPE
+                # TG: The assignment above does not resolve to the underlying
+                # type if it is a class-level property, and for whatever reason
+                # the __get__ and __set__ methods are not automatically called.
```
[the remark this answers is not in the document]
trevorgokey: I found a solution which seems better in every way; see `VirtualSiteHandler.VirtualSiteTypeSelector`

openforcefield/tests/test_forcefield.py
```diff
@@ -1557,6 +1594,437 @@ def test_lookup_parameter_handler_object(self):
...
+    ############################################################################
+    # Divalent virtual site test data
+    ############################################################################
+
+    # A TIP5 definition from OpenMM
+    opts = OrderedDict(
+        {
+            "xml": xml_ff_virtual_sites_divalent_match_all,
+            "smi": None,
+            "assert_physics": (
```
mattwthompson, line 1941: Might it also be useful to verify the mass here? Although it may not be worth it if you have to reach somewhere other than `NonbondedForce`

openforcefield/tests/test_forcefield.py
```diff
@@ -1557,6 +1594,437 @@ def test_lookup_parameter_handler_object(self):
...
+    # Test the wildcard match
+    opts = OrderedDict(
+        {
+            "xml": xml_ff_virtual_sites_bondcharge_match_once,
+            "smi": "[C][C]",
+            "assert_physics": (
+                # charge        sigma         epsilon
+                (+0.1 * q_unit, None, None),
+                (+0.1 * q_unit, None, None),
+                (-0.2 * q_unit, 0.1 * s_unit, 0.1 * e_unit),
```
mattwthompson, line 1870: I'm generally confused about this and lot of the other cases in similar ways

* ~How did the carbons end up with half the virtual site's charge, and how did the virtual site's charge go negative?~ Just kidding, needed to look at the spec again
* How, if the virtual sites XMLs are being tacked on to Parsley, do the heavy atoms lack LJ parameters?

openforcefield/topology/molecule.py
```diff
@@ -669,16 +774,17 @@ def molecule_virtual_site_index(self):
         #       Deleting atoms/molecules in the Topology would have to invalidate the cached index.
         return self._molecule.virtual_sites.index(self)
 
-    @property
-    def molecule_particle_index(self):
-        """
-        The index of this VirtualSite within the the list of particles in the parent ``Molecule``.
-        Note that this can be different from ``molecule_virtual_site_index``.
+    # @property
+    # def molecule_particle_index(self):
```
mattwthompson, line 778: Please either remove this or make it a `NotImplementedError` stub; it looks un-used now but I'm not sure if that will always be the case

openforcefield/topology/molecule.py
```diff
@@ -739,8 +845,8 @@ def __repr__(self):
 
     def __str__(self):
         # TODO: Also include particle_index, which molecule this atom belongs to?
-        return "<VirtualSite name={} type={} atoms={}>".format(
-            self.name, self.type, self.atoms
+        return "<VirtualSite name={} type={} atoms={} particles={}>".format(
+            self.name, self.type, self.atoms, self.n_particles
```
mattwthompson, line 849: Virtually identical to `__repr__`, except out of sync now. I have little preference on which to use, but they should be consistent

```python3
In [18]: VirtualSite(atoms=[mol.atoms[0]])
Out[18]: VirtualSite(name=None, type=VirtualSite, atoms=[Atom(name=, atomic number=8)])

In [19]: str(VirtualSite(atoms=[mol.atoms[0]]))
Out[19]: '<VirtualSite name=None type=VirtualSite atoms=[Atom(name=, atomic number=8)] particles=1>'

```

mattwthompson (commented):
Alright, I've gotten through another partial review. Things are looking pretty good! I have yet to finish all of `molecule.py`, which will be the focus of the next iteration, in addition to giving the notebook a once-over.

A few of my comments here can be disregarded; while catching up on the implementation details, I confused myself and ran in circles a bit.

openforcefield/tests/test_forcefield.py
```diff
@@ -1557,6 +1594,437 @@ def test_lookup_parameter_handler_object(self):
...
+    ############################################################################
+
+    # First test that one vsite is added, and that the last parameter
+    # is chosen over the top-most wildcard match
+    opts = OrderedDict(
+        {
+            "xml": xml_ff_virtual_sites_bondcharge_match_once,
+            "smi": "[N]#[N]",
+            "assert_physics": (
+                # charge        sigma         epsilon
```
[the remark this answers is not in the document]
trevorgokey: We can.. but OpenMM returns them exactly in this order as a tuple, so doing a tuple == tuple was straightforward, and how it is done in other tests which were used as a template.

openforcefield/topology/topology.py
```diff
@@ -143,11 +186,52 @@ def __keytransform__(self, key):
...
 # =============================================================================================
 
+
+# =============================================================================================
+# TopologyAtom
+# =============================================================================================
+
+
+class TopologyParticle(Serializable):
+    # TODO: Should this be added?
```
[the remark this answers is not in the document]
trevorgokey: I forgot to delete this, but glad I left it up for you to catch. Right now Jeff is leaning on not having it. Anything that returns anything like a TopologyParticle just explicitly returns a TopologyAtom or TopologyVirtualParticle. 
I do think having a hierarchy like you said makes sense, but such a refactor seems unlikely.

openforcefield/topology/topology.py
```diff
@@ -577,6 +721,40 @@ def from_dict(cls, d):
         raise NotImplementedError()  # TODO
 
 
+# =============================================================================================
+# TopologyVirtualParticle
+# =============================================================================================
+
+
+class TopologyVirtualParticle(TopologyVirtualSite):
```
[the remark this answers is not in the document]
trevorgokey: It is mostly to inherit all of the members from virtual site. That said, a virtual particle is not necessarily an independent object; it is coupled to its parent vsite. The coupling is a novelty in the toolkit, so I am sure more thought can be put into how to handle it

I think refactoring makes sense, but it is more of a entire architecture question. Ideally, changing would mean implementing `TopologyParticle`, but then one would say everything should be refactored to use it. Then, I would essentially copy all of the `TopologyVirtualSite` API into `TopologyParticle`, and then just have it call super().API_calls, which is why I did the clunky thing to inherit in the first place.

openforcefield/topology/topology.py
```diff
@@ -694,6 +849,25 @@ def atom_start_topology_index(self):
         # Return cached value
         return self._atom_start_topology_index
 
+    @property
+    def virtual_particle_start_topology_index(self):
```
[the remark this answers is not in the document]
trevorgokey: I think it's a very useful method to have. Since virtual sites have a variable particle offset, if I am given a random molecule in the topology, it is impossible to use any heuristic to know the value this property provides. This is mostly due to how we define the ordering of atoms vs vsites. For example, right now if I have some molecule in a topology with indices 2,3,4, it is possible that the vsite indices start at 10,002. I think we should expose this functionality.

It also follows the naming convention of similar API calls.

openforcefield/tests/test_utils_structure.py
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
trevorgokey: Nope, will remove.

openforcefield/tests/test_forcefield.py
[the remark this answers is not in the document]
mattwthompson: Ah, in that case tuples are fine

openforcefield/topology/topology.py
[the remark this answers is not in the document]
mattwthompson: That sounds fine for the short term (a.k.a. here). If this becomes a pain point we can think more about a refactor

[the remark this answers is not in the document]
mattwthompson: (But do delete this and close the associated issue)

openforcefield/topology/topology.py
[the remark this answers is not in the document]
mattwthompson: This seems fine for now. I do think that having separate classes for the "sites" and particles is a nice feature here, though I'll admit I was skeptical at first.

This is definitely not blocking, just something to consider if the two classes diverge in the future and/or there is a compelling reason to implement `TopologyParticle` down the road

openforcefield/topology/topology.py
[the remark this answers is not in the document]
j-wags: Yeah -- Basically what Trevor said -- in the future we may want a TopologyParticle class, but right now the lack of it is just kinda an unpythonic inconvenience (some iterators will return heterogenous types, but they all have mostly the same API)

openforcefield/tests/test_forcefield.py
```diff
@@ -1557,6 +1594,437 @@ def test_lookup_parameter_handler_object(self):
...
+            ),
+        }
+    )
+    bond_charge_parameters_args.append(opts)
+
+    # Test the wildcard match
+    opts = OrderedDict(
+        {
+            "xml": xml_ff_virtual_sites_bondcharge_match_once,
+            "smi": "[C][C]",
```
j-wags, line 1865: Since atom indexing of molecules from SMILES are arbitrary, and here we're checking atom indices in the final system, it could be good to make a `create_ethane` method (like `create_ethanol`) to seed this test.

openforcefield/tests/test_molecule.py
```diff
@@ -1757,39 +1757,42 @@ def test_add_virtual_site_units(self, molecule):
         atom3 = molecule.atoms[2]
         atom4 = molecule.atoms[3]
 
+        atoms = (atom1, atom2)
+
         # Try to feed in unitless sigma
         with pytest.raises(Exception) as excinfo:
```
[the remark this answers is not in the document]
j-wags: I could take this or leave it -- I think this PR is a positive move for our overall code quality/functionality, so an exception refactor would be nice, but the scope is already quite large.

openforcefield/tests/test_molecule.py
```diff
@@ -1811,28 +1814,31 @@ def test_add_bond_charge_virtual_site(self, molecule):
...
-                                                             charge_increments=unit.Quantity(np.array([0.1, 0.2, 0.3]),
+                                                             charge_increments=unit.Quantity(np.array([0.1, 0.2]),
                                                                                              unit.elementary_charge)
                                                              )
         vsite2 = molecule.virtual_sites[vsite2_index]
+        assert atom2 in vsite2.atoms
+        assert atom3 in vsite2.atoms
+        assert vsite2 in atom2.virtual_sites
+        assert vsite2 in atom3.virtual_sites
+        assert vsite2.distance == distance
```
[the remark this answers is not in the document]
j-wags: Agree that this would be a good thing to add -- like a test to ensure that vsite setters do something sane with Quantities that have the wrong units (maybe raise an error), and to make sure that the getter returns a value that was just set.

These are kinda trivial tests and are really unlikely to actually catch a problem, so it's OK with me to leave them as TODO items if you get short on time.

mattwthompson (approved):
I think I've spent enough time putzing around with little details in the code; I haven't found anything alarming, so let's get this shipped out for people to use. Then, we'll have better feedback to go off of for the future iterations of this functionality.

 I'll leave it to your discretion to tidy up the little things like some of my earlier comments and commented-out code in the notebook (which still runs fine for me, for what it's worth) before merging. Oh, and make sure tests pass, of course 🛳️

openforcefield/tests/test_molecule.py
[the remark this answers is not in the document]
trevorgokey: I am going to forgo this for now.

openforcefield/tests/test_molecule.py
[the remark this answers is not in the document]
trevorgokey: I actually think the Quantity tests were already implemented previously, so I just had to tidy them up a bit.

openforcefield/utils/toolkits.py
```diff
@@ -35,6 +35,7 @@
     "UndefinedStereochemistryError",
     "GAFFAtomTypeWarning",
     "ToolkitWrapper",
+    "BuiltInToolkitWrapper",
```
trevorgokey, line 38: @j-wags Is this an old piece that should be removed?

mattwthompson: That might look out of place, but it should stay (I don't think `__all__` got updated when that class was added. I do `from openforcefield.utils.toolkits import *` a lot, so that's a minor point of frustration)

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -4282,6 +4290,842 @@ def create_force(self, system, topology, **kwargs):
...
+    Convert and validate the virtual site type specified in the VirtualSite element
+
+    Parameters
+    ----------
+    vsite_type : Any
+        The virtual site type to validate
+
+    Returns
+    -------
+    policy : str
```
j-wags, line 4304: Just so it doesn't get overlooked before the merge -- Some copy-paste errors in docstrings here.

j-wags (commented):
> I might need some feedback on where to put the parameter handler validators (the 3 validate_virtual_site* functions in parameters.py, 
> Some context for the validators: I just added vsite testing stuff to test_parameters.py and so it made me realize I needed to validate everything explicitly when using the API. It didn't really work to put the validators inside the VSPH, so I just lobbed them outside.

What's the blocker preventing us from from putting them in `VirtualSiteHandler`? You should still be able to import them in tests using
```
from openforcefield.typing.engines.smirnoff.parameters import VirtualSiteHandler
VirtualSiteHandler.validate_virtual_site_type_option(...)
```

> and perhaps some feedback on how to test OpenMM energies when vsites are involved. I know there are already some nice elaborate OpenMM system energy checkers, but they don't really work when vsites are present. I basically copied my vsite showcase utility functions into tests/utils.py, but not sure if that was completely wise. The problem is that the OpenMM util functions depend on a single positions vector for two independent systems, and this does not work for vsites. Maybe not a point currently given the PR scope, but I think there is some room for improvement in there somewhere. 

I think these are good where you put them, and agree that they can be separate from the original functions. If you have time, I'd like to go through them with you to ensure they're all still necessary.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1701,20 +1704,21 @@ def _add_parameters(self, section_dict, allow_cosmetic_attributes=False):
...
 
-        element_name = None
-        if self._INFOTYPE is not None:
-            element_name = self._INFOTYPE._ELEMENT_NAME
-
         for key, val in smirnoff_data.items():
-            # Skip sections that aren't the parameter list
-            if key != element_name:
-                continue
             # If there are multiple parameters, this will be a list. If there's just one, make it a list
```
j-wags, line 1708:
```suggestion

```

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1701,20 +1704,21 @@ def _add_parameters(self, section_dict, allow_cosmetic_attributes=False):
...
+            if self._INFOTYPE is not None:
+                element_name = self._INFOTYPE._ELEMENT_NAME
+                # Skip sections that aren't the parameter list
+                if key != element_name:
+                    break
+
```
j-wags, line 1714:
```suggestion
            # If there are multiple parameters, this will be a list. If there's just one, make it a list
```

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -4282,6 +4289,842 @@ def create_force(self, system, topology, **kwargs):
         system.addForce(gbsa_force)
 
 
+def validate_virtual_site_type_option(vsite_type):
```
j-wags, line 4292: After our discussion, if this doesn't fit into the `ParameterAttribute.converter` framework., I think it's OK if these the methods remain where they are, though you should make them private (`_validate_virtual...`).

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -4282,6 +4289,842 @@ def create_force(self, system, topology, **kwargs):
...
+        vsite_type = str(vsite_type)
+    except ValueError:
+        raise
+
+    if vsite_type in [
+        "BondCharge",
+        "MonovalentLonePair",
+        "DivalentLonePair",
+        "TrivalentLonePair",
+    ]:
```
j-wags, line 4322: It would be nice if this could be dynamically loaded from the find-able virtualsite subclasses (something like `[vst.TAGNAME for vst in all_subclasses(VirtualSiteHandler.VirtualSiteType)]`

trevorgokey: Done!

openforcefield/typing/engines/smirnoff/parameters.py
trevorgokey: Got converters working!

j-wags (approved):
LGTM, @trevorgokey

> ...
> 
> On this pale fear seized every one; they were so frightened that their arms dropped from their hands and fell upon the ground at the sound of the goddess's voice, and they fled back to the city for their lives. But Ulysses gave a great cry, and gathering himself together swooped down like a soaring eagle. Then the son of Saturn sent a thunderbolt of fire that fell just in front of Minerva, so she said to Ulysses, "Ulysses, noble son of Laertes, stop this warful strife, or Jove will be angry with you."
> 
> Thus spoke Minerva, and Ulysses obeyed her gladly. Then Minerva assumed the form and voice of Mentor, and presently made a covenant of peace between the two contending parties.
> 
> 
> THE END
>

trevorgokey:
Merged! Amazing. Thanks @j-wags and @mattwthompson for all the discussions and feedback!

PR #550, opened 2020-03-25
Implement get_parameter function (botched version)

### LINKED ISSUE #541: Implement the `ParameterHandler.get_parameter` function
It would be great to have the `ParameterHandler.get_parameter(...)` function implemented:

https://github.com/openforcefield/openforcefield/blob/0e5ff41b72ce4dcb4c0e9cd431faa28c7b7cd1ab/openforcefield/typing/engines/smirnoff/parameters.py#L1421-L1436

As general note - instead of just calling `pass` it would be great instead to raise a `NotImplementedError` for yet to be implemented functions / features.

mattwthompson:
Fixes #541, if I correctly interpreted the intent of the function from the docstring. The test could be expanded some, right now it is fairly basic and might miss some cases.

### DIFF
--- a/openforcefield/tests/test_parameters.py
+++ b/openforcefield/tests/test_parameters.py
@@ -388,11 +388,19 @@ class TestParameterHandler:
         """Test that ParameterHandler.get_parameter can lookup function
         """
         from simtk import unit
-        bh = BondHandler(skip_version_check=True)
+        bh = BondHandler(skip_version_check=True, allow_cosmetic_attributes=True)
 
         bh.add_parameter({'smirks': '[*:1]-[*:2]',
                           'length': 1*unit.angstrom,
-                          'k': 10*unit.kilocalorie_per_mole/unit.angstrom**2})
+                          'k': 10*unit.kilocalorie_per_mole/unit.angstrom**2,
+                          'id': 'b0'})
+        bh.parameters[0].add_cosmetic_attribute('foo', 'bar')
+
+        # Check base behavior
+        params = bh.get_parameter({'smirks': '[*:1]-[*:2]'})
+
+        assert params[0].length == unit.Quantity(1.0, unit.angstrom)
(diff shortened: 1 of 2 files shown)


openforcefield/tests/test_parameters.py
```diff
@@ -384,6 +384,21 @@ def test_add_delete_cosmetic_attributes(self):
...
+        bh = BondHandler(skip_version_check=True)
+        bh.add_parameter({'smirks': '[*:1]-[*:2]',
+                          'length': 1*unit.angstrom,
+                          'k': 10*unit.kilocalorie_per_mole/unit.angstrom**2})
+
+        params = bh.get_parameter({'smirks': '[*:1]-[*:2]'})
+
+        assert params[0].length == unit.Quantity(1.0, unit.angstrom)
+        assert params[0].k == unit.Quantity(10.0, unit.kilocalorie_per_mole/unit.angstrom**2)
+
```
j-wags: Could you add checks to ensure that:
* a query with no matches returns an empty list
* searching for a nonexistent attr does not raise an exception
* this does indeed behave as an OR query (so, add a statement that would fail an AND)
* this works on optional and cosmetic attributes (like `id` and `blah` respectively)

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1429,11 +1429,15 @@ def get_parameter(self, parameter_attrs):
...
-        list of ParameterType objects
+        params : list of ParameterType objects
             A list of matching ParameterType objects
         """
-        # TODO: This is a necessary API point for Lee-Ping's ForceBalance
-        raise NotImplementedError()
+        params = list()
+        for attr, value in parameter_attrs.items():
+            for param in self._parameters:
+                if getattr(param, attr) == value:
```
j-wags: Would this double-add parameters if a parameter matched on two queried values? It would be good to ensure we only return unique parameters once.

PR #551, opened 2020-03-27
Implement get_parameter function

### LINKED ISSUE #541: Implement the `ParameterHandler.get_parameter` function

mattwthompson:
Same as #550 (see some comments there), except not from my fork. Fixes #541

### DIFF
--- a/openforcefield/typing/engines/smirnoff/parameters.py
+++ b/openforcefield/typing/engines/smirnoff/parameters.py
@@ -1434,7 +1434,13 @@ class ParameterHandler(_ParameterAttributeHandler):
         """
         params = list()
         for attr, value in parameter_attrs.items():
-            for param in self._parameters:
+            for param in self.parameters:
+                if param in params:
+                    continue
+                # TODO: Cleaner accessing of cosmetic attributes
+                # See issue #338
+                if param.attribute_is_cosmetic(attr):
+                    attr = '_' + attr
                 if hasattr(param, attr):
                     if getattr(param, attr) == value:
                         params.append(param)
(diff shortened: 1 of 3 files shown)


mattwthompson:
@j-wags I think this is ready for review

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -1429,11 +1431,43 @@ def get_parameter(self, parameter_attrs):
...
+        for attr, value in parameter_attrs.items():
+            for param in self.parameters:
+                if param in params:
+                    continue
+                # TODO: Cleaner accessing of cosmetic attributes
+                # See issue #338
+                if param.attribute_is_cosmetic(attr):
+                    attr = '_' + attr
+                if hasattr(param, attr):
+                    if getattr(param, attr) == value:
```
jaimergp: This two lines can be reduced to a single check thanks to the three-argument `getattr`. If you agree this adds clarity, you can rewrite it as: 

```python
if getattr(param, attr, None) == value:
    params.append(param)
```

This will work as long as `None` is not an expected value. Otherwise you need to find another sentinel.

PR #554, opened 2020-03-30
Add support for GH Actions CI

jaimergp:


jaimergp:
I think that last commit made it work... Questions / comments:

* Manually exporting OE_LICENSE (global vars don't have access to $HOME at definition time). I think we could use one of the official locations so the oe toolkit finds it without the env var? Also, @SimonBoothroyd figured another workaround but I can't remember which one.
* I have added the `-v` flag to pytest. Is that something you deliberately avoided before?
* What do you think about the py36 timebomb? Too much? 
* If you don't answer to these questions I will sneak a failure step that depends on a random seed set to blow up 1% of the time for no reason. Unless its April 1st, in which case it fails 101% of the time. Looking forward to that day, by the way. PREPARE.

jaimergp:
I had a dream last night where I realized that the Python 3.6 checks can be done on a weekly basis using a separate cron job that only run on `master`, no PRs...

That's the kind of thoughts I find myself having these days. Seriously, 2020... 🤣

jthorton:
Just been looking through this for tips on how to convert to GH actions, just want to check the python 3.6 builds as the tests say 3.7 for the environment so am not sure it's using the right version for some reason, could be wrong though.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,115 @@
...
+      env:
+        ENC_OE_LICENSE: ${{ secrets.OE_LICENSE__EXP_JUN2020 }}
+      run: |
+        echo "${ENC_OE_LICENSE}" > $HOME/oe_license.txt
+    - name: Prepare test environment
+      shell: bash
+      run: |
+        . devtools/gh-actions/initialize_conda.sh
+        conda activate
+        python devtools/scripts/create_conda_env.py -n=test -p=$PYTHON_VER devtools/conda-envs/test_env.yaml
```
jthorton: I think this should be `$PYVER `?

jaimergp: Ohh, excellent catch! So sorry, this could have been such a terrible mistake in `master`. Thanks a bunch!

SimonBoothroyd:
> Manually exporting OE_LICENSE (global vars don't have access to $HOME at definition time). I think we could use one of the official locations so the oe toolkit finds it without the env var? Also, @SimonBoothroyd figured another workaround but I can't remember which one.

I think you can just use `${{ github.workspace }}/oe_license.txt` or alternatively `${{ github.home }}/oe_license.txt`

I think the other thing I do differently is I just store the encrypted OpenEye license in the repo itself (similar to how it is now) and store the decryption key as a secret similar to what is done on travis - see https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets

Another thing you may consider adding is a [quick check](https://github.com/openforcefield/openff-evaluator/blob/17bb98c83f260f33fd3115378a00b91296c8fb5a/.github/workflows/continous_integration.yaml#L80) to make sure that OpenEye can find the license (and that it is in date) - if it can't it can lead to some nasty silent failures IIRC.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,115 @@
...
+            openeye: true
+
+          - os: macOS-latest
+            rdkit: true
+            openeye: false
+
+    env:
+      ORGNAME: omnia
+      PACKAGENAME: openforcefield
+      OPENEYE_CHANNEL: -i https://pypi.anaconda.org/OpenEye/simple
```
SimonBoothroyd: I think this is unused?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,115 @@
...
+      PIP_ARGS: "-U"
```
SimonBoothroyd: possibly also unused?

SimonBoothroyd:
It may also be good to delete `.travis.yaml` in this PR.

.github/workflows/CI.yml
jaimergp: Let's see if it breaks if I remove it :D

jaimergp:
Is this warning new?

```
PytestUnknownMarkWarning: Unknown pytest.mark.slow - is this a typo? 
You can register custom marks to avoid this warning - 
for details, see https://docs.pytest.org/en/latest/mark.html
```

jaimergp:
> I think you can just use `${{ github.workspace }}/oe_license.txt` or alternatively `${{ github.home }}/oe_license.txt`

Awesome!

> I think the other thing I do differently is I just store the encrypted OpenEye license in the repo itself (similar to how it is now) and store the decryption key as a secret similar to what is done on travis - see https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets

Actually, I am not fully convinced that's necessarily better/safer? If we want to be extra paranoid, having the encrypted file publicly available in the repo allows anybody to download it and try brute forcing it (if there's such bored people in the globe...). If we directly encrypt the license then the only way it can be leaked is via a bug in GitHub security. Anyway, none is ideal or safe, but encoding the full license in GH allows us to save one step and one script. I am happy to discuss this further if needed, though!

> Another thing you may consider adding is a [quick check](https://github.com/openforcefield/openff-evaluator/blob/17bb98c83f260f33fd3115378a00b91296c8fb5a/.github/workflows/continous_integration.yaml#L80) to make sure that OpenEye can find the license (and that it is in date) - if it can't it can lead to some nasty silent failures IIRC.

Cool, added! Thanks!

j-wags:
> Anyway, none is ideal or safe, but encoding the full license in GH allows us to save one step and one script.

I feel the same way. Since GHA lets us store the entire OE license text in a single step, I think it gives us a solution with less moving parts We also get security in some edge cases -- When I was working with the OpenFF dangerbot, GHA raised a ton of alarms when the "secret" API token was committed to the README in plaintext. So I wouldn't _rely_ on it, but there may be some nice features that can tell us if something goes horribly wrong.

j-wags:
> I have added the -v flag to pytest. Is that something you deliberately avoided before?

As we added the permutation and molecule set tests, I think we were hitting the travis log size limit, so we had removed the `-v` flag. GHA appears to have a much higher limit, so I support putting it back in.

j-wags:
> Is this warning new?
> ```
> PytestUnknownMarkWarning: Unknown pytest.mark.slow - is this a typo? 
> You can register custom marks to avoid this warning - 
> for details, see https://docs.pytest.org/en/latest/mark.html
> ```

It isn't new, but it is a problem. I bet this happened because we used to invoke our tests from `openforcefield/tests`, but then we added doctests and example notebook tests, and so we had to start calling pytest from the top-level directory. However, we didn't move `conftest.py` (which defines `pytest.mark.slow`) to the top-level directory with it. 

@jaimergp, could you move `openforcefield/tests/conftest.py` up to the top-level directory?

SimonBoothroyd:
It would probably be good to also remove the now unused `oe_license.txt.enc`

j-wags:
> It would probably be good to also remove the now unused oe_license.txt.enc

Ooh, yes, that would be really helpful as well. And it will un-break the `License` badge in the top-level README.

SimonBoothroyd:
> Actually, I am not fully convinced that's necessarily better/safer?

I don't think it is either - just making a note of the differences in approach for reference!

jaimergp:
> > Is this warning new?
> > ```
> > PytestUnknownMarkWarning: Unknown pytest.mark.slow - is this a typo? 
> > You can register custom marks to avoid this warning - 
> > for details, see https://docs.pytest.org/en/latest/mark.html
> > ```
> 
> It isn't new, but it is a problem. I bet this happened because we used to invoke our tests from `openforcefield/tests`, but then we added doctests and example notebook tests, and so we had to start calling pytest from the top-level directory. However, we didn't move `conftest.py` (which defines `pytest.mark.slow`) to the top-level directory with it.
> 
> @jaimergp, could you move `openforcefield/tests/conftest.py` up to the top-level directory?

I don't think that's the recommended approach, according to [docs](https://docs.pytest.org/en/2.7.3/plugins.html?highlight=re#conftest-py-local-per-directory-plugins). How do you feel about creating a copy for each directory?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,113 @@
...
+    runs-on: ${{ matrix.cfg.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - 3.6
+          - 3.7
+          # Uncomment when ready ;)
+          # - 3.8
+        cfg:
```
mattwthompson, line 30: Is it possible to give each job in the matrix a more descriptive name, or have the env variables printed in this output? I like how travis makes it clear which is True and which is False

image

image

.github/workflows/CI.yml
```diff
@@ -0,0 +1,113 @@
...
+          PYTEST_ARGS+=" --ignore=docs --ignore=devtools --doctest-modules"
+        elif [[ "$OPENEYE" == true ]]; then
+          PYTEST_ARGS+=" --ignore=docs --ignore=devtools"
+          PYTEST_ARGS+=" --ignore=examples/check_dataset_parameter_coverage"
+          PYTEST_ARGS+=" --ignore=examples/QCArchive_interface"
+        fi
+        pytest $PYTEST_ARGS
+
+    - name: Codecov
+      uses: codecov/codecov-action@v1
```
mattwthompson, line 113: Does this mean codecov runs on all tests? If we trigger it to only run on a single `"$RDKIT" == true && "$OPENEYE" == true` wouldn't that cover everything?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,113 @@
...
+        ulimit -a
+    - name: Fix conda permissions in MacOS
+      shell: bash
+      if: startsWith(matrix.cfg.os, 'macOS')
+      run: |
+        sudo chown -R $(id -u):$(id -g) ${CONDA}
+    - name: Configure conda
+      shell: bash
+      run: |
+        . devtools/gh-actions/initialize_conda.sh
```
mattwthompson, line 70: What is the purpose of this `.`? Is it telling GHA to run it out of a particular directory? It obviously works fine, I am just curious.

.github/workflows/CI.yml
j-wags: Surprisingly, no. If both toolkits are available, OpenEye will be used preferentially. So there will be lower coverage of RDKit functionality in that build.

.github/workflows/CI.yml
jaimergp: `.` is an alias to `source`. We are importing that initializing script in each step because each GHA step is a separate bash process.

.github/workflows/CI.yml
jaimergp: We can add an `if: <condition>` check to the step if you want! We can also run the coverage on a separate job to make sure it's only run once.

.github/workflows/CI.yml
jaimergp: I don't think we have control over that autogenerated name :( Let me check.

jaimergp: Oh, I stand corrected, I think we can use a [`name:`](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idname) label to override that. Let me push! This is a great addition!

jaimergp: How does it look now? 😃

.github/workflows/CI.yml
jaimergp: Added this in last push to see if it works. Let me know if you agree or if I should revert to no-condition status.

mattwthompson: @j-wags I'm thinking of something else, not the % of unit tests that it runs on (although that's relevant). I meant if codecov runs on each CI run of the matrix, i.e. all 8 or so checks that CI runs vs. on just one, the reports are probably getting uploaded the same way each time. [Here](https://github.com/mosdef-hub/mbuild/blob/70ee2db5c017aa8dc7737214ea5294fd4ec2d049/azure-pipelines.yml#L78-L84) is what I had in mind (it's AZP but you get the idea, and it is exactly what @jaimergp suggested). I _think_ that codecov will just display whatever report it was sent last, so maybe the last CI check to run and be uploaded?

If all checks were the same, I would push us to pick an arbitrary one and use that as a reference, just to not upload 8x as many reports as we would care about. But, given that the toolkit has slightly different behavior with each other toolkit, we should decide which version we care about. I figure the True/True case is what matters. If we were to care about the code coverage when all RDKit-only or OpenEye-only tests are skipped, that wouldn't be too representative of the coverage of the project as a whole. Here, the advantage of picking just one run in the CI matrix is that we can control which we care about and it should behave the same way each time.

.github/workflows/CI.yml
mattwthompson: Beautiful! Thanks you!

.github/workflows/CI.yml
jaimergp: Oh, you are right, we are still executing several actions. I might just pop it out to a separate job.

j-wags:
> I don't think that's the recommended approach, according to docs. How do you feel about creating a copy for each directory?

Hmm, I had looked at that page the other day, and it seems that, since `pytest` is invoked from the top level of the repo (which isn't part of a package/module), we don't get a recursive check for `conftest.py` in subdirectories. So, even if we did have multiple `conftest.py`s, they wouldn't be found as long as we invoke `pytest` from the top directory.

**That said**, the `--runslow` flag is only important in `openforcefield/tests`. So maybe we want to invoke pytest twice:
* Once in `examples`, with `--nbval-lax` keyword
* Once in `openforcefield`, with `--doctest-modules`, and without `--runslow` except in the cron job

The second invocation would pick up `conftest.py` automatically. 

Regarding the warnings, I spoke with @mattwthompson and @dotsdl this morning to debug something related to this, and we concluded that the `warning`s are unavoidable due to how the `--runslow` flag is implemented. QCFractal uses the same thing, and the only reason they don't get the warnings is because all of their CI specifies the `--runslow` option. 

I realize that the discussion of this PR is turning into entirely new features (like adding a `--runslow` cron build). I'd be happy to consider this refactor in a separate PR so that we can keep the scope limited. The warnings are unavoidable and were already in our Travis builds, thus nothing is technically broken here.

.github/workflows/CI.yml
j-wags: > I think that codecov will just display whatever report it was sent last, so maybe the last CI check to run and be uploaded?

Ah, I see. This seems to be the important question. I'm hoping that codecov reports the _union_ of coverage from all tests on a specific commit, instead of having multiple builds for the same commit clobbering each other's reports. 

And indeed, [Codecov's docs](https://docs.codecov.io/docs/merging-reports) say that they do merge reports from multiple builds on the same commit. As further reassurance, it does seem to be grouping [all 8 builds of our recent commits](https://codecov.io/gh/openforcefield/openforcefield/commit/5e6b6dd52980a5e9a0423589c2a6370420e03694/build), and presumably it's merging them for the final readout.

So, if I'm interpreting this correctly, we _do_ want unconditional upload of the coverage report from all builds. But I'm happy to chat tomorrow if I'm misinterpreting the issue.

jaimergp:
> That said, the --runslow flag is only important in openforcefield/tests. So maybe we want to invoke pytest twice

Ok, no probs! I was thinking of splitting that huge pytest run in several steps anyway. I will propose my plan for that in another issue.

I have reverted the codecov filters so I think this is ready to go unless I am missing something.

.github/workflows/CI.yml
jaimergp: Fixed!

mattwthompson: I think you're right - may have been different in the past, else I'm not sure where I got that from

mattwthompson:
I think it's good to go

jaimergp:
I'll merge as soon as I get @j-wags definitive green light!

jaimergp:
Bye Travis!

PR #558, opened 2020-04-01
Order VirtualSites at the end of the Topology particle indexing system

j-wags:
Our `Topology` object aims to allow the user to use the same atom/particle indexing both before and after parameterization. This is important in cases where the user has a set of coordinates for their input molecules, and wants to apply those coordinates to the resulting OpenMM System without needing to reindex them. This PR fixes a flaw with the old indexing system.

 Previously, the `Topology.particles`, `TopologyAtom.topology_particle_index`, and `TopologyVirtualSite.topology_particle_index` properties would assume the following particle indexing for TopologyAtoms and TopologyVirtualSites:
```
for top_mol in topology.topology_molecules:
    for top_atom in top_mol.atoms:
        yield top_atom
    for top_vs in top_mol.virtual_sites:
        yield top_vs
```

This ordering has the advantage that it keeps the TopologyAtoms and TopologyVirtualSites for a single TopologyMolecule in a contiguous range of topology particle indices. However, @trevorgokey and I found during discussion of VirtualSiteHandler that, since SMARTS-based virtual site assignment happens **during** system creation, adding a VirtualSite to the first molecule in a Topology would increment all subsequent TopologyMolecules' particle indices by one. As a consequence, implementing VirtualSiteHandler would make us unable to guarantee that a user could map input coordinates to a resulting System's particle indexing system, since the force field could add an unknown number of VirtualSites to the particle list, shuffled in with actual atoms. 

Instead, this PR implements the following ordering:
```
for top_mol in topology.topology_molecules:
    for top_atom in top_mol.atoms:
        yield top_atom
for top_mol in topology.topology_molecules:
    for top_vs in top_mol.virtual_sites:
        yield top_vs
```

This indexing system guarantees that `TopologyAtom.topology_particle_index` will be equal to the resulting OpenMM System's particle indexing system.  


j-wags:
Thanks for checking this out, @trevorgokey!

Issue #555, opened 2020-03-30
Visualization of OpenFF Molecules


mattwthompson:
**Is your feature request related to a problem? Please describe.**
Not really a problem, just the inconvenience of not being able to visualize molecules without converting to something external

**Describe the solution you'd like**
I would like there to be a simple way to visualize an OpenFF molecule object while fiddling around in a notebook. I don't think this is a critical feature, but it would be a nice feature for developers (i.e. diagnosing why half of a torsion drive failed) and users (think docs, examples, future workshops, etc.) alike.

**Describe alternatives you've considered**
There's surely some way to convert to NGLView or another library with viz support, but I would like to avoid 2-3 conversions and just do a `mol.visualize()` call similar to the [Compound.visualize()](https://github.com/mosdef-hub/mbuild/blob/70ee2db5c017aa8dc7737214ea5294fd4ec2d049/mbuild/compound.py#L1245) method in mBuild. It is effectively a thin wrapper around NGLView/py3Dmol

**Additional context**
the above visualization within MoSDeF used NGLView for years but it can be flaky and [isn't so focused](https://github.com/mosdef-hub/mbuild/issues/529#issuecomment-488539719) on peeking at at a molecular snapshot, so @ahy3nz refactored it to have a switchable-ish backend and added support for py3Dmol. https://github.com/mosdef-hub/mbuild/pull/530 I think this is an approach we could mimic.

Note that you can fire a molecule off to RDKit and get a 2-D representation, but I would really like to also be able to see a 3-D representation. 

![image](https://user-images.githubusercontent.com/7935382/77966443-a960ff80-72a8-11ea-84d8-ce22f7f28ee2.png)

This points back to some discussion about what is necessarily encoded in this object, i.e. if this visualization should fail in the absence of conformers or fall back to some 2-D representation. Please correct me if I missed something, but I believe coordinates are only stored if there is a conformer stored?

This also may be worth discussing in the context of other objects, but Molecule is the easy starting point.

jaimergp:
+1 on a switchable 3D visualization backend that falls back to 2D representation if no conformers are available at that point.

jthorton:
This would great if we had something that came with the toolkit for now if you are on master and have qcelemental installed you can use `mol.to_qcschema()` if your molecule has a conformer.

PR #560, opened 2020-04-01
Add molecule.visualize method

mattwthompson:
Tackling #555 

The easiest way seem to be writing out to disk and letting NGLview load it back in, or hook into RDKit's default visualization for the 2-D case. This is the same approach that [QCElemental tasks](https://github.com/MolSSI/QCElemental/blob/b17840b71054a8fe261ac68d0ca7274f200200e4/qcelemental/models/molecule.py#L424). Other backends like py3Dmol may be worth adding if we are still unhappy with the reliability of NGLview.

Here is a PDF export of a proof of concept (I can't export the NGLview widget, so you'll have to either take my word for it or copy these commands into a notebook yourself)
[viz.pdf](https://github.com/openforcefield/openforcefield/files/4417890/viz.pdf)

To do:
* Is there a straightforward way of visualizing an OpenEye molecule? The best I found is [about 4 years without an update](https://anaconda.org/openeye/openeye-oenotebook)
* Figure out if/how to deal with multiple conformers
* Import NGLview in a more elegant way - I don't like the idea of adding a dep for one corner of the codebase


jaimergp:
You can save the disk trip by exporting the content string. In the end, most of `nglview` integrations happen via PDB serialization. 

For the 2D depiction with OpenEye, as long as you can obtain the image (PNG or SVG) from the backend, then showing that in a notebook is a matter of using the correct `display.<TAG>` object.

jaimergp:
I am cleaning a notebook I had laying around! It'll be ready in a bit.

jaimergp:
There you go. I had to zip the ipynb because GH won't allow the extension otherwise.
It contains an example for rdkit, openeye and nglview.

[offnglview.zip](https://github.com/openforcefield/openforcefield/files/4418099/offnglview.zip)

mattwthompson:
Thank you! I will try to implement your suggestions tomorrow

jthorton:
This is looking great I had a play with the notebook and I like the idea of looking at the conformers in one view this could be great for looking at torsiondrives and optimisation trajectories!

jaimergp:
Oh, does the trajectory player work for you? I wasn't sure about that one! Can you share a screenshot?

jaimergp:
Ah, I see, aromatic rings don't have too many conformers, huh...

Updated notebook with a fix for the trajectory. Now it does work!

trajectory


[offnglview.zip](https://github.com/openforcefield/openforcefield/files/4420343/offnglview.zip)

jthorton:
Yeah, that seemed to be the issue looks great!

mattwthompson:
I've added the OpenEye viz in your notebook and also changed it to write to a PDB memfile instead of an SDF file on disk. Works well for all three backends, except I don't get the cool play button on my machine and haven't for a few years. I tried re-installing things, switching browsers. I still get the render, just only one frame

image

Probably no harm in shipping it anyway, since there are other backends and NGLview does seem to work for a good number of people. However, I would like to have some safeguards in place to prevent anybody from thinking that the `OFFTrajectory` is a core class, at least while we're only using it as glue to NGLview, before/if anything else cool happens with it

jaimergp:
Oh, yeah, good point: `OFFTrajectory` should be something private such as `_OFFTrajectoryNGLView`.

For the player, are you using the 2nd notebook I uploaded? I had a bug in the frame retriever (you have to index `self.trajectory.conformers[index]`, not `self.trajectory` directly.

jaimergp:
Also, this wiki is golden: https://github.com/SBRG/ssbio/wiki/Troubleshooting#tips-for-nglview

mattwthompson:
Yeah, I'm using your second notebook with

```
    def get_coordinates(self, index):
        return self.trajectory.conformers[index] / u.angstrom
```

but the NGLview demo also doesn't work for me (if one of the other cells was meant to be a trj, not just one frame)

mattwthompson:
Not clean enough for review, which hopefully involves somebody getting a trajectory on their own machine, but it works in basic principle for these backends. Thank you @jaimergp for the notebooks, this PR is mostly your work

image

openforcefield/topology/molecule.py
```diff
@@ -4590,6 +4591,72 @@ def add_conformer(self, coordinates):
...
+                oedepict.OEPrepareDepiction(oemol)
+                img = oedepict.OEImage(width, height)
+                display = oedepict.OE2DMolDisplay(oemol, opts)
+                oedepict.OERenderMolecule(img, display)
+                png = oedepict.OEWriteImageToString("png", img)
+                return Image(png)
+        else:
+            raise ValueError('Could not find an appropriate backend')
+
+# TODO: Gracefull skip this class if nglview is not installed
```
jaimergp, line 4632: Maybe having a conditional class definition works here? I don't know how others feel about it but, maybe:

```
try:
    import nglview as nv
    class _OFFTrajectoryNGLView(nv.Trajectory):
        ... # real class here
except ImportError:
    class _OFFTrajectoryNGLView:
        def __init__(self, *args, **kwargs):
            raise NotImplementedError("You need NGLView to use this class")
```

jaimergp: I am tackling this, by the way. Will push later.

jaimergp:
I have tried locally. Two observations:

- Trajectory player will only appear if n_frames > 2.
- RDKit generates a `bad conformer ID` if it's used to write the PDB. Might be a bigger issue not related to visualization.

jaimergp:
> * Trajectory player will only appear if n_frames > 2.

This was in bug in the v 2.6 series. Upgrading to nglview v2.7 fixed it!

jaimergp:
Did some debugging here:

[off-depiction-debug.zip](https://github.com/openforcefield/openforcefield/files/4426937/off-depiction-debug.zip)

openforcefield/topology/molecule.py
```diff
@@ -4680,7 +4680,7 @@ def get_structure_string(self):
         memfile.seek(0)
         block = memfile.getvalue()
         # FIXME: Prevent multi-model PDB export with a keyword in molecule.to_file()?
-        models = block.split('END\n')
+        models = block.split('END\n', 1)
```
jaimergp, line 4683: I'd like some discussion on this workaround. 

(I think) `nglview` accepts multimodel PDBs, but I think it only considers the first one by default. Do we want to pass only the first one to begin with? Is  `END` the right keyword here? Or is it `ENDMDL`? OpenEye generates multiple conformers with `END` tags to separate them.

openforcefield/topology/molecule.py
jaimergp: Check last commits!

openforcefield/topology/molecule.py
mattwthompson: I don't have a recommendation here but you seem to be right about what nglview [does](https://github.com/arose/nglview/issues/765#issuecomment-459862450) and OpenEye does write it with `END`, not `ENDMDL`

Can RDKit not write it, or am I calling something wrong here?
```
In [21]: mol = Molecule.from_smiles('CCOCOCOCCOCCOCOCC')                                       

In [22]: mol.generate_conformers()                                                             

In [23]: mol.to_file('rdkit.pdb', file_format='pdb', toolkit_registry=RDKitToolkitWrapper())   
<ipython-input-23-6634d46173bd> in <module>
----> 1 mol.to_file('rdkit.pdb', file_format='pdb', toolkit_registry=RDKitToolkitWrapper())

~/software/openforcefield/openforcefield/topology/molecule.py in to_file(self, file_path, file_format, toolkit_registry)
   3577         if type(file_path) == str:
   3578             # Open file for writing
-> 3579             toolkit.to_file(self, file_path, file_format)
   3580         else:
   3581             toolkit.to_file_obj(self, file_path, file_format)

~/software/openforcefield/openforcefield/utils/toolkits.py in to_file(self, molecule, file_path, file_format)
   2006         # open a file object and pass to the object writer
   2007         with open(file_path, 'w') as file_obj:
-> 2008             self.to_file_obj(molecule=molecule, file_obj=file_obj, file_format=file_format)
   2009 
   2010     def canonical_order_atoms(self, molecule):

~/software/openforcefield/openforcefield/utils/toolkits.py in to_file_obj(self, molecule, file_obj, file_format)
   1979         try:
   1980             writer = self._toolkit_file_write_formats[file_format](file_obj)
-> 1981             writer.write(rdmol)
   1982             writer.close()
   1983         # if we can not write to that file type catch the error here

ValueError: Bad Conformer Id

File: openforcefield/topology/molecule.py [hunk 3]
```diff
@@ -4592,70 +4593,95 @@ def add_conformer(self, coordinates):
...
 
-# TODO: Gracefull skip this class if nglview is not installed
-import nglview as nv
-import uuid
-class _OFFTrajectoryNGLView(nv.Trajectory):
+
+try:
+    from nglview import Trajectory as _NGLViewTrajectory
+except ImportError:
+    _NGLViewTrajectory = object
```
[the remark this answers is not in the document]
jaimergp: We still want the full module to be importable even when `nglview` is not there. `_OFFTrajectoryNGLView` will only be called when we try to visualize something with nglview, so maybe we can skip the alternative definition... I just don't feel comfortable with defining classes only when certain conditions are met, so I'd rather have the empty shell than nothing at all. But this is only a personal opinion, so please let's discuss this with the rest of the team! I don't have any strong opinions!

openforcefield/topology/molecule.py
jaimergp: Yep, we have a bug there, as I commented in https://github.com/openforcefield/openforcefield/pull/560#issuecomment-608375591. I didn't have the time to double check  this can happen in other parts of the library (99% yes), so I guess we should raise an issue for this bug.

openforcefield/topology/molecule.py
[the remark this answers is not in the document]
mattwthompson: That's a good point, I didn't even think of the fact that this would mean the module is not importable without `nglview` installed. I changed the exception to be raised only when the class would be instantiated, not when it's defined, so that should only be hit if a user is trying to use it.

jthorton:
might be worth having a function like this as well to automatically display the molecule in notebooks https://github.com/MolSSI/QCElemental/blob/b17840b71054a8fe261ac68d0ca7274f200200e4/qcelemental/models/molecule.py#L1001

mattwthompson:
Hey, that's pretty cool (and straightforward enough to implement):

image

I added a simple version of this in 700d51e1cdb25189e3b57bca1066e5e4eebaf511 just so people can see it; we can revert this if we don't want it here

davidlmobley:
This looks pretty sweet. Thanks for all the work on this!

jaimergp:
So this looks great so far, and I would agree to merge but:

* #568 is a blocking issue
* there's no test ensuring that we are not hitting #568-like regressions. In other words, we should include a test addressing 568 as part of this PR

jthorton:
#568 is now closed and has the test so this should be ready to finish up.

jaimergp:
Yes! That worked! Thank you @jthorton!

mattwthompson:
Thanks everybody for the contributions - it's nice that even an individual PR be a collaborative effort!

I missed updating the release notes in #569 but snuck them in here. I think this is good to merge but I also believe a formal review would be in order

jaimergp:
Sure, I'll run a couple of examples on a notebook. Which reminds me... do we want to add this functionality to the existing notebooks?

jaimergp:
While testing I found that `_ipython_display_` was not working as expected, so I went ahead and fixed it to the best of my knowledge (after finding example implementations [here](https://github.com/ipython/ipython/issues/10491) and [here](https://github.com/ipython/ipython/issues/10247)).

Then since I already had the notebook, why not adding some markdown and put it under `examples/`?

Let me know what you think!

mattwthompson:
No guarantees tests will pass now, but I found a couple little things:

* The notebook itself is great, and I think does an excellent job of showing off how the functions can be used. However, testing each backend explicitly does not work so well with the test matrix (in the cases of RDKit and OpenEye) or with NGLview at all, since I don't think that's installed (I know we don't have it as a dependency, and I think that's best, and I don't think any of our deps pull it in). So I added a comment for nbval to skip those tests. The basic `m.visualize()`/`display(m)` cells are still tested and hopefully will catch problems if this funciton breaks.

* We were saying that we would fall back to OpenEye if RDKit was not installed, but the logical didn't really capture that. Now, if the backend is set to RDKit (the default value) but it's not available, it will switch to OpenEye. I think this is a quirk of selecting a default value, so I don't think we need to think about the reverse. And in the case of having conformers, I'm ok with requiring the user to specify NGLview each call is ok.

openforcefield/topology/molecule.py
```diff
@@ -4697,6 +4699,132 @@ def add_conformer(self, coordinates):
...
+    @property
+    def n_frames(self):
+        return len(self.molecule.conformers)
+
+    def get_structure_string(self):
+        memfile = StringIO()
+        self.molecule.to_file(memfile, "pdb")
+        memfile.seek(0)
+        block = memfile.getvalue()
+        # FIXME: Prevent multi-model PDB export with a keyword in molecule.to_file()?
```
dotsdl, line 4824: I think this is fine for this PR. @mattwthompson can you open up a new issue for adding this functionality in the PDB exporters?

mattwthompson: I believe this was fixed in https://github.com/openforcefield/openforcefield/pull/579 but we didn't update this comment. Somebody back me up on this actually being fixed before I merge?

jaimergp: I'd say 579 solved a different issue (RDKit not being able to generate multi-model PDBs). What I noticed there was an API-level way to tell the toolkit we don't want all conformers written to the PDB, just the first one. Something like this:

```python
self.molecule.to_file(memfile, "pdb", first_model_only=True)
```

or 
```python
self.molecule.to_file(memfile, "pdb", conformers=[0])
# conformers can be None/'all', or a list of indices or slice?
```

jaimergp:
> We were saying that we would fall back to OpenEye if RDKit was not installed, but the logical didn't really capture that. Now, if the backend is set to RDKit (the default value) but it's not available, it will switch to OpenEye. I think this is a quirk of selecting a default value, so I don't think we need to think about the reverse. And in the case of having conformers, I'm ok with requiring the user to specify NGLview each call is ok.

Thinking of this, if the user explicitly specifies `rdkit` as the backend (instead of the default), it can silently fallback to OpenEye without saying anything and maybe that can be misleading for somebody...

I suggest two alternative solutions:

A. The easy mode. Throw a warning [here](https://github.com/openforcefield/openforcefield/pull/560/commits/8d188dafb02bd7ba3c6785d246360e98ace544ca#diff-dcc45fa83688f86668e384f78c5ab22aR4753) so the user knows we are not silently ignoring commands, but providing a valid alternative.
B. Refactor `.visualize()` into different `._visualize_<backend>` submethods, and have it generate the adequate chain of attempts based on the toolkit availability. This will need more thought and I think we would only be covering corner-cases we can always fix in the future.

openforcefield/topology/molecule.py
mattwthompson: Moved this to #585

mattwthompson:
My preference now is definitely for the easier of the two options here, although I think refactoring this into more explicitly use separate functions for each backend would also be valuable

mattwthompson:
Okay, if these tests pass I think I've fixed everything

Is something up with codecov? It's not showing up as a status check and the [site](https://codecov.io/gh/openforcefield/openforcefield/commits) is .... yikes

jaimergp:
GH has had some problems today (https://www.githubstatus.com/), so maybe that has impacted them too?

openforcefield/topology/molecule.py
```diff
@@ -4697,6 +4700,137 @@ def add_conformer(self, coordinates):
...
+        ----------
+        backend : str, optional, default='rdkit'
+            Which visualization engine to use. Choose from:
+            - rdkit
+            - openeye
+            - nglview (conformers needed)
+        width : int, optional, default=500
+            Width of the generated representation (only applicable to
+            backend=openeye)
+        height : int, optional, default=500
```
jaimergp, line 4717: Ooops, this should be `300`

```suggestion
        height : int, optional, default=300
```

openforcefield/topology/molecule.py
```diff
@@ -4697,6 +4700,137 @@ def add_conformer(self, coordinates):
...
+                    'conformers.'
+                )
+        if backend == 'rdkit':
+            if RDKIT_AVAILABLE:
+                from rdkit.Chem.Draw import IPythonConsole
+                return self.to_rdkit()
+            else:
+                warnings.warn(
+                    'RDKit was requested as a visualizaiton backend but '
+                    'it was not found to be installed. Falling back to '
```
jaimergp, line 4756: Sorry for the very minor comment (I promise this is the last one!), but... what about rewording this as "trying to use OpenEye..."? If we say we fall back to OpenEye but then this is not available, it would be weird UX, I think.

openforcefield/topology/molecule.py
```diff
@@ -4697,6 +4700,137 @@ def add_conformer(self, coordinates):
...
+                    'Visualizing with NGLview requires that the molecule has '
+                    'conformers.'
+                )
+        if backend == 'rdkit':
+            if RDKIT_AVAILABLE:
+                from rdkit.Chem.Draw import IPythonConsole
+                return self.to_rdkit()
+            else:
+                warnings.warn(
+                    'RDKit was requested as a visualizaiton backend but '
```
jaimergp, line 4755:
```suggestion
                    'RDKit was requested as a visualization backend but '
```

mattwthompson:
All good suggestions, thank you!

mattwthompson:
Thanks @dotsdl and @jaimergp for the commits and reviews!

Issue #547, opened 2020-03-20
JSON serialization of Molecule broken


daveminh:
**Describe the bug**
JSON serialization of Molecule broken

**To Reproduce**
Steps to reproduce the behavior. A minimal reproducing set of python commands is ideal.
<img width="650" alt="image" src="https://user-images.githubusercontent.com/6304824/77211596-abb1a580-6ac1-11ea-840d-3bea21874724.png">

**Computing environment (please complete the following information):**
# packages in environment at /miniconda2/envs/openmm:
#
# Name                    Version                   Build  Channel
amberlite                 16.0                     pypi_0    pypi
ambertools                17.0                     pypi_0    pypi
appnope                   0.1.0           py37hc8dfbb8_1001    conda-forge
attrs                     19.3.0                     py_0    conda-forge
backcall                  0.1.0                      py_0    conda-forge
bleach                    3.1.3              pyh8c360ce_0    conda-forge
boost                     1.72.0           py37hbf1eeb5_0    conda-forge
boost-cpp                 1.72.0               hdf9ef73_0    conda-forge
bson                      0.5.9                      py_0    conda-forge
bzip2                     1.0.8                h0b31af3_2    conda-forge
ca-certificates           2019.11.28           hecc5488_0    conda-forge
cairo                     1.16.0            hec6a9b0_1003    conda-forge
certifi                   2019.11.28       py37hc8dfbb8_1    conda-forge
cffi                      1.14.0           py37h356ff06_0    conda-forge
chardet                   3.0.4           py37hc8dfbb8_1006    conda-forge
codecov                   2.0.15                     py_1    conda-forge
coverage                  5.0.4            py37h9bfed18_0    conda-forge
cryptography              2.8              py37he655712_2    conda-forge
curl                      7.68.0               h8754def_0    conda-forge
cycler                    0.10.0                     py_2    conda-forge
cython                    0.29.15          py37h570ac47_1    conda-forge
dbus                      1.13.6               h2f22bb5_0    conda-forge
decorator                 4.4.2                      py_0    conda-forge
defusedxml                0.6.0                      py_0    conda-forge
entrypoints               0.3             py37hc8dfbb8_1001    conda-forge
exampleplugin             1.0                      pypi_0    pypi
expat                     2.2.9                h4a8c4bd_2    conda-forge
fftw                      3.3.8           nompi_h44ae4c8_1110    conda-forge
fftw3f                    3.3.4                         2    omnia
[...]

PR #569, opened 2020-04-07
Add round-trip serialization tests for other formats

### LINKED ISSUE #559: Test round-trip serialization
In order to avoid issues like #547 popping up, we should take a closer look at our serialization and, in particular, implement round-trips from our data structures to various formats, i.e.

https://github.com/openforcefield/openforcefield/blob/af7c2f862c42ab486a3bf1ec90d36920862faba3/openforcefield/tests/test_molecule.py#L258-L259

Counter-point: How much work would it be to make each `to_dict()` method adhere to the various formats compared to refactoring everything to be built on top of [pydantic](https://pydantic-docs.helpmanual.io/)? I don't have a good feel for how much work each of these would take, so it is hard for me to estimate. What I do know is that 
* we don't want bugs like this 
* a later refactor to pydantic or something similar seems likely, if not inevitable

So I'm not sure how to prioritize this at the moment - @j-wags @dotsdl @jthorton and others what do you think?

Related #531 and probably others

mattwthompson:
Fixes #559 

Most of this is copy-pasted code from an existing test. Testing serialization of all mini drug bank molecules for all formats is probably not necessary, it could be scaled back if we want tests to run quick.

Right now, JSON is failing (known #547) and TOML is not implemented, but everything else seems to work fine on these molecules. Next steps will be to find out if/how to fix these and do it.


mattwthompson:
Ah, that makes sense. I'll admit I was lazy and assume the ❌ were from the earlier URL issues. I think I've implemented your suggestions; feedback welcome

These tests could be cleaned up a little - off the top of my head, I'm not sure how to have the same line of code iterate over `mini_drug_bank()` and a `Molecule.from_smiles('CCO')` without doing the latter N times. On the other hand, this isn't an egregious case of copy-and-paste-ish code.

I had done the `from_smiles('CCO')` because preference was to not raise these failures for the entire set of drugbanks molecules, but I guess CI time is pretty much free. Again, there may be a more elegant way to do this!

jthorton:
I think we can also export XML but not create from it yet if we should add tests for that?

Also with the serialization bug for json could this be fixed by using the `tolist` method instead of going to bytes that way we should keep the shape of the array as well.  This would only be a small change to the to and from_dict method I think.

j-wags:
> I think we can also export XML but not create from it yet if we should add tests for that?

Oh, great call. We should add tests to/from XML

> Also with the serialization bug for json could this be fixed by using the tolist method instead of going to bytes that way we should keep the shape of the array as well. This would only be a small change to the to and from_dict method I think.

I took my cues for the array serialization form here: https://stackoverflow.com/questions/30698004/how-can-i-serialize-a-numpy-array-while-preserving-matrix-dimensions#30699208

I'm not sure that `tolist` is exactly what we want here -- I think there's some lossiness when converting between numpy objects and python builtins.

On the other hand, trying to "faithfully" serialize and deserialize a numpy array brings us through this annoying bytes-based intermediate, which _doesn't actually serialize_ in some formats, so... any solution that _actually serializes_  is good by me.

Honestly, `serialized = pickle.dumps(numpy_array, protocol=0)` (from the link above) might gve us the string-serializable representation we want.

The important places to modify for this would be in `openforcefield/utils/utils.py`'s `{de,}serialize_numpy` functions, and `FrozenMolecule._initialize_from_dict`

mattwthompson:
> We should add tests to/from XML

I can test that `to_xml` doesn't raise an error, but `from_xml` is not implemented and would probably be necessary to test the validity of the XML.

> JSON serialization bug

I prefer this be fixed in a separate PR; here I just wanted to add the tests in a small-ish PR

mattwthompson:
I've added a test for `to_xml` and updated docstrings to be more clear and with fewer typos (though something probably slipped through)

Issue #381, opened 2019-07-12
Can't Parameterize Molecules With Zero Partial Charges


SimonBoothroyd:
**Describe the bug**
Trying to parameterize a molecule which should have zero partial charges will raise an exception.

**To Reproduce**
```
from openforcefield.topology import Molecule, Topology
from openforcefield.typing.engines.smirnoff import ForceField

molecule = Molecule.from_smiles('N#N')
topology = Topology.from_molecules([molecule])

force_field = ForceField('smirnoff99Frosst.offxml')
force_field.create_openmm_system(topology)
```

**Output**
```
    force_field.create_openmm_system(topology)
  File "/XXX/lib/python3.7/site-packages/openforcefield/typing/engines/smirnoff/forcefield.py", line 1128, in create_openmm_system
    parameter_handler.create_force(system, topology, **kwargs)
  File "/XXX/lib/python3.7/site-packages/openforcefield/typing/engines/smirnoff/parameters.py", line 2447, in create_force
    temp_mol.compute_partial_charges_am1bcc(toolkit_registry=toolkit_registry)
  File "/XXX/lib/python3.7/site-packages/openforcefield/topology/molecule.py", line 1991, in compute_partial_charges_am1bcc
    self
  File "/XXX/lib/python3.7/site-packages/openforcefield/utils/toolkits.py", line 2964, in call
    return method(*args, **kwargs)
  File "/XXX/lib/python3.7/site-packages/openforcefield/utils/toolkits.py", line 1252, in compute_partial_charges_am1bcc
    "Partial charge calculation failed. Charges from compute_partial_charges() are all 0."
Exception: Partial charge calculation failed. Charges from compute_partial_charges() are all 0.
```

**Computing environment (please complete the following information):**
 - Operating system: MacOS
 - Toolkit Version: 0.4.0

davidlmobley:
That's slightly tricky since normally we want to make sure partial charges are not all zero. 🤔 

Maybe we want to switch to checking that the partial charge calculation ran and assigned charges, but not check whether they are nonzero...? Or, check that the partial charges are nonzero except if they should be zero (not sure how to check if they should be zero though; it's possible one could do this using symmetry groups somehow, or otherwise just have a list of cases where charges should be zero...?).

dotsdl:
Hey @trevorgokey, I committed to taking this issue on along with #305 and #379. Is there any progress you'd like to report to help me get started, or is this perhaps complete?

j-wags:
[...]

PR #574, opened 2020-04-10
Removed check for all partial charges zero after assignment by quacpac

dotsdl:
Addresses #381.

We check for `quacpac` failure earlier, and we need to tolerate cases where the partial charges are legitimately all zero, such as for `N#N`.

The following now works without error:

```python

molecule = Molecule.from_smiles('N#N')
topology = Topology.from_molecules([molecule])

force_field = ForceField('test_forcefields/smirnoff99Frosst.offxml')

system
```
And we can see that we get zero partial charges on all atoms as expected for this molecule:
```python

molecule = Molecule.from_smiles('N#N')
topology = Topology.from_molecules([molecule])

molecule.generate_conformers()
molecule.compute_partial_charges_am1bcc()

molecule.partial_charges
````
yields:
```
Quantity(value=array([0., 0.]), unit=elementary charge)
```


dotsdl:
Thanks @mattwthompson!

I'm waiting for a response on [this question](https://github.com/openforcefield/openforcefield/issues/381#issuecomment-612158980). I do wonder if there are any cases where `quacpac_status` is `True` but charge assignment actually failed. If we have no reason to be concerned there, then we can proceed with this PR.

For tests, do we have any nontrivial example molecules that AM1BCC fails to parameterize, bonus points if they are known to have all zero partial charges? That last part may be overkill. I do know that say, a lithium ion will trigger failure of `quacpac`.

davidlmobley:
I've never seen a case where quacpac returned True but charge assignment failed. One benefit of commercial code is that this type of thing tends to be well-behaved and problems will be dealt with severely.

However:
> For tests, do we have any nontrivial example molecules that AM1BCC fails to parameterize, bonus points if they are known to have all zero partial charges? That last part may be overkill. I do know that say, a lithium ion will trigger failure of quacpac.

I do not think we have any molecules that AM1BCC fails to parameterize. So far, except for the case of eg monatomic ions and metals, this seems to be a good "chemistry sanity check"; I am not currently aware of any cases of "well formed" molecules where the OpenEye AM1BCC fails. It does fail on some protonation states/tautomers we've occasionally enumerated but those tend to be "obviously wrong" ones. In other words, all the failures I've examined so far (and I've looked at a ton of them) have seemed to be cases of it filtering out bad chemistry.

davidlmobley:
(Let me know if you want me to pull some of these for you; I could dig some up from my notes somewhere.)

dotsdl:
Thanks @davidlmobley! I think that's more than satisfactory an explanation for us to proceed. I don't see the need for an explicit unit test here, as this runtime check we are removing appears both unnecessary and unreliable.

I'll add a changelog entry, and that should be all that's needed for this. @mattwthompson are you satisfied with this approach?

mattwthompson:
Yes, I think this should be good to go; if a user would have to intentionally try very hard to break AM1BCC, that seems like something we don't need to worry about now.

dotsdl:
Great! Thanks @mattwthompson! Merging!

Issue #576, opened 2020-04-13
Conda install (@ 2020, Apr. 13) automatically install version 0.04 (the oldest version)


Kael-the-Mad:
**Describe the bug**
Installation Version Error

**To Reproduce**
On a CentOS 7.6 x64 system, with newly installed Anaconda3, install the package following the instruction will lead to version 0.04 installed. 
This could be an Anaconda (Cloud) setting problem.

(Time: 2020, Apr 13)

jaimergp:
This can totally be due to the new `ambermini` restrictions introduced in https://github.com/conda-forge/ambertools-feedstock/pull/29.

Can you paste the output of `conda list`?

Kael-the-Mad:
Here is the output of `conda list`:
# packages in environment at /home/kael/anaconda3:
#
# Name                    Version                   Build  Channel
_ipyw_jlab_nb_ext_conf    0.1.0                    py37_0  
_libgcc_mutex             0.1                        main    conda-forge
alabaster                 0.7.12                   py37_0  
ambertools                19.0                          0    ambermd
anaconda                  2020.02                  py37_0  
anaconda-client           1.7.2                    py37_0  
anaconda-navigator        1.9.12                   py37_0  
anaconda-project          0.8.4                      py_0  
argh                      0.26.2                   py37_0  
asn1crypto                1.3.0                    py37_0    conda-forge
astroid                   2.3.3                    py37_0    conda-forge
astropy                   4.0              py37h7b6447c_0  
atomicwrites              1.3.0                    py37_1  
attrs                     19.3.0                     py_0    conda-forge
autopep8                  1.4.4                      py_0    conda-forge
babel                     2.8.0                      py_0    conda-forge
backcall                  0.1.0                    py37_0  
backports                 1.0                        py_2    conda-forge
backports.functools_lru_cache 1.6.1                      py_0    conda-forge
backports.shutil_get_terminal_size 1.0.0                    py37_2  
backports.tempfile        1.0                        py_0    conda-forge
backports.weakref         1.0.post1       py37hc8dfbb8_1001    conda-forge
beautifulsoup4            4.8.2                    py37_0    conda-forge
bitarray                  1.2.1            py37h7b6447c_0  
bkcharts                  0.2                      py37_0  
blas                      1.0                         mkl  
bleach                    3.1.0                    py37_0  
blosc                     1.16.3               hd408876_0  
bokeh                     1.4.0                    py37_0    conda-forge
[...]

PR #577, opened 2020-04-13
Test installation from released conda packages

jaimergp:

Addressing #576


j-wags:
This is awesome, Jaime. Thanks for starting to set this up.

One question: Will this fail if the conda install can't proceed? Or should we add a pytest run to this?

jaimergp:
If [this line](https://github.com/openforcefield/openforcefield/pull/577/files#diff-e515a69767202017c75cbe7633201ed4R60) fails, the workflow will be reported as failed. But it might be also a false positive!

One of my concerns (and the reason why I decided to set this up) it's that dependencies such as `ambertools` introduce restrictions on their dependencies too, and that creates conflicts with `openforcefield` itself. Once way `conda` can solve this is by using older packages which didn't feature such restrains. My idea here is to test that the installed version (as reported by conda list) matches the latest released version (as featured in the git tag), but I got slightly frustrated as I faced silly problems that made the testing code inelegant 😬 

I might try again during the week, but feel free to do it if you want this merged sooner!

j-wags:
Not yet ready for review. I'm having trouble with the shell macros that
* Ensure the conda package installed is the same version as the highest repo tag
* Pull down tests corresponding to that tag (and not just whatever's in master, since that will have new tests)

If someone wants to bang on this a bit more I think it's close to ready.

.github/workflows/conda.yml
```diff
@@ -0,0 +1,106 @@
+name: Conda
+
+on:
+  pull_request:
```
mattwthompson, line 4: I did this so it would actually run while working on it, this should be dropped before merging

.github/workflows/conda.yml
```diff
@@ -0,0 +1,116 @@
...
+      OE_LICENSE: ${{ github.workspace }}/oe_license.txt
+      PACKAGE: openforcefield
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Wait for 2 hours if triggered by a release
+        if: (github.event_name == 'release') || (github.event_name == 'prerelease')
+        run:
+            sleep 7200
```
mattwthompson, line 41: This wait time is conservative but I'd like to maximize the chance that a release actually is up on Anaconda Cloud (build time + ~1  hour for their CDN to update + a larger buffer)

It should not run into walltime limits (6h and 72h limits for single jobs & workflows): https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits

j-wags: I could take this or leave it. It's a nice concept in theory, but will likely spell confusion in the case of a rolled-back or held-up release ("we rolled everything back an hour ago -- what's this failing test? Is something really wrong?")

mattwthompson: Good point - if we only have this running as a ~nightly job, we'd still be able to catch breaking changes in dependencies but it would not interfere with the release process.

j-wags (commented):
The `@requires_openeye` tests don't seem to be running for some reason.

[In a normal OE build, we see 16185 tests pass](https://github.com/openforcefield/openforcefield/pull/694/checks?check_run_id=1038650722#step:12:500)

In both the [with-OE](https://github.com/openforcefield/openforcefield/runs/1056761540?check_suite_focus=true#step:10:384) and [without-OE](https://github.com/openforcefield/openforcefield/pull/577/checks?check_run_id=1056761580#step:10:384) environments here, we see 1513 tests pass.

The conda install shows OE getting installed, so maybe the OE license isn't being decrypted correctly/to the right file?

j-wags:
Hmm, [the value of the `OE_LICENSE` environment variable](https://pipelines.actions.githubusercontent.com/u2Wf9sTwFK9NZj8rRqcrkPO1UF3tPeGIdu4NysSHujaRyWbUAY/_apis/pipelines/1/runs/980/signedlogcontent/4?urlExpires=2020-09-03T02%3A38%3A46.6694340Z&urlSigningMethod=HMACV1&urlSignature=587l71rn90KIg0MbGSNp%2BAPEQ6uPoSTCKvL8kx4%2FCFc%3D) seems to be reasonable
```
2020-09-01T14:29:34.0442693Z   OE_LICENSE: /home/runner/work/openforcefield/openforcefield/oe_license.txt
```

Not sure what else it might be.

mattwthompson:
Nice catch! Since this approach involves manually cloning the repo, it's probably decrypting it to the right place earlier in the workflow but not where it unpacks the installer.

mattwthompson:
Looks like I never set `$OPENEYE` correctly; see [here](https://github.com/openforcefield/openforcefield/pull/577/commits/cd3a6feecea2a509460f1d5e30bb753b31218a00) or, if the log doesn't persist, below:

```
Run uname -a
  uname -a
  df -h
  ulimit -a
  shell: /bin/bash -l {0}
  env:
    CI_OS: ubuntu-latest
    OPENEYE: 
    PYVER: 3.7
    OE_LICENSE: /home/runner/work/openforcefield/openforcefield/oe_license.txt
    PACKAGE: openforcefield
    CONDA_PKGS_DIR: /home/runner/conda_pkgs_dir
```

If the last commit fixes this, I will mandate that including the "Additional info about the build" step be included in all of our actions. If that wasn't included, I would have never caught this, since GitHub never yelled about evaluating a conditional against a variable that doesn't seem to exist

jaimergp:
There are two ways to re-run failed workflows:

1. This button on the `Details` page of every run:
image

2. Pushing an empty commit with `git commit --allow-empty -m "Re-run workflows"`. No need to add spaces anywhere!

j-wags:
Thank you for the correction, and not judging me on my ignorance @jaimergp :-)

mattwthompson:
If it looks like I've given up on this and somebody else would like to pick it up: the above issue of OpenEye tests not running seems to persist

mattwthompson:
I _think_ this is finally working now
* All ✅ 
* ~15,000 tests run on OpenEye-enabled builds
* The printed versions versions seem to match

I'll leave in all the messy debugging lines during review for the sake of being explicit, but they'll be removed later.

.github/workflows/conda.yml
j-wags: Hm, I think this merging this PR has a lot of value even without figuring out exactly how it will slot into our release process. So I'd push to merge without resolving this question.

But, that said, I think our release process should probably unconditionally push a package to `omnia/label/rc`, and the release tests should target the package built on that label. So I _do_ think there should be release automation, but it should be a different workflow, which may or may not use this file.

.github/workflows/conda.yml
j-wags:
```suggestion
```

j-wags (approved):
I think this is ready to go! Thanks a million to @jaimergp and @mattwthompson for leading this in various stages :-) This should make it MUCH easier and less stressful to have high uptime, and hopefully this can serve as a template for other repos!

PR #578, opened 2020-04-13
Ignore DOI url tests

j-wags:
DOI.org is now protected behind cloudflare DDOS protection. This PR skips tests of DOI.org links.


j-wags:
Thanks for the quick review, @mattwthompson!

PR #579, opened 2020-04-14
Write multi-conformer PDBS with RDKit

### LINKED ISSUE #568: RDKit cannot write multi-model PDB files
I'm not the PDB wizard that others here are, but @jaimergp [suggested](https://github.com/openforcefield/openforcefield/pull/560#issuecomment-608375591) it's a bug and it seems to be the case. It's probably not too important in the big picture.

MWE, which works fine with `OpenEyeKitToolkitWrapper`:
```
In [21]: mol = Molecule.from_smiles('CCOCOCOCCOCCOCOCC')                                       

In [22]: mol.generate_conformers()                                                             

In [23]: mol.to_file('rdkit.pdb', file_format='pdb', toolkit_registry=RDKitToolkitWrapper())   
<ipython-input-23-6634d46173bd> in <module>
----> 1 mol.to_file('rdkit.pdb', file_format='pdb', toolkit_registry=RDKitToolkitWrapper())

~/software/openforcefield/openforcefield/topology/molecule.py in to_file(self, file_path, file_format, toolkit_registry)
   3577         if type(file_path) == str:
   3578             # Open file for writing
-> 3579             toolkit.to_file(self, file_path, file_format)
   3580         else:
   3581             toolkit.to_file_obj(self, file_path, file_format)

~/software/openforcefield/openforcefield/utils/toolkits.py in to_file(self, molecule, file_path, file_format)
   2006         # open a file object and pass to the object writer
   2007         with open(file_path, 'w') as file_obj:
-> 2008             self.to_file_obj(molecule=molecule, file_obj=file_obj, file_format=file_format)
   2009 
   2010     def canonical_order_atoms(self, molecule):

~/software/openforcefield/openforcefield/utils/toolkits.py in to_file_obj(self, molecule, file_obj, file_format)
   1979         try:
   1980             writer = self._toolkit_file_write_formats[file_format](file_obj)
-> 1981             writer.write(rdmol)
   1982             writer.close()
   1983         # if we can not write to that file type catch the error here

ValueError: Bad Conformer Id
```
_Originally posted by @mattwthompson in https://github.com/openforcefield/openforcefield/pull/560/files_

jthorton:
- [X] Fix #568 

This should fix RDKit and allow it to write multi-conformer PDB files, also added a test for this.

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -55,12 +55,12 @@ Behavior changed
 API-breaking changes
 """"""""""""""""""""
 - `PR #558 <https://github.com/openforcefield/openforcefield/pull/558>`_: Removes
-  ``TopologyMolecule.topology_particle_start_index``, since the :py:class`Topology <openforcefield.topology.Topology>`
+  ``TopologyMolecule.topology_particle_start_index``, since the :py:class:`Topology <openforcefield.topology.Topology>`
   particle indexing system now orders :py:class`TopologyVirtualSites <openforcefield.topology.TopologyVirtualSite>`
   after all atoms.
-  :py:meth`TopologyMolecule.topology_atom_start_index <openforcefield.topology.TopologyMolecule.topology_atom_start_index>`
+  :py:meth:`TopologyMolecule.topology_atom_start_index <openforcefield.topology.TopologyMolecule.topology_atom_start_index>`
   and
-  :py:meth`TopologyMolecule.topology_virtual_site_start_index <openforcefield.topology.TopologyMolecule.topology_virtual_site_start_index>`
+  :py:meth:`TopologyMolecule.topology_virtual_site_start_index <openforcefield.topology.TopologyMolecule.topology_virtual_site_start_index>`
   are still available to access the appropriate values in the respective topology indexing systems.
 - `PR #508 <https://github.com/openforcefield/openforcefield/pull/508>`_:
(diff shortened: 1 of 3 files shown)


jthorton:
This has made me realise I don't think we test all of the possible file output formats which I could add test for here if required?

mattwthompson:
Which formats are you referring to? We are in the process of adding some in #569

jthorton:
Ahh yes I mean backend toolkit formats more like PDB and SDF I know SDF is more heavily tested now that the property PR is merged.

j-wags:
> This has made me realise I don't think we test all of the possible file output formats which I could add test for here if required?

I'm starting to appreciate how "output formats" are _complicated._ That is, writing to a particular output format isn't just "making a valid file within that format's specification", but rather "deciding on exactly which characteristics of a molecule we guarantee to preserve in a round trip to that format". 

So, this doesn't mean "don't add tests for new formats" -- after all, we do "support" them in the API, and testing is better than no testing. I mean something more like "don't be surprised if this task is a rabbit hole, and we end up with a big table of what we do and don't preserve through trips to different formats"

So, this PR shouldn't add tests for every output format, because it will take a long time.

j-wags:
Oh, and please update the release notes before you merge!!

jthorton:
Good point I had a feeling it may be a bigger job than I expected Ill get this finished up today.

PR #583, opened 2020-04-21
Fix toolkit inheritance in classmethods

jaimergp:
I am subclassing `Molecule` in `openkinome/kinoml` to add some provenance functionality to the initializers (I want to preserve the original input SMILES, for example), but I found that the types were not consisting and that some attributes I have added in the subclass could be lost randomly.

I found out the problem resides in how `Molecule` delegates some calls (`classmethods`) to the underlying toolkit. We need to ensure that the same `Molecule` constructor (often itself, but it can also be a subclass!) is used when we need to create a copy of the current molecule or initialize a new one from another source (smiles, inchi, file, etc). Basically, I browsed for occurrences of `Molecule(...)` in `toolkits.py` and then tried to tied to patch all the methods that rely on those, directly or indirectly.

I think I have solved my problem in `kinoml`, and the changes should be transparent to the current toolkit because if `_cls` is None, it defaults to the OpenForceField Molecule anyway.

Tests and documentation for developers are pending, but let's see if this is a good idea or not first!


jaimergp:
One of the most common errors is:

```
AttributeError: 'FrozenMolecule' object has no attribute 'add_atom'
```

Since now the toolkits reuse the class of the passed molecules (look for `_cls=molecule.__class__`) to create copies of the passed molecule (as IMO it should), this introduces problems when `FrozenMolecule` requires further intermediates that need some kind of modification (hence the missing `add_atom`).

openforcefield/topology/molecule.py
```diff
@@ -3288,8 +3292,8 @@ def to_iupac(self):
         from openeye import oeiupac
         return oeiupac.OECreateIUPACName(self.to_openeye())
 
-    @staticmethod
-    def from_topology(topology):
+    @classmethod
+    def from_topology(cls, topology):
```
j-wags, line 3296: Reference molecules in a topology are (currently) always `FrozenMolecule`s, so this should remain as a staticmethod of the `FrozenMolecule` class.

j-wags: Actually, I take that back. We can always upcast it to whatever the user wants, as long as the constructor for the subclass can take a `FrozenMolecule` as input.

j-wags:
Thanks. I think this is a good change, and it is essential for extensibility of our Molecule class.

A few points, in descending order of concreteness:
* Looks like we've got two possible solutions to the `FrozenMolecule.add_atom` issue. As a downstream developer, you may have some helpful perspective on which is better for you:
    1) Have `from_rdkit` and `from_openeye` ALWAYS make an OFF `Molecule`, and then cast the final result to whatever subclass is specified in `cls` right before returning (`return cls(mol)`)
    2) Just use the private API to FrozenMolecule -- `FrozenMolecule._add_atom` etc. (this seems dangerous, since people making subclasses won't expect to need to change this) 
* In the current implementation, class is always a private argument (`_cls`). I think this should be public (`cls` or the more descriptive `return_molecule_class`)
* For `ToolkitWrapper.from_X` methods, I agree that the default value for the `cls` kwarg should be `openforcefield.topology.Molecule`
* `Molecule` and `ToolkitWrapper`/`Registry` are highly interconnected. I wonder if there's some architectural change we should make in the `ToolkitX` classes to go along with this. Generally, the design pattern I'd like to move towards for methods like `Molecule.some_method` (which just immediately redirect to `ToolkitRegistry.call`-->`ToolkitWrapper.some_method`) is for `ToolkitWrapper.some_method` to be an _instance method_, so that we can drop in `CachingToolkitWrapper` classes later. So, we'll want to ensure that the `type` objects passed in through the new `class` kwargs are hashable, but I think they are, so that should be fine.

jaimergp:
Thanks for the review, Jeff!

> Looks like we've got two possible solutions to the FrozenMolecule.add_atom issue.

I'd say that if a class is requesting FrozenMolecules but then it needs to use `add_atom`, then it shouldn't be using FrozenMolecules to begin with.

> or the more descriptive return_molecule_class

Agreed! I didn't want to introduce API changes (or additions) as part of the initial draft. 

> I wonder if there's some architectural change we should make in the ToolkitX classes to go along with this

I will need more details on the last point to comment on that.

j-wags:
Per conversation between @jaimergp and myself this morning, there's no perfect solution to this, but the best thing to do is to change `from_openeye`, `from_rdkit` , and any other `ToolkitWrapper` methods that need to build up molecules atom-by-atom, to call the *private* `_add_atom` method. Other than that, the changes in this PR look good. I'll accept this once tests pass, and slate for inclusion in 0.7.1, due in 4 or 5 weeks.

mattwthompson:
* ced31fa (merging `master` back in) was a pretty hefty and conflicted merge commit, so if something looks out of place it's likely an error on my part that should be reverted to make sure I didn't delete useful code.
* Tests are still failing but not as much as they were a few hours ago. The `ChargeIncrementModelHandler` PR in particular changed a lot of the partial charge internals. 
* (out of scope) Would these changes need to be made to other classes, i.e. `Topology` to ensure this sort of behavior in subclasses? 
* I share @j-wags's feeling that there may be a better way to get toolkit wrappers and topology objects and methods to play well with each other, ideally also removing some duplicated code, but I have no suggestions on how to execute that right now

> the least bad way to get this working was to use the private _add_atom method, so that we're able to build up FrozenMolecules using from_rdkit and such

As far as I can tell, there are only two such calls to `.add_atom` in the toolkit wrappers, and they're using the public methods? Does this only not seem like a problem since it's not explicitly tested for?


To get this merged, I think we should target the following

- [ ] All existing toolkit tests passing
- [ ] Relevant developer docs (a paragraph or two including a combination of the what/why of this PR, information on how to do something similar if it pops up in other classes, and/or anything a developer subclassing `Molecule` should know)
- [ ] Add some tests that check the behavior a `Molecule` subclass

would the above satisfy what `openkinome` and `kinoml` need @jaimergp?

jaimergp:
I am going to run our test suite with this new commit and report back, thank you so much!

mattwthompson:
Merged `master` back in after the big lintings - I tried to be careful but it's possible I botched a few lines along the way

mattwthompson:
Many tests are still failing, probably due to incorrect behavior in one or a small number of functions. It's not obvious to me at the moment, but if somebody wants to dig into it, it's definitely related to charge assignment. This should be close to passing, I just can't isolate the issue at the moment.

j-wags:
@jaimergp -- Just a ping on this. We'd be happy to merge if this covers your use case. Just let us know :-)

jaimergp:
I have been overriding the conda releases with pip installs from this branch for a while now, so it _should be_ ok! I will do a deeper assessment in the following days just to make sure, and after that we can merge. Thanks!

j-wags:
I'm gonna go ahead and move this towards review+merge, since it seems like the changes are working for you :-)

jaimergp:
Yes, sorry @j-wags! We have been working with this branch for a while and no issues so far!

j-wags:
Still working on documenting this, but tests are _functionally_ passing right now. The QCFractal errors are because the MolSSI server is down for maintenance.

j-wags (approved):
I've patched up the developer docs and test coverage, and fixed a few molecule creation pathways that didn't yet use the new model. Thanks everyone who worked on this!

j-wags:
@mattwthompson Could you take a quick peek at the new tests added to `test_molecule`? They're my biggest contribution to this PR, and so it'd be good to have a second set of eyes/sanity check on them.

mattwthompson:
Will do!

openforcefield/tests/test_molecule.py
```diff
@@ -2882,3 +2887,87 @@ def test_visualize_openeye(self):
         mol = Molecule().from_smiles("CCO")
 
         assert isinstance(mol.visualize(backend="openeye"), IPython.core.display.Image)
+
+
+class MyMol(FrozenMolecule):
+    """
+    Lightweight FrozenMolecule subclass for molecule-subclass tests below
+    """
+
```
mattwthompson, line 2896: I don't know if it consequential that this test class inherits from `FrozenMolecule` vs. `Molecule`; I can't immediately think of a way to break it.

```
In [12]: class MyOtherMol(Molecule):
    ...:     pass
    ...:

In [13]: isinstance(MyOtherMol.from_smiles("CCO"), Molecule)
Out[13]: True

In [14]: isinstance(MyOtherMol.from_smiles("CCO"), FrozenMolecule)
Out[14]: True
```

docs/releasehistory.rst
```diff
@@ -22,6 +22,18 @@ API-breaking changes
   there are no special behaviors that are accessed in the case of partially-licensed OpenEye backends. The
   new behavior of this method is the same as if the default value above is always provided.
 
+Behavior Changed
+""""""""""""""""
+- `PR #583 <https://github.com/openforcefield/openforcefield/pull/583>`_: Methods
+  such as :py:meth:`Molecule.from_rdkit <openforcefield.topology.Molecule.from_rdkit>`
+  and :py:meth:`Molecule.from_openeye <openforcefield.topology.Molecule.from_openeye>`,
+  which delegate their internal logic to :py:class:`ToolkitRegistry <openforcefield.utils.ToolkitRegistry>`
+  functions, now guarantee that they will return an object of the correct type. Previously,
```
mattwthompson, line 31: nit: my first reading of this make me wonder if the changed behavior was applicable to `Molecule` itself, not only its subclasses
```suggestion
  functions, now guarantee that they will return an object of the correct type when being called on ``Molecule``-derived classes. Previously,
```

mattwthompson (approved):
LGTM - I only briefly skimmed files not named `test_molecule.py`; not sure how much changed in the toolkit wrappers since I last looked at these files a few months ago

openforcefield/tests/test_molecule.py
j-wags: It's a bit consequential since `FrozenMolecule` doesn't expose `add_atom`, but instead only exposes the private `_add_atom` method (same with `_add_bond` and `_add_conformer`). So I think it's better this way, since using `FrozenMolecule` for these tests actually makes them stricter than just vanilla `Molecule`.

PR #586, opened 2020-04-23
Add warnings to smirnoff99Frosst FFs to not use them in production

j-wags:
This PR adds warnings as comments to several OFFXML files in the `test_forcefields` directory.

Due to some history, we have a few files named as variants of `smirnoff99Frosst` in this repo. Some folks use these for tests and examples, so we shouldn't remove them. However, they have been hand-modified, and their contents don't match up to any published smirnoff99Frosst FFs. This makes them unsuitable for use in scholarly work.


j-wags:
That's not a bad idea, @mattwthompson, but I am concerned that we are too warning-heavy already. Maybe we should have a process where, if we're adding a warning somewhere in the codebase, we must remove a warning from somewhere else. I'll merge this for now, but would be open to suggestions for how to better drive correct user behavior.

Issue #561, opened 2020-04-02
Split CI tests


jaimergp:
# Is your feature request related to a problem? Please describe.

Current CI logs for the actual test is a single category with a loooong list of items addressing different aspects of the codebase and its quality: unit tests, documentation, examples, notebooks...

# Describe the solution you'd like

GH Actions provide UI elements to easily distinguish categories and concerns. We can have multiple _workflows_ (separate YML files), _jobs_ (separate elements in the PR UI), and _steps_ (collapsible blocks within each job). Each of these elements gets a failed/success status, so we could easily see what part is failing. We can benefit from that visual aid if we split our current monolithic "test everything" step. My proposal (hierarchy is workflow > step)

* **Testing**. This workflow will have several _jobs_, one for each entry in the build matrix. Each _job_ would have these testing _steps_.
    * Unit tests: guarantee the library is working as expected (this is the main category, and the only one that runs across all entries in the build matrix)
    * Run examples and notebooks
* **Documentation**. Single job per workflow, so it runs only once per push. Steps would include:
    * Ensure sphinx builds (no deploy, just run the build part to make sure we don't break it silently)
    * Ensure all links are alive (might fail for network errors and this is not critical)
* **Style**. Single job per workflow, so it runs only once per push. Steps would include:
    * Run some linters (already covered by LGTM?)
    * Run formatting checks

# UI examples

Different workflows / jobs:

![image](https://user-images.githubusercontent.com/2559438/78231445-18ba3780-74d3-11ea-9938-6e6c37981da7.png)

Different steps within a job:

![image](https://user-images.githubusercontent.com/2559438/78231473-253e9000-74d3-11ea-8301-e51d7a1fb5b7.png)

jaimergp:
I'll use the upcoming PR to add the README badges too, which I forgot in #554 😬

mattwthompson:
Looks like we mostly pulled this off:

* **Testing**: Tests are split up between unit tests, example notebooks, checking README links, and running code snippets in docs: https://github.com/openforcefield/openforcefield/blob/77b17c71d58f5e95808d0d984a42bf812a2b4040/.github/workflows/CI.yml
* **Documentation**: Not automated here, as far as I can tell.
* **Linting**: Minimal linting action (using only `pip`!): https://github.com/openforcefield/openforcefield/blob/77b17c71d58f5e95808d0d984a42bf812a2b4040/.github/workflows/lint.yml

mattwthompson:
[...]

PR #588, opened 2020-04-28
Split CI in different steps and jobs

jaimergp:

Addresses #561 

Proposed split is:
* Matrix job (runs on several configs):
    * Unit tests + Examples (two separate steps for easy diagnosis)
* Single config job (Ubuntu, Py37, with rdkit and openeye): 
    * Lint checks (new! run `flake8`, `isort --check` and `black --check`, just to see how it feels)
    * Documentation (test readme links and build sphinx locally)

## Questions

- [ ] Do we want Black checks? If so, line length? Maybe the default 88?
- [ ] Do we want Flake8 checks? It's catching a bunch of (important, I'd say) stuff. We'd need to agree on a good config.
- [ ] Do we want to turn Sphinx warnings (there are a lot!) into errors?


mattwthompson:
>  Do we want Black checks? If so, line length? Maybe the default 88?
 Do we want Flake8 checks? It's catching a bunch of (important, I'd say) stuff. We'd need to agree on a good config.

I would personally like at least `flake8` to be implemented eventually and I agree that we need agree on the details before linting the existing code or enforcing it as a check. For example, I like prefer `black -S` but I may be alone in that. For now, can linters just operate on the diff?

Can we reach a consensus during the developer week?

SimonBoothroyd:
> For example, I like prefer black -S but I may be alone in that. For now, can linters just operate on the diff?

My +1 is to just stick directly to the black defaults and not make any config changes. "Black is opinionated so you don’t have to be."

openforcefield/tests/conftest.py
```diff
@@ -19,6 +19,15 @@
...
 import pytest
 
+def pytest_configure(config):
+    """
+    Initialization hook to register custom markers without a pytest.ini
+    More info: https://docs.pytest.org/en/latest/reference.html#initialization-hooks
+    """
+    config.addinivalue_line(
+        "markers", "slow: marks tests as slow (deselect with `-m 'not slow'`)"
+    )
```
j-wags, line 29: 🦜

j-wags: (To clarify -- The parrot is a very good thing)

mattwthompson:
Okay, I've scaled this back a little just to get it over the finish line (linting will still happen, just splititng this off into smaller PRs)

* Uses miniconda action
* Install RDKit vs OpenEye array using separate files, not lines in script
* Includes `openeye-toolkits=2019` pin
* Split off tests into a few smaller steps

It's tempting to further split off the unit tests (one for each of the main modules) but I can load up the unit test run without my browser freezing up, so consider that a win.

j-wags:
I realized that this is a lot of really good stuff, but I have just been flying from fire to fire the last few weeks and haven't reviewed this or had bandwidth to discuss the questions above. These changes will make everyone's lives a lot easier, so I don't want to block this any further.

I'm removing myself as reviewer. @mattwthompson can approve+merge this at his discretion! Thanks everyone for the hard work here.

PR #591, opened 2020-05-03
Add minimization and CLI to conformer energy example

j-wags:
- [x] Add minimization to conformer energy notebook
- [x] Make conformer energy notebook handle SDF input consistently, _even in 0.6.0 toolkit,_ by forcing use of RDKit backend for molecule I/O
- [x] Have notebook output CSV with energies and RMSDs
- [x] Offer CLI to notebook funcitonality (by copying code into a .py script and adding argparse)
- [ ] Update changelog


examples/conformer_energies/conformer_energies.py
```diff
@@ -0,0 +1,93 @@
...
+    from openforcefield.typing.engines.smirnoff import ForceField
+    forcefield = ForceField('openff_unconstrained-1.1.0.offxml')
+    # Loop over molecules and minimize each conformer
+    for molecule in molecules:
+        # If the molecule doesn't have a name, set mol.name to be the hill formula
+        if molecule.name == '':
+            molecule.name = Topology._networkx_to_hill_formula(molecule.to_networkx())
+            print('%s : %d conformers' % (molecule.name, molecule.n_conformers))
+            # Make a temporary copy of the molecule that we can update for each minimization
+        mol_copy = Molecule(molecule)
```
jchodera, line 40: You might refactor everything from here into a method `compute_conformer_energies(molecule)` since that would be a simple, reusable method that others can copy and use.

examples/conformer_energies/conformer_energies.py
```diff
@@ -0,0 +1,93 @@
...
+        system = forcefield.create_openmm_system(off_top)
+        # Use OpenMM to compute initial and minimized energy for all conformers
+        integrator = openmm.VerletIntegrator(1*unit.femtoseconds)
+        platform = openmm.Platform.getPlatformByName('Reference')
+        omm_top = off_top.to_openmm()
+        simulation = openmm.app.Simulation(omm_top, system, integrator, platform)
+
+        # Print text header
+        print('Conformer         Initial PE         Minimized PE       RMS between initial and minimized conformer')
+        output = [['Conformer','Initial PE (kcal/mol)','Minimized PE (kcal/mol)','RMS between initial and minimized conformer (Angstrom)']]
```
jchodera, line 53: You might have the method return the conformer energies and then just iterate through them here in writing out a new SDF.

examples/conformer_energies/conformer_energies.py
```diff
@@ -0,0 +1,93 @@
+import argparse
+from openforcefield.topology import Molecule, Topology
+from openforcefield.utils import RDKitToolkitWrapper
+from rdkit.Chem import rdMolAlign
+import numpy as np
+from simtk import openmm, unit
+
+def main(filename):
```
jchodera, line 8: Maybe this can be `compute_conformer_energies_from_file(filename)`?

examples/conformer_energies/conformer_energies.py
```diff
@@ -0,0 +1,93 @@
...
+            # Write the results out to CSV
+        with open(f'{molecule.name}.csv', 'w') as of:
+            for line in output:
+                of.write(','.join(line)+'\n')
+                # Clean up OpenMM Simulation
+        del simulation, integrator
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description="Perform energy minimization on a molecule, potentially with many conformers. For each conformer, this script will provide the initial energy, minimized energy, and RMSD between the initial and minimized structure (both as STDOUT and a csv file). The minimized conformers will be written out to SDF.")
+    parser.add_argument('-f', '--filename', help='Name of an input file containing conformers')
```
jchodera, line 90: I'd allow arguments like `--molecules molecules.sdf` for input molecules, `--output output.{sdf|csv}`, and optionally `--forcefield openff-1.0.0`.

mattwthompson:
Runs for me out of the box; not sure if the order of magnitude difference in minimized potential energies is a concern?


```
ruxolitinib : 10 conformers
Parametrizing ruxolitinib (may take a moment to calculate charges)
Conformer         Initial PE         Minimized PE       RMS between initial and minimized conformer
    1 /    10 :   34.921 kcal/mol    4.845 kcal/mol     0.836 Angstroms
    2 /    10 :   43.903 kcal/mol    6.026 kcal/mol     0.514 Angstroms
    3 /    10 :   79.537 kcal/mol    5.875 kcal/mol     0.924 Angstroms
    4 /    10 :   40.710 kcal/mol    5.825 kcal/mol     0.736 Angstroms
    5 /    10 :   41.273 kcal/mol    5.928 kcal/mol     0.976 Angstroms
    6 /    10 :   43.646 kcal/mol    4.272 kcal/mol     0.645 Angstroms
    7 /    10 :   42.257 kcal/mol    5.953 kcal/mol     1.118 Angstroms
    8 /    10 :   32.198 kcal/mol    4.852 kcal/mol     0.563 Angstroms
    9 /    10 :   38.401 kcal/mol    4.839 kcal/mol     1.028 Angstroms
   10 /    10 :   41.009 kcal/mol    3.888 kcal/mol     0.653 Angstroms

jchodera:
> not sure if the order of magnitude difference in minimized potential energies is a concern?

Nope! That's to be expected.

j-wags:
Thanks for the review @jchodera. I think I'm out of gas today, but I'll try to get to this tomorrow afternoon. If anyone has some energy today, I'd be happy for them to add those changes.

@mattwthompson Yeah, I left a comment about that in the code. The absolute energies are really different between OE and AmberTools, but the relative energies are pretty close. I checked into the details of the OpenMM system, and they both seem to be using non-periodic electrostatics, and most atoms have partial charges within 0.1 elementary charge of each other. So, I'm not totally sure what's up. It could be that rounding errors in AmberTools are adding up to leave a noticeable net charge on the system, and that's raising the absolute potential energy.

j-wags:
I've updated the example to store the molecule in package data, and copy it to the example folder at runtime. This is important because the example tests do not run in the individual example folders. 

The next thing to do is to disable testing on this example in builds without RDKit. I'm not sure if this is possible, and the example runs nearly identically with OpenEye, so we could `try`/`except` the toolkitwrapper import and molecule loading, so that it _will run with openeye if there no other option,_ just a bit more dangerously until 0.7.0.

mattwthompson:
Okay, I've gotten it to a state in which the tests all pass. It took a little hardcoding but it will be instructive for future tests that may require, as this does, particular files/arguments passed and skips based on toolkit installs. Next I'll do the suggest refactors and it should be good to review.

j-wags:
Awesome. Thanks for all the work on this, @mattwthompson!

mattwthompson:
So I worked a little bit on some of the more robust CLI suggestions from @jchodera and I'm currently inclined to push for this to be reviewed and merged as-is. This is to split up this PR into smaller, more distinct objectives, and so that we can reach more of a consensus on how we want to build CLI functionality moving forward, in order to do it in a more common and unified way. The `argparse` work here is a good starting point in the sense that it took a relatively small amount of work and got it to work on different molecules.

jchodera:
@mattwthompson : I absolutely agree that this simple example should just use `argparse`.

I only suggested we may want to think about our CLI strategy for the premier CLI-enabled tools we build across OFF like `qcsubmit`, since if we design our approach well, we can re-use it in other areas to bolt on a CLI onto our well-crafted APIs as well. But examples like this are just meant to be lightweight copy-and-paste examples where users of the code are not expected to have to spend any time whatsoever to understand how the code works, so `argparse` is the right way to go!

j-wags:
I've made some of the internal refactors, but thinking through all the code paths for optional CSV/SDF output in all its combinations might take a little while. I'm going to merge this so we can get it into partners' hands sooner, but we can always add more options (like file prefix and FF selection) later.

PR #593, opened 2020-05-11
schema extras

jthorton:
This allows extras to be passed to the qcschema which will help with running MM through QCEngine.


openforcefield/tests/test_molecule.py
```diff
@@ -1341,7 +1341,7 @@ def test_to_qcschema(self):
         with pytest.raises(InvalidConformerError):
             qcschema = ethanol.to_qcschema(conformer=1)
         # now make a valid qcschema and check its properties
-        qcschema = ethanol.to_qcschema()
+        qcschema = ethanol.to_qcschema(extras={"test_tag": "test"})
```
j-wags, line 1344: Could you also add another few lines to ensure that we get the intended behavior if `extras` isn't specified?

jthorton: Good call Ill get them added.

PR #594, opened 2020-05-12
Pin sphinx 2.x.x since 3.x.x incompatible with m2r

j-wags:
Per https://github.com/sphinx-doc/sphinx/issues/7420

Quick fix to get docs building again. I'll add making a more permanent fix to our task list.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,7 +32,7 @@ import openforcefield
 #
 # Sphinx pin is because of m2r compatibility
 # https://github.com/openforcefield/openforcefield/pull/594
-needs_sphinx = '2'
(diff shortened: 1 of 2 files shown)


j-wags:
Docs build status is here https://readthedocs.org/projects/open-forcefield-toolkit/builds/11019249/

j-wags:
I've been searching through RTD's documentation, and it doesn't seem to offer an easy way to pin a sphinx version. The closest we could get is switching from a "Sphinx" build to a "Conda" one, but that would be a bit more work. This branch's history is pretty dirty so I'll close this and we can implement a better solution down the line.

PR #595, opened 2020-05-12
Drop temporary_directory/temporary_cd for tempfile

### LINKED ISSUE #503: [LINT] 'temporary_directory' duplicates 'tempfile.TemporaryDirectory'
**Describe the bug**
The `utils.temporary_directory` duplicates the `tempfile.TemporaryDirectory` context manager without providing the same level of flexibility.

`utils.temporary_directory` should likely be removed in favour of `tempfile.TemporaryDirectory`.

mattwthompson:
Fixes #503 

Arguably instead of just removing the functions, they could be replaced with some sort of `DeprecationWarning`. But downstream users are probably not importing these functions directly. Open to suggestions here


SimonBoothroyd:
@mattwthompson (cc @j-wags ) I don't think the `TemporaryDirectory` context manager also cd's into the temporary directory, so I think the changes you've made where you now remove the temporary cd will just create files in the cwd.

Also snippets like

```
        with tempfile.TemporaryDirectory() as tmpdir:
            outfile = 'temp_molecule.' + file_format
            self.to_file(molecule, outfile, file_format)
            file_data = open(outfile).read()
```

would likely be better to just use a [`NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile).

openforcefield/tests/test_examples.py
```diff
@@ -17,11 +17,12 @@
 import re
 import subprocess
 import textwrap
+import  tempfile
```
mattwthompson, line 20: Oops, thought I fixed this typo

mattwthompson:
```
>>> with tempfile.TemporaryDirectory() as tmpdir:
...     Molecule.from_smiles('CCO').to_file('tmp.sdf', file_format='sdf')
... 
>>> os.listdir()
['foo.txt', 'tmp.sdf', 'bar.txt']
```

That's right. I was fairly surprised the tests passed but that shouldn't have been the criteria anyway. I'll fix this tomorrow.

j-wags:
Ahhh, simon is correct above, just saw this and came to say the same thing -- There are still some places where we need to cd into the temporary directory. I'm adding `temporary_cd` back in in #471

SimonBoothroyd:
@j-wags https://github.com/SimonBoothroyd/nonbonded/blob/621151c747925aefeb368d0a2bc99c4305de6dd7/nonbonded/library/utilities/utilities.py#L34 may be helpful - it optionally temporarily cd's into a temporary directory.

j-wags:
Ah, that's really helpful. For now I'm going to revert it to two nested contexts (first make directory, then enter it), since that's more explicit about what's happening. I think this will continue to be a tricky area of the codebase, so "more explicit" will be good here.

Issue #324, opened 2019-05-17
cyclohexane_water packmol box leads to unstable simulation


j-wags:
## Expected Behavior
Parameterizing `systems/packmol_boxes/cyclohexane_water.pdb` (a box with one cyclohexane and a few hundred waters) should lead to a straightforward simulation.

## Actual Behavior

Taking steps of 1 or 2 femtoseconds, hydrogens on the cyclohexane molecule move rapidly and, after <10 steps, cause the simulation to crash. Investigating the parameters by eye shows the correct bond connectivity, with reasonable `k` and `d` values, and a quick look at angles and torsions doesn't show anything out of place either. The starting geometry of the cyclohexane looks reasonable, and measuring the bonds and angles numerically shows values close to equilibrium.

Taking steps of 0.1 femtoseconds, the simulation completes, but the cyclohexane has far more kinetic energy than other molecules in the system. Video below:

![cyclohexane_water_0p1fs](https://user-images.githubusercontent.com/16329175/57899927-c10bd280-7813-11e9-9f95-dd85d594a545.gif)

Removing all molecules except the cyclohexane from the starting PDB leads to an unremarkable simulation that is stable with steps of 2 fs.

Removing all molecules except the cyclohexane and one water from the starting PDB leads to an unremarkable simulation that is stable with steps of 2 fs.

## Ideas

Given that the simulation is totally stable when less particles are present, this may be a heating/thermostat issue. One thought is that the only `PeriodicTorsion`s in the system are in the cyclohexane, and maybe that's leading to an unequal distribution of energy.

## Minimal Reproducible Example
```
from simtk.openmm.app import PDBFile
from openforcefield.utils import get_data_file_path
pdb_file_path = get_data_file_path('systems/packmol_boxes/cyclohexane_water.pdb')
pdbfile = PDBFile(pdb_file_path)
from openforcefield.topology import Molecule
cyclohexane = Molecule.from_smiles('C1CCCCC1')
water = Molecule.from_smiles('O')
from openforcefield.topology import Topology
from openforcefield.typing.engines.smirnoff import ForceField

# Create the Open Force Field Topology from an OpenMM Topology object.
omm_topology = pdbfile.topology
off_topology = Topology.from_openmm(omm_topology, unique_molecules=[cyclohexane, water])

# Load the smirnof99Frosst force field.
forcefield = ForceField('smirnoff99Frosst.offxml')

# Parametrize the topology and create an OpenMM System.
[...]

PR #597, opened 2020-05-14
Minimize PACKMOL-generated systems against Parsley 1.1.0

### LINKED ISSUE #557: Replace sample systems with energy-minimized versions
Some (at least one, probably all) of the systems in `openforcefield/openforcefield/data/systems/packmol_boxes` are in a high energy state, i.e. just the PACKMOL output without any energy minimization. It would lower friction while using these for testing/fiddling to replace these files with systems that are minimized with Parsley; even though it is best practice to minimize a system before integrating, it can help while setting things up to have this done beforehand.

See #324 for the backstory

mattwthompson:
Fixes #557 

[Here](https://gist.github.com/mattwthompson/4fb7dfa779734011e81019134ce764c0) is the script I used to modify these files. It mostly takes @j-wags's snippet in #324 and does the energy minimization. There's a lot of validation that could be done to ensure that the configuration is ok; all I did was run an OpenMM simulation for 1000 steps to ensure it doesn't immediately, or nearly immediately, crash, which was the origin of the issue.

I'm also sharing my script because it could be used, with some modifications, to minimize configurations against particular force fields. This is out of the scope of this PR but that could be a valuable utility at some point.


mattwthompson:
I embedded it as a hyperlink, although it's easy for that go missed if the color of the text is not too brights

https://gist.github.com/mattwthompson/4fb7dfa779734011e81019134ce764c0

davidlmobley:
Ooh this is great, thanks!

Issue #599, opened 2020-05-18
Overriding `__getitem__` in ForceField


mattwthompson:
**Is your feature request related to a problem? Please describe.**
It would be nice to have cleaner lookups from `ForceField` objects to `ParameterHandler` objects, and possibly also directly to the individual parameters
```python3
>>> ff = ForceField('openff-1.1.0.offxml')
>>> ff['Bonds']
<openforcefield.typing.engines.smirnoff.parameters.BondHandler object at 0x7faeffb64b90>
>>> # Also could do
>>> # ff['Bonds']['b12']
# returns the particular parameter object
```

**Describe the solution you'd like**
This can be done by overriding `__getitem__` of `ForceField` and its individual `ParameterHandler` objects. Although some care is needed to ensure strange things don't happen. 

**Describe alternatives you've considered**
`ff.get_parameter_handler('Bonds')` works fine, as does `ff.get_parameter_handler('Bonds').get_parameter({'id': 'b12'})`, but the syntax sugar of `ff['Bonds']['b12']` would be preferred.

**Additional context**
`ParameterHandlers.get_parameter()` doesn't simply take in a value, it takes in a dict mapping that allows it to look up against SMIRKS patterns, ids, or in principle whatever other attribute is stored.

I saw this brought up in #310 and it has probably been mentioned elsewhere.

@j-wags says:

>Note, though, that ForceField.get_parameter_handler has a very important behavior that is different form a lookup -- If the parameterhandler for the requested SMIRNOFF section tag doesn't exist, it is initialized by get_parameter_handler, whereas I'd expect a lookup for the same  tag to fail.

j-wags:
Just speculating -- In the distant future (not now), this could be extended to either accept a `string` or a `ParameterHandler` type/subclass. But that change could be made in a reverse-compatible way, so that shouldn't influence our design choices now. The behavior of a type-based lookup might be complex (what if they have the same tagname, but a different actual class?) so we can punt on that.

mattwthompson:
Just to document the status:

#600 implements `__getitem__` for the `ForceField` class, but that only gets you the corresponding `ParameterHandler`, not an actual parameter. Behavior like `ff['Bonds']['b12']` is not possible but something like `ff['Bonds'].get_parameter({'id': 'b12'})`. The next step, overriding `ParameterHandler.__getitem__` is a little trickier but should be feasible, however, I consider it a lower priority. Let's leave this issue open until then.

PR #600, opened 2020-05-18
Implement ForceField.__getitem__ override

mattwthompson:
Implements #599. I tried to cover everything I could think of here and implement @j-wags's suggestions. I'm not implementing `ParameterHandler.__getitem__` for now.

### DIFF
--- a/openforcefield/tests/test_forcefield.py
+++ b/openforcefield/tests/test_forcefield.py
@@ -1125,6 +1125,8 @@ class TestForceField():
         bonds = forcefield['Bonds']
         with pytest.raises(NotImplementedError):
             forcefield[bonds]
+        with pytest.raises(NotImplementedError):
+            forcefield[type(bonds)]
         
 class TestForceFieldChargeAssignment:
 
--- a/openforcefield/tests/test_forcefield.py
+++ b/openforcefield/tests/test_forcefield.py
@@ -1110,13 +1110,14 @@ class TestForceField():
 
         assert handlers_before == handlers_after
 
-    def test_unregistered_parameter_handler_lookup(self):
+    @pytest.mark.parametrize('unregistered_handler', ['LibraryCharges', 'foobar'])
+    def test_unregistered_parameter_handler_lookup(self, unregistered_handler):
         """Ensure __getitem__ lookups do not register new handlers"""
         forcefield = ForceField('test_forcefields/smirnoff99Frosst.offxml')
-        unregistered_handler = 'LibraryCharges'
 
         assert unregistered_handler not in forcefield._parameter_handlers
-        forcefield[unregistered_handler]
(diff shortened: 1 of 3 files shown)


openforcefield/tests/test_forcefield.py
```diff
@@ -1096,8 +1096,36 @@ def test_nonbonded_method_resolution(self,
...
+        assert unregistered_handler not in forcefield._parameter_handlers
+        forcefield[unregistered_handler]
+        assert unregistered_handler not in forcefield._parameter_handlers
+
+    def test_lookup_parameter_handler_object(self):
+        """Ensure __getitem__ raises NotImplemented when passed a ParameterHandler object"""
+        forcefield = ForceField('test_forcefields/smirnoff99Frosst.offxml')
+        bonds = forcefield['Bonds']
+        with pytest.raises(NotImplementedError):
+            forcefield[bonds]
```
j-wags, line 1127:
```suggestion
            forcefield[type(bonds)]
```

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -1246,3 +1246,11 @@ def _get_parameter_handler_class(self, tagname):
             msg += "Known parameter handler class tags are {}".format(self._parameter_handler_classes.keys())
             raise KeyError(msg)
         return ph_class
+
+    def __getitem__(self, val):
+        """Syntax sugar for lookikng up a ParameterHandler."""
+        if isinstance(val, str):
+            if val in self._parameter_handlers:
+                return self.get_parameter_handler(val)
+        elif isinstance(val, ParameterHandler):
```
j-wags, line 1255: Ah, I was thinking that this could be the class/type itself, instead of an _instance_ of the class/type.

mattwthompson: I went ahead and covered both since I explicitly don't want to handle those cases now without more thinking. Let me know if that seems ok

openforcefield/tests/test_forcefield.py
```diff
@@ -1096,8 +1096,38 @@ def test_nonbonded_method_resolution(self,
...
 
+        assert handlers_before == handlers_after
 
+    def test_unregistered_parameter_handler_lookup(self):
+        """Ensure __getitem__ lookups do not register new handlers"""
+        forcefield = ForceField('test_forcefields/smirnoff99Frosst.offxml')
+        unregistered_handler = 'LibraryCharges'
+
+        assert unregistered_handler not in forcefield._parameter_handlers
+        forcefield[unregistered_handler]
```
dotsdl, line 1119: Do we want this silently returning `None`, or should it throw something like a `KeyError`?

mattwthompson: Good catch - tried to address it in ab2587a

Issue #522, opened 2020-02-19
Defining a specification for atomic map indices


j-wags:
**Is your feature request related to a problem? Please describe.**
Both [OETK](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEAtomBase.html#OEChem::OEAtomBase::GetMapIdx) and [RDKit](https://www.rdkit.org/docs/cppapi/classRDKit_1_1Atom.html#a9c57d20fc12b53253f089dc120382508) support atom map indices. Many users find having a secondary indexing system useful for a variety of reasons. I'm generally against adding complexity, but @jthorton is making a good case for why supporting map indices will be helpful as we develop an automated QCA submission tool and bespoke workflow. 

**Describe the solution you'd like**
We should decide on a "specification for atomic map indices". This specification should be a set of valid behaviors for map indices on a molecule. Hopefully it will line up well with what OETK and RDKit do, but that's not necessarily the biggest goal. What I want to settle on is an answer to "what behaviors do we _allow_?". Can map indices be redundant? Can they be negative?  I'll put a more complete list of decisions below. As I'm learning with partial charges right now, if we make these decisions now, then life will be easy later.

Note that this may be separate from an "API for atomic map indices" -- at this time I don't really care whether we access it as `offmol.properties['atom_map']` or `offmol.atoms[0].map_index`. 

For reference, here's the relevant portions of the [Daylight SMILES specification](https://www.daylight.com/dayhtml/doc/theory/theory.smiles.html), which is where all this mapped-SMILES stuff began to the best of my knowledge. This may help us understand what behavior RDK and OETK _should_ support. 

Section 3.5
>  Atom maps are non-negative integer atom modifiers. They follow the ":" character within an atom expression. They must be the last modifier within the atom expression: 

>  Atom maps are an atomic property. They can legally appear in a SMILES for any atom, whether or not it is part of a reaction. Atom with atom map labels in a molecule SMILES are considered valid; the atom maps are ignored for molecule processing. Absolute and unique SMILES generated by the system for molecules never include atom maps. 

Section 3.5.1
>  Within the SMILES language, atom maps are represented as a non-negative numeric atom modifier following the ":" character (e.g. [CH3:2] is a carbon in class 2). 

[...]

PR #601, opened 2020-05-18
Fix ring SMIRKS parsing

### LINKED ISSUE #592: Improper torsion SMIRKS around trivalent N in proline is mis-percieved as Proper
**Describe the bug**
In porting AMBER FF14SB into SMIRNOFF format, we found that the SMARTS for the improper torsion on the trivalent nitrogen in Proline was being mis-identified as a proper torsion. MWE below.

The issue is with the improper around the N in this structure. I think I've labeled the tagged atoms correctly here. (Note that this is actually a _substructure_ SMARTS, so the structure drawer added implicit Hs to satisfy the valence at the termini of the amino acid, which aren't actually in the SMARTS)
Screen Shot 2020-05-08 at 9 29 03 AM

**To Reproduce**
```
from openforcefield.typing.chemistry import ChemicalEnvironment
a = ChemicalEnvironment('[H][C@:4]1(C(C([C:3]([N:2]1[C:1](=O)C([H])([H])[H])([H])[H])([H])[H])([H])[H])C=O')
print(a.getType())
```
> ProperTorsion

j-wags:
- [x] Fix #592 

I've added tests to highlight the proline problem, as well as a more challenging one to close a 5-membered phosphate ring with all double bonds (to test that a double bond ring closure can be perceived)

I also tried merging in the changes from the [Chemper version of this file](https://github.com/MobleyLab/chemper/blob/master/chemper/graphs/environment.py) (which was updated recently), but that doesn't seem to fix the problem either.


j-wags:
I've decided to just refactor this to use ToolkitWrappers. The Chemper class that this is originally based on is made for _mutable_ chemical environments that can be programmatically manipulated, which means that a lot of the core functionality is implemented outside OE and RDKit. The only part of that functionality that we actually use is validating SMARTS, and OE and RDK can do a fine job of that, so there's no need to keep all this homebrew code around.

The functionality that I'm removing will still be available in Chemper, which is already several commits ahead of the versions of its files copied into OFFTK anyway.

j-wags:
This PR is now ready for review. The release notes cover the major changes made. Basically, the `environment.py` file taken from Chemper some time ago has some powerful functionality that the OFF Toolkit doesn't need, and which is complex enough that we've started finding some limitations. Since the essential parts of it can be implemented in a few lines of OE or RDKit, I've removing the majority of `environment.py` and replaced it with a few ToolkitWrapper methods that validate different aspects of SMIRKS.

dotsdl:
[[Toolkit] Proline SMARTS issue](https://trello.com/c/hSsoDgLx/25-toolkit-proline-smarts-issue)

mattwthompson:
I don't think there's a way to convey "I have accepted the review request and will get to it this week" in the UI here, so I am doing that in a comment

dotsdl:
Should have indicated I was reviewing as well. Looking at the changes and intending to give you some feedback.

j-wags:
Oh, cool. Since @dotsdl is already on this, I've removed @mattwthompson as a reviewer. Thanks both!

openforcefield/utils/toolkits.py
```diff
@@ -3054,6 +3100,52 @@ def to_inchikey(self, molecule, fixed_hydrogens=False):
...
+        from openforcefield.typing.chemistry import SMIRKSParsingError
+        ss = Chem.MolFromSmarts(smarts)
+
+        if ss is None:
+            raise SMIRKSParsingError(f"RDKit was unable to parse SMIRKS {smarts}")
+
+        unique_tags = set()
+        connections = set()
+        for at1 in ss.GetAtoms():
+            if at1.GetAtomMapNum() == 0:
```
dotsdl, line 3137: Are lines like these meant to simply ignore cases where an atom has `:0` as its tag, since that's not part of our spec?

openforcefield/utils/toolkits.py
```diff
@@ -3054,6 +3100,52 @@ def to_inchikey(self, molecule, fixed_hydrogens=False):
...
+                continue
+            unique_tags.add(at1.GetAtomMapNum())
+            for at2 in at1.GetNeighbors():
+                if at2.GetAtomMapNum() == 0:
+                    continue
+                cxn_to_add = sorted([at1.GetAtomMapNum(), at2.GetAtomMapNum()])
+                connections.add(tuple(cxn_to_add))
+        connections = tuple(sorted(list(connections)))
+        unique_tags = tuple(sorted(list(unique_tags)))
+        return tuple(unique_tags), tuple(connections)
```
dotsdl, line 3147:
```suggestion
        return unique_tags, connections
```

dotsdl: Since these are each already tuples, this saves us a small bit of processing.

openforcefield/utils/toolkits.py
```diff
@@ -1833,6 +1833,52 @@ def assign_fractional_bond_orders(self, molecule, bond_order_model=None, use_con
...
+        tagged_atom_connectivity : tuple of tuples of int, shape n_tagged_bonds x 2
+            A tuple of tuples, where each inner tuple is a pair of tagged atoms (tag_idx_1, tag_idx_2) which are
+            bonded. The inner tuples are ordered smallest-to-largest, and the tuple of tuples is ordered
+            lexically. So the return value for an improper torsion would be ((1, 2), (2, 3), (2, 4)).
+
+        Raises
+        ------
+        SMIRKSParsingError
+            If OpenEye toolkit was unable to parse the provided smirks/tagged smarts
+        """
```
dotsdl, line 1859: See comments below on `rdkit` version; same comments apply.

dotsdl:
I like what you've done here. If we can easily outsource parsing of something fairly complex such as SMARTS/SMIRKS to the supported toolkits, as you've done here, it's a big reduction of our maintenance burden. Very clean integration, too. :grin:

openforcefield/tests/test_chemicalenvironment.py
```diff
@@ -21,198 +28,68 @@ def test_createEnvironments(self):
...
+                               'ProperTorsion', TorsionChemicalEnvironment],
+                              ["[#6$([#6X4](~[#7!-1,#8!-1,#16!-1,#9,#17,#35,#53])(~[#8]~[#1])):1]1=CCCC1", 'Atom', AtomChemicalEnvironment],
+                              ["[*:1]-[#7X3:2](-[#6a$(*1ccc(-[#8-1X1])cc1):3])-[*:4]", 'ImproperTorsion',
+                               ImproperChemicalEnvironment],
+                              ["[#6X4:1]1~[*:2]~[*$(*~[#1]):3]1", 'Angle', AngleChemicalEnvironment],
+                              ["[$([#7]1~[#6]-CC1)]", None, ChemicalEnvironment],
+                              ["[$(c1ccccc1)]", None, ChemicalEnvironment],
+                              # The next two tests are for ring-closing bonds
+                              ["[H][C@:4]1(C(C([C:3]([N:2]1[C:1](=O)C([H])([H])[H])([H])[H])([H])[H])([H])[H])C=O",
+                               'ImproperTorsion', ChemicalEnvironment],
```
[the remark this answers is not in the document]
j-wags: Nice close reading. In this case, we want to test that "a smirks which could satisfy a subclass of `ChemicalEnvironment` also satisfies the base class". It took me a little bit to wrap my head around it too. I've added a comment to the test case to avoid future confusion.

openforcefield/utils/toolkits.py
j-wags: Yes. IIRC, RDKit assigns map values of 0 to 
* atoms that don't have map indices at all, and
* atoms like `[#6:0]`
so we play it safe and assume that atoms with map index==0 aren't supposed to be tagged at all.

FYI: Atom map indices are a bit fuzzy right now, because they don't have chemical meaning, but different toolkits have different behaviors around them. So one place we deal with them is in SMIRKS (the topic here), but there's an even bigger question about how to handle them in molecules. I think we know enough about _our_ use cases to lay down a map index spec in OFFMols that will work for the rest of time, but we need to finish the discussion about intended behavior in #522 before we implement it.

Issue #598, opened 2020-05-18
Docs are broken


j-wags:
The new Sphinx has an API change that doesn't work with the now-unmaintained m2r package: https://github.com/sphinx-doc/sphinx/issues/7420

~I tried to fix this in #594 but failed. This issue is to either find a replacement for m2r or pin sphinx to 2.x.x~

Replacing m2r didn't get our funciton/method docs building. This issue has been rescoped to that.

dotsdl:
Taking this on; I'll review #594 and jump off from there.

dotsdl:
See [this comment](https://github.com/openforcefield/openforcefield/pull/602#pullrequestreview-419531903) on path forward beyond just a temporary fix.

j-wags:
Looks like someone has started hosting a fork of the abandoned `m2r` project (called `m2rr`). It's pypi-only right now, but that's not a bad start. https://github.com/qhua948/m2rr

j-wags:
@SimonBoothroyd fixed this in #744!

PR #602, opened 2020-05-19
Added shim that allows us to use m2r components, but recommonmark mainly

dotsdl:
Addresses #598 

This PR fixes the doc build, which uses Sphinx 3.x.  `m2r`, which we used for markdown handling, appears unmaintained and fails when building the docs under this version of Sphinx.

I used [this approach](https://github.com/readthedocs/recommonmark/issues/191#issuecomment-622369992) to use the components we need from `m2r` (`.. mdinclude::` directive), but `recommonmark` for everything else.

I would like a second set of eyes on the built docs to ensure everything is being generated as expected.

Also made some changes to makefile in line with latest sphinx (3.x) config generation. This may not necessarily be desirable; review requested.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -313,20 +313,19 @@ texinfo_documents = [
 
 
 # workaround for m2r broken on sphinx >= 3.x
-#def monkeypatch(cls):
-#    """ decorator to monkey-patch methods """
-#    def decorator(f):
-#        method = f.__name__
-#        old_method = getattr(cls, method)
-#        setattr(cls, method, lambda self, *args, **kwargs: f(old_method, self, *args, **kwargs))
-#    return decorator
-#
-## workaround until https://github.com/miyakogi/m2r/pull/55 is merged
-#@monkeypatch(sphinx.registry.SphinxComponentRegistry)
-#def add_source_parser(_old_add_source_parser, self, *args, **kwargs):
-#    # signature is (parser: Type[Parser], **kwargs), but m2r expects
-#    # the removed (str, parser: Type[Parser], **kwargs).
-#    if isinstance(args[0], str):
-#        args = args[1:]
-#    return _old_add_source_parser(self, *args, **kwargs)
-#
+def monkeypatch(cls):
+    """ decorator to monkey-patch methods """
+    def decorator(f):
+        method = f.__name__
+        old_method = getattr(cls, method)
+        setattr(cls, method, lambda self, *args, **kwargs: f(old_method, self, *args, **kwargs))
+    return decorator
+
+# workaround until https://github.com/miyakogi/m2r/pull/55 is merged
+@monkeypatch(sphinx.registry.SphinxComponentRegistry)
+def add_source_parser(_old_add_source_parser, self, *args, **kwargs):
+    # signature is (parser: Type[Parser], **kwargs), but m2r expects
+    # the removed (str, parser: Type[Parser], **kwargs).
+    if isinstance(args[0], str):
+        args = args[1:]
+    return _old_add_source_parser(self, *args, **kwargs)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,13 +21,15 @@ import os
 import sys
 sys.path.insert(0, os.path.abspath('.'))
 
+import sphinx
+
 # bootstrap theme
 import sphinx_bootstrap_theme
 
 import openforcefield
 
-from recommonmark.transform import AutoStructify
-from m2r import MdInclude
+#from recommonmark.transform import AutoStructify
+#from m2r import MdInclude
 
 # -- General configuration ------------------------------------------------
 
@@ -49,7 +51,8 @@ extensions = [
     'sphinx.ext.viewcode',
     'sphinx.ext.intersphinx',
     'nbsphinx',
-    'recommonmark', # render markdown
+    #'recommonmark', # render markdown
+    'm2r', # render markdown
     ]
 
 autosummary_generate = True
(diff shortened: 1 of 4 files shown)


dotsdl:
[[Toolkit] Docs are broken](https://trello.com/c/rPLZWTVz/22-toolkit-docs-are-broken)

docs/conf.py
```diff
@@ -26,6 +26,9 @@
 
 import openforcefield
 
+from recommonmark.transform import AutoStructify
```
mattwthompson, line 29: This was not pulled down for me with `conda env create --name openff-docs --file environment.yml`

mattwthompson:
This builds locally for me and looks good (after grabbing `recommonmark`). I looked over a handful of pages and nothing seemed out of the ordinary - anything in particular to check?

j-wags:
Looking at this now -- I've tagged this branch for builds on RTD, so we should be able to see status here: https://readthedocs.org/projects/open-forcefield-toolkit/builds/

dotsdl:
Just tried switching `m2r_parse_relative_links` to `True` as @j-wags suggested in [here](https://github.com/openforcefield/openforcefield/pull/602#pullrequestreview-415684493). Appears to produce the [same effect](https://open-forcefield-toolkit.readthedocs.io/en/doc-fix/releasehistory.html#id35). :cry:

dotsdl:
Tried [this approach](https://github.com/miyakogi/m2r/issues/51#issuecomment-618285433) locally just now. Doesn't require `recommonmark`, but also doesn't fix the issue with broken API links. Going to push this so our dependencies remain the same, though it doesn't advance our approach much further.

j-wags:
I wonder if it'd be worthwhile to take another go at pinning to an old version of sphinx. If we can force it down to `2.4.4`, then all of our problems are solved (for a while).

dotsdl:
> I wonder if it'd be worthwhile to take another go at pinning to an old version of sphinx. If we can force it down to 2.4.4, then all of our problems are solved (for a while).

Just tried to set `sphinx == 2.4` in conda `docs/environment.yml`, but it looks like [the build](https://readthedocs.org/projects/open-forcefield-toolkit/builds/11100212/) installs sphinx after running this, overriding that specification. Any ideas on other knobs we can turn here?

jaimergp:
Maybe this helps: https://github.com/readthedocs/readthedocs.org/issues/5810

j-wags:
We might be able to force a "conda" style build instead of a "sphinx" style one, and then just manually install the dependencies, including `sphinx=2`, as conda packages.

dotsdl:
> We might be able to force a "conda" style build instead of a "sphinx" style one, and then just manually install the dependencies, including `sphinx=2`, as conda packages.

We are already doing a `conda` style build. I have [added an explicit sphinx pin](https://github.com/openforcefield/openforcefield/pull/602/files#diff-b0278a395ac6448087984d5186dbe855R11) to our `docs/environment.yml`, but this gets overwritten by [the build](https://readthedocs.org/projects/open-forcefield-toolkit/builds/11124310/), which runs the following right after installing our environment:

```
> conda install --yes --quiet --name doc-fix mock pillow sphinx sphinx_rtd_theme
Collecting package metadata (current_repodata.json): ...working... done
Solving environment: ...working... done

## Package Plan ##

  environment location: /home/docs/checkouts/readthedocs.org/user_builds/open-forcefield-toolkit/conda/doc-fix

  added / updated specs:
    - mock
    - pillow
    - sphinx
    - sphinx_rtd_theme


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    ca-certificates-2020.1.1   |                0         125 KB
    certifi-2020.4.5.1         |           py37_0         155 KB
    mock-4.0.2                 |             py_0          32 KB
    openssl-1.1.1g             |       h7b6447c_0         2.5 MB
    pillow-7.1.2               |   py37hb39fc2d_0         603 KB
    sphinx-3.0.3               |             py_0         1.1 MB
    sphinx_rtd_theme-0.4.3     |             py_0         4.4 MB
    ------------------------------------------------------------
                                           Total:         9.0 MB

The following NEW packages will be INSTALLED:

  mock               pkgs/main/noarch::mock-4.0.2-py_0
  sphinx_rtd_theme   pkgs/main/noarch::sphinx_rtd_theme-0.4.3-py_0

The following packages will be UPDATED:

  sphinx                     conda-forge::sphinx-2.4.0-py_0 --> pkgs/main::sphinx-3.0.3-py_0

The following packages will be SUPERSEDED by a higher-priority channel:

  ca-certificates    conda-forge::ca-certificates-2020.4.5~ --> pkgs/main::ca-certificates-2020.1.1-0
  certifi            conda-forge::certifi-2020.4.5.1-py37h~ --> pkgs/main::certifi-2020.4.5.1-py37_0
  openssl            conda-forge::openssl-1.1.1g-h516909a_0 --> pkgs/main::openssl-1.1.1g-h7b6447c_0
  pillow             conda-forge::pillow-7.1.2-py37h718be6~ --> pkgs/main::pillow-7.1.2-py37hb39fc2d_0


Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done 
```
I don't see any additional options in the [RTD conda docs](https://docs.readthedocs.io/en/stable/guides/conda.html) for overriding this behavior.

dotsdl:
Besides the `.. mdinclude` directive, is there anything else out of `m2r` that we actually need? If not, we could move [the spec](https://github.com/openforcefield/openforcefield/blob/master/The-SMIRNOFF-force-field-format.md) into the doc tree and try sticking with `recommonmark`. This may sidestep the problem entirely.

dotsdl:
Cool, went with one of the workarounds in `conf.py` that appeared to work well enough, and left the other one behind as a comment for now until we have a more permanent solution. As it happens, it appears that API references for methods, but not for e.g. classes, are broken, so we haven't lost all of the internal linking functionality we'd like.

PR #607, opened 2020-05-27
Switch element parsing from OpenMM to QCElemental

mattwthompson:
I'm not convinced this is a good feature but I'm pushing it in case others find it valuable. I'm inclined to build element parsing in the System object off of QCElemental instead of OpenMM since it is much more lightweight and narrow. I wanted to see if this would be a big change for the toolkit and it doesn't seem to be.

There isn't anything particularly wrong with OpenMM's element module (I don't think - correct me if I'm wrong) and I don't think there's a major new feature that's enabled by this. The above benefits aside, the main change here is better compatibility with a future pydantic/QCElemental refactor. A downside is that this would add QCElemental as a hard dependency while removing OpenMM is not going to happen anytime soon, if ever. 

I don't intend for this to be included in 0.7.0 and it would be fine if this hangs until we work on a refactor that this would actually help.

### TODO

- [ ] Maybe alias `element.symbol` to `symbol` with a `DeprecationWarning` since it is an API break, although minor.
- [ ] Figure out a `from_mass` lookup, which may require monkeypatching or changing QCElemental's `periodictable` module.
- [ ] Add `QCElemental` as a dependency in the conda recipe

### DIFF
--- a/openforcefield/tests/test_molecule.py
+++ b/openforcefield/tests/test_molecule.py
@@ -30,6 +30,7 @@ from tempfile import NamedTemporaryFile
 import numpy as np
 import pytest
 from simtk import unit
+import qcelemental as qcel
 
 from openforcefield.topology.molecule import Molecule, Atom, InvalidConformerError
 from openforcefield.utils import get_data_file_path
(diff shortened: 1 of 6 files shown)


jchodera:
Switching element parsing to QCElemental sounds like a good choice for all the reasons you articulate!

Unrelated to the current code changes, but related to the suggestion of basing the future System object off QCElemental/QCSchema: QCSchema made the decision to bake in some implicit units that are standard for quantum chemistry, but act as an impediment to conversion among packages. We may ultimately adopt a standardized unit system for serialization, but thinkiing about how to make unit system conversion easy is an important thing to consider from the start. For future unit handling, I'd suggest `pint`, which was developed after `simtk.unit` and is now much more of a standard in the Python field:
https://pint.readthedocs.io/

If we need something that works at a lower level (e.g. C), there are other standard unit-handling libraries.

openforcefield/tests/test_molecule.py
```diff
@@ -219,6 +220,27 @@ def mini_drug_bank(xfail_mols=None, wip_mols=None):
...
+def known_openmm_elements():
+    """
+    Fixture containing elements known to both OpenMM's Element class and
+    QCElemental.
+    """
+    from simtk.openmm.app import element
+    elements = [getattr(element, name) for name in dir(element) if (type(getattr(element, name)) == element.Element)]
+
+    # The below runs into a problem with deuterium (fails name assertion)
+    elements.remove(element.deuterium)
```
j-wags, line 233: Deuterated drugs are a pretty big deal. I don't think that molecular simulations would realistically show a difference between H and D, but we should keep track of the lack of deuterium handling in an Issue so we remember this is here.

j-wags:
Thanks for the breakdown. After some thought, I think this is a medium-value gain coupled to a high-value loss. I'd propose that we "park" this change until the 1.0 release, which will be a singular event with all sorts of large API breaks, but also a well-structured migration guide.

mattwthompson:
This is old enough now that merging upstream changes isn't really worth it. Implementing this change would be better done by starting fresh.

PR #608, opened 2020-05-27
Fix codecov

mattwthompson:
We haven't had a coverage report properly uploaded (with data) to codecov since March 31


mattwthompson:
The comment above won't work because the commit on `master` this PR is based off of does not have codecov data associated with it (h/t @SimonBoothroyd) but I think it's safe to say this PR does what is intended, i.e. removing a test file drops the coverage:

image



(will undo the previous commit before merging)

PR #613, opened 2020-06-01
Skip cron jobs on forks

### LINKED ISSUE #612: Disable nightly builds on forks
The CI GitHub Action is set up to run nightly cron jobs on forks and these always fail due to OpenEye licensing. The owner for the forks (say, me for my fork) gets a nightly email about a job failing, which is not a good user experience for anybody who comes across this repo and forks it. We should simply not run these cron jobs on forks while the license issue exists and only add it back if we can fin a way to safely run OpenEye builds on forks.

mattwthompson:
Fixes #612, copied from https://github.community/t/do-not-run-cron-workflows-in-forks/17636

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,6 +17,7 @@ on:
 
 jobs:
   test:
+    if: (github.event_name == 'schedule' && github.repository == 'openforcefield/openforcefield') || (github.event_name != 'schedule')
     name: Test on ${{ matrix.cfg.os }}, Python ${{ matrix.python-version }}, RDKit=${{ matrix.cfg.rdkit }}, OpenEye=${{ matrix.cfg.openeye }}
     runs-on: ${{ matrix.cfg.os }}
     strategy:


mattwthompson:
I pushed this to the `master` branch of my fork to test this, I will report tomorrow if it runs tonight

jaimergp:
Oh, nice catch. Let's see if it works!

mattwthompson:
I didn't propertly merge it before last night, but I fixed it and changed it to run hourly on my fork and the last few builds have been skipped. Looks like it's working as intended there: https://github.com/mattwthompson/openforcefield/actions?query=event%3Aschedule

so it should (:tm:) work here

jaimergp:
Awesome! I wonder if this should be backported to MolSSI's template, at least as a commented out line?

mattwthompson:
I'm not sure - I don't think our case of a license that only the upstream repo can access isn't super common in the field. But, that aside, it can be annoying for a user to get nightly emails for something they're not responsible for maintaining. At least a commented line or something documented in the history there would be nice; I don't have a strong feeling for or against it being the default behavior.

mattwthompson:
Good to merge this one too @jaimergp?

jaimergp:
Sure, go ahead!

PR #616, opened 2020-06-03
Update add_parameter functionality

### LINKED ISSUE #605: Expand add_parameter Functionality

The `ParameterHandler.add_parameter` function should:

- allow a `Parameter` object to be passed in rather than just a dictionary.
- allow the user to optionally specify whether the new parameter should be added before and / or after certain existing parameters so that it gets applied in the correct circumstances 

e.g.

```
vdw_handler = off_force_field.get_parameter_handler("vdW")
vdw_handler.add_parameter(my_new_parameter, after="[#1:1]-[#6X4]", before="[#1:1]-[#6X4](-[#7,#8,#9,#16,#17,#35])-[#7,#8,#9,#16,#17,#35]")
```

mattwthompson:
Resolves #605 
- [x] Support taking in `ParameterType` objections
- [x] Support specifying location through index
- [x] Valiate that args are actually SMARTS?


mattwthompson:
I think this is ready for review (but we'll see what the tests say).

* The `before`/`after` args can either be ints corresponding to locations in the parameter list or SMIRKS patterns describing existing parameters. In the latter, they are converted over to an index through a lookup, so the final determination of where to put the new parameter is ultimately done with ints.

* Either a dict or already-constructed `ParameterType` object can be passed in. If a dict is passed in, it's immediately turned into a `ParameterType`.

* I _believe_ SMIRKS validation is already done when the objects are constructed.

* I've put together a pretty long test, but there are so many combinations of inputs and possibilities for users to do weird thing that I'm sure things aren't covered. I'd hope the coverage is decent, though, and I believe at least the desired MWE in the issue is supported.

* I think the language here is a bit cumbersome; when I call `add_thing(thing=mything, before=val)` I read it in my head as "put this new thing `mything` before `val`" whereas the lanuage @SimonBoothroyd proposed, as is implemented here, is more like "put this new thing in the list, the value before where I want it to be is `val`". I'm not sure how best to structure the function signature but would like some feedback on what provides the best user experience. I don't currently think the functionality  is lacking, it's just how it's presented that confuses me.

openforcefield/tests/test_parameters.py
```diff
@@ -292,6 +292,60 @@ class MyParameterType(_ParameterAttributeHandler, MixIn):
...
+        dict_to_add_by_index = {'smirks': '[#1:1]-[#8:2]', 'length': self.length, 'k': self.k, 'id': 'd2'}
+
+        param_to_add_by_smirks = BondHandler.BondType(
+            **{'smirks': '[#6:1]-[#6:2]', 'length': self.length, 'k': self.k, 'id': 'p1'}
+        )
+        param_to_add_by_index = BondHandler.BondType(
+            **{'smirks': '[#6:1]=[#8:2]', 'length': self.length, 'k': self.k, 'id': 'p2'}
+        )
+
+        with pytest.raises(ValueError):
```
j-wags, line 327: I found it easy to lose track of why we expected an error here. Could you add comments for each failure mode in this test to explain why the failure happens?

openforcefield/tests/test_parameters.py
```diff
@@ -292,6 +292,60 @@ class MyParameterType(_ParameterAttributeHandler, MixIn):
...
+
+        bh.add_parameter(dict_to_add_by_index, after=2)
+
+        assert [param.id for param in bh._parameters] == ['b1', 'b2', 'd2', 'd1', 'b3']
+
+        bh.add_parameter(parameter=param_to_add_by_smirks, before='[*:1]-[*:2]')
+
+        assert [param.id for param in bh._parameters] == ['b1', 'p1', 'b2', 'd2', 'd1', 'b3']
+
+        bh.add_parameter(parameter=param_to_add_by_index, after='[*:1]#[*:2]')
```
j-wags, line 345: Could you change this one to have _both_ a `before` and an `after`, which are several elements apart, and then have the assertion ensure the new parameter went somewhere in that range?

SimonBoothroyd:
> As a newcomer, I would think "before=X" means "put this new thing somewhere above X in the OFFXML or before X in a list"

This is what I was proposing. The value should appear later in the list than the `after` value and earlier in the list than the `before` value.

mattwthompson:
As evidence of how confused I managed to make myself while working on this, the docstrings do it the "right way" but the code didn't! It should be fixed and consistent now. And to help with this, I've added a example in the docstring. I'm unable to properly build docs locally so I can't verify that it will build correctly

j-wags:
Ahh, so we were all thinking the same thing all along (except the code, somehow). Huzzah for asynchronous communication!

Let me know when to re-review. Should be quick!

j-wags:
Ah, the doc example is failing, since it's not a fully functional script (`bh` is never defined).

https://github.com/openforcefield/openforcefield/pull/616/checks?check_run_id=858973823#step:8:26288

mattwthompson:
Well, this sent me down quite the rabbit hole of the internals of pytest (trying to build the docs was the wrong approach). Which, of course, isn't fully captured in the diff being pretty simple. It seems obvious in hindsight, but you can run the docstring tests (relying on a standard library module `doctest`) by pointing it to a module instead of a test path - something like `pytest -v --doctest-modules openforcefield/typing/engines/smirnoff/parameters.py`.

CI is failing on README links - of course - but should pass after a restart.

mattwthompson:
Oh, and please review again at your leisure @j-wags!

Issue #621, opened 2020-06-15
CI Failing on Master


SimonBoothroyd:
**Describe the bug**
The CI tests which include `rdkit` are currently failing on master since 4 days ago. This roughly coincides with a new release of [Rdkit on conda-forge](https://anaconda.org/conda-forge/rdkit/files).

mattwthompson:
I've found that QCEngine's deps (https://github.com/MolSSI/QCEngine/issues/252, fixed in https://github.com/MolSSI/QCEngine/pull/253) which happened to introduce an API-breaking change at a time near RDKit's recent release. Everything is passing for me locally now.

I fear this will require a QCEngine bump before 0.7.0. Pinning an implicit dependence to a lower version is the only alternative I can think of, and that is not appealing.

mattwthompson:
And just to resolve my confusion ("why does this QCEngine issue only show up in RDKit tests?") the example is skipped when RDKit is not installed, and this is encoded in the GHA script:

https://github.com/openforcefield/openforcefield/blob/3de67ca96265a3953c008e8e244c380d4aa81f2e/.github/workflows/CI.yml#L105-L115

mattwthompson:
Fixed for now (save one codecov step, the recent cron CI passed). Thanks for catching this @SimonBoothroyd!

PR #622, opened 2020-06-15
Pin py-cpuinfo until there's a QCEngine release

j-wags:
Needed to get tests passing, per https://github.com/MolSSI/QCEngine/issues/252
- [x] Short-term fix for #621

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -14,6 +14,7 @@ dependencies:
   - nbval
   - codecov
   - coverage < 5.0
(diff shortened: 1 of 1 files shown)


mattwthompson:
Should we 
* Pin  `QCEngine`  here to `<=` current version since the patch will only work on the most recent version of `py-cpuinfo`, or
* l'll update the PR to handle both versions of `py-cpuinfo` ?

j-wags:
Ah, good call. I'll do the former.

j-wags:
Thanks @mattwthompson . Merged and added a note to unpin in trello

PR #623, opened 2020-06-19
Use org-wide oe license for CI

j-wags:
Looks like the previous OE license we were using expired. I'm updating this to use the new, org-wide license, which should be good for another year.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -79,9 +79,9 @@ jobs:
     - name: Decrypt OpenEye license
       shell: bash
       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -79,9 +79,9 @@ jobs:
     - name: Decrypt OpenEye license
```
mattwthompson, line 79: Nitpick: Might be a better verb to use here than "decrypt", even though that's what's happening on a server somewhere (and this PR would be a good place to update this)

j-wags:
Thanks. Merged!

PR #634, opened 2020-07-02
Fix bad string case when calling RDKitToolkitWrapper.from_file directly

### LINKED ISSUE #628: Sanitize case of file extensions

Maybe not a bug per se, but definition a little annoyance to the user
```python 3
In [25]: RDKitToolkitWrapper()._toolkit_file_read_formats
Out[25]: ['SDF', 'MOL', 'SMI']

In [26]: RDKitToolkitWrapper().from_file(file_path=get_data_file_path('molecules/ethanol.sdf'), file_format='sdf')
Out[26]: []

In [27]: RDKitToolkitWrapper().from_file(file_path=get_data_file_path('molecules/ethanol.sdf'), file_format='SDF')
Out[27]: [Molecule with name '*****' and SMILES '[H]C([H])([H])C([H])([H])O[H]']
```

This is because the supported file formats are specified in all caps and not lowered elsewhere:
https://github.com/openforcefield/openforcefield/blob/131d86c7cacc3bbd98d10114a3b326ec1d7db9f9/openforcefield/utils/toolkits.py#L2212

Some careful use of `file_format.lower()` or `.upper()` when parsing arguments, and also updating bookkeepping elsewhere (various `ToolkitWrapper._toolkit_file_read_formats` and probably some other things to be discovered in re-running unit tests). I don't think there's a downside to this; if one is given `cool_molecule.mol2` and `cool_molecule.MOL2` I cannot see a reason why they should be processed as different file formats.

One _could_ just capitalize this argument on each call, but I think it would improve user experience to just sanitize the input. (Read: I kept trying `file_format='sdf'` and it wasn't until I dug into the internals that I noticed it's coded to only support `file_format='SDF'`)

mattwthompson:
Fixes #628, although this is pretty localized to the edge case of calling `RDKitToolkitWrapper.from_file` since normally  you'd want to do `Molecule.from_file(...)`

Pushing as separate commits to show (if it works the same as locally) that the bug only impacted the RDKit wrapper, since `OpenEyeToolkitWrapper.from_file` doesn't use `file_format` at all.

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -7,6 +7,16 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
 * ``minor`` increments add features but do not break API compatibility
 * ``micro`` increments represent bugfix releases or improvements in documentation
 
+0.7.1 - Current development
+---------------------------
+
+Bugfixes
+""""""""
+- `PR #634 <https://github.com/openforcefield/openforcefield/pull/634>`_: Fixes a bug in which calling 
+  :py:meth:`RDKitToolkitWrapper.from_file <openforcefield.utils.toolkits.RDKitToolkitWrapper.from_file>` directly
+  would not load files correctly if passed lowercase `file_format`. Note that this bug did not occur when calling
+  :`Molecule.from_file` <openforcefield.topology.molecule.Molecule.from_file>`.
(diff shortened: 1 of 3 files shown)


mattwthompson:
CI failure was stochastic (one single badge check failed)

PR #640, opened 2020-07-11
Use frappuccino to check for API changes

mattwthompson:
https://pypi.org/project/frappuccino/

### DIFF
--- a/.github/workflows/api-breaks.yml
+++ b/.github/workflows/api-breaks.yml
@@ -1,0 +1,62 @@
+name: CI
+
+on:
+  push:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  test:
+    if: (github.event_name == 'schedule' && github.repository == 'openforcefield/openforcefield') || (github.event_name != 'schedule')
+    name: Test for API changes
+    runs-on: ${{ matrix.cfg.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version:
+          - 3.7
+        cfg:
+          - os: ubuntu-latest
+
+    env:
+      CI_OS: ${{ matrix.cfg.os }}
+      PYVER: ${{ matrix.python-version }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install most recent version from conda
+      uses: goanpeca/setup-miniconda@v1
+      with:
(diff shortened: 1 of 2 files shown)


mattwthompson:
This package doesn't report exit status (yet) but the gist of it is accessible in the log (` Compare API of released and development versions`)

For example, right now the API changes are 

```
2020-07-11T16:28:20.1005785Z The following items are new:
2020-07-11T16:28:20.1006013Z     + openforcefield.utils.toolkits.inspect
2020-07-11T16:28:20.1006251Z     + openforcefield.typing.engines.smirnoff.parameters.DuplicateParameterError
2020-07-11T16:28:20.1006523Z     + openforcefield.utils.toolkits.ToolkitRegistry.deregister_toolkit(self, toolkit_wrapper)
2020-07-11T16:28:20.1006689Z 
2020-07-11T16:28:20.1006878Z The following signatures differ between versions:
2020-07-11T16:28:20.1007017Z 
2020-07-11T16:28:20.1007679Z     - openforcefield.typing.engines.smirnoff.parameters.ParameterHandler.add_parameter(self, parameter_kwargs)
2020-07-11T16:28:20.1008926Z     + openforcefield.typing.engines.smirnoff.parameters.ParameterHandler.add_parameter(self, parameter_kwargs='None', parameter='None', after='None', before='None')

```

And you can also grab different versions, i.e. here's the mess of changes between 0.6.0 and 0.7.0: https://gist.github.com/mattwthompson/b852e5b2748db27ca119bd0035010dcd

jchodera:
Wow, this is awesome! Would it be worth adding this to the cookiecutter?

mattwthompson:
It's more or less a personal project and maybe not mature enough right now to roll out with the cookiecutter, by default, in my opinion. It could be valuable as option for people who do care about API changes - probably opt-in since it would be annoying for a newer developer being yelled at by CI while they're rapidly changing the API of a fresh project. I opened it here since the toolkit values API stability (or at least would benefit from detailed information about changes) whereas it seems like most other projects in the community don't value it so much (outside of a few pillars, most are 0.x with an ~experimental API)

mattwthompson:
Closing this in favor of housing it in https://github.com/openforcefield/status  later if useful in the release process.

PR #645, opened 2020-07-14
Pin OpenEye toolkits to 2019

mattwthompson:
@j-wags I know we are aiming to roll out compatibility with the new `2020.0.4` version of the OpenEye toolkits, but I would like to keep the master branch stable until then


mattwthompson:
Pinging the codecov badge failed, but only on the `RDKit=false, OpenEye=true` builds. Quite the false alarm - restarting

PR #646, opened 2020-07-14
Add functionality to ignore/strip short-lived stereochemistry

j-wags:
Differences in tertiary nitrogen stereochemistry shouldn't be important in parameterization. This PR will add methods to strip this information or disregard it in certain contexts. 
Screen Shot 2020-07-17 at 10 48 06 AM

The information about the nitrogen "stereocenter" in the above molecule SHOULD be stored (becuase otherwise we'll fail OEMol->OFFMol->OEMol roundtrips), but it SHOULD NOT be considered significant by the `__eq__` operator.

So, I propose that we add `Molecule.strip_atom_stereochemistry(smarts)`, which expects a SMARTS tagging a single atom, and will remove it stereo info (if set). So, the case above could be handled by `offmol.strip_atom_stereo('[N+0X3:1](-[*])(-[*])(-[*])')`. This SMARTS should only tag a single atom (can check this using `ChemicalEnvironment.validate`). Any atom indices found by running `ToolkitRegistry.call('find_smarts_matches', ...)` on this molecule should have their stereochemical information cleared, for example using `offmol.atoms[idx].stereochemistry=None`.

We should add a new kwarg to `Molecule.are/is_isomorphic` called `strip_pyrimidal_n_atom_stereo=True`. If `True`, we make a COPY of the molecules and strip N stereo information from the copies. Since the `__eq__` operator just calls these methods, our `__eq__` operator will therefore disregard tertiary N stereo.

Tests for this should compare two OFFMols made from SMILES and ensure that 
```
Molecule.from_smiles('CCC[N@](C)CC') == Molecule.from_smiles('CCC[N@@](C)CC')
```
 but that the same test FAILS if we explicitly call `is/are_isomorphic` with `strip_n_atom_stereo=False`.

We should test that this method DOES NOT mangle:
* _Quaternary_ N+1s (with 4 distinct heavy substituents) should be disregarded
* Aromatic tertiary N+1s (like the NH in [histidine's sidechain ring](https://upload.wikimedia.org/wikipedia/commons/thumb/1/18/Histidin_-_Histidine.svg/1200px-Histidin_-_Histidine.svg.png)). Though these are planar so there shouldn't be stereo info to strip, so we may not be able to test this. 

**Additional context**
We could make a kwarg to `create_openmm_system` to say `tertiary_nitrogen_stereochemistry_is_distinguishing`, but this would add a lot of complexity and it doesn't seem like a lot of users would desire that.

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -10,9 +10,27 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
 0.7.1 - Current development
 ---------------------------
 
+
+Behavior changed
+""""""""""""""""
+- `PR #646 <https://github.com/openforcefield/openforcefield/pull/646>`_: Checking for molecule
+  equality (using thr ``==`` operator now disregards all pyrimidal nitrogen stereochemistry
+  by default. To re-enable, use
+  :py:meth:`Molecule.are_isomorphic <openforcefield.topology.Molecule.are_isomorphic>`
+  with the ``strip_pyrimidal_n_atom_stereo=False`` keyword argument.
+- `PR #646 <https://github.com/openforcefield/openforcefield/pull/646>`_: Adds
+  an optional ``toolkit_registry`` keyword argument to
+  :py:meth:`Molecule.are_isomorphic <openforcefield.topology.Molecule.are_isomorphic>`,
+  which identifies the toolkit that should be used to search for pyrimidal nitrogens.
+
+
 Bugfixes
 """"""""
-- `PR #634 <https://github.com/openforcefield/openforcefield/pull/634>`_: Fixes a bug in which calling 
+- `PR #646 <https://github.com/openforcefield/openforcefield/pull/646>`_: Fixes a bug where
+  :py:meth:`Molecule.chemical_environment_matches <openforcefield.topology.Molecule.chemical_environment_matches>`
+  was not able to accept a :py:class`ChemicalEnvironment <openforcefield.typing.chemistry.ChemicalEnvironment>` object
+  as a query.
+- `PR #634 <https://github.com/openforcefield/openforcefield/pull/634>`_: Fixes a bug in which calling
   :py:meth:`RDKitToolkitWrapper.from_file <openforcefield.utils.toolkits.RDKitToolkitWrapper.from_file>` directly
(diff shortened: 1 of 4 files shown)


mattwthompson:
New tests are passing, an old test is failing for reasons I won't be able to get into today. I've ruled out it being flaky, but don't know if it's due to an upstream change or a problematic behavior here


``TestRDKitToolkitWrapper.test_non_standard_inchi_round_trip[molecule342]``

This molecule has SMILES `'[H][O][C@]1([C]([H])([H])[H])[C](=[C]=[C]([H])/[C](=[C]([H])/[C]([H])=[...([H])([H])[H])[C]([C]([H])([H])[H])([C]([H])([H])[H])[C]([H])([H])[C@@]([H])([O][C](=[O])[C]([H])([H])[H])[C]1([H])[H]'`

mattwthompson:
There are roughly tens or hundreds of similar comparisons that fail in various OpenEye comparisons. This is probably due to the fact that molecules are modified in the process of comparison (without modifying arguments to non-default values). @j-wags suggests we deepcopy the molecules and run comparisons on those so that the original molecules are not modified.

mattwthompson:
Locally, deepcopying has fixed the ~100 failed tests in the `OpenEyeToolkitWrapper` tests.

Still getting the same error as above, although on a different molecule now (`openforcefield/tests/test_toolkits.py::TestRDKitToolkitWrapper::test_non_standard_inchi_round_trip[molecule119]`)

j-wags:
Alright, these commits should do it. Will merge when tests pass.

j-wags:
As for why `Drugbank_2684` was failing roundtrip InChI tests in RDKit, here's my post from Slack:

> Good news -- It was always broken! The molecule was always getting scrambled during the InChI roundtrip, we just didn't look very close before. Now that we try to strip pyrimidal nitrogen stereochemistry, the toolkit is realizing that the molecule is so badly mangled that it can't even perform a SMARTS search.
> So, we throw it onto the large pile of molecules that fail InChI roundtrips. Remember, everyone: InChI is not a safe way to serialize an OFFMol.

j-wags:
Welp, something's broken with the QCArchive interface. I've pinged @jthorton, who should know more about it, but I won't let it hold up the release. Thanks for the hard work on this, @mattwthompson -- Merging!

Issue #644, opened 2020-07-14
OpenEye atom stereochemistry assumptions failed.


SimonBoothroyd:
**Describe the bug**
An `Exception` is raised when creating a molecule with SMILES `CC/C=C\CCO` and then writing to SDF.

**To Reproduce**

`conda install --name off070 -c conda-forge -c omnia -c openeye openforcefield openeye-toolkits`

and then:

```
from openforcefield.topology import Molecule
openff_molecule = Molecule.from_smiles("CC/C=C\CCO")
openff_molecule.to_file("tmp.sdf", "SDF")
```

**Output**

```
    openff_molecule.to_file(file.name, "SDF")
  File "xxx/python3.7/site-packages/openforcefield/topology/molecule.py", line 3073, in to_file
    toolkit.to_file(self, file_path, file_format)
  File "xxx/python3.7/site-packages/openforcefield/utils/toolkits.py", line 500, in to_file
    oemol = self.to_openeye(molecule)
  File "xxx/python3.7/site-packages/openforcefield/utils/toolkits.py", line 973, in to_openeye
    'Programming error: OpenEye bond stereochemistry assumptions failed.'
Exception: Programming error: OpenEye bond stereochemistry assumptions failed.
```

**Computing environment (please complete the following information):**

<details>
  <summary>`conda list`</summary>

```
amberlite                 16.0                     pypi_0    pypi
ambertools                17.0                     pypi_0    pypi
appnope                   0.1.0           py37hc8dfbb8_1001    conda-forge
arpack                    3.7.0                hbb27dc6_1    conda-forge
attrs                     19.3.0                     py_0    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
bleach                    3.1.5              pyh9f0ad1d_0    conda-forge
boost                     1.72.0           py37hbf1eeb5_0    conda-forge
boost-cpp                 1.72.0               hef959ae_2    conda-forge
brotlipy                  0.7.0           py37h9bfed18_1000    conda-forge
bson                      0.5.9                      py_0    conda-forge
bzip2                     1.0.8                h0b31af3_2    conda-forge
ca-certificates           2020.6.20            hecda079_0    conda-forge
cairo                     1.16.0            ha8983da_1005    conda-forge
certifi                   2020.6.20        py37hc8dfbb8_0    conda-forge
cffi                      1.14.0           py37h356ff06_0    conda-forge
[...]

PR #647, opened 2020-07-14
Make OE re-percieve aromaticity in `to_openeye`

j-wags:
Quick fix for #644 

The quick fix simply has Molecule.to_openeye disregard the OFFMol's existing is_aromatic flags and re-perceives them. In almost all cases, this should be fine.

For more details, see: https://github.com/openforcefield/openforcefield/issues/644#issuecomment-658422943

### DIFF
--- a/openforcefield/tests/test_molecule.py
+++ b/openforcefield/tests/test_molecule.py
@@ -708,9 +708,9 @@ class TestMolecule:
                               'DrugBank_3726', 'DrugBank_3844', 'DrugBank_3930', 'DrugBank_4161',
                               'DrugBank_4162', 'DrugBank_4778', 'DrugBank_4593', 'DrugBank_4959',
                               'DrugBank_5043', 'DrugBank_5076', 'DrugBank_5176', 'DrugBank_5418',
-                              'DrugBank_5737', 'DrugBank_5902', 'DrugBank_6304', 'DrugBank_6305',
-                              'DrugBank_6329', 'DrugBank_6355', 'DrugBank_6401', 'DrugBank_6509',
-                              'DrugBank_6531', 'DrugBank_6647',
+                              'DrugBank_5737', 'DrugBank_5902', 'DrugBank_6295', 'DrugBank_6304',
+                              'DrugBank_6305', 'DrugBank_6329', 'DrugBank_6355', 'DrugBank_6401',
+                              'DrugBank_6509', 'DrugBank_6531', 'DrugBank_6647',
 
                               # These test cases are allowed to fail.
                               'DrugBank_390', 'DrugBank_810', 'DrugBank_4316', 'DrugBank_4346',
--- a/openforcefield/utils/toolkits.py
+++ b/openforcefield/utils/toolkits.py
@@ -1331,7 +1331,7 @@ class OpenEyeToolkitWrapper(ToolkitWrapper):
         for atom in molecule.atoms:
             oeatom = oemol.NewAtom(atom.atomic_number)
             oeatom.SetFormalCharge(atom.formal_charge.value_in_unit(unit.elementary_charge)) # simtk.unit.Quantity(1, unit.elementary_charge)
-            oeatom.SetAromatic(atom.is_aromatic)
+            #oeatom.SetAromatic(atom.is_aromatic)
             oeatom.SetData('name', atom.name)
             oeatom.SetPartialCharge(float('nan'))
             oemol_atoms.append(oeatom)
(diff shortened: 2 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -91,7 +91,7 @@ jobs:
         conda activate test
         if [[ "$RDKIT" == true ]]; then conda install rdkit; fi
         if [[ "$OPENEYE" == true ]]; then
-          conda install openeye-toolkits=2019
+          conda install openeye-toolkits
```
mattwthompson, line 94: I could see the case for `openeye-toolkikts >=2020` ... probably better to leave it off - how much do we worry about users having old versions of the toolkit installed? If they _did_ (and they shouldn't) then the errors may be cryptic.

openforcefield/utils/toolkits.py
```diff
@@ -1331,7 +1331,7 @@ def to_openeye(molecule, aromaticity_model=DEFAULT_AROMATICITY_MODEL):
         for atom in molecule.atoms:
             oeatom = oemol.NewAtom(atom.atomic_number)
             oeatom.SetFormalCharge(atom.formal_charge.value_in_unit(unit.elementary_charge)) # simtk.unit.Quantity(1, unit.elementary_charge)
-            oeatom.SetAromatic(atom.is_aromatic)
+            #oeatom.SetAromatic(atom.is_aromatic)
```
mattwthompson, line 1334: Please add a note here if we intend to un-comment it in the future, or (better, I think) remove this altogether since it does not fit into the medium- or long-term plans you proposed

openforcefield/utils/toolkits.py
```diff
@@ -1347,12 +1347,14 @@ def to_openeye(molecule, aromaticity_model=DEFAULT_AROMATICITY_MODEL):
             oebond = oemol.NewBond(oemol_atoms[atom1_index],
                                    oemol_atoms[atom2_index])
             oebond.SetOrder(bond.bond_order)
-            oebond.SetAromatic(bond.is_aromatic)
+            #oebond.SetAromatic(bond.is_aromatic)
```
mattwthompson, line 1350: Same as above

openforcefield/tests/test_molecule.py
```diff
@@ -708,9 +708,9 @@ def test_to_from_iupac(self, molecule):
                               'DrugBank_3726', 'DrugBank_3844', 'DrugBank_3930', 'DrugBank_4161',
                               'DrugBank_4162', 'DrugBank_4778', 'DrugBank_4593', 'DrugBank_4959',
                               'DrugBank_5043', 'DrugBank_5076', 'DrugBank_5176', 'DrugBank_5418',
-                              'DrugBank_5737', 'DrugBank_5902', 'DrugBank_6304', 'DrugBank_6305',
-                              'DrugBank_6329', 'DrugBank_6355', 'DrugBank_6401', 'DrugBank_6509',
-                              'DrugBank_6531', 'DrugBank_6647',
+                              'DrugBank_5737', 'DrugBank_5902', 'DrugBank_6295', 'DrugBank_6304',
```
mattwthompson, line 711: A bigger issue to tackle than in here, but is there documentation about why these lists of molecules are problematic and why?

.github/workflows/CI.yml
j-wags: This should actually be reverse-compatible! So there's no need to pin OE or worry about older versions :-)

openforcefield/utils/toolkits.py
j-wags: Added a TODO item to revisit the relationship between OFFMols' and OEMols' aromatic flags in a96274f

openforcefield/utils/toolkits.py
j-wags: Added a TODO item to revisit the relationship between OFFMols' and OEMols' aromatic flags in a96274f

openforcefield/tests/test_molecule.py
j-wags: No. This block is a reservoir of technical debt. Dealing with edge cases in OE and RDK is kinda like running a series of seismographs. When they have big releases it's just like "oh, a bunch of stuff behaves differently now", but to actually go and check+fix every road and power line along the fault would be a full-time job in itself. 

Even worse, I'm sure some of the cases that previously violated one of our assumptions are being _fixed_ in some of these updates. But to catch those we'd need a whole matrix of true/fale negatives/positives. So, it's something that a really cheminformatics-inclined person could set up in the future. But we try to avoid the rabbit hole of doing cheminformatics here.

davidlmobley: I just made a comment in a related issue about aromaticity models, which is probably the reason behind SOME of these.

davidlmobley: That was over here, FTR: https://github.com/openforcefield/openforcefield/issues/644#issuecomment-666605131

PR #648, opened 2020-07-14
Remove utils.structure module

mattwthompson:
If this works, can be pushed into either 0.7.1 or 0.8.0

https://github.com/openforcefield/openforcefield/issues/333


mattwthompson:
I went ahead and removed the deprecation warning that was applied to this function. If we still wanted to remove it we'd need to either drop or update `examples/inspect_assigned_parameters/inspect_assigned_parameters.ipynb`; just leaving it and removing the warning is the path of least resistance. The blocking test is passing locally now

PR #649, opened 2020-07-15
Fix problem in searching for stereo-specific SMARTS

j-wags:
- [x] Add `OEPrepareSearch` command so we standardize aromaticity model prior to searching
- [x] Update RDKit so we can do stereo-specific SMARTS searches
- [x] fix DuplicateParameterError message in `add_parameter`.

During FF porting, we found that searches of the entire AMBER FF14SB OFFXML were raising the following oechem warning (x1000) when glycine was involved in a search:  

```
Warning: OESubSearch::Match() is unable to match atom stereo in the target (default_name) for patterns with atom stereo, call OEPrepareSearch on the target first
```

So, I added `OEPrepareSearch` to `OpenEyeToolkitWrapper.find_smarts_matches`, and the error went away. 


j-wags:
Since the original "bug" here is that OE prints a warning 3000 times and then the task completes successfully, I think this would be a lot of trouble to make a unit test for. Further, I can't explain (or succinctly reproduce) what raised this problem in the first place (it seemed to be "doing a SMARTS search of glycine using some subset of SMARTS in a 2 MB OFFXML.")

So, instead, I'll add tests to ensure that OE and RDK can do chiral substructure searches at all. My priorities are:
* (best) only match exact chiralities
* (acceptable) disregard chirality in substructure searching

We should still aspire to hit the "best" solution here, since that will help in a few ambiguous cases with the protein force field (in case we ever want to support R amino acids?) and the bespoke workflow (molecules that are symmetric except for stereochemistry).

j-wags:
Ok, looks like the last few commits should hit the (best) target here. I'll update releasenotes.

PR #650, opened 2020-07-15
Lint codebase and add lint checks to CI

mattwthompson:
I'm not sure if the action will run off of the branch; those checks may need to be verified locally
- [x] Lint codebase with `black` and `isort`
- [x] Manually fix things that `flake8` complains about
- [x] Set up linter checks as a GitHub Action
- [x] Update setup.cfg
- [x] Update dev docs


j-wags:
Note: Since this will cause massive merge conflicts with every branch, we'll hold off on this until the number of open PRs is lower.

mattwthompson:
Agreed -  #615 specifically should be merged in beforehand, with care - and I'm happy to be responsible for gently merging things back into here.

j-wags:
Per discussion in this morning's call. let's open a new PR to implement black (and isort, if possible). We have a limited number of open PRs so I'd be happy to merge black at this point and resolve changes as needed.

Issue #637, opened 2020-07-04
API docs are broken


jchodera:
**Describe the bug**
The [0.7.0 API docs](https://open-forcefield-toolkit.readthedocs.io/en/0.7.0/api/generated/openforcefield.topology.Molecule.html#openforcefield.topology.Molecule) don't render and link to method API docs as they should.

**To Reproduce**
Go to https://open-forcefield-toolkit.readthedocs.io/en/0.7.0/api/generated/openforcefield.topology.Molecule.html#openforcefield.topology.Molecule and try to click on the method names.

j-wags:
Closing as duplicate to #602, where I approved breaking these links in order to unblock the 0.7.0 release (since the sphinx 3 release had broken everything, and RTD doesn't let us downgrade sphinx). 

Looks like someone has started hosting a fork of the abandoned `m2r` project (called `m2rr`). It's pypi-only right now, but that's not a bad start. https://github.com/qhua948/m2rr

PR #660, opened 2020-07-23
Use m2r2 for markdown doc rendering

dotsdl:
Addresses #598.

I added [m2r2](https://pypi.org/project/m2r2/) as a replacement for `m2r` in our documentation build. Also attempted to fix the table for `<ProperTorsions>` in a nice way, but not sure I'm happy with the result. Trying to make something that works reasonably both rendered and raw in markdown; feedback desired.


j-wags:
Oh my goodness -- If the `- pip` section in the environment `yml` works, I'm going to cry. I spent a few hours with this before the 0.7.1 release and couldn't figure out how in the world to add a pip dependency to the docs env.

j-wags:
Testing this build here: https://readthedocs.org/projects/open-forcefield-toolkit/builds/11524054/

mattwthompson:
The rendered markdown passes the eye test (in particular https://open-forcefield-toolkit.readthedocs.io/en/doc-fix/smirnoff.html) but I'm not sure what all to look for. There's a TOC on the master build, should that be here, too? https://open-forcefield-toolkit.readthedocs.io/en/master/smirnoff.html

Re: the table, my personal preference would be close to what it was prior, which was probably something similar to the `vdW` section:
image

which looks fine, not perfect, in the raw form

```
| vdW section tag version | Tag attributes and default values                                                                                                                                                                                 | Required parameter attributes                 | Optional parameter attributes |
|-------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------|-------------------------------|
| 0.3                     | `potential="Lennard-Jones-12-6`, `combining_rules="Lorentz-Berthelot"`, `scale12="0"`, `scale13="0"`, `scale14="0.5"`, `scale15="1.0"`, `cutoff="9.0*angstrom"`, `switch_width="1.0*angstrom"`, `method="cutoff"` | `smirks`, `epsilon`, (`sigma` OR `rmin_half`) | `id`, `parent_id`             |
```

I'm not creative enough to think of a way around the verbosity. One concern I have with the changes here is that, at first glance, the rows imply connections between values in adjacent columns that I believe are not intended.

j-wags:
Fixing the `ProperTorsions` section of the spec docs would be nice here, and is worth a merge by itself. 

> Re: the table, my personal preference would be close to what it was prior, which was probably something similar to the vdW section:

I agree with Matt on this. The way that the proposed changes put the ProperTorsion attributes on different rows is a bit rough, and I'd prefer them the same way as we have vdW attributes. I'll approve this PR once that change is made. 

> There's a TOC on the master build, should that be here, too?

The `master` build is ancient... the real bleeding-edge docs are on `latest`. In fact, I'll just go ahead and hide `master` altogether. 

I was hoping that this would recover the API docs from #637, and the links to them from the release notes. But I'm happy to take little bites of this problem to bring back functionality.

dotsdl:
Table reverted to comma-delimited, horizontal format. Building now!

dotsdl:
Okay, this is now consistent with the VDW table @mattwthompson pointed out above; that was a good catch.

PR #661, opened 2020-07-24
Add conformer generation CLI utility

### LINKED ISSUE #641: Make a conformer generation CLI tool
People like the conformer energies script. We should make one that also _generates_ conformers. However, we don't want people to confuse OpenFF force fields' performance with that of the original conformer generation method (since right now we do not have a native conf. gen method).

Create a script, `generate_conformers.py`, that can be invoked as follows:

`python generate_conformers.py -m mymol.smi -f openff-1.2.0 --rdkit --rms_cutoff 0.25`

Required:
* `--rdkit` or `--openeye` is a required argument specifying which underlying toolkit to use for initial conformer generation
* `-m`, `--molecule filename` is the molecule file to load
* `-f`, `-forcefield openff-x.y.z` is the FF to load (with or without the `.offxml` part, maybe also accepting "X.Y.Z" string)

Optional
* `-r`, `-rms_cutoff` (default `0.25`, angstroms) -- The redundancy cutoff between _pre-minimized_ conformers to pass to the `generate_conformers` method
* `-p`, `-prefix mymol` (default `molecule.name` for each loaded molecule if defined, otherwise `moleculeN`, where N is the order is was read from the file) -- string prefix for outputted files (eg `mymol_conf1.sdf`).
* `--constrained` (default `False`/unset) -- Whether to use `openff-unconstrained-X.Y.Z`. By default we use the unconstrained FF

Notes:
* If possible, there should be either _no_ or _a very generous_ energy cutoff during initial conformer _generation_ (that way, RDKit won't disqualify a "good" conformer early on because of its built-in FF)
* ~This should use a "conformer collapsing" scheme similar/identical to `conformer_energies.py`~ This becomes complicated if the repeats are non-contiguous, so let's just have an optional `--deduplicate-conformers` argument instead, and by default just run naively on all input mols. 
* If charges are defined on the molecule when loaded (so, like, if the `<atom.dprop.PartialCharges>` flag is set), then those should be used for the calculation (passed to the `charge_from_molecules` kwarg)
* The conformers should be output as separate SDFs with SD data pairs for 
    * `absolute_energy` in kcal/mol
    * conformer generation method (openeye version or rdkit version)
    * ~partial charges (so, use `ForceField.to_openmm_system`'s `return_topology` kwarg, and attach the final coordinates to `ret_topology._reference_molecules[0]`, then save that to file)~ My mistake -- this data isn't available currently.

Tests:
* loading one molecule from a `.sdf` file WITHOUT charges
* loading one molecule from a `.sdf` file WITH charges
*  (OE only) loading one molecule from a `.mol2` file WITH charges
*  (OE only) loading one molecule from a `.mol2` file WITHOUT charges (is this even possible)
* loading a molecule with a defined `name` (from any format) and ensuring the output file has that prefix
* loading a molecule with a defined `name` (from any format), and providing a `-f` option and ensuring the output file has the `-f` prefix
* loading a molecule with a defined `name` (from any format) and ensuring the output file has that prefix
* loading multiple molecules from a `.sdf` file
* loading one molecule from SMILES in a `.smi` file
* loading multiple molecules from SMILES in a `.smi` file
* loading a molecule with ambiguous stereo from SMILES and enumerating stereoisomers
* loading a molecule with defined stereo from SMILES and preserving that stereochemistry

Undecided/Implementer's choice:
* What to do if prefix is provided, but multiple molecules are in file?

Notable "Don'ts":
* Don't try to provide "relative energies", since our conformer collapsing routine isn't reliable, and people will get mixed up about which "minimum" the energy is relative to.
* Don't make the `-f` argument optional, since then people wouldn't know which FF is being used on the backend. If we want to remove that argument, we could make a variant of this script for each FF (`generate_conformers_1.2.0.py`)
* If we want to collapse conformers, we could do so by appending the SDF texts to each other. But I think this would be a bad built-in behavior because OFFTK doesn't natively handle multiconformer SDFs. 
* I thought about whether we should make a CLI entrypoint for this script, or build it into a OFF Toolkit module, and I don't think we should. Keeping it as just a python script makes the contents more visible to our users (some of whom are reusing our code already), and makes it easier to update/version (since it's not hidden somewhere in an install directory)

mattwthompson:
WIP, resolves #641

* The module structure is temporary and likely to live elsewhere in the future
* I'd like us to think about how to test these utilities more rigorously than we might normally do in a unit test, i.e. add some safeguards that the results are reasonable in addition to ensuring that meet the requirements of the spec

### TODO:
- [ ] De-duplicate conformers after OpenMM minimization
- [ ] Fix `return_topology` issue with partial charges

### Tests to include

Function calls
- [x] loading one molecule from a .sdf file WITHOUT charges
- [x] loading one molecule from a .sdf file WITH charges
- [x] (OE only) loading one molecule from a .mol2 file WITH charges
- [ ] (OE only) loading one molecule from a .mol2 file WITHOUT charges (is this even possible)
- [ ] loading a molecule with a defined name (from any format) and ensuring the output file has that prefix
- [ ] loading a molecule with a defined name (from any format), and providing a -f option and ensuring the output file has the -f prefix
- [ ] loading a molecule with a defined name (from any format) and ensuring the output file has that prefix
- [ ] loading multiple molecules from a .sdf file
- [x] loading one molecule from SMILES in a .smi file
- [x] loading multiple molecules from SMILES in a .smi file
- [x] loading a molecule with ambiguous stereo from SMILES and enumerating stereoisomers
- [x] loading a molecule with defined stereo from SMILES and preserving that stereochemistry

Direct CLI calls
- [ ] Ensure `--prefix` overrides original filename
- [ ] Ensure conformers are enumerated
- [ ] Ensure stereoisomers are enumerated?
- [ ] Ensure SD data tags are accurately written
- [ ] Test different forms of `--forcefield`
- [ ] Test different forms of `--toolkit`
- [ ] Gracefully handle cases  of arguments requesting unavailable resources (force field versions, un-installed/licensed toolkits)
- [ ] Sanity checks on piping several calls together


mattwthompson:
Moving this over to `openff-cli` seems to have worked well so I'll go ahead and close this here.

Issue #664, opened 2020-07-27
Ensuring non-RDKit tests in CI do not install RDKit


mattwthompson:
RDKit is installed in `rdkit: false, openeye: true` CI runs. See one such log [here](https://github.com/openforcefield/openforcefield/runs/912672348?check_suite_focus=true). At first, I thought this may be my fault; in #588 I switched the RDKit and OpenEye toolkit installs from lines in the install script to separate environment files. It's not intuitive that this would be the case, though, since those files did not change and not enough time has placed to fall back to the classic "something happened upstream" explanation. I looked into some older CI logs (see [here](https://github.com/openforcefield/openforcefield/runs/883749877?check_suite_focus=true) from more than a week ago) and this was happening before - see in that log that RDKit was installed and its tests were not skipped. Were there safeguards previously in place to ensure RDKit was not installed? 

It's not clear to me why the typical user (i.e. outside of debugging) would intentionally not have RDKit installed and it is shipped as an explicit dependency in the conda packaging. (The opposite is not true; some users may not have OpenEye licenses and it is shipped as an optional-but-recommended sort of dependency.) So I'm assuming that the purpose of these tests is to force toolkit methods to call OpenEye functions. Since OpenEye is listed higher in `GLOBAL_TOOLKIT_REGISTRY`, the intended functions will still get called, but RDKit will still be hanging out in the background instead of not installed. The edge case of a test really, really needing RDKit to not be installed (see [here](https://github.com/openforcefield/openforcefield/pull/615#discussion_r461141438) won't work, however.

If my thinking above is accurate, I will update the test environment to ensure the right toolkits are installed and add a check.

PR #667, opened 2020-07-30
Verify toolkit installation

mattwthompson:
Attempt at fixing #664

- [x] Determine how to conditionally uninstall a package (that may or may not be found) without sending a bad exit code


mattwthompson:
@jaimergp this has me stumped and I could use your help here whenever you had some bandwith. I expected `'RDKit unexpectedly found'` to be caught (see linked issue) but the one-liner isn't even finding OpenEye installed. It shows up in the conda list and nothing fishy has been happening with the license lately, I don't think. Locally, those four one-liners behave as expected (installing/uninstalling each toolkit along the way). It doesn't strike me as an issue with GitHub Actions, conda, or our own software.

jaimergp:
Are we sure is it not due to licensing? This line is oddly suspicious given the expiry date (Jun 2020):

https://github.com/openforcefield/openforcefield/blob/e923cd65152d7dbb91ae3f0e3c2a38bf1874364f/.github/workflows/CI.yml#L94

mattwthompson:
That's probably it - I botched this in #588; the lifetime of that PR extends to before you update the org secret and because that file was changed so much, patch did not get merged in.

mattwthompson:
Okay, that's a relief. Now the failures are only as expected (`RDKit=false, OpenEye=true` because RDKit sneaks in when it maybe should be installed)

jaimergp:
You can `conda remove --force rdkit` in that step condition if needed.

mattwthompson:
I got hung up on how to get that command to behave when the package it's trying to remove might not be there. Not pretty but a bash if does the job

mattwthompson:
Okay, this serves the purpose I intended it to serve. There is probably some golf that could condense it down to fewer characters, but I think it's readable as is.

jaimergp:
LGTM!

PR #670, opened 2020-08-03
Assign molecule charges on ref mols when return_topology=True

j-wags:
Now that we're making more extensive use of the `return_topology` kwarg to `create_openmm_system`, I thought it'd make sense to also assign the assigned charges to the reference molecules. This PR adds that behavior, as well as a test that it works for our current charge methods.

### DIFF
--- a/openforcefield/tests/test_forcefield.py
+++ b/openforcefield/tests/test_forcefield.py
@@ -1883,6 +1883,36 @@ class TestForceFieldChargeAssignment:
             q, sigma, epsilon = nonbondedForce.getParticleParameters(particle_index)
             assert q != 0 * unit.elementary_charge
 
+    @pytest.mark.parametrize('charge_method,additional_offxmls', [ ('ToolkitAM1BCC', []),
+                                                                   ('LibraryCharges', [xml_ethanol_library_charges_ff]),
+                                                                   ('ChargeIncrementHandler', [xml_charge_increment_model_formal_charges]),
+                                                                   ('charge_from_molecules', [])])
+    def test_charges_on_ref_mols_when_using_return_topology(self, charge_method, additional_offxmls):
+        """Ensure that charges are set on returned topology if the user specifies 'return_topology=True' in
+        create_openmm_system"""
+        from simtk.openmm import NonbondedForce
+
+        mol = create_acetate()
+        ff = ForceField('test_forcefields/smirnoff99Frosst.offxml', *additional_offxmls)
+        charge_mols = []
+        if charge_method == 'charge_from_molecules':
+            mol.partial_charges = np.array([-1.3, -0.2, -0.1, 0., 0.1, 0.2, 0.3]) * unit.elementary_charge
+            charge_mols = [mol]
+        omm_system, ret_top = ff.create_openmm_system(mol.to_topology(),
+                                                      charge_from_molecules=charge_mols,
+                                                      return_topology=True)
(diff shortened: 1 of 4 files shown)


openforcefield/tests/test_forcefield.py
```diff
@@ -1883,6 +1883,36 @@ def test_library_charges_dont_parameterize_molecule_because_of_incomplete_covera
...
+
+        # Make sure the charges on the molecule are all nonzero, and that the molecule's
+        # partial_charges array matches the charges in the openmm system
+        all_charges_zero = True
+        for particle_index, ref_mol_charge in enumerate(ref_mol_from_ret_top.partial_charges):
+            q, sigma, epsilon = nonbondedForce.getParticleParameters(particle_index)
+            assert q == ref_mol_charge
+            if q != 0. * unit.elementary_charge:
+                all_charges_zero = False
+        assert not(all_charges_zero)
```
mattwthompson, line 1915: Interesting, I didn't know `not` could be a function as well an operator

openforcefield/tests/test_forcefield.py
```diff
@@ -1883,6 +1883,36 @@ def test_library_charges_dont_parameterize_molecule_because_of_incomplete_covera
...
+                                                      charge_from_molecules=charge_mols,
+                                                      return_topology=True)
+        nonbondedForce = [f for f in omm_system.getForces() if type(f) == NonbondedForce][0]
+        ref_mol_from_ret_top = [i for i in ret_top.reference_molecules][0]
+
+        # Make sure the charges on the molecule are all nonzero, and that the molecule's
+        # partial_charges array matches the charges in the openmm system
+        all_charges_zero = True
+        for particle_index, ref_mol_charge in enumerate(ref_mol_from_ret_top.partial_charges):
+            q, sigma, epsilon = nonbondedForce.getParticleParameters(particle_index)
```
mattwthompson, line 1911: This only compares OFFMol <-> OMM System partial charges for the first molecule in the topology, but a) this covers my use case and b) I'm pretty sure that's likely, but perhaps not guaranteed, to be fine for the rest of the system. I can't think of a reason why the 0th molecule in the iteration would pass but the Nth would fail? It may be worth adding a note or thinking about adding more tests later.

openforcefield/tests/test_forcefield.py
j-wags: There must have been some trauma in my early life around PEMDAS, because I assume monkeys-paw-tier malicious intent by the python interpreter when evaluating any logical statement I write. As you may have seen elsewhere in the codebase, I am big on parentheses.

openforcefield/tests/test_forcefield.py
j-wags: Yeah, I was tempted to actually make this a copy of the extensive `charge_method_hierarchy` test, which makes a system containing five molecules, each of which ends up being treated by a different charge method. But as much as I like the thoroughness of that test, it is starting to depart the land of "simple unit tests", and I've mangled previous projects with test-complexity-creep. So, I want to leave the "multiple molecule" and "mix-of-charge-methods" situations untested, and assume that the charge assignment machinery is modular enough to guarantee the correct behavior.

mattwthompson: >  it is starting to depart the land of "simple unit tests", 

Agreed!

j-wags:
Thanks, Matt. I've updated the release notes and docs, and will go ahead and merge!

Issue #674, opened 2020-08-07
Have toolkit wrappers know version of the toolkit they're wrapping


mattwthompson:
**Is your feature request related to a problem? Please describe.**
As far as I can tell, `Toolkitwrapper` and therefore `ToolkitRegistry` objects don't know the version of the toolkit that they're wrapping. Would these be useful features?

**Describe the solution you'd like**

* When instantiating a wrapper class, the version detected from the wrapped toolkit is saved
* Each wrapper knows the class, and registries can also query that for each wrapper

Something like 
```python3

In [1]: from openforcefield.utils.toolkits import *

In [2]: reg = ToolkitRegistry(toolkit_precedence=[OpenEyeToolkitWrapper, RDKitToolkitWrapper, AmberToolsToolkitWrapper])

In [3]: # OpenEyeToolkitWrapper().wrapped_version() would return '2020.0.4'

In [4]: # RDKitToolkitWrapper().wrapped_version() would return '2020.03.4'

In [5]: # reg.wrapped_versions() would return something like ['2020.0.4', '2020.03.04', 'whatever_ambertools_says']
```
**Describe alternatives you've considered**
It's easy to enough to just import each of them and check their versions, but this would provide a single interface to them. It may also help debugging weird behavior, i.e. if the wrapper finds a different version than they think they have installed through `conda list`; this would directly store what the wrapper finds.

**Additional context**
This is motivated by writing a CLI, in which I think it would be useful to be able to access the versions of installed toolkits by only passing around a `ToolkitRegistry` object

j-wags:
Ooh, that would be super handy. The only change I'd suggest it to make `reg.wrappped_versions` return a dict of `{'OpenEyeToolkitWrapper': '2020.0.4', ...}`, so that users know which version corresponds to which toolkit. 

I think something like this was desirable for QCSubmit's operation provenance, where if QCSubmit, for example, ran `Molecule.generate_conformers(provenance=True)`, part of the return value would be the toolkit(s) that were used and their versions. So this would be a great step towards finally implementing that.

mattwthompson:
I don't think anything in the original post is missed. This is from the 0.8.0 tag:

```
In [1]: from openforcefield.utils.toolkits import *

In [2]: reg = ToolkitRegistry(toolkit_precedence=[OpenEyeToolkitWrapper, RDKitToolkitWrapper, AmberToolsToolkitWrapper])
   ...:

[...]

PR #675, opened 2020-08-07
Add wrapped toolkit versions to toolkit wrappers and registry

mattwthompson:
#674 


openforcefield/utils/toolkits.py
```diff
@@ -365,6 +383,11 @@ def _check_n_conformers(
         else:
             warnings.warn(wrong_confs_msg, IncorrectNumConformersWarning)
 
+    def __repr__(self):
+        return (
+            f"ToolkitWrapper around {self.toolkit_name} version {self.toolkit_version}"
+        )
```
j-wags, line 389: 👍

openforcefield/utils/toolkits.py
```diff
@@ -4081,6 +4111,10 @@ def __init__(self):
                 f"available. {self._toolkit_installation_instructions}"
             )
 
+        out = subprocess.check_output(["antechamber", "-L"])
+        ambertools_version = out.decode("utf-8").split("\n")[1].split()[3].strip(":")
+        self._toolkit_version = ambertools_version
```
j-wags, line 4116: This isn't perfect, but I do agree that it beats the traditional "we don't have any clue which version of ambertools is down there" (and its neighbor "`conda install -c conda-forge ambertools` --> You've installed AmberTools17, from pypi")

openforcefield/utils/toolkits.py
```diff
@@ -4212,7 +4246,9 @@ def assign_partial_charges(
         # TODO: How should we implement find_executable?
         ANTECHAMBER_PATH = find_executable("antechamber")
         if ANTECHAMBER_PATH is None:
-            raise IOError("Antechamber not found, cannot run charge_mol()")
+            raise AntechamberNotFoundError(
+                "Antechamber not found, cannot run charge_mol()"
+            )
```
j-wags, line 4251: Make sure to note the change in exception types in the `Behavior Changed` section of the release notes.

openforcefield/utils/toolkits.py
```diff
@@ -247,6 +255,16 @@ def is_available():
         """
         return NotImplementedError
 
+    @property
+    def toolkit_version(self):
+        """
+        Return the version of the wrapped toolkit
```
j-wags, line 261: Can we say anything about the type that's returned here? If we can hold ourselves to it, maybe specify that it'll always be `str`, and add that to the test so future developers know if they've broken that contract.

openforcefield/utils/toolkits.py
mattwthompson: Agreed; I've added a TODO to note that this isn't perfect and we welcome an improvement

openforcefield/utils/toolkits.py
mattwthompson: Generally it's strings all the way down, but I didn't implement a version for `BuiltInToolkitWrapper`. Blocking, or should it default to something like `'0.0.0'` since we're probably not interested in versioning it?

```
In [4]: [(type(key), type(val)) for key, val in GLOBAL_TOOLKIT_REGISTRY.registered_toolkit_versions.items()]
Out[4]: [(str, str), (str, str), (str, str), (str, NoneType)]

PR #677, opened 2020-08-10
Lint with black and set up automated black check

mattwthompson:
Split out from #650. `isort` and `black` have some small disagreements that could be resolved here but I would rather deal with those later while resolving some `flake8` complaints. The changes here are roughly the minimum to get this adopted as quickly as possible, and I tried to keep the commit history discrete if we wish to roll anything back.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,9 +2,6 @@ name: lint
 
 on:
   push:
-    paths:
-      - 'openforcefield/**.py'
-      - 'setup.cfg'
 
 jobs:
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,28 @@
+name: lint
+
+on:
+  push:
+    paths:
+      - 'openforcefield/**.py'
+      - 'setup.cfg'
+
+jobs:
+
+  lint:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v1
+        with:
+          python-version: '3.7'
+
+      - name: Install linters
+        run: |
+          pip install black
+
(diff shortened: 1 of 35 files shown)


docs/developing.rst
```diff
@@ -266,11 +266,16 @@ The naming conventions of classes, functions, and variables follows `PEP8 <https
...
-Anything not covered above is up to personal preference.
-To remove the human friction from code formatting, we will likely adopt a standard formatter like `black <https://github.com/psf/black>`_ in the near future.
+There is a step in CI that uses these tool(s) to check for a consistent style. To ensure that changes follow these standards, you can install and run these tool(s) locally:
+
+.. code-block:: bash
 
+    $ conda install black -c conda-forge
+    $ black openforcefield
+
+Anything not covered above is up to personal preference.
```
SimonBoothroyd, line 281:
```suggestion
Anything not covered above is currently up to personal preference, but may change as new linters are added.
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,28 @@
+name: lint
+
+on:
+  push:
+    paths:
+      - 'openforcefield/**.py'
+      - 'setup.cfg'
```
SimonBoothroyd, line 7:
```suggestion
```

Can probably drop these lines as the linting step is normally very fast.

mattwthompson:
Thanks @SimonBoothroyd!

j-wags:
Looks great. Thanks for making the call and merging this in -- I've been hesitating too much on this and it's good to just drop the hammer and deal with the consequences instead of leaving it hanging forever!

j-wags:
Looks great. Thanks for making the call and merging this in -- I've been hesitating too much on this and it's good to just drop the hammer and deal with the consequences instead of leaving it hanging forever!

PR #678, opened 2020-08-10
Add ForceField.deregister_parameter_handler

j-wags:
- [x] Adds `ForceField.deregister_parameter_handler`, which accepts either a string, ParameterHandler type, or ParameterHandler instance. 


j-wags:
@SimonBoothroyd Could you review and let me know if this covers your use case?

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -795,6 +795,22 @@ def get_parameter_io_handler(self, io_format):
...
+        Delete a parameter handler specified by tag name, class, or instance.
+
+        Parameters
+        ----------
+        handler: str, openforcefield.typing.engines.smirnoff.ParameterHandler-derived type or object
+            The handler to delete.
+        """
+        if isinstance(handler, ParameterHandler):
+            tagname = handler.TAGNAME
+        elif isinstance(handler, type):
```
SimonBoothroyd, line 809: Perhaps:

```suggestion
        elif issubclass(handler, ParameterHandler):
```
?

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -795,6 +795,22 @@ def get_parameter_io_handler(self, io_format):
...
+            tagname = handler._TAGNAME
+        else:
```
SimonBoothroyd, line 811: You might consider adding a fallback with a more informative exception 
```suggestion
        elif isinstance(tagname, str):
            tagname = handler
        else:
            raise InvalidHandlerKey(handler) 
```
however you could also just use a built-in exception like `NotImplementedError` for now if you don't have time!

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -795,6 +795,22 @@ def get_parameter_io_handler(self, io_format):
 
         return io_handler
 
+    def delete_parameter_handler(self, handler):
```
SimonBoothroyd, line 798: You may consider naming this `remove_parameter_handler` to be consistent with built-in methods like `List.remove`. Or `deregister_parameter_handler` to match the `register_parameter_handler` method, although (separate PR) adding `add_parameter_handler` and adding a deprecation warning to `register_parameter_handler` to remove in 0.9.0 would be my +1 to get a more consistent API feel.

j-wags: This has made me realize that we lack rules and consistency for when to use `register/deregister` vs `add/delete/remove`. The only other use of `remove/delete` I can see is `ParameterHandler.delete_cosmetic_attribute`, but I don't recall putting much thought into that name.

From first principles, `register/deregister` clearly applies to _registries_, but it's not clear to me whether registries should hold _types_ or _instances_. For example, `ForceField.parameter_handler_classes` is a registry of types -- Basically anything that quacks like a `ParameterHandler` can plug in there and the `ForceField` can initialize them at will. But `ToolkitRegistry.toolkits`is a registry of _instances_, which just seems to be different in some ways, but still works with the pattern. 

After some thought, I think your second suggestion is right -- I'll plan to change it to `deregister_parameter_handler`.

Issue #514, opened 2020-02-12
Molecule.from_iupac Returns Empty Molecule on Failure


SimonBoothroyd:
**Describe the bug**
The `Molecule.from_iupac` function returns an empty molecule when `oeiupac.OEParseIUPACName(oemol, iupac_name)` fails rather than raising an exception.

**To Reproduce**
```
from openforcefield.topology import Molecule
assert len(Molecule.from_iupac(".BETA.-PINENE").atoms) > 0
```

**Output**
```
Warning: Failed to parse name: .BETA.-PINENE
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
```

jchodera:
Good catch!

What kind of exception should we return? A [`ValueError`](https://docs.python.org/3/library/exceptions.html#ValueError)?

SimonBoothroyd:
I wonder if it would be good to start moving towards more custom exceptions for public API points. Perhaps something like:

```
class MoleculeParseError(Exception):
    """An exception raised when a cheminformatics toolkit
    is unable to parse a molecule from a particular source.
    """

    def __init__(self, message, toolkit_registry):
        """
        Parameters
        ----------
        toolkit_registry: ToolkitRegistry
            The registry which attempted to parse the molecule
            source.
        """

        super(MoleculeParseError, self).__init__(message)
        self.toolkit_registry = toolkit_registry
```

j-wags:
Yes, we should move toward custom exceptions -- It's not standard practice
to build off of Python built-in exceptions. At some point we should define
a logical OFFTK exception hierarchy and then do a refactor so we only throw
those.

PR #682, opened 2020-08-18
Update IUPAC functionality (refactor to toolkit wrapper, handle parsing failures)

mattwthompson:
- [x] Move the meat of `Molecule.{to|from}_iupac` to `OpenEyeToolkitWrapper.{to|from}_iupac`
- [x] Handle the "easy" part of #514 (cc @simonboothroyd) but does not tackle the goal of a more elegant exception hierarchy
- [x] Note that in https://github.com/openforcefield/openforcefield/commit/76a8aff3e3017ff864206970a3b71ff953e7f9b2, the tests behave quite differently than the similar tests in `TestMolecule`. I have no idea why, as the code is nearly identical.
- [x] Remove most IUPAC tests from `test_molecule.py` ? -- See below, `MiniDrugBank` was loaded in from separate files


openforcefield/topology/molecule.py
```diff
@@ -3490,16 +3490,23 @@ def chemical_environment_matches(
 
         return matches
 
-    # TODO: Move OE-dependent parts of this to toolkits.py
     @classmethod
     @OpenEyeToolkitWrapper.requires_toolkit()
```
j-wags, line 3494:
```suggestion

```

We should be able to ditch this now that it routes to a ToolkitRegistry. Trying to call the method will result in an informative ToolkitRegistry error like "I tried to do `from_iupac` but I only know about RDKitToolkitWrapper and AmberToolsToolkitWrapper and neither of them can do `to_iupac`"

openforcefield/topology/molecule.py
```diff
@@ -3522,20 +3529,31 @@ def from_iupac(cls, iupac_name, **kwargs):
...
-        oeiupac.OEParseIUPACName(oemol, iupac_name)
-        oechem.OETriposAtomNames(oemol)
-        result = oechem.OEAddExplicitHydrogens(oemol)
-        if result == False:
-            raise Exception("Addition of explicit hydrogens failed in from_iupac")
-        molecule = cls.from_openeye(oemol, **kwargs)
         return molecule
 
-    # TODO: Move OE-dependent parts of this to toolkits.py
     @OpenEyeToolkitWrapper.requires_toolkit()
```
j-wags, line 3555:
```suggestion

```

openforcefield/tests/test_molecule.py
```diff
@@ -847,95 +847,37 @@ def test_to_from_rdkit(self, molecule):
...
+        cholesterol = Molecule.from_smiles(
+            "C[C@H](CCCC(C)C)[C@H]1CC[C@@H]2[C@@]1(CC[C@H]3[C@H]2CC=C4[C@@]3(CC[C@@H](C4)O)C)C"
         )
-        assert molecule.is_isomorphic_with(
-            molecule_copy, atom_stereochemistry_matching=not undefined_stereo
+
+        cholesterol_iupac = cholesterol.to_iupac()
+
+        assert Molecule.are_isomorphic(
+            cholesterol, Molecule.from_iupac(cholesterol_iupac)
```
j-wags, line 880: 👍

openforcefield/utils/toolkits.py
```diff
@@ -160,6 +160,10 @@ class ChargeCalculationError(MessageException):
     pass
 
 
+class InvalidIUPACNameError(MessageException):  # TODO: Should this live in molecule.py?
```
j-wags, line 163:
```suggestion
class InvalidIUPACNameError(MessageException):
```

I think this is the right place for it.

j-wags (commented):
I'm pretty happy with the state of this, but the quote:

>  Note that in 76a8aff, the tests behave quite differently than the similar tests in TestMolecule. I have no idea why, as the code is nearly identical.

is a bit concerning. I know you did this work >2 weeks ago, but do you recall what was different about the behavior? These broad-molecule-set tests are really valuable to us and I'd like to ensure we aren't losing any generality here.

mattwthompson:
Looks like the difference in behavior is likely due to the two tests loading in the MiniDrugBank from different files (`molecules/MiniDrugBank.sdf` vs `molecules/MiniDrugBank_tripos.mol2`). Thanks to @pavankum for finding this.

I'm not sure the best path forward. It would probably be less of a headache to have the unit tests load the dataset from the same file, so that parsing, etc. errors can be caught in a more unified way. But it might also be worth knowing the cases in which these errors pop up and how they differs between the two files. I believe these are pits of complexity and therefore out of scope of this PR, but we should be aware of them.

mattwthompson:
To elaborate and be more explicit,

* I don't consider this blocking, as fixing it would be hugely out of scope and is more or less a known pile of technical debt
* I'm not sure it's worth passing on to @pavankum as it stands here; one one hand, it's something we need to do and could be a useful point to dig into some of the internals and interoperability, but on the other hand it's not a top priority and it can be a bottomless pit of complexity

j-wags (approved):
Oh, wow, I never realized that the two tests load MiniDrugBank from different sources. That's a bit of a mess. Thanks @mattwthompson and @pavankum for digging into this :-)

Anyway, this PR looks great to me. Be sure to update the release notes before the merge!

PR #684, opened 2020-08-20
Update default ToolkitRegistry() beahvior

### LINKED ISSUE #359: ToolkitRegistry() should initialize an empty registry
@jchodera, the current default constructor for `ToolkitRegistry` will attempt to initialize OE, Amber, and RDK toolkits. The command to initialize an empty ToolkitRegistry is `ToolkitRegistry(toolkit_precedence=[])`. Revisiting this months later, this is really unintuitive, and I think we should change it (`ToolkitRegistry()` should initialize an empty registry). Opening a new issue.

_Originally posted by @j-wags in https://github.com/openforcefield/openforcefield/issues/346#issuecomment-502410801_

mattwthompson:
Fixes #359. Changed behavior includes

* `ToolkitRegistry()` now returns an empty registry, which was discussed in the linked issue.
* `ToolkitRegistry(register_imported_toolkit_wrappers=True)` with `None`-like `toolkit_precedence` now returns a registry with precedence matching `GLOBAL_TOOLKIT_REGISTRY` (`[OpenEyeToolkitWrapper, RDKitToolkitWrapper,
            AmberToolsToolkitWrapper, BuiltInToolkitWrapper]`) instead of arbitrary/semi-random order.
* `ToolkitRegistry(register_imported_toolkit_wrappers=True)` with specified `toolkit_precedence` now returns a registry obeying the order of `toolkit_precedence`, including only toolkit wrappers that can be imported.


mattwthompson:
Looking over this again with fresh eyes, I think the first bullet point is the most important change and it could be the only change. I care less about the  interactions between the `register_imported_toolkit_wrappers` and arguments `toolkit_precedence` - further, I'm not sure about the value of `ToolkitRegistry(register_imported_toolkit_wrappers=True)` compared to `GLOBAL_TOOLKIT_REGISTRY`, which will just try to register everything it can. I believe `all_subclass` returns all subclasses that _could_ be imported, not those that _are_ imported; if this is the case, the `register_imported_toolkit_wrappers` should either be renamed or removed since its name is inaccurate and/or functionality duplicated elsewhere.

j-wags:
Thinking way back, the behavior I expected for the ToolkitRegistry constructor was to be able to say 
```
tkr = ToolkitRegistry([AmberToolsToolkitWrapper, OpenEyeToolkitWrapper, RDKitToolkitWrapper])
```
with no positional arguments. I'd advocate changing to support that use case (ie. making `toolkit_precedence` be the first keyword argument, so you can also use it as if it were the first positional argument) This would still be a small API break, but a worthwhile one.

mattwthompson:
I'm in support of hiding it, and would also be open to removing it. Is there a current use case of "try to import what you can" that isn't covered by `GLOBAL_TOOLKIT_REGISTRY` (and likely to survive the refactors happening over there)? This seems like an edge case borne out of the lifecycle of this code, not one focused on tackling user experience.


> Thinking way back, the behavior I expected for the ToolkitRegistry constructor was to be able to say

All sorts of dangerous behavior pops up now and would be fixed by your suggestion


```python3
In [2]: toolkit_registry = ToolkitRegistry(OpenEyeToolkitWrapper)

In [3]: toolkit_registry.registered_toolkits
Out[3]:
[<openforcefield.utils.toolkits.OpenEyeToolkitWrapper at 0x142d72950>,
 <openforcefield.utils.toolkits.RDKitToolkitWrapper at 0x14a6b92d0>,
 <openforcefield.utils.toolkits.AmberToolsToolkitWrapper at 0x14a6b9690>,
 <openforcefield.utils.toolkits.BuiltInToolkitWrapper at 0x142d83dd0>]

In [4]: ToolkitRegistry('foobar')
Out[4]: <openforcefield.utils.toolkits.ToolkitRegistry at 0x10d9d0090>

In [10]: tkr = ToolkitRegistry([AmberToolsToolkitWrapper, OpenEyeToolkitWrapper, RDKitToolkitWrapper])

In [11]: tkr.registered_toolkits
Out[11]:
[<openforcefield.utils.toolkits.OpenEyeToolkitWrapper at 0x14a8bcc10>,
 <openforcefield.utils.toolkits.RDKitToolkitWrapper at 0x14a8bce50>,
 <openforcefield.utils.toolkits.AmberToolsToolkitWrapper at 0x14a8bc950>,
 <openforcefield.utils.toolkits.BuiltInToolkitWrapper at 0x10da9e750>]

```

So, yeah. That's getting changed.

openforcefield/tests/test_toolkits.py
```diff
@@ -2817,6 +2817,55 @@ def test_check_n_conformers(self):
 class TestToolkitRegistry:
     """Test the ToolkitRegistry class"""
 
+    def test_register_empty_toolkit(self):
```
mattwthompson, line 2820: This test could potentially be dropped if we want `ToolkitRegistry()` to not be allowed, i.e. cannot construct an empty one. Personally I think the flexibility of being able to start with an empty registry is worthwhile and the default registration of toolkits should be handled with whatever update to `GLOBAL_TOOLKIT_*` we figure out.

mattwthompson:
Did we ever reach a resolution on this?

j-wags (approved):
This looks great to me. Thanks, @mattwthompson. Before we merge, could you add a big note about this right at the top of the "behavior changed" section of the release notes? This will probably be the biggest change that will affect downstream dependencies.

PR #685, opened 2020-08-21
Add ForceField.__hash__

### LINKED ISSUE #642: Create a `ForceField.hash` method

I've received several requests for a FF hashing method. This would seem to be pretty simple (just write to an XML string and hash it), but we should keep in mind the following complications: 
* The method will need to strip cosmetic attributes
* The method will need to deterministically order attributes (so a bond's `k` and `length` have to come out in the same order, every time, and we should document what that order is, and add tests to ensure it doesn't change)
* The method will need an option for whether to strip `id` and `parent_id` fields, since they're part of the spec, but are nonfunctional.
* We may want to include the Toolkit's version in the name of the hash (or otherwise store it in a nearby data field) in case different toolkit versions end up hashing things differently and we only catch it later.

We could try to hash the SMIRNOFF data dictionary itself, but I think the OFFXML representation would be stabler.

mattwthompson:
Resolves #642, pending some more thorough testing of the details

I took some liberty with decisions here, quoting from the linked issue

> The method will need to deterministically order attributes (so a bond's k and length have to come out in the same order, every time, and we should document what that order is, and add tests to ensure it doesn't change)

This is straightforward at a low level: since in Python 3.6+ `dict` maintains order, you can just recursively iterate through its sorted items and add things back in. I found a MWE on stackoverflow and modified it to (I think) work for SMIRNOFF data dicts. I noticed that a lot of `OrderedDict`s are used throughout these classes - is there something about them we need that is not served by the default dictionary? The friction between the two types is really low, so I'm not sure I would advocate for a change. Mostly curious if there's something I'm not considering.

Need to add some tests but it's possible that the ordering is already enforced with `to_dict()`. See [here](https://gist.github.com/mattwthompson/fdd7bf713e017f725bfb2628dbfd744e). It may still be worth it to re-order again, though.

> The method will need an option for whether to strip id and parent_id fields, since they're part of the spec, but are nonfunctional.

Generally I think it is bad practice to add options to hashes since you want it to be deterministic and rigid. I'm also not sure if it would work with the built-in `hash` function takes no keywords (see simple behavior [here](https://gist.github.com/mattwthompson/c4d94ff4deabd358f9bcaa0e294b4e44) - I don't feel good about `hash(obj)` not working and forcing users to use `obj.__hash__(*args)`, or especially them doing different things).

It looks like I chose to strip these off here, but I don't have a strong opinion on it. Can switch it back if we want to keep them in.

> We may want to include the Toolkit's version in the name of the hash (or otherwise store it in a nearby data field) in case different toolkit versions end up hashing things differently and we only catch it later.

Generally I don't find this so useful; when I hash things I'm comparing two objects in memory from the same interpreter/kernel/etc. Comparing hashes of different force fields seems like something that could be expected to be dangerous, perhaps unadvisable in general. (If we wanted to provide some safeguards for users who choose to do this anyway, maybe there is a middle ground like forbidding the behavior of custom hashes from changing between patch and minor releases.) I think, also, most of the actual calls to `__hash__` are various things that happen under the hood, where different versions of the toolkit running side-by-side can't happen. There are surely plenty of use cases I'm not considering, since I'm only going off of my experience here. 

Hash functions conventionally gobble up data and return an int. I suppose, if we wanted to, we could cast the int to str and tack on some metadata, but I would like to be really of the value added and potential pitfalls (in particular, those specific to the use cases we care about) before venturing out into non-standard territory.

- [ ] Add [tests](https://github.com/openforcefield/openforcefield/tree/master/openforcefield/tests) (needs more)


openforcefield/tests/test_forcefield.py
```diff
@@ -1545,6 +1545,50 @@ def test_lookup_parameter_handler_object(self):
...
+
+        forcefield.get_parameter_handler("Bonds").add_cosmetic_attribute("foo", 4)
+        assert "foo" in forcefield["Bonds"]._cosmetic_attribs
+
+        hash_with_cosmetic = hash(forcefield)
+
+        assert hash_with_cosmetic == hash_without_cosmetic
+
+    length = 1 * unit.angstrom
+    k = 10 * unit.kilocalorie_per_mole / unit.angstrom ** 2
```
j-wags, line 1562: Is there a need for these to be class variables? We could move them inside the method below if not.

j-wags:
>  I noticed that a lot of OrderedDicts are used throughout these classes - is there something about them we need that is not served by the default dictionary? 

Nope. They're fully legacy, when we weren't sure if we'd hang on to python 3.5

> [I]t's possible that the ordering is already enforced with to_dict(). See here. It may still be worth it to re-order again, though.

I think that aspects of the current implementation will ensure that `to_dict` will consistently order attributes for now, but we don't make any guarantee about that consistency in future development. For example, the eventual Pydantic refactor of ForceField may or may not order things consistently. I'd prefer not to require that `to_dict` will apply any ordering, so I like the explicit re-ordering logic in this PR.

> Generally I think it is bad practice to add options to hashes since you want it to be deterministic and rigid.

Yeah, good call. We can add a finer-grained `hash(*args)` method later if folks really want it.

> (regarding incorporating toolkit version in the hash) Generally I don't find this so useful; when I hash things I'm comparing two objects in memory from the same interpreter/kernel/etc... I would like to be really of the value added and potential pitfalls (in particular, those specific to the use cases we care about) before venturing out into non-standard territory.

Yeah, you've convinced me. The vast majority of hashings are using the same kernel, so inter-kernel consistency isn't needed here.

At its core, I'd like `__hash__` equality to indicate that two ForceFields encode identical chemistry/physics, independent of nonfunctional metadata. Folks who want to compare "strict" equality, including metadata and attribute ordering and all, can still access that using the string equality operator. 

So, I think the implementation here is pretty solid (though I'd also propose removing the `Author` and `Date` fields). Once you're comfortable with it, enable tests and ping me on Slack for final review :-)

openforcefield/tests/test_forcefield.py
mattwthompson: No need, just preference to avoid repeated code. Can be removed if desired (either practical downside or stylistic preference)

mattwthompson: Ah, on second reading, I can see why this looks weird. At some point I think it would be useful to do this for a bunch of these tests that regularly construct objects with the same values in order to reduce repeated code. But that's out of scope here and doing it for a single test doesn't provide the benefits of doing it everywhere.

openforcefield/tests/test_utils.py
```diff
@@ -97,3 +97,19 @@ def test_dimensionless_units():
...
+    from collections import OrderedDict
+
+    from openforcefield.typing.engines.smirnoff import ForceField
+    from openforcefield.utils.utils import sort_smirnoff_dict
+
+    forcefield = ForceField("test_forcefields/smirnoff99Frosst.offxml")
+    smirnoff_dict = forcefield._to_smirnoff_data()
+
+    # Ensure data is not created or destroyed
+    # dict.__eq__ does not check order
```
j-wags, line 112: Interesting. I didn't know that!

openforcefield/typing/engines/smirnoff/forcefield.py
```diff
@@ -1422,3 +1422,32 @@ def __getitem__(self, val):
...
+
+    def __hash__(self):
+        """Deterministically hash a ForceField object
+
+        Notable behavior:
+          * `author` and `date` are stripped from the ForceField
+          * `id` and `parent_id` are stripped from each ParameterType"""
+
+        # Completely re-constructing the force field may be overkill
+        # compared to deepcopying and modifying?
```
j-wags, line 1434: I think this is fine for now. I can't think of a strong argument either way.

j-wags (approved):
Looks great. Apologies for the delay on this. Be sure to update the release notes!

PR #690, opened 2020-08-26
Re-lint with black 20.8b0

### LINKED ISSUE #689: Linting checks failing after black udpate
`black` just bumped to a new version and has made some different decisions on how a few things should be formatted. This provides a good opportunity to decide how we want to handle this not only today but in the future, since it likely won't be the first time we encounter this issue. I can think of a few ways (thinking out loud here, not really suggesting that these are all good ideas):

1. Whenever this happens, make a PR that lints again with the changes. Advantages include maintaining "out of the box" behavior `pip/conda install black`) for all developers and CI services and keeping up-to-date with whatever features are added in new versions. The disadvantage is that is it a small maintenance burden (perhaps once or twice a year).

2. Pin linter(s) to some version(s). This has the advantage of not needing to re-lint things ever (in principle) and no direct maintenance burden. The downsides are forcing users and CI services to install old versions, which is only going to make conda environments more of a headache to deal with over time.

3. Pin linter(s) to some version(s) and periodically update. This seems like the worst of both worlds - forcing non-standard installs, making us decide when to update, and still doing some manual work.

4. Kludge together custom settings for each linter to minimize re-formatting. The upside is fewer/no linting PRs, but the downsides are numersou - probably not possible, requires a lot of work that we aren't so interested in doing, and likely destroys the value of "out of the box" linting.

I think 1. is the best and the others are all bad, to varying degrees, so I'll be opening a PR in a moment to do 1.

The timing seems unfortunate, given that we only added these checks a week or two ago. But it looks like it was just bad luck - this is the second major bump in about 16 months: https://pypi.org/project/black/#history. If we continue to add linters, this may become more like once every few months, but right now I don't expect `black` and `isort` to make significant changes with much frequency nor do I expect us to end up using 5+ linters with frequent behavior changes.

mattwthompson:
Resolves #689 


mattwthompson:
`black` changelog here: https://github.com/psf/black/blob/master/CHANGES.md#208b0

Many changes look like they're fixing actual formatting issues, mostly around docstrings

PR #693, opened 2020-08-27
Reset codecov settings

mattwthompson:
When tidying up #588, I screwed up a merge conflict or two and nuked the codecov settings and the reports have been useless since then. See [here](https://github.com/openforcefield/openforcefield/commit/f591ba2cd7f25ddc6687b7cbfd726f1a8a5deeb9#diff-128469c2fd54cf32412f10096b91495aL119-L122) for what I'm reverting here. Frustratingly, the action still acts like it's finding the file ["Found 1 reports"](https://github.com/openforcefield/openforcefield/runs/1036980150#step:15:29)


mattwthompson:
Some of the reports seem to be propagating through with actual data: https://codecov.io/gh/openforcefield/openforcefield/commits

The test failures here are in some notebooks, which seem to be genuine errors, although I have not investigated them yet.

Once this last CI run finishes and uploads reports, it should finish up issue (that I introduced)

.github/workflows/CI.yml
```diff
@@ -52,6 +52,7 @@ jobs:
       PYVER: ${{ matrix.python-version }}
       OE_LICENSE: ${{ github.workspace }}/oe_license.txt
       PACKAGE: openforcefield
+      PYTEST_ARGS: -r fE --tb=short --cov=openforcefield --cov-config=setup.cfg --cov-append --cov-report=xml
```
mattwthompson, line 55: `setup.cfg` or another config file is probably a better home for these settings, but the docs I was digging through were hard to parse and this should work well enough

j-wags (approved):
Excellent! Thanks for digging into this, Matt!

PR #700, opened 2020-09-01
Add constructor automation

mattwthompson:
- [ ] Triggered by release or manually
  - [x] Build conda installer.
  - [x] Install single-file installer.
  - [x] Assert versions correct
- [x] Test conda installer.
- [ ] Upload conda installer to hosting destination (Github release assets).
  - [x] Tarball uploaded to the action's artifacts (via `actions/upload-artifact`)
  - [ ] Tarball moved to some release's assets (to avoid a human needing to download and upload 4 files, several 100 MB each)


mattwthompson:
Current status: having trouble sourcing the new environment since the installer acts like a new install of conda and normally a user would close and restart their shell before activating the new environment

.github/workflows/installer.yml
```diff
@@ -0,0 +1,133 @@
...
+
+        pwd
+        ls
+        ls build
+
+    - name: Upload tarball as artifact
+      uses: actions/upload-artifact@v2
+      with:
+        path: toolkit-installer-constructor/build/openforcefield-*.sh
+
```
mattwthompson, line 86: Something about this isn't working, and I'm strangely having a difficult time keeping track of paths while moving around in these scripts.  The logs each pretty much look like this:
```
Run actions/upload-artifact@v2
  with:
    path: toolkit-installer-constructor/build/openforcefield-*.sh
    name: artifact
    if-no-files-found: warn
  env:
    CI_OS: ubuntu-latest
    PYVER: 3.7
    CONDA_PKGS_DIR: /home/runner/conda_pkgs_dir
##[warning]No files were found with the provided path: toolkit-installer-constructor/build/openforcefield-*.sh. No artifacts will be uploaded.
```

In recent memory, the following paths have failed:
* `/build/openforcefield-*.sh`
* `toolkit-installer-constructor/build/openforcefield-*.sh`
* `../toolkit-installer-constructor/build/openforcefield-*.sh`

The wildcard is fine: https://github.com/actions/upload-artifact#upload-using-a-wildcard-pattern

j-wags: I think you're missing one level of hierarchy. Here's what the directory structure looks like after the SFI constuctor runs:
```
jeffreywagner@JW-MBP$ cd projects/OpenForceField/toolkit-installer-constructor/
jeffreywagner@JW-MBP$ ls
README.md       cookiecutter    run.py
build           environment.yml temp_miniconda
jeffreywagner@JW-MBP$ cd build/
jeffreywagner@JW-MBP$ ls
openforcefield-0.7.1-py3.6-linux-64 openforcefield-0.7.1-py3.7-linux-64
openforcefield-0.7.1-py3.6-osx-64   openforcefield-0.7.1-py3.7-osx-64
jeffreywagner@JW-MBP$ cd openforcefield-0.7.1-py3.6-osx-64/
jeffreywagner@JW-MBP$ ls
build.sh                                   post_install.sh
construct.yaml                             test
openforcefield-0.7.1-py36-MacOSX-x86_64.sh tmp
jeffreywagner@JW-MBP$ 
```

mattwthompson: You're right - I was missing a dir with the same name as the script
```
Successfully created '/Users/runner/work/openforcefield/toolkit-installer-constructor/build/openforcefield-0.7.1-py3.6-osx-64/openforcefield-0.7.1-py36-MacOSX-x86_64.sh'.
```
```
Successfully created '/home/runner/work/openforcefield/toolkit-installer-constructor/build/openforcefield-0.7.1-py3.6-linux-64/openforcefield-0.7.1-py36-Linux-x86_64.sh'.
```

j-wags: Huzzah! 

Also, each run of the constructor builds _all platforms_, which has started failing recently (pacakges built on mac won't work on linux). 

So we'll want to make sure that mac machines only upload mac artifacts and linux machines only upload linux artifacts. That said, we probably _don't_ need separate runs for py36 and 37. So the matrix could just be {mac, linux}.

mattwthompson:
Theres _plenty_ of cleaning up that can be done, but it's possible that this actually works. I'd caution against trusting the ✅ 100%, as there have been many runs that report a passing status but include a step that reported a non-zero exit code. So that's one of the reasons why there's grossly verbose output in some places.

That being said, this should be ready for review. The only missing feature is finding a way to attach the file to the most recent releases' assets (either through an artifact or by some more direct means). I just haven't looked into it yet.

j-wags:
It looks like the artifacts aren't being uploaded:
https://github.com/openforcefield/openforcefield/runs/1120226023#step:8:10

j-wags:
Alright, I've finally managed to [get _one_ installer uploaded](https://github.com/openforcefield/openforcefield/runs/1179249830?check_suite_focus=true#step:8:11), though I still don't see it in the `artifacts` section of the Action. The relevant restrictions for the `upload-artifact` action appear to be as follows:

* No putting `..` in the upload path
* Linux build file paths begin with `/home`, Mac with `/Users`
* No putting `/*/...` at the start of the upload path, since it'll hit system directories and crash the action
* If the workflow fails, the artifacts are not made available (?)

j-wags:
Ok, I'm getting the installers getting uploaded now. I think there's some sort of (long) delay before they appear for download. They also all share the default name (`artifact`), which means they clobber each other upon upload.

j-wags:
Welp, I got pretty far, but now I'm encountering what appears to be a legit technical issue (something about libgfortran on mac). Putting this on the backburner for now.  

https://github.com/openforcefield/openforcefield/runs/1179413012?check_suite_focus=true#step:7:388

j-wags:
Oof. I get the same problem when I run this locally on my mac. Opening a new issue.

j-wags:
Removing the conda version pin and using the latest versions of conda/constructor works on my laptop, but the mac builds here are [still getting constructor version 3.0.X instead of 3.1.X](https://github.com/openforcefield/openforcefield/runs/1183746821?check_suite_focus=true#step:6:52). I may need to update the conda version from the miniconda action.

j-wags (approved):
Alright! This seems to be working. Given that there are other steps in the release process that we're actively hammering out I've disabled the `release+2 hours` automatic builds. This workflow will only run when manually triggered.

In the near future, we'll probably hit our artifact storage size limit. There's some sort of action like `purge-artifacts` that we can integrate when that happens.

PR #701, opened 2020-09-01
Set up notebooks on mybinder.org

### LINKED ISSUE #255: Update all examples to run using Binder

mattwthompson:
Resolves #255 via mybinder.org. BinderHub is an option if we want or need to self-host.

Some examples need to be double-checked (mostly those that require use of optional dependencies) but most worked roughly out of the box. Find them here: https://mybinder.org/v2/gh/openforcefield/openforcefield/binder?filepath=%2Fexamples%2F

Before merging, the link should be changed to point to `master` branch instead of this feature branch.

- [x] Disable "Join this repo's Video Chat" option that I definitely did not opt in to (see [here](https://github.com/jupyterhub/mybinder.org-deploy/issues/1562#issuecomment-693486021))
- [ ] Update README


mattwthompson:
To document the current status: something is off with the conda environment that's loaded into the notebooks. I've tried a few things but have yet to find a solution. Under the hood, it's all open-source infrastructure built into docker images, so it should be possible to build into this whatever we need. I'll get back to this eventually, but this is also free for fresh eyes to pick up.

image

j-wags:
FWIW, I don't think the "video chat" thing is a hard blocker. If we can resolve the other issues I'd be OK with merging this.

mattwthompson:
This seems to be doing well, I've gone through most examples on the link in my first comment and nothing looks out of place. I think the `postBuild` stuff I was working on was just the wrong idea from the start; I don't know what I was in the middle of troubleshooting, but removing that file was all that I needed to do here to get things working.

Once the dust settles from the release, let's restart the CI jobs here. The changes in the actual codebase, though, are pretty much completely independent of the actual software, so there should be no reason for CI to fail. Approving this PR should involve at least opening a few of the examples on mybinder and assuring they run well (I'd say the QCArchive, visualization, and SMIRNOFF simulation examples would be a bare minimum).

j-wags:
The notebooks run beautifully -- Even the visualization! This is awesome.

Before I approve, could we also add binder links to each README?

j-wags (approved):
Beautiful. Thanks, Matt!

mattwthompson:
Thanks @j-wags!

PR #705, opened 2020-09-04
Add fractional bond order interpolation for bond terms

mattwthompson:
This PR mostly steals the machinery @dotsdl built in #582 for proper torsions and applies them to bonds. Most of that PR was done generally enough that it's easy to re-use, i.e. once the parameters are loaded up, the interpolation function doesn't need to know if that `k` is eventually going to be applied to a torsion or bond

There isn't an issue here, but it's on the roadmap on Confluence.

- [x] Reliably re-compute bond orders inside of `BondHandler` as a forceful initial implementation (later: better handling of the fractional bond order assignment methods and the associated edge cases; we have notes on this)
- [x] Interpolate `length` in addition to `k`
- [x] Build up a `TestBondHandler` class to explicitly test relevant behavior
- [ ] Does adding the headers in `<Bonds>` break the SMIRNOFF spec? Existing [files](https://gist.github.com/mattwthompson/afcd29a98e3a67068baff7f1b18d49ed) seem to specify `None`
- [x] Improve docstrings for above `ParameterAttribute` subclasses
- [x] Fix serialization
- [x] Raise custom exceptions where possible
- [x] Update [Force Field Release Checklist](https://github.com/openforcefield/openforcefields/wiki/Release-Process/31af13698a5f5c1cf39500acc1a5d664cf3c7426) with instructions about `Bonds` and `ProperTorsions` headers. 
- [x] `partial_bond_orders_from_molecules` for bonds
  - [ ] Any other test to add (important, un-covered behavior)?
  - [x] Reading from XML strings
  - [x] Sanity checks against hand-calculated k values
  - [x] Clearing user-defined bond orders (in favor of whatever `.assign_fractional_bond_orders` says

Split-off tasks and other things to do later:
- [ ] Other, more intricate combinations of different bond order methods, precedence for dealing with pre-specified bond orders, etc.
- [ ] Add example notebook  (for this and torsions based on feedback by people using it; could also be split off)
- [ ] Refactor the different `_{get|process}_{indexed|mapped|indexed_mapped}_parameter_attributes()` functions (could be split off into a subsequent PR)
- [ ] Basic regression tests
  - [ ] Some way to ensure `BondHandler` isn't causing a performance regression via accidentally re-computing fractional bond orders
  - [ ] Any regressions from changes to `ProperTorsionHandler`?


openforcefield/data/test_forcefields/smirnoff99Frosst.offxml
```diff
@@ -9,7 +9,7 @@
         <!-- https://github.com/openforcefield/smirnoff99Frosst/                                         -->
         <!-- If you absolutely must refer to this file in scholarly work, indicate the file name and     --> 
         <!-- Open Force Field Toolkit version that was used                                              -->
-	<Bonds version="0.3" potential="harmonic" fractional_bondorder_method="None" fractional_bondorder_interpolation="linear">
```
mattwthompson, line 12: Not sure how to get the `ParameterAttribute` behavior to be compatible with existing headers like the old line, but also allow tacking on an XML with the new line

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2356,6 +2448,21 @@ def create_force(self, system, topology, **kwargs):
...
+        for match in bond_matches.values():
+            if match.parameter_type.k_bondorder:
+                uses_interpolation = True
+                break
+
+        # Re-computing fractional bond orders is expensive and should be skipped if not necessary
+        if uses_interpolation:
+            for top_mol in topology.topology_molecules:
+                top_mol.reference_molecule.assign_fractional_bond_orders(
+                    bond_order_model=self.fractional_bondorder_method.lower()
```
mattwthompson, line 2463: Might be better to do the `.lower()` downstream, i.e. to avoid 

```
In [4]: mol.assign_fractional_bond_orders('AM1-Wiberg')
<ipython-input-4-8eb006fac6a1> in <module>
----> 1 mol.assign_fractional_bond_orders('AM1-Wiberg')

~/software/openforcefield/openforcefield/topology/molecule.py in assign_fractional_bond_orders(self, bond_order_model, toolkit_registry, use_conformers)
   2644                 self,
   2645                 bond_order_model=bond_order_model,
-> 2646                 use_conformers=use_conformers,
   2647             )
   2648         elif isinstance(toolkit_registry, ToolkitWrapper):

~/software/openforcefield/openforcefield/utils/toolkits.py in call(self, method_name, raise_exception_types, *args, **kwargs)
   4965                     for exception_type in raise_exception_types:
   4966                         if isinstance(e, exception_type):
-> 4967                             raise e
   4968                     errors.append((toolkit, e))
   4969

~/software/openforcefield/openforcefield/utils/toolkits.py in call(self, method_name, raise_exception_types, *args, **kwargs)
   4961                 method = getattr(toolkit, method_name)
   4962                 try:
-> 4963                     return method(*args, **kwargs)
   4964                 except Exception as e:
   4965                     for exception_type in raise_exception_types:

~/software/openforcefield/openforcefield/utils/toolkits.py in assign_fractional_bond_orders(self, molecule, bond_order_model, use_conformers)
   2278         else:
   2279             raise ValueError(
-> 2280                 f"Bond order model '{bond_order_model}' is not supported by OpenEyeToolkitWrapper. "
   2281                 f"Supported models are ['am1-wiberg', 'pm3-wiberg']"
   2282             )

ValueError: Bond order model 'AM1-Wiberg' is not supported by OpenEyeToolkitWrapper. Supported models are ['am1-wiberg', 'pm3-wiberg']
```

dotsdl:
Working on this now. I owe you this review @mattwthompson. Apologies for the delay.

openforcefield/tests/test_forcefield.py
```diff
@@ -3291,21 +3295,70 @@ def test_parameterize_protein(self, toolkit_registry, registry_description):
...
 
         mol = get_molecule()
 
-        forcefield = ForceField(
-            "test_forcefields/smirnoff99Frosst.offxml", xml_ff_torsion_bo
-        )
+        forcefield = ForceField("test_forcefields/smirnoff99Frosst.offxml", xml_ff_bo)
         topology = Topology.from_molecules(mol)
 
         omm_system = forcefield.create_openmm_system(
```
dotsdl, line 3364: If we choose to make Bonds respect `partial_bond_orders_from_molecules` (I think we should), we'll want to test that pre-specified bond orders are respected by the bond parameter assignment as they are for proper torsions.

openforcefield/tests/test_forcefield.py
```diff
@@ -3291,21 +3295,70 @@ def test_parameterize_protein(self, toolkit_registry, registry_description):
...
+        (
+            "get_molecule",
+            "k_torsion_interpolated",
+            "k_bond_interpolated",
+            "central_atoms",
+        ),
         [
-            (create_ethanol, 4.953856, (1, 2)),
-            (create_reversed_ethanol, 4.953856, (7, 6)),
+            (create_ethanol, 4.953856, 42208.5401, (1, 2)),
```
dotsdl, line 3348: Are these values based on what comes out of AM1-Wiberg, or based on what's specified for those molecules in the functions above?

openforcefield/data/test_forcefields/smirnoff99Frosst.offxml
dotsdl: Ah, I see what you mean here now. This is tricky. Our `openff-1.2.1.offxml` also specifies `fractional_bondorder_method="None", so can't stack in these params with this either.

Best approach might be to built tests without these incompatible FFs for now? I know that' not a satisfying solution.

openforcefield/tests/test_forcefield.py
```diff
@@ -3291,21 +3295,70 @@ def test_parameterize_protein(self, toolkit_registry, registry_description):
...
+        """Test the fractional bond orders are used to interpolate k values as we expect
+        Until toolkit-specific differences in fractional bond order generation (for harmonic bond forces) are handled,
+        this test is pinned to using OpenEye."""
 
         mol = get_molecule()
 
-        forcefield = ForceField(
-            "test_forcefields/smirnoff99Frosst.offxml", xml_ff_torsion_bo
-        )
+        forcefield = ForceField("test_forcefields/smirnoff99Frosst.offxml", xml_ff_bo)
```
dotsdl, line 3361: I understand the reasons for the changes to the `smirnoff99Frosst.offxml` now. Not sure of a good solution. @j-wags thoughts?

dotsdl (commented):
Great work @mattwthompson! I think there are two main things to be addressed:
1. Respect of `partial_bond_orders_from_molecules` in `BondHandler`.
2. Bond length interpolation; either we don't support it and it needs to change in the spec, or we implement it as written in the spec.

I don't have a good answer for what we should do for testing with e.g. `smirnoff99Frosst.offxml` given that it explicitly specifies `fractional_bondorder_method=None`. I'm interested in @j-wags thoughts on how to proceed forward there.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2305,17 +2373,41 @@ class BondType(ParameterType):
...
-        k = ParameterAttribute(unit=unit.kilocalorie_per_mole / unit.angstrom ** 2)
+        k = ParameterAttribute(
+            default=None, unit=unit.kilocalorie_per_mole / unit.angstrom ** 2
+        )
+
+        # fractional bond order params
+        k_bondorder = MappedParameterAttribute(
+            default=None, unit=unit.kilocalorie_per_mole / unit.angstrom ** 2
+        )
+
```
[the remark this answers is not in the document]
mattwthompson: ❗ Good catch! I totally missed this, will try to implement.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2356,6 +2448,21 @@ def create_force(self, system, topology, **kwargs):
         # Add all bonds to the system.
```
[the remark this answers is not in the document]
mattwthompson: Yes, but to try and keep the initial implementation small we decided to only support this behavior of automatically re-computing everything. The next step will be to support more efficient things like intaking bond orders defines in the the input molecule(s).

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2356,6 +2448,21 @@ def create_force(self, system, topology, **kwargs):
...
+        # Determine if any parameters rely on bond order interpolation, but
+        # only if they were found as matches
+        uses_interpolation = False
+        for match in bond_matches.values():
+            if match.parameter_type.k_bondorder:
+                uses_interpolation = True
+                break
+
+        # Re-computing fractional bond orders is expensive and should be skipped if not necessary
+        if uses_interpolation:
```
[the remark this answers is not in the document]
mattwthompson: Agreed, I have added a note

openforcefield/data/test_forcefields/smirnoff99Frosst.offxml
mattwthompson: Whatever the best solution happens to be, I doubt it includes breaking the old files that have been stable for a long time. I'll set up separate files/snippets for testing purposes here, but I think a spec change and corresponding up/downconverter will be inevitable.

openforcefield/tests/test_forcefield.py
mattwthompson: These are based on AM1-Wiberg, since (as above) we are currently throwing out the bond orders on the input molecules. I've updated the docstring to clarify and, because I found myself confused about the back-of-the-envelope calculation, tried to document it here as well.

openforcefield/tests/test_forcefield.py
```diff
@@ -3497,10 +3590,16 @@ def test_fractional_bondorder_calculated_rdkit(
...
+            force
+            for force in omm_system.getForces()
+            if isinstance(force, openmm.HarmonicBondForce)
+        ][0]
+
         ret_mol = list(ret_top.reference_molecules)[0]
         bond = ret_mol.get_bond_between(*central_atoms)
 
-        # ambertools appears to yield around 1.0009 for this bond
+        # ambertools appears to yield around 0.99529355 for this bond
```
mattwthompson, line 3602: I don't know why, but all day I have been getting the new value (0.99529355) on my machine. On a Linux cluster, it's 1.0009. Relevant snippet:


```
from openforcefield.tests.test_forcefield import create_ethanol
mol = create_ethanol()
from openforcefield.utils.toolkits import *
AmberToolsToolkitWrapper().assign_fractional_bond_orders(mol)
mol.get_bond_between(1, 2).fractional_bond_order
```

openforcefield/tests/test_forcefield.py
```diff
@@ -3459,29 +3531,50 @@ def test_fractional_bondorder_superseded_by_standard_torsion(
...
+            "k_torsion_interpolated",
+            "k_bond_interpolated",
+            "bond_length_interpolated",
+            "central_atoms",
+        ),
         [
-            (create_ethanol, 4.953856, (1, 2)),
-            (create_reversed_ethanol, 4.953856, (7, 6)),
+            (create_ethanol, 4.953856, 42266.9635, 1.39991, (1, 2)),
+            (create_reversed_ethanol, 4.953856, 42266.9635, 1.39991, (7, 6)),
```
mattwthompson, line 3543: This test is failing for the `create_ethanol` case _but not_ the `create_reversed_ethanol` case. Restarting the tests, if the failure is not stochastic I will need to go down a path of "why is AmberTools producing different results with seemingly-identical inputs?"

https://github.com/openforcefield/openforcefield/runs/1201420191?check_suite_focus=true#step:12:400

mattwthompson: Raised an issue here, since I think it's a bug https://github.com/openforcefield/openforcefield/issues/736

openforcefield/data/test_forcefields/smirnoff99Frosst.offxml
j-wags: > The smirnoff99Frosst.offxml is an artifact of a particular point in time, so we should leave it sit.

So, this file isn't actually an important artifact. It's a dangerously-named test file that's been hand-modified and that I get a ton of grief about whenever I try to kill it off. I would be happy to just rename it to `test_forcefields/smirnoff_test.offxml` or something else so that we don't have to worry about people using it in production.

But in the short term, let's just work on getting this running. I'd prefer to follow the pattern of [`ProperTorsions` in this file](https://github.com/openforcefield/openforcefield/blob/9815deeb1546dfbbd1a6a9317f557e7455b10663/openforcefield/data/test_forcefields/smirnoff99Frosst.offxml#L141), and change this to something like:

```
	<Bonds version="0.3" potential="harmonic">
```

and then let the initialized handlers load the default value (`AM1-Wiberg`).

j-wags:
(blocking)

> This is tricky. Our openff-1.2.1.offxml also specifies `fractional_bondorder_method="None"`, so can't stack in these params with this either.

Oof. Big oof. 

I've been thinking about this for a while now and I'm not sure about the cleanest way forward. I'll come back to this tomorrow, but my current thoughts are that:

* We can trivially update `test_forcefields/smirnoff99Frosst.offxml` to make `Bonds` declare `fractional_bondorder_interpolation="AM1-Wiberg"` (remember, this is a test file, we can do whatever we want)
* The problem here comes when we _mix_ FFs with different bond FBO methods. We don't currently _need_ to mix a `fractional_bondorder_method="None"` FF with an `AM1-Wiberg` one. Future FF releases that implement FBO bonds can simply specify it. So maybe this isn't a real problem?
* The 0.3 `Bonds` spec commits us to keep the default value for `fractional_bondorder_method` as `None`. I know that at least one other group (OE) is implementing our spec so we can't play around with this. 
* We could to increment the `Bonds` tag version to 0.4, and make the value for `fractional_bondorder_method="AM1-Wiberg"` by default. This would be a convenience to users and developers so that they don't get blindsided by `BondHanders` getting initialized with `FBO=None` by default. But users who are FF developers would probably be able to modify a `ForceField` with like, `ForceField.get_parameter_handler('Bonds').fractional_bondorder_method='AM1-Wiberg'`
* Regardless of what we choose, we should make subsequent `openff-X.Y.Z` releases specify Bonds and ProperTorsions tags have `fractional_bondorder_method="AM1-Wiberg"`. 


Anyway, maybe this isn't that big of a deal. I'll chime back in on this once I've slept on it.

mattwthompson:
> We can trivially update `test_forcefields/smirnoff99Frosst.offxml` to make Bonds declare fractional_bondorder_interpolation="AM1-Wiberg" (remember, this is a test file, we can do whatever we want)


>  I would be happy to just rename it to `test_forcefields/smirnoff_test.offxml` or something else

Agree on both


> The 0.3 Bonds spec commits us to keep the default value for fractional_bondorder_method as None. [So ... ] we can't play around with this.

Agree that changing a default on an existing spec is a non-starter

> Regardless of what we choose, we should make subsequent openff-X.Y.Z releases specify Bonds and ProperTorsions tags have fractional_bondorder_method="AM1-Wiberg".

Agree here as well

> We could to increment the Bonds tag version to 0.4,

Are you referring to a spec update, and what would be implemented in the next force field release? Is there an issue with bumping just one handler and no others?

The blocking issue, as I understand it, is that we want people to be able to take a Parsley line force field off the shelf and appending to it something that uses bond order interpolation for bonds. So long as we'd like to avoid mixed `ParameterAttribtue`s, this would require up-converting the Parsley handler from `fractional_bondorder_method="None"` to `fractional_bondorder_method="AM1-Wiberg"`. This is dangerous territory, but I would be ok with either up-conversion or doing some intelligent and highly constrained adding of parameters (i.e. _provided that there are no terms that use interpolation_, like in an existing Parsley `BondHandler`, change `.fractional_bondorder_method`). I see this as a sort of accidental by-product of explicitly setting `fractional_bondorder_method="None"` when, with the benefit of hindsight, this probably should have just been un-specified and left to fall back to the default value. This hack is therefore a sort of cleaning up of an existing issue that we own (one that would have been hard to see without a crystal ball) and avoids fudging the spec. 
 This does feel like a grey area of best practices, so I'd like to consider any edge cases that will cause issues in the future. I can't think of any now, but that doesn't mean they don't exist.

Other cases it doesn't really make sense to consider (now)
* A force field with `<Bonds ... fractional_bondorder_method="None" ...>` but with some parameters that use interpolation. This doesn't make sense and should error out.
* Mixed bond order assignment methods within force fields (i.e. bonds and torsions mis-matching) and between force fields

<details>
<summary>Click for more un-edited, disorganized thoughts</summary>
After some thought, I think I'm in favor of bumping the spec for this handler to adjust the default value of `fractional_bondorder_method`  and also adding in the above hack. This would only work with a hack to force compatibility with the existing releases, but I don't think it's problematic. 

Let me first complicate things by expanding a few of these issues out and then see if they can be collapsed back down

* Two separate issues here are what a version of the SMIRNOFF spec considers a default value and what is actually supported by the toolkit. somebody is using one of our existing force fields that does not include interpolation parameters (which currently includes all released force fields)

* In `ProperTorsionsHandler`, the spec was updated to include a version 0.4, but the most recent [release](https://github.com/openforcefield/openforcefields/blob/8555206ec382b50f0ce3b9ccc9789ac3803b81a3/openforcefields/offxml/openff-1.2.1.offxml#L146) specifies the method that should be used, so it didn't necessarily need to be bumped from 0.3. The difference with `BondHandler` is that all releases specify `<Bonds version="0.3" potential="harmonic" fractional_bondorder_method="None"`.

The use cases I would like to see solved include:

1. Taking a Parsley line force field off the shelf and appending to it something that uses bond order interpolation for bonds. So long as we'd like to avoid mixed `ParameterAttribtue`s, this would require up-converting the Parsley handler from `fractional_bondorder_method="None"` to `fractional_bondorder_method="AM1-Wiberg"`. This is dangerous territory, but there should be no issue with this up-coversion in this case, or similar cases _provided that there are no terms that use interpolation._
</details>

j-wags:
> Are you referring to a spec update, and what would be implemented in the next force field release? Is there an issue with bumping just one handler and no others?

Not a top-level spec update. Just the Bonds section version. 

> This is dangerous territory, but I would be ok with either up-conversion or doing some intelligent and highly constrained adding of parameters

I'd thought about this, but silent upconversion is likely to be a big problem; we'd fail OFFXML --> FF ->> OFFXML round trips if we upconvert `None` to` AM1-Wiberg`. Alternately, we could silently interpret `None` as `AM1-Wiberg` in handler compatibility checks and in `create_force` , but this seems like kicking the can down the road. So I think that incrementing the Bonds section version is the right way forward here. 

> use cases I would like to see solved include: 
> Taking a Parsley line force field off the shelf and appending to it something that uses bond order interpolation for bonds.

I think it's best practice to _not_ do this, since `None` and `AM1-Wiberg` really should be incompatible. I'll update the FF release checklist to ensure future FF releases specify `Bonds fractional_bondorder_method="AM1-Wiberg"`, and since the 1.3.0 release should come soon, our latest production FF will soon be Bond-WBO-friendly.

So, I'll bump the Bonds section version. I'm noticing that our ProperTorsionHandler was never updated to explicitly allow reading section version 0.4, so I might fold that change (and tests) in here as well. 

Also, I think we DO want to support `partial_bond_orders_from_molecules`, and I can't immediately tell if that's already implemented. So I'll add that as well if it's not in use.

mattwthompson:
> I think it's best practice to not do this,

Fully agree that it's the wrong thing to do in general, and that your proposed change for 1.3.x is right move. What I'm iffy on is if this should be allowed for the existing versions (1.0.0 through 1.2.1), since the specification of `"None"` is more or less an accident. It's a fine solution to just direct users to always use the most recent version, if a developer wants to go back to do and use an older version, they can hand-modify it accordingly and accept the consequences (of which I don't think there should be any?)

Agree that bumping the `<Bonds>` section is the best choice here, just getting tied up with some of the details.

> I think we DO want to support partial_bond_orders_from_molecules, and I can't immediately tell if that's already implemented

It's not; I intentionally kicked the can down the road on that to set some bounds on this PR, so it's not implemented here. Feel free to implement it here, hopefully it can closely mirror the existing functionality in `ProperTorsionHandler`.

davidlmobley:
Looks like I agree with where this issue is going -- getting the fractional bondorder method updated in 1.3 and following.

j-wags:
Great. I'm implementing the new behaviors.

> What I'm iffy on is if this should be allowed for the existing versions (1.0.0 through 1.2.1), since the specification of "None" is more or less an accident. It's a fine solution to just direct users to always use the most recent version, if a developer wants to go back to do and use an older version, they can hand-modify it accordingly and accept the consequences (of which I don't think there should be any?)

Unfortunately, I think the correct thing to do is to require users to hand-modify openff-1.2.1 and earlier. I'm realizing the benefit of being a stickler about API stability, and probably spec adherence as an extension of that.

jchodera:
I'm *super excited* for this feature because it will allow us to finally use [this dataset](https://github.com/openforcefield/qca-dataset-submission/tree/master/submissions/2019-07-02%20VEHICLe%20optimization%20dataset)!
image

openforcefield/tests/test_parameters.py
```diff
@@ -1068,6 +1069,43 @@ def test_bondtype_to_dict(self):
...
+        length = 1.4 * unit.angstrom
+        k1 = 101 * unit.kilocalorie_per_mole / unit.angstrom ** 2
+        k2 = 202 * unit.kilocalorie_per_mole / unit.angstrom ** 2
+
+        with pytest.raises(SMIRNOFFSpecError):
+            BondHandler.BondType(
+                smirks="[*:1]-[*:2]",
+                length=length,
+            )
+
```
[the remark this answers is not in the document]
j-wags: Added :-)

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2242,6 +2477,59 @@ def get_unitless_values(attr):
...
+    @staticmethod
+    def _check_partial_bond_orders_from_molecules_duplicates(pb_mols):
+        if len(set(map(Molecule.to_smiles, pb_mols))) < len(pb_mols):
+            raise ValueError(
+                "At least two user-provided fractional bond order "
+                "molecules are isomorphic"
+            )
+
+    @staticmethod
+    def _assign_partial_bond_orders_from_molecules(topology, pbo_mols):
```
j-wags, line 2489: I've moved these two methods to the base class since they're used by both BondHandler and ProperTorsionHandler. I thought about keeping them as functions at the module level, but their use is so specifically tied to the behavior of ParameterHandlers that I think they fit best as a staticmethod of the class.

openforcefield/typing/engines/smirnoff/parameters.py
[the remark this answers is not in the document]
j-wags: Length is now interpolatable

openforcefield/typing/engines/smirnoff/parameters.py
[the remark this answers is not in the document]
j-wags: `partial_bond_orders_from_molecules` is now supported in BondHandler.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2619,58 +3011,6 @@ def check_handler_compatibility(self, other_handler):
...
         )
 
-    def check_partial_bond_orders_from_molecules_duplicates(self, pb_mols):
-        if len(set(map(Molecule.to_smiles, pb_mols))) < len(pb_mols):
-            raise ValueError(
-                "At least two user-provided fractional bond order "
-                "molecules are isomorphic"
-            )
-
-    def assign_partial_bond_orders_from_molecules(self, topology, pbo_mols):
```
j-wags, line 2629: Moved to the base class

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2523,31 +2939,6 @@ def create_force(self, system, topology, **kwargs):
 
 # =============================================================================================
 
-# TODO: This is technically a validator, not a converter, but ParameterAttribute doesn't support them yet (it'll be easy if we switch to use the attrs library).
-def _allow_only(allowed_values):
```
j-wags, line 2527: Moved this to a "utilities" section at the top of the file.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2837,59 +3180,6 @@ def _assign_fractional_bond_orders(
                 k / idivf,
             )
 
-    @staticmethod
-    def _linear_interpolate_k(k_bondorder, fractional_bond_order):
```
j-wags, line 2841: Moved to module level, renamed, and grouped with other utilities.

j-wags:
Just went through a first pass of implementing a Bonds 0.4 spec. This:
* has default `fractional_bondorder_method="AM1-Wiberg"`
* has default `potential="(k/2)*(r-length)^2"` (I've implemented a workaround to make this equivalent to `harmonic`. Generally we need to keep pushing toward symbolic potentials when we make spec changes)

I've also
* Added a bunch of tests for the new 0.4 Bonds defaults, compatibility, and just tests for tests' sake. 
* Bumped the `test_forcefields/smirnoff99Frosst` `Bonds` and `ProperTorsions` sections to version 0.4, with appropriate `potential` values
* Started using `ParameterAttribute` setter validation to ensure that only valid `potential` and `fractional_bondorder_interpolation` values can be set, which should the toolkit fail informatively _before_ it spends time trying to perform parameterization.
* Re-brands `ProperTorsionHandler._linear_interpolate_k` as a `parameters.py` module-level function`_linear_inter_or_extrapolate`, since we're now using it to interpolate for lengths, and probably more things in the future. 

To do
* This implementation will break if Bonds and ProperTorsions ask for DIFFERENT `fractional_bondorder_methods`, since it will blindly trust the values from earlier in the workflow. I could either
    * Clear the fractional bond orders after each handler completes, so the next handler can't make assumptions
    * Overload the Topology (or each Molecule) with some information about how its bond orders were assigned, so subsequent Handlers can use that information

I'll sleep on that second part and decide tomorrow.

j-wags:
Alright, I'm going with "unconditionally do not trust". I tried adding a sort of hacky "memory" of which fractional bondorder method was used on each reference molecule in the topology, but it got really complicated really fast. Ideally this will be cleanly handled by caching toolkit wrappers in the future.

j-wags (approved):
I approve this PR, pending 
* a final go/no-go/round of review from Matt
* Matt writing up releasehistory

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -82,6 +85,20 @@ class SMIRNOFFSpecError(MessageException):
     pass
 
 
+class FractionalBondOrderInterpolationMethodUnsupportedError(MessageException):
+    """
+    Exception for when an unsupported fractional bond order interpolation assignment method is called.
+    """
+
+    pass
```
mattwthompson, line 93: Not really consequential, but these `pass` lines unnecessary/for readability only. As long as a class has a docstring, it works

```python3
>>> class PointlessClass:
...
  File "<stdin>", line 2

    ^
IndentationError: expected an indented block
>>> class PointlessClass:
...     """Docstrings for pointless classes? Sure, why not"""
...
>>> PointlessClass()
<__main__.PointlessClass object at 0x7f9fb4366eb8>

File: openforcefield/typing/engines/smirnoff/parameters.py [hunk 17]
```diff
@@ -2811,18 +3176,21 @@ def _assign_fractional_bond_orders(
...
                 )
             else:
-                raise Exception(
-                    "Fractional bondorder treatment {} is not implemented.".format(
-                        self.fractional_bondorder_method
+                # TODO: This code is effectively unreachable due to the the _allow_only converter used in this
+                #       ParameterAttribute's definition, which only allows "linear". Remove?
+                raise FractionalBondOrderInterpolationMethodUnsupportedError(
+                    "Fractional bondorder interpolation method {} is not implemented.".format(
+                        self.fractional_bondorder_interpolation
```
mattwthompson, line 3193: I'm fine with this code in here as an extra safeguard against bad behavior.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2353,6 +2713,32 @@ def create_force(self, system, topology, **kwargs):
...
+        # TODO: Jeff tried implementing a way to mark how bond orders were assigned on the
+        # topology, but realized that there's already a hierarchy of assignment
+        # methods. That is, if a molecule was assigned using PBOs_from_mols, then
+        # a different fractional bondorder method SHOULD NOT attempt
+        # recalculation, whereas if the previous method was simply DIFFERENT,
+        # then the old results should be erased/cached and overwritten with the
+        # new ones. It will be easier to handle this at the level of caching
+        # the results of molecule.assign_fractional_bond_orders
+        for top_bond in topology.topology_bonds:
+            top_bond.bond.fractional_bond_order = None
```
mattwthompson, line 2727: Just to check my understanding: we're removing all of the bond orders in the toplogy, but `partial_bond_orders_from_molecules` will work since that argument takes in different objects? And there's enough deepcopying happening that we don't risk clearing `partial_bond_orders_from_molecules` objects here?

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -2333,14 +2685,22 @@ def check_handler_compatibility(self, other_handler):
...
         )
 
+        # potential="harmonic" and potential="(k/2)*(r-length)^2" should be considered identical
+        self_has_harmonic_potential = self.potential == 'harmonic' or self.potential == "(k/2)*(r-length)^2"
+        other_has_harmonic_potential = other_handler.potential == 'harmonic' or other_handler.potential == "(k/2)*(r-length)^2"
+        if not(self_has_harmonic_potential and other_has_harmonic_potential):
+            if self.potential != other_handler.potential:
+                raise IncompatibleParameterError(
+                    f"potential values are not identical. "
+                    f"(handler value: {self.potential}, incompatible value: {other_handler.potential}")
```
mattwthompson, line 2702: Worth considering in other spec discussions if `harmonic` should be deprecated at some point after a larger push to explicitly encode expressions

mattwthompson (commented):
The UI foolishly sees this as "my" pull request, so I cannot approve with the pretty green check mark, but the comment serves as a substitute. I made a few small comments, nothing blocking. Adding the changelog now, then merge at will

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Oh, silly me. I'll leave them for consistency with the widespread (mis)use currently in the codebase, but I'd welcome a separate PR to pull all instances of this out at once.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Thanks. I'm wondering if it's a bad pattern to validate things multiple times like this (since then, if we wanted to change the behavior in the future, we'd have to track down _all_ the separate instances of validation). But I'll leave this in and we can learn what this pattern leads to.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: I'd advocate for silent upconversion over deprecation, but thematically yes, I'd like the word "harmonic" to stop appearing in our FFs.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Good question. Important implementation points are:
* The Topology is deepcopy-ed right at the beginning of create_openmm_system, so we're not messing around with the original. 
* Right now, `partial_bond_order_from_molecules` ("PBO mols") is a free-standing list of molecules in kwargs. So, when we wipe the partial bond orders on the `Topology`,  PBO mols aren't affected. We do a _little_ bit of redundant work by re-assigning them for each handler that may need the info from the PBO mols (meaning more isomorphism checks), but this shouldn't be a huge deal.

j-wags:
Thanks, Matt. Merging!

PR #710, opened 2020-09-08
Remove uses of AlkEthOH in docs/examples, replace s99F with Parsley

j-wags:
- [x] Bring in changes from #703 
- [x] replace several uses of smirnoff99Frosst with Parsley (openff-1.0.0)


README.md
```diff
@@ -63,15 +63,13 @@ molecule = Molecule.from_file(sdf_file_path)
 from openforcefield.topology import Topology
 topology = Topology.from_molecules(molecule)
 
-# Load the smirnoff99Frosst SMIRNOFF force field definition
+# Load the latest OpenFF force field definition
 from openforcefield.typing.engines.smirnoff import ForceField
-forcefield = ForceField('test_forcefields/smirnoff99Frosst.offxml')
+forcefield = ForceField('openff-1.2.0.offxml')
```
mattwthompson, line 68: Updating this should probably be added to the `openforcefields` release process? A function that automatically finds the most recent one comes with some downsides (i.e. `from openforcefield.typing.engines.smirnoff.forcefield import get_available_force_fields, ForceField; sorted([ForceField(file).date for file in get_available_force_fields()])`)

j-wags: [It's already part of the checklist](https://github.com/openforcefield/openforcefields/wiki/Release-Process#7-if-this-is-a-major-release-update-ffs-in-open-force-field-toolkit-examples). We're just always in a rush so we don't do it. 

There's some degree of judgement as to which FF releases are "major" enough to warrant updates to examples. And critically, there is lots of surrounding context -- Like, a bunch of things I fixed in this update had markdown saying "this line loads smirnoff99Frosst", and the actual code loaded Parsley. Further, it required a bit of judgement to know when to use `openff-1.0.0.offxml`, `openff_unconstrained-1.0.0.offxml`, or `openff-1.2.0.offxml`. 

Not sure how to design a process to simplify/automate this. Really the examples contain more information than code, and keeping them maintained is a complex/human intensive process.

mattwthompson: Ah, well, my suggestion's already covered.

I'd not push for updating the examples more often than a "major" release line (i.e. not updated for Parsley updates, but updated for the bump to Sage). I would not trust the automation I can think of building, and I definitely don't want to do that by hand.

The-SMIRNOFF-force-field-format.md
```diff
@@ -285,7 +285,7 @@ This decoupling of how parameters are assigned for each term provides a great de
 
 Below, we describe the specification for each force field term definition using the XML representation of a SMIRNOFF force field.
 
-As an example of a complete SMIRNOFF force field specification, see the prototype [SMIRNOFF99Frosst offxml](https://github.com/openforcefield/openforcefield/blob/master/openforcefield/data/test_forcefields/smirnoff99Frosst.offxml).
+As an example of a complete SMIRNOFF force field specification, see [a recent force field in the "Parsley" line (openff-1.2.0.offxml)](https://github.com/openforcefield/openforcefields/blob/1.2.0/openforcefields/offxml/openff-1.2.0.offxml).
```
mattwthompson, line 288: I modified `test_readme_links` to point to this file and found the following links to be broken

https://open-forcefield-toolkit.readthedocs.io/en/latest/converting.html
https://github.com/pandegroup/openmm/blob/topology/wrappers/python/simtk/unit/unit_definitions.py
https://mulan.swmed.edu/group/gaff.php
http://rdkit.org/ (HTTP?)

Might be nice to add that automated test, but not blocking.

mattwthompson (approved):

The-SMIRNOFF-force-field-format.md
j-wags: Thanks for checking. Replaced like 30 `http://`s with `https://`s. Found the new GAFF link. Pointed to more up-to-date/existant examples.

PR #711, opened 2020-09-08
Fix box_vectors setter

### LINKED ISSUE #638: Topology.box_vectors setter allows dangerous usage
Box vectors is expected to be a 3x3 array. However, no error is raised if the the user provides a 1x3 list (assuming an already-defined orthonormal coordinate system)

**Current behavior**
```
from simtk import unit

mol = Molecule.from_smiles('C')
top = mol.to_topology()
top.box_vectors = [[1,0,0], [0,1,0], [0,0,1]] * unit.angstrom
top.to_openmm()
```
works fine, but 
```

mol = Molecule.from_smiles('C')
top = mol.to_topology()
top.box_vectors = [1., 1., 1.] * unit.angstrom
top.to_openmm()
```
doesn't raise an error until the `top.to_openmm()` line

```
> > Traceback (most recent call last):
>   File "reproduce.py", line 14, in <module>
>     top.to_openmm()
>   File "/Users/jeffreywagner/projects/OpenForceField/openforcefield/openforcefield/topology/topology.py", line 1905, in to_openmm
>     omm_topology.setPeriodicBoxVectors(self.box_vectors)
>   File "/Users/jeffreywagner/miniconda3/envs/off-dev/lib/python3.7/site-packages/simtk/openmm/app/topology.py", line 245, in setPeriodicBoxVectors
>     if not is_quantity(vectors[0][0]):
>   File "/Users/jeffreywagner/miniconda3/envs/off-dev/lib/python3.7/site-packages/simtk/unit/quantity.py", line 733, in __getitem__
>     assert not is_quantity(self._value[key])
> TypeError: 'float' object is not subscriptable
```

An informative error should be raised if someone tries to set `top.box_vectors` with a 1x3 array

mattwthompson:
Fixes #638 

This was annoying me today while working on some other interoperability. Previously, the setter attempted to enforce shape `(3, )` and the docstrings were ambiguous about box vectors vs. box lengths. vs. other representations of periodicity. The changes here assume we want the box vectors to be `(3, 3)` and not store box _lengths_ there.

I still hold the opinion that `Topology` should only know _if_ it is periodic, and holding the values is a task for the system object. But now's not the time to make that switch.


j-wags:
By my understanding this would be an API break, so I'd like to avoid it unless it fully blocks something we need to do. At the same time, I do agree that this is a crummy API point, and that we should only have a Topology know _whether_ it's periodic.

Would either of these work:
* Changing the documentation/adding helpful error messages explaining the proper usage?
* Parking this change until the 1.0.0 release, and leaving a path to provide this with inputs of EITHER shape 3x1 or 3x3?

I could also just be convinced to swallow the API break, if this is holding back System work.

mattwthompson:
There are two issues right now:

1. It's not clear what `box_vectors` is supposed to be, in terms of an internal representation. I've always thought of box vectors as a 2-D matrix of shape `(3, 3)`, as is the case in OpenMM. The `{to|from}_opemm` functions imply that it is 2-D, but that's not documented clearly in the property and setter.
2. The current setter _requires_ that the input be `(3, )`. That's be fine if it was a setter for `box_lengths`, but it's a setter for `box_vectors.

> Parking this change until the 1.0.0 release, and leaving a path to provide this with inputs of EITHER shape 3x1 or 3x3?

I think we can do this now; always store the vectors as 2-D, but, if a 1-D array is passed to the setter, convert that to the corresponding 2-D matrix with some warning

```python3
from simtk import unit
import numpy as np
box_lengths = [4, 5, 6] * unit.nanometer
print("lengths: ", box_lengths)
box_vectors = box_lengths * np.eye(3)
print("vectors: ", box_vectors)
```

out:

```
lengths:  [4, 5, 6] nm
vectors:  [[4. 0. 0.]
 [0. 5. 0.]
 [0. 0. 6.]] nm
```

This would enable compatibility with current workflows that pass a 1-D array to the setter, but maintains functionality with the expected shape (2-D) elsewhere. Your snippet in the linked issue shows that setting it with a 1-D prohibits conversion out to OpenMM; on that basis I'd call this a bugfix, not an API break, since it would make a dangerous call no longer dangerous. This also enables the 1.0.0 behavior to be decided on elsewhere (i.e. adding a `is_periodic` property and deprecating `box_vectors`)

j-wags (approved):
Apologies for my earlier confusion. You're right -- This squarely solves the problem in the issue. Thanks for implementing the reverse-compatible logic as well.

This looks good to me. And I agree that we could replace this with `is_periodic` in the 1.0.0 release. Let's talk about the jurisdiction of each object at our next System design meeting :-)

PR #713, opened 2020-09-14
Access miniconda action from conda-incubator

mattwthompson:
The ownership of the `setup-miniconda` action recently changed from its main author to the `conda-incubator` organization, so GitHub Actions runners fail to access it. This change should point them to the right URL.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -57,7 +57,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: goanpeca/setup-miniconda@v1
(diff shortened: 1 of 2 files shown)


mattwthompson:
Stub issue for this: https://github.com/conda-incubator/setup-miniconda/issues/70

j-wags (approved):
Looks good. Build time jumped a bit for some reason, though it may have just been random. Apologies for the delay and thanks for tackling this!

mattwthompson:
If build times become an issue down the line, we can check back on on the status of using `mamba` here. The last I heard, there were some minor differences in how it solved environments, which we'd want to avoid in order to best reproduce user experience.

PR #718, opened 2020-09-18
Add Molecule.rings, Molecule.n_rings, Bond/Atom.is_in_ring

### LINKED ISSUE #679: Add `atom.is_in_ring` and `bond.is_in_ring`
We’ll need an RDKit-backed or toolkit-agnostic `atom/bond.is_in_ring` for fragmenter migration

* Maybe SMARTS syntax using rN can do this? Or does that only check for specific values of N? 
* Maybe there’s also some method call involving SSSR that can do this?
* Maybe NetworkX can do this? find_cycles or something?
    * Are there some cycle sizes that are ignored or treated specially by OETK or fragmenter? What about 3- or 4-membered rings? We should aim to replicate this behavior precisely in the first iterations.
* Do we anticipate needing any special behavior for macrocycles? Or is that outside the scope of this work?

mattwthompson:
Resolves #679 

I got most of the way toward implementing this properly before running into a chemical perception snafu (as one does). See comments for a couple questions that need to be resolved.

- [x] Add a test for a highly conjugated ring system that `nx.cycle_basis` would fail on (see [here](https://github.com/mosdef-hub/foyer/issues/170#issuecomment-394820633))


openforcefield/topology/molecule.py
```diff
@@ -429,6 +435,24 @@ def is_bonded_to(self, atom2):
...
         return False
 
+    @property
+    def is_in_ring(self):
+        """
+        Return whether or not this atom is in a ring(s) (of any size)
+
+        """
+        if self._molecule is None:
+            raise NotAttachedToMoleculeError(
```
mattwthompson, line 445: There are a few similar places that this exception could go (currently `ValueError`s)

openforcefield/topology/molecule.py
```diff
@@ -1381,6 +1405,23 @@ def molecule_bond_index(self):
...
+                "This Bond does not belong to a Molecule object"
+            )
+
+        # Note that the definition here needs to be different than in Atom.is_in_ring,
+        # since a bond's two atoms may be in a ring, but _not the same_ ring;
+        # A MWE is biphenyl (see unit test)
+        # bonds_in_any_ring = [val[0] for val in self._molecule.chemical_environment_matches('[*;R*:1]~[*;R*:2]')]
+        return self.atom1.is_in_ring and self.atom2.is_in_ring
```
mattwthompson, line 1423: A better approach (if this is something that chemical perception can do well) may be to just find all of the rings in a structure, listed them separately ring by ring, and iterate through them one at a time. This would get around the issue of a bond between atoms in different rings, but not in a ring itself.

jthorton: I think this is something that networkx might be able to do well and should be easy to implement as we already have the `to_networkx` method. I also think it might be more useful to know which ring an atom or bond is in rather than just if it is in a ring especially in the case of fragmenter.

mattwthompson: It's not 100% functional yet but I've gone down this path, which includes storing rings as a nested list in `Molecule._rings`. It's not really stress-tested nor is the performance evaluated, but it worked pretty well for some simple cases, like catching the non-ring bond in biphenyl. Thanks for the suggestion, and let me know if this helps your tools or if there are any useful test cases we should keep in mind here!

maxentile: Nice!

A possibility for a test here might be to compare the networkx cycle basis with RDKit RingInfo AtomRings  https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.html#rdkit.Chem.rdchem.RingInfo

On the 1156 molecules in [the AlkEthOH rings set](https://github.com/openforcefield/open-forcefield-data/blob/master/Model-Systems/AlkEthOH_distrib/AlkEthOH_rings.smi), these two methods agreed with each other.

```python
def _canonicalize(rings):
    """remove atom-ordering and ring-ordering, for the purpose
    of checking whether two collections of rings describe the same collections
    of atoms
    
    >>> rings = [[5, 4, 3, 2, 1, 0], [8, 7, 6, 4, 3]]
    >>> _canonicalize(rings)
    [(0, 1, 2, 3, 4, 5), (3, 4, 6, 7, 8)]
    """
    return sorted([tuple(sorted(ring)) for ring in rings])

def check_nx_cycle_basis_agrees_with_rdkit_ringinfo(offmol):
    # enumerate rings using networkx
    nx_rings = nx.cycle_basis(offmol.to_networkx())
    
    # enumerate rings using rdkit
    rdmol = offmol.to_rdkit()
    ring_info = rdmol.GetRingInfo()
    rd_rings = ring_info.AtomRings()
    
    # check if equivalent up to ordering
    return _canonicalize(nx_rings) == _canonicalize(rd_rings)
```

mattwthompson: Thanks for the write-up! That's really encouraging that it actually works on a relevant test set (more than the first three molecules to pop in my head)

One edge case I have dealt with in the past that worries me here is [_highly_ conjugated ring networks](https://github.com/mosdef-hub/foyer/issues/170#issuecomment-394779512). I used a SMILES drawer to make a not-very-physical slab of seven rings attached together (`C1C2C(CCC1)CC5C3C2CCC7C3C4C(CCC6C4C5CCC6)CC7`), and can get it to fail

```
In [9]: rdinfo.AtomRings()
Out[9]:
((0, 1, 2, 3, 4, 5),
 (6, 7, 8, 9, 1, 2),
 (10, 11, 12, 13, 8, 9),
 (16, 17, 18, 19, 14, 15),
 (21, 22, 23, 18, 19, 20),
 (24, 25, 12, 13, 14, 15),
 (7, 8, 13, 14, 19, 20))

In [10]: mol.rings
Out[10]:
[[9, 8, 7, 6, 2, 1],
 [12, 25, 24, 15, 14, 13],
 [19, 18, 17, 16, 15, 14],
 [20, 21, 22, 23, 18, 17, 16, 15, 14, 13, 8, 7],
 [20, 19, 14, 13, 8, 7],
 [9, 10, 11, 12, 13, 8],
 [5, 4, 3, 2, 1, 0]]
```

Networkx's [criteria](https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.algorithms.cycles.cycle_basis.html) for the solution is ever so slightly different than how we would define rings, and there is a [different](https://github.com/mosdef-hub/foyer/blob/12e9e6d567be321a9d977c335c9f6422af04887b/foyer/smarts_graph.py#L244) implementation of the same functionality out there to deal with that. It's really unlikely to be an error that causes problems in the data we're working with now, but there's no way to predict when this discrepancy might pop up, so I had planned to switch it over to the longer approach.

However, this is giving me pause to consider which path is the better solution here, given that RDKit seems to have done the heavy lifting already.

mattwthompson: Another data point that would be useful here is whether or not we want these objects to generally be aware of their own rings (something that's implied by the `self._rings` caching), i.e. if that functionality is useful outside of the scope of the feature request here. If so, there could be a `ToolkitWrapper` function that implements this (presumably there is an OpenEye equivalent to this RDKit function). But if not, we don't need to go about implementing several different things on the backend for this purpose.

jthorton: It looks like the `nx.minimum_cycle_basis` function would help with this specific molecule and still works in general as it passes the test above from @maxentile. I think having the rings property on the molecule is really useful and I can imagine this being used in various areas outside of fragmenter. For example, this method will probably be used to check if dihedrals are in rings when submitting torsiondrives to qcarchive.

mattwthompson: Thanks for finding that - it seems to work perfectly and  I've implemented that and `Molecule.rings`/`Molecule.n_rings` with some basic tests

It's out of scope here, but it might be interesting to compare networkx, RDKit, and OpenEye methods for a larger data set since comparing methods wouldn't require labeling numbers of rings by hand

maxentile: >However, this is giving me pause to consider which path is the better solution here, given that RDKit seems to have done the heavy lifting already.

I did a quick random check to see if there are molecules in the [Enamine Diverse REAL drug-like set](https://enamine.net/library-synthesis/real-compounds/real-compound-libraries) where RDKit ring info disagrees with nx.minimum_cycle_basis. Of ~3000 random molecules checked, ~400 of them appear to have a discrepancy between nx.minimum_cycle_basis and RDKit ring info.

I haven't inspected these yet to say anything about how relevant the chemistry is, or whether RDKit vs. Networkx is giving the desired result, but in case this is useful for testing, here are the smiles encountered so far that cause discrepancies : https://gist.github.com/maxentile/64558c9e64b3394191cdd92594de760b

jthorton: Thanks for that I pulled out some molecules randomly and they seem to have a group like the molecule here and the networkx method seems to just miss one of the possible rings in the molecule. 
Screenshot 2020-09-22 at 17 49 37

mattwthompson: Looks like a common feature in that data set is that RDKit finds more rings:

```python3
In [11]: def n_rdkit(mol):
    ...:     rdmol = mol.to_rdkit()
    ...:     info = rdmol.GetRingInfo()
    ...:     rings = info.AtomRings()
    ...:     return len(rings)
    ...:

In [12]: for mol in mols:  # mols is from Fass's gist above
    ...:     assert n_rdkit(mol) > mol.n_rings  # using the networkx implementation
    ...:     # no exceptions raised

```

Pulling out a random molecule, I see a similar moiety to what @JoshHorton points out (not sure what its proper name is):

image


so on this basis I think we should trust RDKit's implementation and avoid networkx's out-of-the-box solutions.

mattwthompson: Well, this has certainly led down a deep rabbit hole, one that I didn't intend to go but should still be worthwhile for the toolkit's functionality.

* The world of graph theory does not seem to define any nouns that are close enough to what a chemist thinks of as a ring, so networkx's convenient functions will unfortunately not work. See above for bicyclics as a case that demonstrates the issue, norbornane being a simple starting point.
* RDKit seems to have done this already, see @maxentile's code above. An RDKit-based solution, I think should be implemented as a `ToolkitWrapper` function, which runs into an issue below.
* Foyer's solution to the problem (which takes only a networkx graph [1] and a setting for the maximum size of a ring it searches for) pretty closely matches RDKit's, but fails on some edge cases with multi-cyclics [2].
* OpenEye, for that matter, seems to [intentionally](https://docs.eyesopen.com/toolkits/python/oechemtk/ring.html#smallest-set-of-smallest-rings-sssr-considered-harmful) not have this functionality. It looks like it can be backed out with some hacks, but this is not a path I'm eager to go down.

I think the two solutions here are
1. Work toward patching the function that foyer uses to behave better in some test dataset of a few thousand molecules.
2. Write a `ToolkitWrapper` function that uses RDKit, and RDKit alone, to determine the rings in a molecule. This would be an issue if, for some reason, a user did not or could not use RDKit.



--
1. As an aside, I think it's really cool that much of the community uses common data structures like networkx graphs and NumPy arrays. I was able to conda install a package, import a function from it and just send it `mol.to_networkx()` out of the box, and it worked.

2. With the SMILES `'[H]C1(C2(C(C1(C(C2([H])[H])([H])SC3(C(OC3([H])[H])([H])[H])[H])[H])([H])[H])[H])[H]'` foyer finds a 10-membered ring by doubling back on itself while traversing the bicyclic group on the left. This may be patchable, I'm not sure without putting my thinking cap on. Moreso, this may be a case of the nasty complexity hinted at in the OpenEye docs linked above.

image

j-wags:
@jthorton This seems pretty detailed already -- Do you mind if I assign you as the final reviewer for this PR?

openforcefield/tests/test_molecule.py
```diff
@@ -2764,6 +2764,43 @@ def test_get_bond_between(self):
         with pytest.raises(NotBondedError):
             mol.get_bond_between(hydrogens[0], hydrogens[1])
 
+    @pytest.mark.parametrize(
+        ("smiles", "n_rings"),
+        [
+            ("CCO", 0),
+            ("c1ccccc1", 1),
+            ("C1CC2CCC1C2", 2),  # This should probably be 3?
```
jthorton, line 2772: I agree strange that RDKit does not pick this up

openforcefield/tests/test_molecule.py
```diff
@@ -2764,6 +2764,43 @@ def test_get_bond_between(self):
...
+    )
+    @requires_rdkit
+    def test_molecule_rings(self, smiles, n_rings):
+        """Test the Molecule.rings property"""
+        assert (
+            n_rings == Molecule.from_smiles(smiles, allow_undefined_stereo=True).n_rings
+        )
+
+    @requires_rdkit
+    def test_is_in_ring(self):
```
jthorton, line 2787: This should probably be done using parametrize as well so we can easily work out which one is failing.

openforcefield/topology/molecule.py
```diff
@@ -4669,6 +4711,34 @@ def get_bond_between(self, i, j):
 
         raise NotBondedError("No bond between atom {} and {}".format(i, j))
 
+    @property
+    def rings(self):
+        if self._rings is None:
+            self._get_rings()
+        return self._rings
+
+    def _get_rings(self, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY):
```
jthorton, line 4720: Should we explicitly use the RDKit toolkit wrapper here to do the call as the method is only coded for that backend? So it would be similar to the `to_openeye` or `to_rdkit` methods, as with this current setup, this would also avoid the case of raising an error that the function could not be called if a user removes the RDKit toolkit wrapper from the `GLOBAL_TOOLKIT_REGISTRY `.

jthorton:
@mattwthompson Thanks for taking this one! it looks good to go I found a few points but only the toolkit call one is a potential issue.

openforcefield/topology/molecule.py
mattwthompson: Not sure; I like the idea of just letting `ToolkitWrapper.call` handle it, since maybe it will be implemented in another wrapper. But that's probably not on the horizon, and the error message doesn't really describe to the user that it's only available with RDKit installed:

```python3
In [1]: from openforcefield.topology import Molecule, Topology


In [2]: from openforcefield.utils.toolkits import *

In [3]: GLOBAL_TOOLKIT_REGISTRY.deregister_toolkit(RDKitToolkitWrapper)

In [4]: Molecule.from_smiles('c1ccc2ccccc2c1').n_rings
<ipython-input-4-33b5c9295de5> in <module>
----> 1 Molecule.from_smiles('c1ccc2ccccc2c1').n_rings

~/software/openforcefield/openforcefield/topology/molecule.py in n_rings(self)
   3272     def n_rings(self):
   3273         """Return the number of rings found in the Molecule"""
-> 3274         return len(self.rings)
   3275
   3276     @property

~/software/openforcefield/openforcefield/topology/molecule.py in rings(self)
   4715     def rings(self):
   4716         if self._rings is None:
-> 4717             self._get_rings()
   4718         return self._rings
   4719

~/software/openforcefield/openforcefield/topology/molecule.py in _get_rings(self, toolkit_registry)
   4737
   4738         """
-> 4739         rings = toolkit_registry.call("find_rings", self)
   4740         self._rings = rings
   4741

~/software/openforcefield/openforcefield/utils/toolkits.py in call(self, method_name, raise_exception_types, *args, **kwargs)
   5023         for toolkit, error in errors:
   5024             msg += " {} {} : {}\n".format(toolkit, type(error), error)
-> 5025         raise ValueError(msg)
   5026
   5027     def __repr__(self):

ValueError: No registered toolkits can provide the capability "find_rings" for args "(Molecule with name '' and SMILES '[H]c1c(c(c2c(c(c(c(c2c1[H])[H])[H])[H])[H])[H])[H]',)" and kwargs "{}"
Available toolkits are: [ToolkitWrapper around OpenEye Toolkit version 2020.1.0, ToolkitWrapper around AmberTools version 20.0, ToolkitWrapper around Built-in Toolkit version None]


In [5]: GLOBAL_TOOLKIT_REGISTRY.register_toolkit(RDKitToolkitWrapper)

In [6]: Molecule.from_smiles('c1ccc2ccccc2c1').n_rings
Out[6]: 2
```

mattwthompson: Okay, after the next commit, these functions require RDKit. Here's what it looks like without (probably shouldn't be a `LicenseError`)

```
In [3]: mol.n_rings
LicenseError                              Traceback (most recent call last)
<ipython-input-3-3710dfea87bf> in <module>
----> 1 mol.n_rings

~/software/openforcefield/openforcefield/topology/molecule.py in n_rings(self)
   3277
   3278         """
-> 3279         return len(self.rings)
   3280
   3281     @property

~/software/openforcefield/openforcefield/topology/molecule.py in rings(self)
   4725         """
   4726         if self._rings is None:
-> 4727             self._get_rings()
   4728         return self._rings
   4729

~/software/openforcefield/openforcefield/utils/toolkits.py in wrapped_function(*args, **kwargs)
    206                         cls._toolkit_name
    207                     )
--> 208                     raise LicenseError(msg)
    209                 value = func(*args, **kwargs)
    210                 return value

LicenseError: This function requires the The RDKit toolkit
```

openforcefield/tests/test_molecule.py
mattwthompson: I think some of the discussion in this thread can be summarizes as follows:

* This method is likely to find all rings for most structures, but will miss some rings when multi-cyclic moieties like in norbornane are present
* This method guarantees (?) that a bond or atom will know _if_ it is in a ring, but possibly not _all_ rings it is in. 

Does this work for your use cases, for now?

mattwthompson:
Thanks @jthorton!

openforcefield/topology/molecule.py
jthorton: Thanks for changing this I think this error is much better than the previous output and agree maybe the error type should change but at least the message is very clear.

mattwthompson:
Looks like CI passed - think this is good to go @jthorton?

jthorton:
@mattwthompson This looks ready to go, thanks again.

mattwthompson:
Thanks again for all the feedback @jthorton and @maxentile! This was a lot of fun to work on

PR #722, opened 2020-09-24
Run slow tests on nightly CI and deployment tests

mattwthompson:
Minor oversight from before

One test is failing locally (`pytest -v --runslow openforcefield/tests/test_molecule.py::TestMolecule::test_compute_partial_charges`), but my environment probably has other stuff floating around in it 


j-wags:
I'm seeing wrong-exception failures when parsing invalid XML strings, but only in linux builds with OETK. But I don't see any recent updates to xmltodict. Strange. Kicking builds again.

j-wags:
Update: This is still happening. Let's see if it trips on the nightly builds.

mattwthompson:
Yeah, that test result and log made me context switch. First impression is that it wouldn't be a quick fix.

mattwthompson:
FWIW, this is what I got locally last night and twice this morning (`pytest -v openforcefield/tests --runslow`)

```
=============================================== short test summary info ===============================================
FAILED openforcefield/tests/test_molecule.py::TestMolecule::test_compute_partial_charges - AttributeError: 'Molecule...
=========== 1 failed, 16200 passed, 9543 skipped, 80 xfailed, 307 xpassed, 16 warnings in 955.08s (0:15:55) ===========
[10:11:20] ERROR:
```

.github/workflows/CI.yml
```diff
@@ -141,10 +140,12 @@ jobs:
...
           PYTEST_ARGS+=" --ignore=openforcefield/tests/test_examples.py"
           PYTEST_ARGS+=" --ignore=openforcefield/tests/test_links.py"
+          # if [[ "$GITHUB_EVENT_NAME" == "schedule" ]]; then
           if [[ "$RDKIT" == true && "$OPENEYE" == true ]]; then
-            PYTEST_ARGS+="--doctest-modules"
+            # Switch back to cron-only before merging
+            PYTEST_ARGS+=" --runslow"
           fi
-          pytest $PYTEST_ARGS $PACKAGE/
+          pytest $PYTEST_ARGS openforcefield/tests/
```
mattwthompson, line 148: There are two changes here, which get mushed together in the history

1. Drop `--doctest-mocules` for now, as it's buggy and/or giving us problems that I'm pushing out of the scope of this PR.
2. Adding `--runslow` to the cron job, although right now it's triggered in the TRUE/TRUE case so it runs on commits.

mattwthompson:
Opening this up for review - it's mostly what it says on the tin, but at least one test needed to be updated, probably because it was hidden by this mark while changes happened elsewhere. I'm getting a single cryptic error when the toolkit tries to parametrize iodide, which seems intentional but is reported as an error. It could be a leaky test, I'm locally getting the same as CI reports.

Some things that might be worth checking
  * Should we be running more tests (i.e. more molecules, more combinations of molecules and charge methods, etc.)?
  * Any other silent failures/silent "should run but doesn't" ?
  * Any other deprecation warnings that aren't intentional and aren't the responsibility of an upstream?

openforcefield/tests/test_molecule.py
```diff
@@ -2694,67 +2694,70 @@ def test_chemical_environment_matches_RDKit(self):
...
+        assert (abs(initial_charges / unit.elementary_charge) > 0.01).any()
+        # Check total charge
+        charges_sum_unitless = initial_charges.sum() / unit.elementary_charge
+        total_charge_unitless = molecule.total_charge / unit.elementary_charge
+        if abs(charges_sum_unitless - total_charge_unitless) > 0.0001:
+            print(
+                "molecule {}    charge_sum {}     molecule.total_charge {}".format(
+                    molecule.name, charges_sum_unitless, total_charge_unitless
+                )
+            )
```
j-wags, line 2751: (not blocking) Should this `print` be commented out? Or will it only trigger if something's wrong?

openforcefield/tests/test_molecule.py
```diff
@@ -2694,67 +2694,70 @@ def test_chemical_environment_matches_RDKit(self):
         )  # this is the wrong stereochemistry, so there shouldn't be any matches
 
     @pytest.mark.slow
-    def test_compute_partial_charges(self):
+    @pytest.mark.parametrize(
```
j-wags, line 2697: Nice refactor.

j-wags (approved):
LGTM! Be sure to remove the cron job. No releasehistory updates needed!

j-wags:
I see what's going on with the iodide thing. It's annoying and will trigger when someone tries to parameterize a system where molecules fail in the `AM1BCCHandler` but get charges assigned using the `ChargeIncrementModelHandler`.

We could make the warning disappear by removing [this line](https://github.com/openforcefield/openforcefield/blob/988c6d627aa27a1cc1b666b68fde345612716175/openforcefield/typing/engines/smirnoff/parameters.py#L4176), but it deserves a bit more thought, so I'm fine to leave it for now.

openforcefield/tests/test_molecule.py
mattwthompson: I'm pretty sure `pytest` only prints prints if a test failed but I'll switch it back to be safe.

mattwthompson:
Test failures are due to something happening with a QCArchive server

Issue #721, opened 2020-09-24
Allow some ChargeIncrementHandler attributes to be automatically set


mattwthompson:
**Is your feature request related to a problem? Please describe.**
@SimonBoothroyd says:
> Would you consider extending the charge increment handler so as to allow certain attributes to be automatically set? Namely so I can just set `charge_increment1` to the desired value and specify that `charge_increment2` should automatically be chosen to be negative that value without have to write `charge_increment2=-X` in the file?
This would simplify optimisations a bit (currently a FB eval attribute needs to be used) and for sure make computing derivatives w.r.t. these parameters easier
Otherwise this behaviour needs to be hardcoded into places like evaluator, and I'm not sure have well the current behaviour plays with autodiffing

**Describe the solution you'd like**
When N atoms are tagged, but N-1 charge increments are specified, the atom without a specified charge increments gets a charge determined by adding up all other charge increments and multiplying by -1. So, a parameter like 

```XML
<ChargeIncrement smirks="[#6X4:1]-[#8:2]" charge1increment="-0.0718" charge2increment="+0.0718"/>`
```
could be equivalently be expressed as
```XML
<ChargeIncrement smirks="[#6X4:1]-[#8:2]" charge1increment="-0.0718"/>
```
while only exposing one value to optimization as `charge2increment` is automatically set. If N-2 are specified, raise an exception (under discussion, see below). 

**Describe alternatives you've considered**
* An approach like above but without the N-1 restriction should be considered. @j-wags summarizes this proposed implementation: 

> Hmm, so Simon's proposing, for a `ChargeIncrementType` with N tagged atoms, allowing the specification of between 1 and N `chargeincrementX` values. Then every atom with an explcitly-specified `chargeincrementX` value is guaranteed to get exactly the specified increment, and all the others get an equal share of the offset required to make the whole thing net-neutral?

which might be useful for ACC fitting. A characteristic of this approach is that the remaining charge is spread out **evenly** among the other atoms, which should be made clear to the user in documentation, and would also not error out in the N-2 case mentioned above. 

[...]

PR #726, opened 2020-09-25
Automatically set missing charge increment

mattwthompson:
First pass at implementing #721, leaving out some important tests and behavior.

- [x] Run tests from XML file(s)
- [x] Update CIMH SMIRNOFF spec version to 0.4


mattwthompson:
This seems like the type of behavior that's really difficult to communicate with only an authoritative paragraph - in other words, I didn't really understand the feature request without the relevant examples.

openforcefield/tests/test_parameters.py
```diff
@@ -1691,14 +1691,27 @@ def test_charge_increment_type_wrong_num_increments(self):
...
+
+        explicit = ChargeIncrementModelHandler.ChargeIncrementType(
+            smirks="[*:1]-[*:2]",
+            charge_increment=[
+                0.1 * unit.elementary_charge,
+                -0.1 * unit.elementary_charge,
+            ],
+        )
+
+        assert inferred.to_dict() == explicit.to_dict()
```
j-wags, line 1714: I'm removing the equality comparison at the end of this test, since I think that a chargeincrement with one unspecified value shouldn't automatically add that value during serialization, or make that value accessible before parameterization.

j-wags:
This should be ready for review when you have time, @mattwthompson. I know you can't technically review your own PR, but if you could peruse the code and leave a approving/change-requesting comment, I can follow up with an actual approval as appropriate.

openforcefield/tests/test_parameters.py
```diff
@@ -2050,15 +2050,25 @@ def test_charge_increment_type_wrong_num_increments(self):
...
-            ci_type = ChargeIncrementModelHandler.ChargeIncrementType(
-                smirks="[#6:1]-[#7:2]-[#6]",
-                charge_increment1=0.05 * unit.elementary_charge,
-            )
+        ci_type = ChargeIncrementModelHandler.ChargeIncrementType(
+            smirks="[#6:1]-[#7:2]-[#6]",
+            charge_increment1=0.05 * unit.elementary_charge,
+        )
 
+    def test_charge_increment_one_ci_missing(self):
```
mattwthompson, line 2058: I agree with your reasoning behind making these models different _pre_-parametrization, but I would like to see a test that ensures their behavior is identical or sufficiently similar. My reading of `test_charge_increment_model_invalid_number_of_cis` is that the two cases are tested only to the extend that they don't error out. `test_charge_increment_model_one_less_ci_than_tagged_atom` probably covers enough behavior to practically safeguard ourselves, but it would be nice to have a test that runs these two `ChargeIncrementType`s through parametrizationa nd ensures they produced the same result.

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -4219,11 +4219,25 @@ def __init__(self, **kwargs):
...
+                    f"ChargeIncrement {self} was initialized with an invalid combination "
+                    f"of tagged atoms and charge increments"
                 )
+            #elif diff == 1:
+            #    # Print a warning/message to user?
+            #    # TODO: Explicitly track un-set charge increments; this approach
+            #    # assumes and necessitates that the missing value is the last value
+            #    import numpy as np
+            #
+            #    self.charge_increment.append(-1 * np.sum(self.charge_increment))
```
mattwthompson, line 4240: I'm fine with not warning the user in this case, and therefore would consider this TODO and comment block resolved

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -4353,9 +4368,24 @@ def create_force(self, system, topology, **kwargs):
...
+                # tagged atom offsets the charge of the others to make the chargeincrement net-neutral
+                if len(charge_increments) - len(atom_indices) == -1:
+                    charge_increment_sum = 0. * unit.elementary_charge
+                    for ci in charge_increments:
+                        charge_increment_sum += ci
+                    charge_increments.append(-charge_increment_sum)
+                elif len(charge_increments) - len(atom_indices) == 0:
+                    pass
+                else:
+                    raise SMIRNOFFSpecError(f"Trying to apply chargeincrements {charge_increment_match.parameter_type} "
```
mattwthompson, line 4383: This exception _and_ the one at L4231 are both raised by tests, according to codecov. I found this counter-intuitive at first, but now I see in one of the tests that you modified a handler after construction achieve this broken state. Nice!

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -4353,9 +4368,24 @@ def create_force(self, system, topology, **kwargs):
                 atom_indices = (
                     charge_increment_match.environment_match.topology_atom_indices
                 )
-                charge_increment = charge_increment_match.parameter_type
+                charge_increments = copy.deepcopy(charge_increment_match.parameter_type.charge_increment)
+
+                # If we've been provided with one less charge increment value than tagged atoms, assume the last
+                # tagged atom offsets the charge of the others to make the chargeincrement net-neutral
+                if len(charge_increments) - len(atom_indices) == -1:
```
mattwthompson, line 4375: This does the check backwards from `diff = n_tags - n_increments` above; I don't care which is used but they should be identical for readability

mattwthompson (commented):
I found a few quibbles, nothing blocking, so consider this comment "approval." Updated and at your discretion! To 0.8.0 we march!

The-SMIRNOFF-force-field-format.md
```diff
@@ -246,15 +249,26 @@ The sum of formal charges for the molecule or fragment will be used to determine
...
+Starting in the 0.4 version of this section, a `ChargeIncrement` may be specified with one less `charge_increment` value than it has tagged atoms.
+The missing `charge_increment` value must be that of the highest tagged atom index. 
+This missing `charge_increment` will be calculated to offset the sum of the other `charge_increment`s in the same `ChargeIncrement` parameter to achieve a net value of 0.
+This allows `ChargeIncrement` parameters to be defined similar to bond charge corrections. 
 
 Note that atoms for which library charges have already been applied are excluded from charging via `<ChargeIncrementModel>`.
 
 Future additions will provide options for intelligently fragmenting large molecules and biopolymers, as well as a `capping` attribute to specify how fragments with dangling bonds are to be capped to allow these groups to be charged.
 
+| ChargeIncrementModel section tag version | Tag attributes and default values                                  | Required parameter attributes                                                                   | Optional parameter attributes |
```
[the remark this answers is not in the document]
j-wags: Interestingly, we set a precedent of doing this without ` marks in other sections, so I'm going to leave this as-is for consistency.

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Yeah, the other exception/test is a _bit_ forced, but I had already gone through the work of making a new exception, and I think it set the record for exception name length, so I opted to keep it.

openforcefield/tests/test_parameters.py
j-wags: Good call. Added in bcc27be

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: 👍  0eb9380

openforcefield/typing/engines/smirnoff/parameters.py
j-wags: Good idea. Changed in 9cc29a5.

j-wags (approved):
Looks great!

PR #729, opened 2020-09-29
Added conda env troubleshooting item to FAQ

dotsdl:


FAQ.md
```diff
@@ -46,3 +46,13 @@ This means we recommend one of the following or equivalent:
 - IUPAC names for the components of your system
 
 Essentially, anything which provides the full identity of what you want to simulate (including stereochemistry) should work, though it may require more or less work to get it into an acceptable format.
+
+## My conda installation of the toolkit doesn't appear to work. What should I try next?
+
+We recommend that you install the toolkit in a fresh conda environment, explicitly passing the channels to be used, in-order:
+
+    conda install -n <my_new_env> -c conda-forge -c omnia openforcefield
```
mattwthompson, line 54: Maybe before this, a line explicitly for creating the env? I worry a user may think they're installing things into a new env while still in the existing env

j-wags: Hm, is there any reason not to use `conda create` here? I haven't actually seen `conda install` used this way before, though I may just not know best practices.

dotsdl: Ah thanks, this is my flub. Yes, we want `conda create` instead. What I get writing this from memory and not quickly testing it. :facepalm:

j-wags (approved):
Not sure how much you'd like to expand this, but I'll give blanket approval for whenever you think it's ready.

dotsdl:
I think that's all I wanted really. Thanks @j-wags, @mattwthompson! Merging!

PR #730, opened 2020-09-30
Add Topology.is_periodic

mattwthompson:
This was discussed in #638 and #711 but I don't think we got around to implementing it. The idea here is to add a boolean `Topology.is_periodic` property that makes more explicit whether or not it's intended to be a periodic or "no cut-off" system, instead of checking `.box_vectors` for `None`, which can either intentionally or accidentally be `None`. This should be compatible with a  future refactor that drops `Topology.box_vectors` if we want them to be property of the system object and not this object, but does not require that change to take place. This is not a blocking feature for system object development, but I think it would be nice to have.

I also added a setter, which will 
* error out with a custom exception if `True` is passed with box vectors not present but
* accept `False` if they are present.

This is behavior I'm not too sure about and welcome feedback on. Maybe the latter should require `.box_vectors` be `None`-ed out first?


mattwthompson:
@j-wags and I met briefly to discuss this and decided that the best thing to do now would be to modify the getter to just return the state of `.box_vectors`, but not reference a cached internal attribute, and the setter will exist but pretty much function as a validator, informing the user of how to actually "set" the periodicity through setting box vectors. This gets around potential issues of saving the periodicity state in multiple locations and ensuring that they can never become out of sync or constructed/converted incorrectly.

We eventually do need to reach a consensus on how to store periodicity in the pydantic models. The current implementation, mocked up as a pydantic model, is basically

```python3
In [8]: class Topology(pydantic.BaseModel):
   ...:     ...
   ...:     box_vectors: Optional[unit.Quantity] = None
   ...:
   ...:     class Config:
   ...:         arbitrary_types_allowed = True
   ...:
```
and having a `None` value does not distinguish between "this system is non-periodic" and "this system is periodic, I just haven't set the box vectors yet." The actual value of the box vectors could be housed long-term in our system object, but that doesn't necessarily mean it can't be duplicated here.

openforcefield/topology/topology.py
```diff
@@ -1149,6 +1156,32 @@ def box_vectors(self, box_vectors):
...
+        is_periodic : bool
+            Whether or not this Topology is periodici
+
+        """
+        if is_periodic and self.box_vectors is None:
+            raise InvalidPeriodicityError(
+                "Cannot set is_periodic to True without box vectors. Consider setting "
+                ".box_vectors first"
+            )
+        self._is_periodic = is_periodic
```
j-wags, line 1183: (blocking) Is is necessary to store `self._is_periodic` at all? The performance gain should be marginal, but the added complexity of having a data member that could fall out of sync seems pretty high.

j-wags (changes requested):
I don't think that the current state of this reflects the outcome of our discussion.

openforcefield/tests/test_topology.py
```diff
@@ -188,6 +190,21 @@ def test_box_vectors(self):
             topology.box_vectors = good_vectors
             assert (topology.box_vectors == good_vectors * np.eye(3)).all()
 
+    def test_is_periodic(self):
+        """Test the getter and setter for is_periodic"""
+        vacuum_top = Topology()
+        assert vacuum_top.is_periodic is False
```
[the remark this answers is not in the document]
mattwthompson: I probably go back and forth - the only thing I'll consciously do is avoid the `if not:` conditional since I've been burned on falsiness many times.

I looked it up, and if the object is reliably a boolean, there is no difference, since something that evaluates to `False` will only ever end up at the same place in memory. The first two answers in this thread are probably the best summary of my findings: both are fine, but `is` is marginally better for the built-in boolean types as it avoids `__eq__` business - but the opposite is true for derived types like NumPy data. 

https://stackoverflow.com/questions/27276610/boolean-identity-true-vs-is-true

openforcefield/topology/topology.py
mattwthompson: It depends on whether or not the difference between `None` and "haven't set it yet" is important. I'll push what it would look like and seek some more input on it.

More generally, it depends on if we want `.is_periodic`, `.box_vectors`, none, or both to live here long-term vs. other solutions, like carrying them alongside other objects in the process of system creation.

mattwthompson: Here's what it looks like: https://gist.github.com/mattwthompson/2e9d71e4e22cdc3f469bd948d8f01765

openforcefield/tests/test_topology.py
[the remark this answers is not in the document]
j-wags: Oh, interesting. Thanks for checking into this! I'll leave it up to implementer's choice then :-)

openforcefield/topology/topology.py
j-wags: Thanks for the gist. It is a bit awkward to have an error that says "Error: go unset this other thing before you set this thing", but it does keep the _actual_ complexity down to what it was before. **I'm in favor of the functionality you show there, and that you have implemented currently**

Until the `System` object work progresses to the point where we have a strong preference/spec for what's allowed further down the pipeline, let's make this PR as "minimally invasive" as possible -- it's not an API break if we ALLOW an object to be in a new state later, but it is if we DISALLOW a previously-OK state.

openforcefield/topology/topology.py
```diff
@@ -1399,6 +1405,38 @@ def box_vectors(self, box_vectors):
...
+        Parameters
+        ----------
+        is_periodic : bool
+            Whether or not this Topology is periodici
+
+        """
+        if is_periodic is self.is_periodic:
+            warnings.warn(
+                f".is_periodicic is already {is_periodic}. " "Setting has no effect."
+            )
```
j-wags, line 1428: (not blocking) I could take or leave the "Everything is OK" warning. Your call. 
```suggestion

```

j-wags (approved):
This looks great. Thanks for the back-and-forth. I've left one suggestion (which is non-blocking), and other than that this should be good to merge once the release notes are updated :-)

openforcefield/topology/topology.py
mattwthompson: Agreed

mattwthompson:
Test failures are due to the broken OpenMM link, which is known and in the process of being resolved.

PR #743, opened 2020-10-09
 Set non-bonded cutoff to ForceField values in non-periodic systems

### LINKED ISSUE #735: Cutoff Not Set for Topology Without Box Vectors
**Describe the bug**

The cutoff is not set for non-bonded forces when a topology does not have box vectors. This can cause issues where the user sets the box vectors after creating the parameterised system.

This _may_ be intended behaviour, but it caused me a lot of confusion in trying to debug my code as there is no documentation for how nonbonded handlers are resolved and what order things must be set in (#380).

Even for systems without PBC however, I would still expect the specified value of the cutoff to be inserted into the system file rather than the OMM default be used.

**To Reproduce**

```
# box vectors.

# Create the FF
force_field = ForceField()

# Add only a VdW handler.
vdw_handler.method = "cutoff"
vdw_handler.cutoff = 6.0 * unit.angstrom
vdw_handler.scale14 = 1.0

vdw_handler.add_parameter(
    {
        "smirks": "[#18:1]",
        "sigma": 1.0 * unit.angstrom
    }
)

# topology.box_vectors = numpy.eye(3) * unit.nanometers

assert numpy.isclose(
    system.getForce(0).getCutoffDistance().value_in_unit(unit.angstrom),
    6.0
)
```

**Output**
`AssertionError` on the last assert.

 - `openforcefield =0.7.2`

mattwthompson:
May resolve .  #735 


j-wags (approved):
Per @mattwthompson's comment in #735 

> No matter what, the cutoff should not silently fall back to OpenMM's default value

I fully agree with this. We'll want to discuss what the proper behavior for creating nonperiodic systems _should_ be, but if the cutoff attribute _must_ be set to a number, then it makes sense to use this number.

Issue #537, opened 2020-03-06
Where did sigma go in parameter handlers?


proteneer:
**Describe the bug**

We're on a rather old version of the off toolkit (May 2019), and we're currently parsing the parameter handlers using:
```
if handler_name == "vdW":
            ...
            for param in handler_params.parameters:
                sigma = to_md_units(param.sigma)
```
For some reason when we updated to the current version, param.sigma, although defined, now defaults to None.

**To Reproduce**

Access the param.sigma field

**Output**

print(param.sigma) shows None

**Computing environment (please complete the following information):**
 - Linux
 - Not using conda

**Additional context**

I think at some point we defaulted to rmin_half as opposed to sigma, and it's only populated using some magical incantation.

jchodera:
We should definitely allow both retrieval and setting of vdW parameters by _either_ `sigma` or the less commonly used (but occasionally convenient)`rmin_half`!

jchodera:
To be clear: Both `sigma` and `rmin_half` are supposed to be properties that can be set/get for vdW parameters. The internal representation that was originally used to define the parameter is remembered so that we don't accidentally read `rmin_half` and write out `sigma`, but we _must_ provide convenience accessors to change or set either.

j-wags:
Ah, I see. It looks like we used to convert everything to be sigma (even if it was input as rmin_half): https://github.com/openforcefield/openforcefield/blob/0.2.0/openforcefield/typing/engines/smirnoff/parameters.py#L1402-L1405

Now, we more strictly respect the original parameter, which is `rmin_half` for both smirnoff99Frosst and the Parsley line.

> we must provide convenience accessors to change or set either.

I agree that we should do this. I'll add `@property`s for getting and setting both vdW parameters (`sigma` and `rmin_half`), with the getters doing conversion in the background if needed. Is there any clear desired behavior for which should be written out? That is, if a user loads Parsley and sets a `sigma`, should that parameter be written out with an `rmin_half` or `sigma`?

I'm in favor of writing out the most recent representation (so, in the above example, `sigma` should be written out)

jchodera:
> Now, we more strictly respect the original parameter, which is rmin_half for both smirnoff99Frosst and the Parsley line.

[...]

PR #750, opened 2020-10-23
Fix missing sigma/rmin_half attrs

mattwthompson:
#537 

TODO:
- [ ] Fix failing virtual site tests
- [ ] More dict tests
- [ ] Possibly switch default rep to display `sigma`, despite [our force fields](https://github.com/openforcefield/openforcefields/blob/92ab687198c02190b2bbe13b69a6604d217f31a1/openforcefields/offxml/openff-1.3.0.offxml) being shipped with `rmin_half`

Boilerplate:

### DIFF
--- a/openforcefield/typing/engines/smirnoff/parameters.py
+++ b/openforcefield/typing/engines/smirnoff/parameters.py
@@ -1058,7 +1058,7 @@ class _ParameterAttributeHandler:
 
         return smirnoff_data
 
-    def to_dict(self, discard_cosmetic_attributes=False):
+    def to_dict(self, discard_cosmetic_attributes=False, duplicate_attributes=None):
         """
         Convert this object to dict format.
 
@@ -1070,6 +1070,9 @@ class _ParameterAttributeHandler:
         ----------
         discard_cosmetic_attributes : bool, optional. Default = False
             Whether to discard non-spec attributes of this object
+        duplicate_attributes : list of string, optional. Default = None
+            A list of names of attributes that redundantly decsribe
+            data and should be discarded during serializaiton
 
         Returns
         -------
@@ -1081,6 +1084,10 @@ class _ParameterAttributeHandler:
         # returned dict (call list() to make a copy). We discard
         # optional attributes that are set to None defaults.
         attribs_to_return = list(self._get_defined_parameter_attributes().keys())
+
+        if duplicate_attributes is not None:
+            for duplicate in duplicate_attributes:
+                attribs_to_return.pop(attribs_to_return.index(duplicate))
 
         # Start populating a dict of the attribs.
         indexed_attribs = set(self._get_indexed_parameter_attributes().keys())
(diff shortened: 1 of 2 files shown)


j-wags:
>  Possibly switch default rep to display sigma, despite our force fields being shipped with rmin_half

Unfortunately, I think we're in a tough spot, since the most fundamental thing the toolkit should support is faithful round-tripping of OFFXML files. So I don't think a solution that converts `rmin_half` to `sigma` in a round trip is ok here. I like the functionality to retrieve either `sigma` or `rmin_half` from the getters, but `to_dict` needs to output whichever of the two was set most recently. 

The only solution I can think of is a hidden attribute that records "which representation was most recently set" for each `vdWType`, or having the `sigma` setter clear the `rmin_half` attribute and vice versa.

mattwthompson:
I can't say I find it to be a satisfying solution, but I also can't think of anything better, and this requires the least sweeping changes to the codebase as is.

> the most fundamental thing the toolkit should support is faithful round-tripping of OFFXML files

Could you expand on the details a little bit? What are the exact requirements for this condition to be satisfied? This appears to currently be be tested, although I'd add that a naive check (below) and the existing serialization tests in `TestForceField` pass for all commits of this branch, so I'm not sure that

> So I don't think a solution that converts rmin_half to sigma in a round trip is ok here. 

is a requirement that's explicitly checked in the tests. I did one test by hand (with recent changes) and the mainline files are exported with `rmin_half`, the only differences between the files being what you'd expect (comments, cosmetic attributes, etc.).

```python
In [1]: from openforcefield.typing.engines.smirnoff import ForceField


In [2]: parsley = ForceField('openff-1.3.0.offxml')

In [3]: parsley.to_file('out.offxml')

In [4]: round_trip = ForceField('out.offxml')

In [5]: parsley == round_trip  # this checks id(obj) so it should never pass
Out[5]: False

In [6]: hash(parsley) == hash(round_trip)
Out[6]: True
```

openforcefield/typing/engines/smirnoff/parameters.py
```diff
@@ -3476,6 +3483,34 @@ def __init__(self, **kwargs):
...
+                super().__setattr__(name, value)
+                super().__setattr__("sigma", value / 2 ** (1 / 6))
+                self._extra_nb_var = "sigma"
+
+            if name == "sigma":
+                super().__setattr__(name, value)
+                super().__setattr__("rmin_half", value * 2 ** (1 / 6))
+                self._extra_nb_var = "rmin_half"
+
+            super().__setattr__(key=name, value=value)
```
j-wags, line 3502: We can get rid of the `super.__setattr__(name, value)` lines inside the `if` blocks here, since the unconditional `__setattr__` should run regardless (I tested this change locally and it seemed fine) 

```suggestion
            super().__setattr__(key=name, value=value)
            if name == "rmin_half":
                super().__setattr__("sigma", value / 2 ** (1 / 6))
                self._extra_nb_var = "sigma"

            if name == "sigma":
                super().__setattr__("rmin_half", value * 2 ** (1 / 6))
                self._extra_nb_var = "rmin_half"

```

j-wags (approved):
This looks good, and I couldn't find a way to confuse this system while playing around locally. Good work implementing a significant new behavior into the existing codebase in like, 30 lines. That's really impressive.  

I've added a suggestion that should bring your line count down lower. Can't let you pad those github stats :-P

Be sure to update the releasenotes before you merge!

PR #756, opened 2020-10-29
Replace private attribute access in various ParameterHandlers' create_force methods

### LINKED ISSUE #733: Handler Internal Parameter Attributes Not Found
**Describe the bug**
The internal values of a parameter handlers attributes are not set until the attributes are first accessed (e.g. `_method` is not set for a `NonbondedForce` until `method` is accessed). This leads to `AttributeError` exceptions when creating forces from handlers created from scratch without accessing each of the attached attributes.

**To Reproduce**

```
# box vectors.
topology.box_vectors = numpy.eye(3) * unit.nanometers

# Create the FF
force_field = ForceField()

# Add only a VdW handler.
vdw_handler.add_parameter(
    {
        "smirks": "[#18:1]",
        "sigma": 1.0 * unit.angstrom
    }
)

```

**Output**

```
  File "repro.py", line 37, in <module>
    main()
  File "repro.py", line 31, in main
  File "/XXX/openforcefield/typing/engines/smirnoff/forcefield.py", line 1297, in create_openmm_system
    parameter_handler.create_force(system, topology, **kwargs)
  File "/XXX/openforcefield/typing/engines/smirnoff/parameters.py", line 3217, in create_force
    if self._method == "PME":
  File "/XXX/openforcefield/typing/engines/smirnoff/parameters.py", line 948, in __getattr__
    f"{self.__class__} object has no attribute '{item}'"
AttributeError: <class 'openforcefield.typing.engines.smirnoff.parameters.vdWHandler'> object has no attribute '_method'
```

 - `openforcefield =0.7.2`

j-wags:
- [x] Fix #733, and fixed several other instances of directly accessing private attributes that should be handled by the `ParameterAttribute` class.
- [x] Clean up FF loading test that was being skipped due to having an identical name

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -35,6 +35,9 @@ Bugfixes
 """"""""
 - `PR #745 <https://github.com/openforcefield/openforcefield/pull/745>`_: Fixes bug when
   serializing molecule with conformers to JSON.
+- `PR #756 <https://github.com/openforcefield/openforcefield/pull/756>`_: Fixes bug when running
+  :py:meth:`vdWHandler.create_force <openforcefield.typing.engines.smirnoff.parameters.vdWHandler.create_force>`
+  using a ``vdWHandler`` that was initialized using the API.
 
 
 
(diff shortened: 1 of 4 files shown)


j-wags:
This should be ready for review. The test failures are all URL timeouts.

mattwthompson (approved):
LGTM and this fixes the MWE locally as well.

It would seem strange for this to be the only handler using this pattern, but I did a quick scan of `parameters.py` and didn't find anything else as offensive. Some of the other handlers already have "build up using the API" tests, which is good.

PR #760, opened 2020-11-04
Lazy import functions

### LINKED ISSUE #155: Move imports inside functions
If we move `import`s into functions, only importing packages when needed, importing and using the toolkit will be a quicker, cleaner user experience for most users. We should apply this principle generally throughout the toolkit (though this is not necessary for the tests).

mattwthompson:
Resolves #155 

I went through most of the core modules and looked for imports that did not need to be global or did not need to be there at all (F401). I used https://github.com/nschloe/tuna to asses which imports are the slowest. The results are pretty stochastic, but the heaviest now are, roughly in order:
1. SciPy stack
2. NetworkX
3. Cheminformatics toolkits
4. OpenMM
5. versioning module

Right now, these core core imports average around 1.0-1.1 seconds on my machine:

```python3
```

The major change here is moving the `nglview` import into a separate module. Previously, it was being checked every time something from `openforcefield.topology` was imported, even though only a fraction of the time will a visualization method be called. This should save around 0.7s seconds in the majority of workflows.

- [ ] Better comparison of import times

### DIFF
--- a/openforcefield/tests/test_examples.py
+++ b/openforcefield/tests/test_examples.py
@@ -12,7 +12,6 @@ Test that the examples in the repo run without errors.
 # GLOBAL IMPORTS
 # ======================================================================
 
-import glob
 import os
 import pathlib
 import re
--- a/openforcefield/tests/test_forcefield.py
+++ b/openforcefield/tests/test_forcefield.py
@@ -3874,13 +3874,7 @@ class TestForceFieldParameterAssignment:
   </Constraints>
 </SMIRNOFF>
 """
-        from simtk.openmm import app
-
-        from openforcefield.tests.utils import (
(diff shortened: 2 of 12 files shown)


mattwthompson:
Out of scope of this PR, but https://github.com/openforcefield/openforcefield/issues/617 may present an alternative to `networkX`.

mattwthompson:
Reaching diminishing returns with more digging, so opening this for review. I'm also punting on a more thorough comparison of import times, but can be done with something like this one-liner (after `pip install tuna`):

```shell
$ python -X importtime -c "from openforcefield.topology import Molecule" 2> import.log; tuna import.log
```

j-wags (approved):
LGTM. Thanks, Matt!

PR #762, opened 2020-11-04
Add hydrogens in Molecule.from_rdkit

### LINKED ISSUE #566: Molecule.from_rdkit() allows input of implicit-hydrogen molecules
**Describe the bug**
The `Molecule.from_rdkit(rdmol)` method will allow implicit-hydrogen molecules to be imported without converting them to explicit-hydrogen molecules.

**To Reproduce**
```python
from rdkit import Chem
molOrig = Chem.MolFromSmiles("CCCCCCCCCC")
ofmol = Molecule.from_rdkit(molOrig)
print(ofmol.atoms)
from rdkit import Chem
molOrig = Chem.MolFromSmiles("CCCCCCCCCC")
ofmol = Molecule.from_rdkit(molOrig)
print(ofmol.atoms)
```
returns a molecule without explicit protons that works in our pipeline just fine, except will always fail to produce anything useful:
**Output**
```
[Atom(name=, atomic number=6), Atom(name=, atomic number=6), Atom(name=, atomic number=6), Atom(name=, atomic number=6), Atom(name=, atomic number=6), Atom(name=, atomic number=6), Atom(name=, atomic number=6), Atom(name=, atomic number=6), Atom(name=, atomic number=6), Atom(name=, atomic number=6)]
```

 - `osx`
 - `openforcefield==0.6.0` and `rdkit==2019.09.03`

**Additional Context**
cc: https://github.com/openmm/openmmforcefields/issues/107

Remedies are either
* throw an exception on import
* add hydrogens explicitly
* don't worry about anything until `Molecule._validate()` is called just before we need a valid molecule (most permissive, hardest to debug issues)

mattwthompson:
Resolves #566 via implementation suggested in comments. Not clear if there are any downstream affects in this toolkit or other projects built around it.


mattwthompson:
Surprisingly, this patch fixes all of the noted InChi round-trips that fail with RDKit - b554400 failed because none of the expected failures triggered, and e8457f8 should fix that by completely removing that clause. I'm opening this for review as it fits the narrow scope of the feature request, but I would not expect it to be safe to merge until we think more seriously about possible downstream effects.

j-wags:
This seems very common-sense, though I'll need to think about it. Currently, [the OpenEyeToolkitWrapper _unconditionally_ adds explicit Hs if the molecule has any implicit ones.](https://github.com/openforcefield/openforcefield/blob/dbc982ecfe39a3f04d4cec2269b5926db88a9b47/openforcefield/utils/toolkits.py#L1297-L1299). So I don't see why RDKitToolkitWrapper shouldn't. Though there might be some institutional knowledge underlying why we didn't do that previously. 

Also, instead of asking for user input on whether `hydrogens_are_explicit`, we might be able to use something like `sum([atom.GetNumImplicitHydrogens() for atom in rdmol.GetAtoms()])` to see whether `AddHs` should be run.

j-wags (changes requested):
(blocking) An undergrad with the Shirts lab reported a chemical perception error that will likely be tracked down to OpenEyeToolkitWrapper's automatic hydrogen addition. If that's the case, then it's possible that these tests are actually enforcing a **wrong** behavior. 

I'm going to work with them on crafting a minimal reproducing case and opening an issue. Let's park this PR until we know what's going on. 

(I'm marking this as "requesting changes" just for recordkeeping, to keep us from accidentally merging)

mattwthompson:
The last commit involved manually fixing several merge conflicts, since this has lingered for months. Passing locally and the diffs against `upstream/master` look the same before and after, so I think it went smoothly.

Do note that **some** tests are still going to fail while #889 is unresolved, but they are not necessarily related to this patch.

j-wags (commented):
Ok, so I tried out a bunch of things today. I'm writing this as a progress report to pick back up on. We can decide to cut 0.9.2 without this PR if we want the other features out sooner.

First, I tested whether `Molecule.from_file` would respect charges from SDFs (tested loading carboxylates vs carboxylic acids from implicit-H SDF). This PR behaved correctly for those, and I couldn't find a good quality SDF where this PR had trouble :-) 


Then I tried things with SMILES:

```
implicit_h_smiles = "CC"
explicit_h_smiles = '[H]C([H])([H])C([H])([H])([H])'
mapped_implicit_h_smiles = '[C:1][C:2]'
mapped_explicit_h_smiles = '[H:3][C:1]([H:4])([H:5])[C:2]([H:6])([H:7])([H:8])'
implicit_undef_stereo_smiles = 'C(F)(Cl)(Br)'
implicit_def_stereo_smiles = '[C@](F)(Cl)(Br)'
```

```
m = Molecule.from_smiles(implicit_h_smiles)
print(m.to_smiles())
```
> `[H][C]([H])([H])[C]([H])([H])[H]`

```
m = Molecule.from_smiles(explicit_h_smiles)
print(m.to_smiles())
```
> `[H][C]([H])([H])[C]([H])([H])[H]`

So far, so good, but then:

```
m = Molecule.from_smiles(mapped_implicit_h_smiles)
print(m.n_atoms)
print(m.to_smiles())
print(m.properties['atom_map'])
```

> `2`
> `[CH3][CH3]`
> `{0: 1, 1: 2}`

```
m = Molecule.from_mapped_smiles(mapped_implicit_h_smiles)
print(m.n_atoms)
print(m.to_smiles())
```

> `2`
> `[CH3][CH3]`

Somehow when map indices get in the mix, the explicit Hs don't get added :-/ 


I'm not sure what I expect from the stereo tests (maybe the H needs to be explicit for the stereo to be valid in the SMILES), but I'm posting them here as a note to myself for when I get back to this. 

Another potential concern is that, while the OE backend automatically puts the hydrogens in physically reasonable places, the RDKit backend silently places them at (0,0,0), which is unphysical and may surprise folks. I don't think this is a blocking concern.

lilyminium:
I ran into something similar when I was trying to add caps to residues. I'm not sure why this is default, but RDKit doesn't think you have implicit hydrogens unless you set atom.SetNoImplicit(False) (i.e., allow implicit hydrogens). 

When I add this to your `from_smiles()` [function](https://github.com/openforcefield/openff-toolkit/blob/ae2a812a4be4184d9b5db71e370341346a3c0343/openff/toolkit/utils/toolkits.py#L3355-L3360), @j-wags's tests give me the desired 8 atoms. 

```diff
        for atom in rdmol.GetAtoms():
+++            atom.SetNoImplicit(False)
            if atom.GetAtomMapNum() != 0:
                # set the map back to zero but hide the index in the atom prop data
                atom.SetProp("_map_idx", str(atom.GetAtomMapNum()))
                # set it back to zero
                atom.SetAtomMapNum(0)
+++        rdmol.UpdatePropertyCache(strict=False)     
```

> the RDKit backend silently places them at (0,0,0)

I don't know how to query positions from OFF Molecules, but does this fix that?

```diff
--- rdmol = Chem.AddHs(rdmol)
+++ rdmol = Chem.AddHs(rdmol, addCoords=True)
```

If you run into complaints about not calling calcImplicitValence, try `rdmol.UpdatePropertyCache(strict=False)`. Hydrogens are added in both `from_smiles` and `from_rdkit`.

Incidentally you might want to update [this line](https://github.com/openforcefield/openff-toolkit/blob/ae2a812a4be4184d9b5db71e370341346a3c0343/openff/toolkit/utils/toolkits.py#L3395-L3397):

```diff
		molecule = self.from_rdkit(
            rdmol, _cls=_cls, allow_undefined_stereo=allow_undefined_stereo,
+++			hydrogens_are_explicit=hydrogens_are_explicit,
        )
```

j-wags:
@lilyminium Your changes above are great. I've put in the `addCoords` kwarg and tested it. The resulting behavior seems good. 

The other suggested changes gave the right behavior for `from_smiles` when given partial atom maps, but then I realized that OpenEyeToolkitWrapper has the same problem. I don't want to expand the scope for this PR to also modify `from_openeye`, so I'll open a new issue for those.

j-wags (approved):
Ok, I found a lot of spooky bugs by turning over this stone, but to the best of my knowledge this PR doesn't add any new ones, and it definitely fixes an important use case. I've added a broad statement of possible impact to the release history, but in my tests I couldn't find any cases where a well molecule does any worse than before. Thanks for the patience. I'm hoping that these can move faster once we have a cheminformatics-wiz in residence! 🧙

Issue #761, opened 2020-11-04
Smirnoff+Amber simulation


aizvorski:
I'm trying to follow the examples/using_smirnoff_with_amber_protein_forcefield/toluene_in_T4_lysozyme.ipynb .  Everything works, except there isn't any obvious way to take the resulting `complex_system` and run it in openmm.  

`complex_system` is an `openmm.System` which is produced from `complex_structure.createSystem` which is a ParmEd `Structure`.  To run an openmm simulation it's also necessary to have an openmm topology and positions.  The complex_structure contains position data since it can save a full structure with `complex_structure.save('file.pdb')`, but there is no obvious way to get those as openmm topology object.

How can I extend the toluene_in_T4_lysozyme example to actually running an openmm simulation?

j-wags:
Hi @aizvorski ,

Thanks for writing in. This code should carry that example forward into a simulation (though, note that the minimization takes a while on my laptop):

```
from simtk import openmm, unit

# Propagate the System with Langevin dynamics.
time_step = 2*unit.femtoseconds  # simulation timestep
temperature = 300*unit.kelvin  # simulation temperature
friction = 1/unit.picosecond  # collision rate
integrator = openmm.LangevinIntegrator(temperature, friction, time_step)

# Length of the simulation.
num_steps = 1000  # number of integration steps to run

# Logging options.
trj_freq = 1  # number of steps per written trajectory frame
data_freq = 1  # number of steps per written simulation statistics

# Set up an OpenMM simulation.
simulation = openmm.app.Simulation(complex_structure.topology, complex_system, integrator)

# Set the initial positions.
positions = complex_structure.positions
simulation.context.setPositions(positions)

# Randomize the velocities from a Boltzmann distribution at a given temperature.
simulation.context.setVelocitiesToTemperature(temperature)

# Configure the information in the output files.
pdb_reporter = openmm.app.PDBReporter('trajectory.pdb', trj_freq)
state_data_reporter = openmm.app.StateDataReporter('data.csv', data_freq, step=True,
                                                   potentialEnergy=True, temperature=True,
                                                   density=True)
simulation.reporters.append(pdb_reporter)
simulation.reporters.append(state_data_reporter)

simulation.minimizeEnergy()
minimized_coords = simulation.context.getState(getPositions=True).getPositions()

simulation.context.setVelocitiesToTemperature(300*unit.kelvin)
simulation.step(10)
```

aizvorski:
[...]

PR #763, opened 2020-11-05
Add "toolkit showcase" example

j-wags:
This is an example I've been meaning to add for a long time -- It's the notebook I go through at talks, which mainly features a way to go from a prepared ligand and protein all the way to a solvated simulation with ions. Several blocks in this notebook show how to do useful steps that haven't been documented anywhere else.

- [x] Address #761 by adding a soup-to-nuts protein simulation setup example
- [x] Polish up text/documentation
- [x] Figure out how to handle nonstandard mdtraj and gromacs deps


mattwthompson:
@j-wags could you add these files to tracking?

* CID_15513.pdb
* minim.mdp
* nvt.mdp
* npt.mdp
* md.mdp

Yoshanuikabundi:
We seem to be missing the `img` folder for this example - `img/openforcefields.png`, `img/xkcd_charge.png`, and `img/dog_food.jpg`. Can obviously remove the references to them but I like pictures! I'm guessing `xkcd_charge.png` is [XKCD 567](https://xkcd.com/567/) but I'm clueless about the others.

Yoshanuikabundi:
WIP notebook on binder: https://mybinder.org/v2/gh/openforcefield/openforcefield/toolkit-showcase?filepath=%2Fexamples%2Ftoolkit_showcase

j-wags:
Notes from today's meeting:

* Let's remove type hints
* Remove `-c omnia` from openmmforcefields install instructions (also maybe just remove the install instructions since we'll now base this off openmmforcefields)
* Since we're no longer using ParmEd for system combination, try using the constrained force field (`openff-1.3.0.offxml`) for the example. This should let us take 2fs timesteps. (this may crash at system export to GROMACS... Let's touch base if so)
* add a note that the padding during water addition is too small for general use, it's just so that we can run a quick example
* Remove ParmEd from water-clash resolution (maybe not needed since we'll solvate the protein+ligand together?)
* Use constraints for hbonds throughout gromacs example

Yoshanuikabundi:
Tried the unconstrained force field... It caused ParmEd to crash at OpenMM export, and in addition GROMACS crashes while GROMPPing, I think because ParmEd didn't write out any bond information at all for hydrogen-involving bonds. Will have another look (and actually record the error messages :/) after I convert it to use `openmmforcefields`

Yoshanuikabundi:
Implemented the switch to `openmmforcefields`, at least in code. Turned out to be pretty simple, only needed to write ~30 lines of glue code to convert an OFFTK `Topology` to an OpenMM `Topology`, which now resides in a function in the utils.py file. This removes the need for the `find_clashing_water` function, so it seemed like a fair trade.

The resulting workflow is much simpler, and the OpenMM side of it doesn't need to touch ParmEd. I also got the GROMACS bit working without the extra warnings, and the whole thing now uses the standard Parsley 1.3.0 force field (rather than the constrained one).

Tomorrow I'll move the descriptions and prose around to update them for the new workflow and then we should be good to review.

j-wags (commented):
Re-add openmmforcefields to environment --> Could add openff-toolkit-examples to binder reqs. 

Try replacing `openff_topology_to_openmm` with `offtop.to_openmm()` (this worked during our call)

In the `omm_forcefield.createSystem` step, let's either:
* add a warning box, stating that normally a SMIRNOFF FF contains information about the cutoff+constraints that should be used. 
* Load the `openff-1.3.0` forcefield explicitly and reference `parsley['Electrostatics'].cutoff` (note though that this won't work for the `openmm.app.PME` or the hbond constraints)

We can keep from double-calling `omm_forcefield.createSystem` if we use a PDB file to go to NGLView right after we parameterize the complex:
```
from simtk.openmm.app import PDBFile
PDBFile.writeFile(modeller.topology, modeller.getPositions(), open('complex.pdb', 'w'))
view = nglview.show_file('complex.pdb')
```
Then the second createSystem and the ParmEd stuff can all go into the GROMACS section. 

> ParmEd can't handle constraints from openmmforcefields

should become 

> ParmEd's GROMACS exporter can't handle constraints from openmm

> The Toolkit will always support all the functional forms in both the SMIRNOFF spec

should become

> The Toolkit will strive to support all the functional forms in both the SMIRNOFF spec

>  Load the ligand molecule and the Parsley force field 

could become 

>  Teach OpenMM about the ligand molecule and the Parsley force field

Point the links to other examples to the _folder containing the example notebook_. 

Pending the changes above,  **I approve this merge!** Since I'm technically the author you'll need to self-approve and then merge, but you can let me know if there's trouble and I'll force it though.

Yoshanuikabundi (approved):
I've implemented all the changes Jeff listed in his review, and I also changed the visualization steps to directly convert OpenMM `Topology` objects to MDTraj ones, which means we don't have to write PDB files to disk for them.

PR #776, opened 2020-11-25
Update Topology.from_mdtraj

### LINKED ISSUE #773: Topology.from_openmm/from_mdtraj don't autoconstruct unique_molecules — docstring wrong?
**Describe the bug**
[Documentation for openforcefield.topology.Topology](https://open-forcefield-toolkit.readthedocs.io/en/latest/api/generated/openforcefield.topology.Topology.html#openforcefield.topology.Topology.from_mdtraj) states that ``unique_molecules`` can be omitted and the list will be auto-constructed, if possible, from ``mdtraj_topology``.

I spent a while trying to get this to work before looking at the source code and realizing that this feature doesn't seem to be present ([ref](https://open-forcefield-toolkit.readthedocs.io/en/latest/_modules/openforcefield/topology/topology.html#Topology.from_mdtraj)). If I'm wrong, please disregard and close this issue. If I'm right, it would be nice if the docstring were changed to reflect this and not lead unwary sojourners astray with false hope.

mattwthompson:
Fixes #773 by
* raising an exception `MissingUniqueMoleculesError` when `unique_molecules=None` but shouldn't be (previously a cryptic error would be raised when trying to iterate over `None` inside of `from_openmm`)
* adding tests for `Topology.from_mdtraj`
* updating docstrings

I'm open to "removing" it from the public API by prepending the method with an underscore; I have no opinion on that decision either way.

- [ ] Can `from_openmm` safely infer molecules from an OpenMM Topology if `unique_molecules=None`? I can't tell if this was ever working

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -67,7 +67,7 @@ Bugfixes
   using a ``vdWHandler`` that was initialized using the API.
 - `PR #776 <https://github.com/openforcefield/openforcefield/pull/776>`_: Fixes a bug in which
   the :py:meth:`Topology.from_openmm <openforcefield.topology.Topology.from_openmm>` and
-  :py:meth:`Topology.from_openmm <openforcefield.topology.Topology.from_openmm>` methods would
+  :py:meth:`Topology.from_mdtraj <openforcefield.topology.Topology.from_mdtraj>` methods would
   dangerously allow `unique_molecules=None`.
 
 
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -32,7 +32,10 @@ Behavior Changed
   running these constructors using subclasses of :py:class:`FrozenMolecule <openforcefield.topology.Molecule>`
   would not return an instance of that subclass, but rather just an instance of a
   :py:class:`Molecule <openforcefield.topology.Molecule>`.
-
+- `PR #753 <https://github.com/openforcefield/openforcefield/pull/753>`_: ``ParameterLookupError``
+  is now raised when passing to
+  :py:meth:`ParameterList.index <openforcefield.typing.engines.smirnoff.parameters.ParameterList>`
+  a SMIRKS pattern not found in the parameter list.
 
 New features
 """"""""""""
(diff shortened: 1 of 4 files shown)


openforcefield/tests/test_topology.py
```diff
@@ -605,6 +606,11 @@ def test_from_openmm(self):
             get_data_file_path("systems/packmol_boxes/cyclohexane_ethanol_0.4_0.6.pdb")
         )
 
+        with pytest.raises(
+            MissingUniqueMoleculesError, match="requires a list of Molecule objects"
+        ):
+            Topology.from_openmm(pdbfile.topology, unique_molecules=None)
```
j-wags, line 612: Could we remove the `unique_molecule=None` part? What we mean to test is that the _default_ argument alone is unacceptable, and some day we might make the default an empty list, but still want to fail if it's empty.

openforcefield/tests/test_topology.py
```diff
@@ -711,6 +717,28 @@ def test_to_from_openmm(self):
...
+
+        pdb_path = get_data_file_path(
+            "systems/test_systems/1_cyclohexane_1_ethanol.pdb"
+        )
+        trj = md.load(pdb_path)
+
+        with pytest.raises(
+            MissingUniqueMoleculesError, match="requires a list of Molecule objects"
+        ):
+            Topology.from_mdtraj(trj.top, unique_molecules=None)
```
j-wags, line 732: ditto as above.

j-wags (approved):
Thanks, Matt!

> Can from_openmm safely infer molecules from an OpenMM Topology if unique_molecules=None? I can't tell if this was ever working

I don't think so. Even if the `mdtraj_topology` object knows bond orders, it will still need to explicitly know its own formal charge. So this is not simple functionality, and it's certainly not implemented yet.

docs/releasehistory.rst
```diff
@@ -62,13 +65,10 @@ Bugfixes
...
-
-Behavior changed
-""""""""""""""""
-- `PR #753 <https://github.com/openforcefield/openforcefield/pull/753>`_: ``ParameterLookupError``
-  is now raised when passing to
-  :py:meth:`ParameterList.index <openforcefield.typing.engines.smirnoff.parameters.ParameterList>`
-  a SMIRKS pattern not found in the parameter list.
+- `PR #776 <https://github.com/openforcefield/openforcefield/pull/776>`_: Fixes a bug in which
+  the :py:meth:`Topology.from_openmm <openforcefield.topology.Topology.from_openmm>` and
+  :py:meth:`Topology.from_openmm <openforcefield.topology.Topology.from_openmm>` methods would
```
mattwthompson, line 70:
```suggestion
  :py:meth:`Topology.from_from_mdtraj <openforcefield.topology.Topology.from_mdtraj>` methods would
```

PR #782, opened 2020-12-08
Remove pins to previous major versions of pytest and coverage

mattwthompson:
Toward the end of #615, pytest bumped a major version (6.x) and some pytest-adjacent packages had yet to roll out updates for compatibility, so I pinned to `pytest <6` (980f21802730e9b1e67a6eb65d6417f5477cd279). That was in July, and I believe everything has been updated since then. During the 0.8.1 release, I noticed this pin was not removed, and this PR does that. My development environment, and probably others', does not have this pin, so I don't expect tests to fail.


mattwthompson:
Something wasn't working with using the cov trinity during the notebook tests, so I dropped it. In principle, any functionality "missed" by notebooks/examples/docs should also be covered elsewhere in unit tests, and in this case it seems to not have an impact on the coverage (for better or worse).

j-wags (approved):
LGTM!

PR #796, opened 2020-12-18
Test on Python 3.8 and 3.9

mattwthompson:
We should test on newer versions of Python before releasing builds that support them. Whether or not we deep the support for new versions official, it'd be easy for users to install them after the migration because OpenMM supports up to 3.9 and most of this stack is `noarch`.

The only dependency I've found that isn't available on Python 3.8+ is `openmmtools`. It is in the group of packages that are being migrated, but has not been released or built for a few months and is one of the few that can't be `noarch`. However, it's not used in any non-deprecated examples, so I'm just dropping it from all environments.

Everything else I looked at should be available on both versions (even if only on `omnia`) because of the shift to `noarch` builds.


mattwthompson:
Tests are failing because recent builds of `smirnoff99frosst` are not `noarch` and falling back to older versions results in something that can't be imported

j-wags:
Building a s99f noarch package now 
https://github.com/omnia-md/conda-recipes/pull/1078

j-wags:
Uploaded the noarch s99F package to `main` and restarted tests on this PR. Also running general `conda` test workflow here: https://github.com/openforcefield/openforcefield/actions/runs/431289262

mattwthompson:
Excellent - those tests are passing (so far) and it's pulling down the build you just uploaded

j-wags (approved):
Cool. No releasehistory update needed. LGTM if tests pass :-)

mattwthompson:
Will merge once tests pass, which might be a wait.

I have to admit, though, that I'm not sure why these failed. The old builds were noarch, using the same tarball, and nothing else I could distinguish as different. I probably won't dig much deeper into this now that it works, but I wanted to not that I don't know _why_ it now works.

mattwthompson:
Getting some timeouts in the macOS builds, restarting

mattwthompson:
Looks like it's a repeatable failure on macOS; Ubuntu builds have all passed

j-wags:
Weird. Seems like a timeout in a coverage test in the non-OE build. Since this passed for OE, it's probably something in `RDKitToolkitWrapper.generate_conformers` or `AmberToolsToolkitWrapper.compute_partial_charges`.

mattwthompson:
That's right - OpenEye builds seem to zoom through the partial charges (~5s):

```
2020-12-18T22:47:14.2407280Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 0 PASSED [ 15%]
2020-12-18T22:47:14.2605743Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 1 PASSED [ 16%]
2020-12-18T22:47:14.2818138Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 2 PASSED [ 17%]
2020-12-18T22:47:14.3036738Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 3 PASSED [ 18%]
2020-12-18T22:47:14.3255454Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 4 PASSED [ 18%]
2020-12-18T22:47:18.5268751Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 5 PASSED [ 19%]
2020-12-18T22:47:22.6058852Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 6 PASSED [ 20%]
2020-12-18T22:47:22.8803088Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 7 PASSED [ 21%]
2020-12-18T22:47:23.0486014Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 8 PASSED [ 22%]
```

and the RDKit-only builds fail after timeout (~2000 seconds, a default set by `nbval`):

```
2020-12-18T22:44:20.3746600Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 0 PASSED [ 15%]
2020-12-18T22:44:20.3970730Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 1 PASSED [ 16%]
2020-12-18T22:44:20.4353020Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 2 PASSED [ 17%]
2020-12-18T22:44:20.4981660Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 3 PASSED [ 18%]
2020-12-18T22:44:20.5135060Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 4 PASSED [ 18%]
2020-12-18T22:44:28.0201830Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 5 PASSED [ 19%]
2020-12-18T23:17:48.5338760Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 6 FAILED [ 20%]
2020-12-18T23:17:48.8818020Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 7 XPASS [ 21%]
2020-12-18T23:17:49.2397440Z examples/check_dataset_parameter_coverage/check_parameter_coverage::ipynb::Cell 8 XPASS [ 22%]
```

However, the charges are set to dummy values and not re-calculated later
```
    # Set partial charges to placeholder values so that we can skip AM1-BCC 
    # during parameterization
    molecule.partial_charges = (np.zeros(molecule.n_atoms) + 0.1) * unit.elementary_charge
```

The failing cell is actually one that uses `multiprocessing` to repeat a serial operation, but in this example the dataset is small so I don't think it's a concern here. So it's arguably droppable, but it would be nice to ensure that cell works if a user wants to use it to push through a larger dataset. A ~1000x performance difference between the two toolkits seems unlikely, doesn't it?

The failing builds are reporting hanging operations on RDKit builds, and only on Python 3.8 and 3.9 and only on macOS. It turns out the default spawn start was [changed](https://docs.python.org/3/whatsnew/3.8.html#multiprocessing), only on this platform, in 3.8 from `fork` (used on other platforms) to `spawn`. I'm forcing it to be `fork` so as to at least be the same on all platforms. `spawn` [might](https://pythonspeed.com/articles/python-multiprocessing/) be better, but it doesn't work out of the box like it should. (If we ever switch to officially supporting Windows, we'll need to [be more careful](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods)).

Issue #712, opened 2020-09-14
Add more options for controlling which tests run


mattwthompson:
It would be nice to have some arguments to pass to `pytest` that can enable more control over which tests run. The current test suite takes something like 10 minutes to run if only doing the unit tests and about twice that if including examples. There is huge variance in different parts of the test suite, i.e. all of `test_parameters.py` takes ~10 seconds and just running the serialization tests in `test_molecule.py` takes about a minute or two despite being stable (on the timescale of a few commits). At different times, one might want to have some of these turned off or on, i.e. to check that a change didn't break the `Molecule` API, but doing so without running all serialization tests for all molecules and file formats.

**Describe the solution you'd like**
Some clever use of `pytest.mark`(https://docs.pytest.org/en/stable/example/markers.html) would enable the toggling on/off ofcategories of tests, i.e.
 * Serialization (through `to_dict` formats and `{to|from}_file` methods)
 * Parametrizing of molecules
 * Cheminformatics round-trips (SMILES, IUPAC, etc.)
  * `Molecule.{to_from}_{rdkit|openeye}`
 * Assigning charges
 * etc.

From a set of options, we can choose the default tests to run for different purposes, i.e. running more tests on nightly cron jobs than normal CI runs. Even if most are on by default, markers could enable focusing on just a set of functionality, i.e. `pytest -v -m serialization` to only run serialization tests.

**Describe alternatives you've considered**
 In some cases, you can point `pytest` to a particular `TestClass` and only run its tests, but that requires a fair amount of intervention and only gives so much control. What often happens is needing to pick between waiting 10+ minutes for all tests to run before pushing a commit(s) or pushing "I think this will work, but I'm not sure" commits and waiting 20-30 minutes for CI to report back.

**Additional context**
The existing `--runslow` option (which was recently fixed #707) captures something similar to this idea, but with less granularity.

mattwthompson:
FWIW, the culprits for examples running fairly slow seem to be those that deal with swapping parameters; each `swap_amber_parameters` notebook takes ~4 mins
```
2020-09-15T14:19:41.5005420Z examples/QCArchive_interface/QCarchive_interface.ipynb ..........F.      [ 10%]
2020-09-15T14:19:52.9318940Z examples/SMIRNOFF_simulation/run_simulation.ipynb .......                [ 16%]
2020-09-15T14:19:55.8584790Z examples/check_dataset_parameter_coverage/check_parameter_coverage.ipynb . [ 17%]
[...]

PR #800, opened 2020-12-30
Testing updates

mattwthompson:
This PR is another pass through the test suite, fixing a handful of that are non-critical but non-negligible

- [x] Update `unittest` code in `test_topology.py`
- [x] Convert `create_molecule` functions to fixtures
  - [ ] Find a way to force these through `@pytest.mark.parametrize` (maybe by re-using molecules with something other than fixtures?)
- [ ] Load MiniDrugBank molecules from a common source (discovered in a recent PR, one which I can't dig up at the moment)
- [x] Add marks for turning on/off some categories of tests (#712) i.e. `pytest -m "not serialization"` will skip tests marked as such
- [x] Remove ` as excinfo` pattern when `excinfo` not used

### DIFF
--- a/openforcefield/tests/conftest.py
+++ b/openforcefield/tests/conftest.py
@@ -458,6 +458,7 @@ def dinitrogen():
     dinitrogen.partial_charges = charges
     return dinitrogen
 
+
 @pytest.fixture(scope="function")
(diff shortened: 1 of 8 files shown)


mattwthompson:
Closing for now, may revisit in part later

Issue #804, opened 2021-01-06
Migrate openforcefield to openff.toolkit


SimonBoothroyd:
## Description

This issues tracks the progress of migrating the `openforcefield` package to the new `openff-toolkit` and the `openforcefields` package to the new `openff-forcefields`. The will involve

* renaming the `openforcefield` repository to `openff-toolkit` and the `openforcefields` repository to `openff-forcefields`.
* making a new release of the conda-packages as `openff-toolkit` and `openff-forcefields`.
* migrating the `openforcefield` package to a namespace package with imports changing from `import openforcefield` to `import openff.toolkit`.

The rationale for this change is to improve clarity of where this package fits into the overall OpenFF ecosystem, and to make it consistent with the other core OpenFF packages (`openff-evaluator`, `openff-recharge`, `openff-system`, `openff-benchmark`...)

## Notes

* The version of the `openforcefields` package is intrinsically tied to the version of the latest force field that it contains (e.g. `openforcefields=1.3.0` contains the `openff-1.3.0.offxml` force field and its predecessors). This currently prevents the `openforcefields` package from being migrated to a full name space package as it would require a new release of the package without a new force field being released. 
  
  The current consensus is to go ahead with the renaming of the `openforcefields` repository, and to release a new conda package under the name of `openff-forcefields=1.3.0`. It will install into a directory called `openforcefields`, and will still be importable by `openforcefields` but will show up in conda list / be conda installable as `openff-forcefields`.
On the next FF release we will simultaneously perform the namespace migration and FF release.

  While not ideal, there isn't really a better alternative without resorting to hacky alternatives (patching the install in conda, creating a tag which doesn't properly follow semver... etc...) or by breaking the parity between the package version and the force field version.

## Migration Plan

The migration should proceed as follows:

1) ~add depreciation warnings to the repository stating that after `0.8.3` the package will need to be imported as `openff.toolkit`, rather than `openforcefield` (#805).~

2) ~cut a new release of the package as `0.8.3` and release on `omnia` **only**. *After this release no new features should be merged into master until the 0.9.0 release has been made.*~

3) ~rename the `openforcefield` repository to `openff-toolkit` and the `openforcefields` repository to `openff-forcefields`.~

4) ~release the `smirnoff99frosst` and `openff-forcefields` packages on `conda-forge` **only**. (tag https://github.com/conda-forge/staged-recipes/pull/13657)~

[...]

PR #805, opened 2021-01-06
Add Deprecation Warning for Namespace Change

SimonBoothroyd:
- [X] Tag issue being addressed (#804)
- [X] Update [changelog](https://github.com/openforcefield/openforcefield/blob/master/docs/releasehistory.rst)

### DIFF
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -16,7 +16,7 @@ From version ``0.9.0`` onwards the toolkit will be importable only as ``import o
 ``from openff.toolkit import XXX``.
 
 **Note** This change will also be accompanied by a renaming of the package from ``openforcefield`` to ``openff-toolkit``,
-and so users need not worry about accidentally pulling in a version with changed imports. Users will have to explicitly
+so users need not worry about accidentally pulling in a version with changed imports. Users will have to explicitly
 choose to install the ``openff-toolkit`` package once released which will contain the breaking import changes.
 
 New features
--- a/docs/releasehistory.rst
+++ b/docs/releasehistory.rst
@@ -23,7 +23,7 @@ New features
 """"""""""""
 - `PR #794 <https://github.com/openforcefield/openforcefield/pull/794>`_: Adds a decorator
   ``@requires_package`` that denotes a function requires an optional dependency.
-- `PR #804 <https://github.com/openforcefield/openforcefield/pull/804>`_: Adds a deprecation warning for the up-coming
+- `PR #805 <https://github.com/openforcefield/openforcefield/pull/805>`_: Adds a deprecation warning for the up-coming
   release of the ``openff-toolkit`` package and it's import breaking changes.
 
 0.8.2 - Bugfix release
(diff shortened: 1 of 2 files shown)


openforcefield/__init__.py
```diff
@@ -1,4 +1,17 @@
...
+
 from ._version import get_versions
 
 __version__ = get_versions()["version"]
 del get_versions
+
+
+warnings.warn(
+    "Importing this package as `import openforcefield.XXX` and "
+    "`from openforcefield import XXX` was deprecated in version `0.8.3`. From version "
```
jaimergp, line 11:
```suggestion
    "`from openforcefield import XXX` was marked for deprecation in version `0.8.3`. From version "
```

docs/releasehistory.rst
```diff
@@ -10,11 +10,21 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
 0.8.3 - Current development
 ---------------------------
 
+This release is a final patch for the ``0.8.X`` series of releases of the toolkit, and also marks the last
+version of the toolkit which will be imported as ``import openforcefield.XXX`` / ``from openforcefield import XXX``.
+From version ``0.9.0`` onwards the toolkit will be importable only as ``import openff.toolkit.XXX`` /
+``from openff.toolkit import XXX``.
+
+**Note** This change will also be accompanied by a renaming of the package from ``openforcefield`` to ``openff-toolkit``,
+and so users need not worry about accidentally pulling in a version with changed imports. Users will have to explicitly
```
mattwthompson, line 19:
```suggestion
so users need not worry about accidentally pulling in a version with changed imports. Users will have to explicitly
```

docs/releasehistory.rst
```diff
@@ -10,11 +10,21 @@ Releases follow the ``major.minor.micro`` scheme recommended by `PEP440 <https:/
...
+choose to install the ``openff-toolkit`` package once released which will contain the breaking import changes.
+
 New features
 """"""""""""
-- `PR #794 <https://github.com/openforcefield/openforcefield/pull/974>`_: Adds a decorator
+- `PR #794 <https://github.com/openforcefield/openforcefield/pull/794>`_: Adds a decorator
   ``@requires_package`` that denotes a function requires an optional dependency.
-
+- `PR #805 <https://github.com/openforcefield/openforcefield/pull/805>`_: Adds a deprecation warning for the up-coming
+  release of the ``openff-toolkit`` package and it's import breaking changes.
```
mattwthompson, line 27:
```suggestion
  release of the ``openff-toolkit`` package and its import breaking changes.
```

openforcefield/__init__.py
```diff
@@ -1,4 +1,17 @@
+import warnings
+
 from ._version import get_versions
 
 __version__ = get_versions()["version"]
 del get_versions
+
+
+warnings.warn(
+    "Importing this package as `import openforcefield.XXX` and "
```
mattwthompson, line 10: Non-blocking, but it would be nice to have more precise language here
* The entire package is never really imported
* Saying "importing this module as ... " is incorrect when only a class is imported
* Likewise, "importing this class/etc. ..." is also incorrect when entire modules are imported

Maybe something more general like "Importing components from this package as ... " / "From version `0.9.0` onwards, components of this package will need to be imported as ..." ?

mattwthompson (approved):
Diff looks good to me with one note. I grabbed this branch locally and it worked as expected

PR #786, opened 2020-12-10
Reliability fixes for benchmarking work and 0.8.2 releasenotes

j-wags:
- [x] Fixes #785 
- [x] Fix Molecule copy initializer (`offmol2 = Molecule(offmol)`) so that `offmol2.properties` doesn't point to the same dict as `offmol.properties`
- [x] Add [tests](https://github.com/openforcefield/openforcefield/tree/master/openforcefield/tests)
- [x] Update docstrings/[documentation](https://github.com/openforcefield/openforcefield/tree/master/docs), if applicable
- [x] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [x] Update [changelog](https://github.com/openforcefield/openforcefield/blob/master/docs/releasehistory.rst)

PR #811, opened 2021-01-12
Fix Antechamber conformer dependence test

### LINKED ISSUE #810: macOS, CI-only failures in AmberTools charge assignment
**Describe the bug**
Some AmberTools+RDKit partial charge assignment tests  are failing in CI as of some point over the weekend. I doubt these are related to either the new patch of AmberTools (https://github.com/conda-forge/ambertools-feedstock/pull/53 or https://ambermd.org/ATPatches.php) as the changelog seems unrelated for all recent patches. I may be missing something obvious but there have not been major changes to charge assignment or toolkit wrapper methods merged lately. The [`macos-latest`](https://github.com/actions/virtual-environments#available-environments) image that GitHub uses is also still on Catalina.

**To Reproduce**
Look at CI logs for failures like [here](https://pipelines.actions.githubusercontent.com/u2Wf9sTwFK9NZj8rRqcrkPO1UF3tPeGIdu4NysSHujaRyWbUAY/_apis/pipelines/1/runs/2319/signedlogcontent/5?urlExpires=2021-01-11T15%3A37%3A06.9164960Z&urlSigningMethod=HMACV1&urlSignature=o1gFTZVU0jCZH1jyI7wsd6MCRPbcHQiuoLy5cnAKaug%3D) (link may not be persistent). However, `pytest -v openforcefield/tests/test_toolkits.py::TestAmberToolsToolkitWrapper::test_assign_partial_charges_conformer_dependence` in a a conda environment that does not have OpenEye Toolkits installed, on macOS, passes locally.

**Output**

Locally: passes

In CI:
```
2021-01-11T00:44:31.6208080Z _ TestAmberToolsToolkitWrapper.test_assign_partial_charges_conformer_dependence[am1bcc] _
2021-01-11T00:44:31.6210370Z openforcefield/tests/test_toolkits.py:2523: in test_assign_partial_charges_conformer_dependence
2021-01-11T00:44:31.6213650Z     assert abs(pc1 - pc2) > 1.0e-3 * unit.elementary_charge
2021-01-11T00:44:31.6215770Z E   assert Quantity(value=0.0, unit=elementary charge) > (0.001 * Unit({BaseUnit(base_dim=BaseDimension("charge"), name="elementary charge", symbol="e"): 1.0}))
2021-01-11T00:44:31.6218670Z E    +  where Quantity(value=0.0, unit=elementary charge) = abs((Quantity(value=-0.1361, unit=elementary charge) - Quantity(value=-0.1361, unit=elementary charge)))
2021-01-11T00:44:31.6221040Z E    +  and   Unit({BaseUnit(base_dim=BaseDimension("charge"), name="elementary charge", symbol="e"): 1.0}) = unit.elementary_charge
2021-01-11T00:44:31.6223990Z _ TestAmberToolsToolkitWrapper.test_assign_partial_charges_conformer_dependence[am1-mulliken] _
2021-01-11T00:44:31.6226360Z openforcefield/tests/test_toolkits.py:2523: in test_assign_partial_charges_conformer_dependence
2021-01-11T00:44:31.6228630Z     assert abs(pc1 - pc2) > 1.0e-3 * unit.elementary_charge
2021-01-11T00:44:31.6230720Z E   assert Quantity(value=0.0, unit=elementary charge) > (0.001 * Unit({BaseUnit(base_dim=BaseDimension("charge"), name="elementary charge", symbol="e"): 1.0}))
2021-01-11T00:44:31.6233570Z E    +  where Quantity(value=0.0, unit=elementary charge) = abs((Quantity(value=-0.254, unit=elementary charge) - Quantity(value=-0.254, unit=elementary charge)))
2021-01-11T00:44:31.6235910Z E    +  and   Unit({BaseUnit(base_dim=BaseDimension("charge"), name="elementary charge", symbol="e"): 1.0}) = unit.elementary_charge
```

 - Operating system: macOS **only**; Ubuntu-based builds are passing
 - Output of running `conda list` (each has been run with both 20.9 and 20.10)
   - Locally: https://gist.github.com/mattwthompson/f1e899d793f44d692b500518e585a07d
   - Copied from a CI log: https://gist.github.com/mattwthompson/3c91ea32c95fc1cf5ed45c01321897c6

**Additional context**
OpenEye should not be involved, or at least the failures are only on RDKit=True, OpenEye=false builds. Using 20.10 vs 20.9 didn't make a difference locally, and I was able to run each in CI by restarting  the workflow before and after 20.10 was rolled out to Anaconda.

I doubt this is a bug in the actual code, but it may point to something that requires further testing or changes to the existing tests.

mattwthompson:
Shotgun attempt, may resolve #810. The existing test moves one of the carbons over +3 A in x, this simply moves the other carbon +3 A in y in an attempt to further screw up the partial charges.


j-wags (approved):
Hurrah! This is a satisfactory solution for me. No bookkeeping needed on this one.

mattwthompson:
@SimonBoothroyd I know you spent some time looking into this - are you satisfied with the assumption that it's as digit rounding issue in some files, or is there any reason to believe there is something else going on? If the former, and if we don't come across any other bad behaviors, I'll just merge this (after 0.9.0).

SimonBoothroyd:
I've not been able to reproduce this issue locally at all (I even tried running on Catalina in addition to Big Sur to match the CI) so it's not clear at all what's going on here...

> are you satisfied with the assumption that it's as digit rounding issue in some files, or is there any reason to believe there is something else going on? 

I don't think it's a rounding issue - if you compare what the charge differences should be locally then there are definitely larger than machine (and even float) precision... but I'm genuinely not sure what the issue could be here...

j-wags:
I was also mega-confused by this, but the fix seems to be be reasonable. 

> I don't think it's a rounding issue - if you compare what the charge differences should be locally then there are definitely larger than machine (and even float) precision

Out of curiosity, what is the magnitude of the difference? 

I'd typed up some hypotheses last night, many of which would be disproven by Matt's working fix, but I'll list them all here for completeness:

* These calculations are being done in a temporary directory. Maybe something in the new CI machines is assigning the same temporary directory to both runs, the directory isn't being deleted, and the second run is picking up the results of the first.
* Some difference in rounding with the new Python versions we support? (disproven, we would have seen this when we added the new Python versions to the matrix)
* Some complex bug in new AmberTools that returns all 0s on those platforms? (disproven, magnitudes are printed by pytest and are non-zero)
* Some problem where the partial charges for both molecules actually reference the same array? Could be related to [this change](https://github.com/openforcefield/openff-toolkit/pull/786/files#diff-2def74dd5edd6688466f856dc7dc664dbc970b47c790cfc84cc4db369e0b802cR2196-R2200) (unlikely, since we should have seen this in CI)
* Some difference in AmberTools that causes geometry optimization to run so long that both arrive at the same structures?
* Some mix of different macos hardware architectures, like maybe some ARM machines are being titrated in, and not being properly segregated from the x86s?

mattwthompson:
I'm going to merge this (if tests continue to pass) but leave the associated issue open; this fix at least serves the short-term interest of getting that test to pass, but the underlying issue (if there is one) is unresolved and cannot be diagnosed.

PR #816, opened 2021-01-15
Match file paths case-insensitive when loading ForceField data

### LINKED ISSUE #815: Allow Smirnoff99Frosst force fields with lower case names
When trying to run the smirnoff99Frosst forcefields via QCArchive the name is lowered somewhere within the internals of QCFractal/QCEngine which means the force field can not be loaded on Linux but this still works on mac as it ignores the capitalization when loading the file.

It would be great if the toolkit could even out this behaviour by replacing any lower case names which reach the toolkit with the correct capitalization before loading the file. I think this could be done with the following code placed here https://github.com/openforcefield/openff-toolkit/blob/bb40b067484f968b39092b8a892bbc12f2d6d010/openff/toolkit/typing/engines/smirnoff/forcefield.py#L1087
```python
import re
source = re.sub("frosst", "Frosst", source)
```

I have tried changing this in other places in the chain but I think it makes the most sense here as this is the main call to load the forcefield file.

mattwthompson:
Resolves #815 (if I can figure out an elegant way to do the matching)


j-wags (approved):
This has been behaving well in the `0.8.4rc` line, so I think this is good to merge! Please update the releasenotes before doing so.

PR #803, opened 2021-01-05
Migrate openforcefield to openff.toolkit

SimonBoothroyd:
## Description

This PR handles the migration over to the new `openff-toolkit` package name and the namespace migration such that the package will be importable as `openff.toolkit` rather than `openforcefield`.

The consensus is that this PR should be merged ready for the first release of this package on C-F, such that the C-F version will only ever contain `openff-toolkit >=0.9.0`, while the omnia channel will only contain the `openforcefield <=0.8.2`

## Todos

- [ ] Tag issue being addressed
- [ ] Add [tests](https://github.com/openforcefield/openforcefield/tree/master/openforcefield/tests)
- [X] Update docstrings/[documentation](https://github.com/openforcefield/openforcefield/tree/master/docs), if applicable
- [X] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [ ] Update [changelog](https://github.com/openforcefield/openforcefield/blob/master/docs/releasehistory.rst)
- [ ] Update the `openforcefields` CI dependencies. This hasn't been changed yet otherwise the CI will fail.

## Notes 

- This PR updates all links from `openforcefield/openforcefield/` to `openforcefield/openff-toolkit` as the repo will also be renamed as part of the package rename. Until the repo is renamed however the link tests are expected to fail.
- The RTD links will not be updated / changed as there is currently not a clean way to rename a RTD site.
- The `swap_existing_ligand_parameters_with_openmmforcefields.ipynb` is expected to fail tests as it relies upon the `openmmforcefields` package which in turn depends upon `openforcefield`.

## Status

- [x] Ready to review
- [ ] Ready to merge

PR #818, opened 2021-01-19
Fix deployment tests, skip tests requiring optional dependencies

mattwthompson:
This PR fixes a small typo in the deployment tests (that would have been quite difficult to see ahead of time in #803). This is contained to just this workflow; **the release is not broken**. When comparing git tags, the runner steps up one directory before running `git ls-remote` to find the tag of the cloned repo, which is then compared to the version found in the Python package. This check works fine and passes. But when going back into the repo, it needs to be `openff-toolkit`, not `openff/toolkit`. So this line is an exception to the general pattern of which path needs to be stepped into.

I'll see if I can run the workflow from this branch to be sure it works.


mattwthompson:
This link is probably not persistent, but the failing check is fixed and the tests are passing so far: https://github.com/openforcefield/openff-toolkit/runs/1729531265?check_suite_focus=true

mattwthompson:
To clarify, this PR does two things
1. Fixes the `openff-toolkit` path while the runner is moving around in the directories
2. Marks some tests as xfail



\# 1 will work before a release, but # 2 will not work until the next release because this action checks out the most recent commit

j-wags (approved):
Looks good, and the `test_molecule` changes are really nice too. For the tests where OE and RDKit gave insignificantly different results (eg different precisions), I've changed them to be separate tests. Thanks, @mattwthompson!

PR #828, opened 2021-02-02
Remove deprecated np.float

mattwthompson:
NumPy 1.20 was released over the weekend and deprecated some aliases for built-in types.

In:

```python3
import numpy as np
np.float
```

Out:
```python3
__main__:1: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
<class 'float'>
```

This PR takes the deprecation warnings on its face and changes some `np.float` to just `dtype=float`. I couldn't find other instances of similar built-in types. Some of the OpenEye API only worked with `dtype=np.float64`, so I left some like that.

This seems innocuous and unlikely to present problems, so we can get ahead of this warning being printed any time our users import `Molecule`.


j-wags (approved):
Hm, I can reproduce the warning in your example (just calling `np.float`), but I can't find a sequence of toolkit calls that raise it. 

Anyway, I don't see much of a chance for harm here, so I'm fine to go ahead with it!

mattwthompson:
FWIW, one way they sneak in is with conformers, I can trigger it with i.e. `pytest -v openff/toolkit/tests/test_molecule.py::TestMolecule::test_add_conformers`

PR #831, opened 2021-02-10
Expose ELF conformer selection through the OpenEye wrapper.

SimonBoothroyd:
## Description

This PR exposed a new `apply_elf_conformer_selection` function through the `OpenEyeToolkitWrapper` in keeping with the same functionality which was added to the RDKit wrapper (#793).

The new function employs the built-in OpenEye ELF methods to select a set of ELF conformers from an existing set stored on an input molecule.

## TODO

## Status

- [x] Ready to review
- [ ] Ready to go

### DIFF
--- a/openff/toolkit/utils/toolkits.py
+++ b/openff/toolkit/utils/toolkits.py
@@ -58,6 +58,7 @@ import importlib
 import inspect
 import itertools
 import logging
+import re
 import subprocess
 import tempfile
 from distutils.spawn import find_executable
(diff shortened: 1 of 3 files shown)


openff/toolkit/utils/toolkits.py
```diff
@@ -2114,6 +2115,86 @@ def generate_conformers(
...
+
+        output_stream = oechem.oeosstream()
+
+        oechem.OEThrow.SetOutputStream(output_stream)
+        oechem.OEThrow.Clear()
+
+        status = oequacpac.OEAssignCharges(oe_molecule, charge_engine)
+
+        oechem.OEThrow.SetOutputStream(oechem.oeerr)
+
```
j-wags, line 2171: (not blocking) Could you add (in a comment here) an example of what the error messages look like? I've found it's super helpful to have a raw text snippet when including parsing code like this, especially if the output format breaks some day and we don't know what input this was assuming.

openff/toolkit/utils/toolkits.py
```diff
@@ -2114,6 +2115,86 @@ def generate_conformers(
...
+        from openeye import oechem, oequacpac
+
+        if molecule.n_conformers == 0:
+            return
+
+        oe_molecule = molecule.to_openeye()
+
+        # Select a subset of the OMEGA generated conformers using the ELF10 method.
+        charge_engine = oequacpac.OEELFCharges(
+            oequacpac.OEAM1Charges(), limit, percentage, True
```
j-wags, line 2160: (blocking) I think traditional ELF10 conformer selection uses MMFF94 charges. https://docs.eyesopen.com/toolkits/python/quacpactk/molchargetheory.html#elf-conformer-selection

j-wags (changes requested):
Looks very nice. One blocking question -- Should the charges to determine the ELF10 conformer be calculated using AM1 or MMFF? I think it's the latter.

SimonBoothroyd:
> Should the charges to determine the ELF10 conformer be calculated using AM1 or MMFF? I think it's the latter.

The `OEAM1Charges` refers to the method to use when computing the underlying charges which will be averaged, rather than the method used in the ELF method which I believe is hardcoded, so would've been safe here. 

Looking over the docs however it seems like OE has added [a new API point](https://docs.eyesopen.com/toolkits/python/quacpactk/releasenotes/version2_1_0.html) which avoids going through the charge API entirely which is much cleaner - I've changed over to this in a6a2f5e

openff/toolkit/utils/toolkits.py
SimonBoothroyd: Commented on this below

openff/toolkit/utils/toolkits.py
SimonBoothroyd: Hmm I'm not actually sure what error message will be raised here (especially with the new API point) - I'll play around a bit and see if I can get one to generate, but otherwise will pass on this until we get a real world example!

j-wags (approved):
Looks great. Thanks!

PR #832, opened 2021-02-11
Expose ELF conformer selection through the Molecule API

SimonBoothroyd:
## Description

This PR adds a new `apply_elf_conformer_selection` function to the Molecule API, exposing the new ELF functionality added in #793 and #831.

## TODOs

## Status

- [X] Ready to go

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -2816,8 +2816,8 @@ class FrozenMolecule(Serializable):
 
         See Also
         --------
-        ``OpenEyeToolkitWrapper.apply_elf_conformer_selection``
-        ``RDKitToolkitWrapper.apply_elf_conformer_selection``
+        OpenEyeToolkitWrapper.apply_elf_conformer_selection
+        RDKitToolkitWrapper.apply_elf_conformer_selection
 
         Parameters
         ----------
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -37,6 +37,7 @@ import operator
 import warnings
 from collections import OrderedDict
 from copy import deepcopy
+from typing import Optional, Union
 
 import networkx as nx
 import numpy as np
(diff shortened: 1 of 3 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -2792,6 +2793,62 @@ def generate_conformers(
...
+        -----
+        * The input molecule should have a large set of conformers already
+          generated to select the ELF conformers from.
+        * The selected conformers will be retained in the `conformers` list
+          while unselected conformers will be discarded.
+
+        See Also
+        --------
+        ``OpenEyeToolkitWrapper.apply_elf_conformer_selection``
+        ``RDKitToolkitWrapper.apply_elf_conformer_selection``
```
j-wags, line 2820: This makes docs fail to build. Not sure why. Yay rst. 
```suggestion
        OpenEyeToolkitWrapper.apply_elf_conformer_selection
        RDKitToolkitWrapper.apply_elf_conformer_selection
```

openff/toolkit/topology/molecule.py
```diff
@@ -2792,6 +2793,62 @@ def generate_conformers(
...
+
+        Parameters
+        ----------
+        toolkit_registry
+            The underlying toolkit to use to select the ELF conformers.
+        percentage
+            The percentage of conformers with the lowest electrostatic interaction
+            energies to greedily select from.
```
j-wags, line 2828: (not blocking) Will there be an API change if we switch to the OE-style "actually use a higher percentage if there aren't many input conformers" approach? Or would we make a separate method for that? We may want to plan ahead so that we end up with a clear API.

j-wags (approved):
LGTM! One nonblocking question about how the high-level API may expand in the future, but feel free to merge as we discuss.

Also, for some reason the docs fail to compile with the 
```
See Also
--------
``OpenEyeToolkitWrapper.apply_elf_conformer_selection``
```
pattern, but they succeed when the backticks are removed. No clue why. 🤷  (cc'ing @Yoshanuikabundi -- One of reStructured Text's odd little behaviors to look out for)

openff/toolkit/topology/molecule.py
SimonBoothroyd: I think the API would remain the same, just a docstring update and a note about the behaviour change during a release would be fine.

PR #833, opened 2021-02-12
Raise error when force field is missing Electrostatics tag

### LINKED ISSUE #716: ForceField with no Electrostatics tag can be used to make a system
**Describe the bug**
A `ForceField` with no `Electrostatics` tag can be used to create an OpenMM `System` with arbitrarily-set electrostatics.

**To Reproduce**

```
mol=Molecule.from_smiles('O')
ff = ForceField('test_forcefields/tip3p.offxml')
ff.create_openmm_system(mol.to_topology())
from pprint import pprint
pprint(XmlSerializer.serialize(sys))
```

```
(?xml version="1.0" ?
 System openmmVersion="7.4.2" type="System" version="1"
   PeriodicBoxVectors
     A x="2" y="0" z="0"/
     B x="0" y="2" z="0"/
     C x="0" y="0" z="2"/
   /PeriodicBoxVectors
   Particles
     Particle mass="15.99943"/
     Particle mass="1.007947"/
     Particle mass="1.007947"/
   /Particles
   Constraints
     Constraint d=".09572000000000001" p1="0" p2="1"/
     Constraint d=".09572000000000001" p1="0" p2="2"/
     Constraint d=".15139006545247014" p1="1" p2="2"/
   /Constraints
   Forces
     **Force alpha="0" cutoff="1" dispersionCorrection="1" 
 ewaldTolerance=".0005" forceGroup="0" ljAlpha="0" ljnx="0" ljny="0" ljnz="0" 
 method="0" nx="0" ny="0" nz="0" recipForceGroup="-1" rfDielectric="78.3" 
 switchingDistance="-1" type="NonbondedForce" useSwitchingFunction="0" 
 version="3"**
       GlobalParameters/
       ParticleOffsets/
       ExceptionOffsets/
       Particles
         Particle eps=".635968" q="-.834" sig=".3150752406575124"/
         Particle eps="0" q=".417" sig="1"/
         Particle eps="0" q=".417" sig="1"/
       /Particles
       Exceptions
         Exception eps="0" p1="0" p2="1" q="0" sig="1"/
         Exception eps="0" p1="0" p2="2" q="0" sig="1"/
         Exception eps="0" p1="1" p2="2" q="0" sig="1"/
       /Exceptions
     /Force
   /Forces
 /System)
```

mattwthompson:
Fixes #716

This caused me problems today - what cutoff/method to use in a force field that only specifies library charges?

The SMIRNOFF spec is not clear on whether or not `Electrostatics` is **required**; I'm not entirely sure if it should be in an abstract sense, but that is currently the only way to specify electrostatics method, 1-4 scaling factor, and (optionally) cutoff, so it's needed in the current code.

The file `test_forcefields/tip3p.offxml` could be updated, since it's not strictly valid out of the box. But it's only used for testing, so that's not necessarily a requirement, and I'm not sure if PME is canon for TIP3P.

- [ ] Update SMIRNOFF spec?
- [x] Update TIP3P file?


openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1281,6 +1289,10 @@ def create_openmm_system(self, topology, **kwargs):
         """
         return_topology = kwargs.pop("return_topology", False)
 
+        if "Electrostatics" not in self.registered_parameter_handlers:
+            # Could do this later, to not raise when systems don't have any charges?
```
mattwthompson, line 1293: There could be a more elegant way to handle this, I guess, if we're concerned about force fields that have no electrostatics. Relates to if that is a requirement in the spec.

j-wags:
I think there are actually two issues at play here:

* A theoretical issue: 
    * If the FF **does** assign partial charges, then it **must** have an `Electrostatics` tag (since otherwise, it's not clear how to calculate energies/forces with those charges).
    * If the FF **doesn't** assign partial charges, then it **can't** (or maybe just shouldn't) have an `Electrostatics` tag

* An implementation issue: In OpenMM, if EITHER a `vdW` OR `Electrostatics` tag is specified, then BOTH should probably be specified (since they're both going to cause a `NonbondedForce` to be created, and will force the default settings of one into another). 

So, I think the right place to raise an error IS in `ForceField.create_openmm_system`, since that's where the user has committed to the implementation limitations of OpenMM. 

In the short term it would be simplest to have `create_openmm_system` just complain unless BOTH a `vdW` and `Electrostatics` tag are in the FF. But if we wanted to be complete, the matrix of allowable outcomes should be:

|                    | no partial charges assigned       | some partial charges assigned | all partial charges assigned       |
|--------------------|-----------------------------------|-------------------------------|------------------------------------|
| No nonbonded tags  | System with no NonbondedForce     | MissingPartialChargesError    | MissingElectrostaticsHandlerError  |
| Electrostatics tag | MissingPartialChargesError        | MissingPartialChargesError    | NonbondedForce with all 0 epsilons |
| vdW tag            | NonbondedForce with all 0 charges | MissingPartialChargesError    | MissingElectrostaticsHandlerError  |
| ES+vdW tags        | MissingPartialChargesError        | MissingPartialChargesError    | NonbondedForce                     |


So, it'll be simplest to just implement the final row and have any other permutation of FF/partial charges raise either the correct sort of error, or a NotImplementedError. @SimonBoothroyd, since you're the main customer, do you foresee testing force fields that don't have vdW or electrostatics any time soon?

mattwthompson:
I'd like to emphasize how much I think we need the spec to be updated to make clear if the `<Electrostatics>` tag is required, and lingering ambiguity will only make the implementation difficult long-term. (It's probably worth exploring this question for other tags, at least `<vdW>`.) As it stands,

> Electrostatic interactions are specified

the implication to me is that it is required, but I don't know. If we can get (or manufacture) clarity on this question, we can be sure that our implementation here and future implementations follow the spec - if SMIRNOFF is flexible to the extent of optional non-bonded tags, then we know we need to cover them, but if it's not, we don't (and shouldn't).

j-wags:
> I'd like to emphasize how much I think we need the spec to be updated to make clear if the <Electrostatics> tag is required, and lingering ambiguity will only make the implementation difficult long-term

It shouldn't be required. It should be valid for a SMIRNOFF FF to contain only bonds, and for someone to go run a simulation using that.

mattwthompson:
> In the short term it would be simplest to have create_openmm_system just complain unless BOTH a vdW and Electrostatics tag are in the FF. But if we wanted to be complete, the matrix of allowable outcomes should be:

I think the best thing to do is just handle the first case now - "stop the bleeding" - and implement the table in a follow-up PR. I've updated this to reflect that, even though the added exceptions should be removed in the next PR.

Because of the requirement

> If the FF does assign partial charges, then it must have an Electrostatics tag 

I also updated `test_forcefields/tip3p.offxml` to have a boilerplate `<Electrostatics>` tag.  The 1-4 scaling factors aren't needed for water, but in principle it's necessary for PME to be specified, so I do think this is a necessary change.

j-wags:
I've thought about this more, and based on 

* hearing about @SimonBoothroyd's use cases on Monday and realizing that, in the exploratory nonbonded work, there is _intentionally_ no vdWHandler/ElectrostaticsHandler (or subclass thereof)
* seeing the number of changes that I had to make to the tests to get them passing again

I think that raising an error here (with no override) will actually be counterproductive. 

What Simon/FF experts need is `CustomNonbondedForces` in cases where the `NonbondedForce` class doesn't squarely fit. Any change other than that is either a move sideways or backwargs, and raising an error with no override will actually stop Simon's work. 

Casual users just want to get some kind of energy and don't care about whether their ES and vdW cutoffs agree. 

So, I'm inclined to just make this a warning, like `MissingElectrostaticsHandlerWarning: No ES cutoff specified. Assuming your ES cutoff is {omm_nonbondedforce.getCutoff()} for no good reason`

I know that this leaves pointy edges in cases where, for example, there's no ESHandler, and only _some_ molecules get partial charges assigned. But that's a separate issue.

Apologies, I know that #716 and the other discussion up until now have been saying "this should be an error", and a warning won't really stop the bleeding, but now that I've thought more, the easy cure for this is worse than the disease.

Any objection? If not, I'll roll forward with this in the next few days.

mattwthompson:
No objections here

j-wags:
I just learned from @jchodera and @peastman that OpenMM can be expected to have the same performance if vdW and electrostatics are in separate `CustomNonbondedForces` as if they're in the same `NonbondedForce`!

So, I think we can take a different angle on this problem -- something like "make `create_openmm_system` have an optional kwarg to create separate forces for vdW and electrostatics, even if they COULD be in the same `NonbondedForce`". Making the old behavior remain the default should preserve reverse-compatibility for folks that were accessing individual Forces of the outputted OpenMM system, but have the override unblock @SimonBoothroyd's work on alternate nonbonded functional forms!

And, in the bigger picture, I need to become more active on the OpenMM issue tracker to ask this sort of thing ahead of time, since this knowledge unblocks a lot of the complexity around this issue, and I should have just asked sooner 🤦

mattwthompson:
A bit of a complication if the idea is to have two `CustomNonbondedForce`: https://github.com/openmm/openmm/issues/1901#issuecomment-329258376

I think this means 4 forces will need to be carried around for most systems - one each for the actual forces, one each for the 1-4 interactions

mattwthompson:
Closing as splitting the mega `NonbondedForce` into separate `CustomNonbondedForce` objects, which better track the sections of a SMIRNOFF force field, looks to be a more promising and sustainable approach. If that ends up not being the case, we can pick this back up (and hope the merge conflicts aren't too nasty).

PR #838, opened 2021-02-16
Consistently use two words for force field in documentation

### LINKED ISSUE #112: Inconsistent with force field as 2 words
The READMEs in this repository are still inconsistent about if force field is one word or two. The Mobley group has been using "force field" and I believe OpenFF agreed on this as well via Slack, but we should make sure we're consistent in our documentation and especially when we write our own name!

Yoshanuikabundi:
Previously, force field was inconsistently rendered as either "forcefield" or "force field". With this PR, documentation, notebook examples and code comments are consistently rendered as two words. This PR also addresses a few incorrectly capitalised and spaced references to the Open Force Field Toolkit, as well as a typo I noticed while going through the docs.

It also adds the auto-generated API documentation folder, `docs/api/generated`, to the `.gitignore` file. This appears to have been an oversight as it is automatically generated when the docs are built and previously could be committed.

Resolves #112.

### Not addressed in this PR:
- Force field is typically rendered correctly according to PEP8 naming rules for class names (`ForceField`), but incorrectly in variable names (`forcefield` rather than the correct `force_field`). Fixing this would constitute a breaking change to the API.
- File names are not changed to preserve URLs
- No regression tests are added. If we later add lints for documentation, one of them can be a spellchecker and we can make sure forcefield is not in the dictionary. Short of that, a regression test would have to be familiar with every syntax in the repository so that it knows how to ignore URLs, filenames, code examples, actual code etc.

### PR Checklist: 
- [ ] ~~Add [tests](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/tests)~~
- [ ] ~~Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable~~


j-wags (approved):
This looks great! A few suggested changes based on markdown/rendering weirdness. But otherwise this is perfect. 

Feel free to merge once you've committed the suggested changes, or if you'd like we can merge at the beginning of our next call.

Issue #792, opened 2020-12-16
Multi-conformer WBO Support


SimonBoothroyd:
**Is your feature request related to a problem? Please describe.**

Currently both the AT and OE backends only use a single conformer to compute the Wiberg bond orders (WBOs). Should multiple conformers be used to compute the WBOs, and if so, how should the WBOs of multiple conformers be pooled / averaged.

This is a question which has been mentioned in several code comments and PR comments, but I think it's worth having a single issue to track this.

Given that Sage is slated to be the first OpenFF to include WBO interpolated parameters it is probably good to start revisiting this and thinking what the 'recommended approach' should be (both for fitting and for the end user).

Related comments this [PR](https://github.com/openforcefield/openforcefield/pull/508#discussion_r379205699):

@j-wags 

> I think conformer averaging would be a nice thing to do. But, like ELF10, it's not clear that there's a single common-sense way to do it. So, I'd like to leave conformer averaging algorithms as something for the community to experiment with. Once someone wants to try it, they can invent a new keyword for bond_order_model, and implement support for it in a ToolkitWrapper.

@davidlmobley:

> Definitely seems like a thing to do separately/unresolved science.

**Describe the solution you'd like**

Some possible solutions would be to:

1. Just use a single conformer (_current default_). This would either be the first conformer provided by a user if multiple conformers are passed to `use_conformers`, or a single conformer generated in-place in `assign_partial_charges`. This would likely be the 'most' conformer dependant and input specific approach to computing the WBOs.

2. Take a simple average of the WBOs for each provided (or in-place generated) conformer.

    a. Take the Boltzmann average of the WBOs for each provided (or in-place generated) conformer as is implemented in [fragmenter](https://github.com/openforcefield/fragmenter/blob/master/fragmenter/chemi.py).

j-wags:
I like the possible solutions listed above, and don't think they're mutually exclusive. 

Option 1 is already in-place, and is triggered by the [`fractional_bondorder_method="AM1-Wiberg"`](https://github.com/openforcefield/openforcefields/blob/2855e9fd81e768e2d53dbe898fda7e42b869bbd7/openforcefields/offxml/openff-1.3.0.offxml#L146) attribute in the `ProperTorsions` or `Bonds` element of a SMIRNOFF force field. As you mention, it has a major shortcoming which is that it's heavily conformer-dependent.

[...]

PR #839, opened 2021-02-16
Add Support for Multi-conformer and ELF10 WBOs

SimonBoothroyd:
## Description

This PR adds support to the OpenEye toolkit wrapper for computing WBOs from multiple conformers, and adds a convenient way to compute them from ELF10 conformers in a single call.

This PR partially addresses #792.

## TODOs

## Status

- [x] Ready to go


j-wags (changes requested):
These changes look technically good, but I think that the result of merging this would put `master` in an unacceptable state -- the difference in `molecule.assign_fractional_bond_orders` behavior with the AmberToolsToolkitWrapper (first conf only) vs. OpenEyeToolkitWrapper (unweighted average over all confs) is substantial. 

So, two concrete asks:

1) Previously, you'd mentioned not including AmberToolsToolkitWrapper in these changes because the conformer minimization issue is unresolved. But I think that problem is orthogonal to this one. Since having both ToolkitWrappers do all-conformer averaging is something we can implement now to prevent the problem above, I think we should add the averaging to `ATTKW.assign_fractional_bond_orders` in this PR as well. 

2) It's likely that the behavior of the method currently called `am1-wiberg-elf10` will change in the future -- maybe, as we find which parameters yield improved charges, the backend implementation will change to have stronger/weaker restraints during minimization, or a different RMS cutoff, or we'll use a fraction other than 2%. The point at which the method becomes "distinct" will be unclear, but at some point we'll have a method which does a substantially different thing than what it does now. This could have undesired effects on force field performance. So we need to reconcile how we want to have names like `am1-wiberg-elf10` refer both to a) code that we can experiment with while avoiding undue concerns about consistency, and b) precise strings that can be included in our next OFFXML and be consistently supported by all future versions of OFFTK.

The second point is hard to solve. Basically, either
* we make the awkward statements "all of the subsequent versions of this implementation are about the same, no need to worry about the charge method changing significantly between releases" and at the same time "this implementation of the charge method is deficient enough that we are investing time and energy into changing it"
* we need to specify _exactly_ what these charge methods _aspire_ to do, and then say that even if our implementation changes from toolkit version to toolkit version, that we're inching ever closer to correctly implementing the aspirational behavior (this may be difficult -- how _exactly_ do we aspire to generate canonical conformers? -- but maybe some lesser level of detail is good enough?)
* we start adding version numbers to everything (eg. `am1-wiberg-elf10-offtk-0.9.1`), which will become a snowball of legacy support

I lean most toward the second option (specify-and-aspire) but am interested to hear your thoughts.

j-wags:
Only the first one is an immediate blocker. 

On the second, my inclination is that we should identify a "aspirational implementation" -- Maybe "whatever ELF10 OE is doing in the 2020 toolkits". But we should write that down somewhere definitive (like the SMIRNOFF spec) by the time this method gets into a released force field.

SimonBoothroyd:
> Previously, you'd mentioned not including AmberToolsToolkitWrapper in these changes because the conformer minimization issue is unresolved. But I think that problem is orthogonal to this one. Since having both ToolkitWrappers do all-conformer averaging is something we can implement now to prevent the problem above, I think we should add the averaging to ATTKW.assign_fractional_bond_orders in this PR as well.

My preference here would have been to just raise a `NotImplementedError` on the AT side when multiple conformers are provided given that the behaviour is not 100% understood. 

My thinking with the conformer minimisation is that it may lead to multiple of a users conformers being collapsed into a single conformer essentially depending on how well the minimiser performs, and thus leading to different behaviour than the user may expect. In practice this likely won't happen (with the exception of strong electrostatic interactions perhaps), but there is a possibility I guess.

I've made the change as requested however in 3f2a37e, with the disclaimer that I wouldn't recommend people use it just yet.

docs/releasehistory.rst
```diff
@@ -34,6 +36,10 @@ Behavior changed
   factor for electrostatic interactions is now properly set by the value specified in the force
   field. Previously it fell back to a default value of 0.83333. The toolkit may now produce
   slightly different energies as a result of this change.
+- `PR #839 <https://github.com/openforcefield/openforcefield/pull/839>`_: The average WBO will now be returned when
+  multiple conformers are provided to ``assign_fractional_bond_orders`` using ``use_conformers`` and an OpenEye
+  toolkit wrapper is used. The behaviour when computing WBOs using the AmberTools toolkit wrapper is currently
+  unchanged.
```
j-wags, line 42:
```suggestion
  multiple conformers are provided to ``assign_fractional_bond_orders`` using ``use_conformers``.
```

j-wags (approved):
Thanks. I agree that it's a mixed bag, but given how stop-and-start development can be around here, I think it's best to keep the toolchains very in sync (otherwise I have a bad habit of forgetting these sorts of details until something really embarrassing happens)

PR #841, opened 2021-02-17
Add missing @requires_openeye to a slowtest that uses mol2

mattwthompson:
The non-OpenEye runs in the cron job are failing because one of the slow tests, which reads a MOL2 file, was not decorated with `@requires_openeye`. This PR fixes that. I thought I covered all cases when testing #722 - I made an environment for each case - but I must have missed it.

Logs (may not be persistent) fail only on the `OpenEye=false` cases and only for this test. This only shows up on the cron job since `--runslow` is off for normal CI runs.

https://github.com/openforcefield/openff-toolkit/actions/runs/573256586


j-wags (approved):
LGTM!

PR #847, opened 2021-02-22
Resolve warnings emitted by Sphinx

### LINKED ISSUE #397: Fix docstrings for ParameterHandlers
**Describe the bug**
Currently, the RTD pages for ParameterHandlers contain the docs for `ParameterAttribute`s, which are uninformative and misleading:
Screen Shot 2019-08-08 at 2 26 54 PM

Yoshanuikabundi:
Previously, building the documentation raised about 400 warnings. This pull request brings that number down to 12.

Many of these warnings (>100) were simple errors in ReStructuredText syntax where the developers' intent was clear.I haven't individually documented these but they frequently caused the rendered documentation to appear incorrect, suggesting that developers are not checking the rendered documentation for the docstrings they write. I suggest we document how to do this clearly in the development guide, add building the documentation to CI, and consider moving away from ReST and towards a simpler format like Markdown.

A major source of warnings (~255 warnings) was that the `ParameterAttribute` class' docstring was reused every time the descriptor was used. This caused warnings owing to the documentation being out of context, and also caused the implementation details of the `ParameterAttribute` class to appear all over the code. This PR adds a `docstring` optional argument to `ParameterAttribute.__init__()` so that by default, a `ParameterAttribute`'s docstring will be empty, but also allowing it to be set by the implementor. This resolves #397.

This PR also removes the `@classmethod` decorator from `utils.toolkits.ToolkitWrapper.toolkit_installation_instructions()`. Previously this method was decorated by both `@property` and `@classmethod`, causing it not to actually work in ordinary Python code as well as documentation warnings. With this PR it is a property that points to the class attribute `ToolkitWrapper._toolkit_installation_instructions`. An alternative fix would be to remove the method and simply rename the attribute to `toolkit_installation_instructions`.

The remaining warnings stem from 2 sources:
1. autodoc tries to import classes like `vdWHandler.vdWType` as though `vdWHandler` was a module when it's actually a class. This accounts for seven warnings. The actual source of this warning is that sphinx-autosummary assumes  `vdWHandler` is a module when it generates the stub files; this can be seen by adding `autosummary_generate_overwrite = False` to `docs/conf.py` and editing one of the auto-generated stub files:
```diff
--- a/api/generated/openff.toolkit.typing.engines.smirnoff.parameters.GBSAHandler.GBSAType.rst
+++ b/api/generated/openff.toolkit.typing.engines.smirnoff.parameters.GBSAHandler.GBSAType.rst
@@ -3,10 +3,10 @@
 
 .. currentmodule:: openff.toolkit.typing.engines.smirnoff.parameters
 
-.. autoclass:: GBSAHandler.GBSAType
+.. autoclass:: openff.toolkit.typing.engines.smirnoff.parameters::GBSAHandler.GBSAType
 
    
-   .. automethod:: __init__
+   .. automethod:: openff.toolkit.typing.engines.smirnoff.parameters::GBSAHandler.GBSAType.__init__
 
    
    .. rubric:: Methods

```
With this change, the corresponding warning goes away. I have filed an [issue](https://github.com/sphinx-doc/sphinx/issues/8919) upstream.

2. The SMIRNOFF spec is in Markdown and converted to RST via M2R2. This raises five spurious warnings that can probably only be fixed upstream in M2R2.

### Checklist
- [ ] ~~Add [tests](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/tests)~~
- [ ] ~~Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable~~

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -63,12 +63,12 @@ autodoc_default_options = {
 # Disable NumPy style attributes/methods expecting every method to have its own docs page
 numpydoc_class_members_toctree = False
 
-_python_doc_base = 'http://docs.python.org/3.6'
+_python_doc_base = 'https://docs.python.org/3.6'
 intersphinx_mapping = {
     _python_doc_base: None,
-    'http://docs.scipy.org/doc/numpy': None,
-    'http://docs.scipy.org/doc/scipy/reference': None,
-    'http://scikit-learn.org/stable': None
+    'https://numpy.org/doc/stable': None,
+    'https://docs.scipy.org/doc/scipy/reference': None,
+    'https://scikit-learn.org/stable': None
 }
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -55,7 +55,7 @@ Open Force Field Toolkit Concepts
 
 ``OFF Topology``
   An object that efficiently holds many OFF ``Molecule`` objects.
-  The atom indexing in a ``Topology`` may differ from those of the underlying ``Molecule``s
+  The atom indexing in a ``Topology`` may differ from those of the underlying ``Molecule``\ s
 
 ``OFF TopologyMolecule``
   The efficient data structures that make up an OFF Topology.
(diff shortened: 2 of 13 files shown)


j-wags (approved):
Looks great. I've tested this locally (tried to set a `ParameterAttribute` docstring and ensured that it appeared in the final docs). This looks wonderful!


> I suggest we document how to do this clearly in the development guide, add building the documentation to CI, and consider moving away from ReST and towards a simpler format like Markdown.

👍 👍 👍 👍 Happy to leave rST behind forever. If only I had more thumbs to give! 👍 👍 👍 👍

Issue #823, opened 2021-01-25
Refactor _get_rings to be a ToolkitWrapper call


mattwthompson:
In #679, I accidentally added direct RDKit imports, breaking the pattern previously used in the Molecule API, which in all other places use `ToolkitWrapper` methods to re-route the imports to that infrastructure. This change is needed to do https://github.com/conda-forge/openff-toolkit-feedstock/issues/1 in the safest manner

SimonBoothroyd:
I think it would be helpful to also refactor the functionality of `is_in_ring` and `n_rings` into toolkit wrappers. This will be important for providing an OE implementation for these functions as OE does not provide an SSR implementation and hence won't be able to provide a `rings` implementation which matches RDKit.

SimonBoothroyd:
@mattwthompson @j-wags I just hit issues with this again - does this have a new milestone as it seems to have been missed from 0.9.2?

mattwthompson:
Could you glance at #855 and let me know if it looks like it's on the right track? It's been long enough I'll have to get into that head space, but if so I can start it back up.

mattwthompson:
@SimonBoothroyd could you also remind me which issues you run into that this would solve? Is it that the different wrappers return different results from ring-finding? Would it be acceptable for `Molecule.{is_in_ring|n_rings}` to _require_ RDKit?

SimonBoothroyd:
#855 looks along the right lines. 

> could you also remind me which issues you run into that this would solve?

Currently RDKit is a required dependency for this functionality, which is not ideal in cases where:

* I don't want to have it installed e.g. I don't need to worry about the TK silently falling back to RD when OE fails. I know I can pass registries or modify the global registry, but this adds a lot of unnecessary verbosity to the code and isn't great a UX
* RDKit fails to create a molecule as in #1181 where OE can successfully.

> Would it be acceptable for Molecule.{is_in_ring|n_rings} to require RDKit?

Is this necessary though? I'm not too bothered about get rings, but both `is_in_ring` and `n_rings` have direct counterparts in OE ([OEAtomIsInRing](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEAtomIsInRing.html#OEChem::OEAtomIsInRing), [OEDetermineRingSystems](https://docs.eyesopen.com/toolkits/python/oechemtk/ring.html#ring-systems-identification))

mattwthompson:
[...]

PR #855, opened 2021-02-26
Refactor ring methods in Molecule class

mattwthompson:
This PR tries to implement solutions to #823. The major change is that several properties `Molecule.rings`, `.n_rings`, `Atom.is_in_ring`, `Bond.is_in_ring`, have been converted to methods, which breaks the existing API. The ways that membership in rings was being checked wasn't necessarily the best, since RDKit exposes a more direct way to do this (I assume OEChem does as well).

The main problem with the earlier implementation is that storing rings in `Molecule._rings` loses the provenance of which toolkit determined the number of rings (RDKit, OpenEye, and other toolkits are not guaranteed to agree). So these methods now generate rings on the fly as needed instead of storing them as private attributes and exposing getters. This should work around edge cases resulting from toolkit differences (unclear how we can test these without OpenEye functionality added). There an obvious downside of this change is wasteful re-computing the rings. I don't know if this is an important performance hit or not.

- [ ] Number of bonds in rings has changed by 1 for some conjugated rings?
- [ ] Any way to keep ring data as properties while respecting toolkit behaviors? (Is this even a good idea?)

### DIFF
--- a/openff/toolkit/tests/test_molecule.py
+++ b/openff/toolkit/tests/test_molecule.py
@@ -3098,9 +3098,8 @@ class TestMolecule:
     @requires_rdkit
     def test_molecule_rings(self, smiles, n_rings):
         """Test the Molecule.rings property"""
-        assert (
-            n_rings == Molecule.from_smiles(smiles, allow_undefined_stereo=True).n_rings
-        )
+        mol = Molecule.from_smiles(smiles, allow_undefined_stereo=True)
+        assert n_rings == mol.get_n_rings()
 
     @pytest.mark.parametrize(
         ("smiles", "n_atom_rings", "n_bond_rings"),
@@ -3115,8 +3114,8 @@ class TestMolecule:
         """Test Atom.is_in_ring and Bond.is_in_ring"""
         mol = Molecule.from_smiles(smiles)
 
-        assert len([atom for atom in mol.atoms if atom.is_in_ring]) == n_atom_rings
-        assert len([bond for bond in mol.bonds if bond.is_in_ring]) == n_bond_rings
+        assert len([atom for atom in mol.atoms if atom.is_in_ring()]) == n_atom_rings
+        assert len([bond for bond in mol.bonds if bond.is_in_ring()]) == n_bond_rings
 
     @requires_pkg("ipython")
     @requires_rdkit
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -5249,12 +5249,11 @@ class FrozenMolecule(Serializable):
             self._get_rings()
         return self._rings
 
-    @RDKitToolkitWrapper.requires_toolkit()
-    def _get_rings(self):
+    def _get_rings(self, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY):
         """
-        Call out to RDKitToolkitWrapper methods to find the rings in this molecule.
+        Call out to ToolkitWrapper methods to find the rings in this molecule.
 
-        Requires the RDKit to be installed.
+        Currently only implemented vai The RDKit, which must be installed.
 
         .. note ::
 
@@ -5262,9 +5261,6 @@ class FrozenMolecule(Serializable):
             function may not be well-behaved and may report a different number
             rings than expected. Some problematic cases include networks of many
             (5+) rings or bicyclic moieties (i.e. norbornane).
-
-        .. todo :: This could be refactored to use ToolkitWrapper.call() to flexibly
-            access other toolkits, if find_rings is implemented.
 
         Returns
         -------
@@ -5274,8 +5270,22 @@ class FrozenMolecule(Serializable):
             found, a single empty tuple is returned.
 
         """
-        toolkit = RDKitToolkitWrapper()
-        rings = toolkit.find_rings(self)
+        if isinstance(toolkit_registry, ToolkitRegistry):
+            rings = toolkit_registry.call(
+                "find_rings",
+                molecule=self,
+            )
+        elif isinstance(toolkit_registry, ToolkitWrapper):
+            toolkit = toolkit_registry
+            rings = toolkit.find_rings(
+                molecule=self,
+            )
+        else:
+            raise InvalidToolkitRegistryError(
+                "Invalid toolkit_registry passed to from_smiles. Expected ToolkitRegistry or ToolkitWrapper. "
+                f"Got  {type(toolkit_registry)}"
+            )
+
         self._rings = rings
 
 
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -436,10 +436,11 @@ class Atom(Particle):
                     return True
         return False
 
-    @property
-    def is_in_ring(self):
+    def is_in_ring(self, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY) -> bool:
         """
         Return whether or not this atom is in a ring(s) (of any size)
+
+        Currently only implemented via The RDKit, which must be installed.
 
         """
         if self._molecule is None:
@@ -447,7 +448,13 @@ class Atom(Particle):
                 "This Atom does not belong to a Molecule object"
             )
 
-        return any([self.molecule_atom_index in ring for ring in self._molecule.rings])
+        _is_in_ring = toolkit_registry.call(
+            "atom_is_in_ring",
+            self._molecule,
+            self.molecule_atom_index,
+        )
+
+        return _is_in_ring
 
     @property
     def virtual_sites(self):
@@ -1646,8 +1653,7 @@ class Bond(Serializable):
             raise ValueError("This Atom does not belong to a Molecule object")
         return self._molecule.bonds.index(self)
 
-    @property
-    def is_in_ring(self):
+    def is_in_ring(self, toolkit_registry=GLOBAL_TOOLKIT_REGISTRY) -> bool:
         """
         Return whether or not this bond is in a ring(s) (of any size)
 
@@ -1657,10 +1663,10 @@ class Bond(Serializable):
                 "This Bond does not belong to a Molecule object"
             )
 
-        for ring in self._molecule.rings:
-            if self.atom1.molecule_atom_index in ring:
-                if self.atom2.molecule_atom_index in ring:
-                    return True
+        if self.atom1.is_in_ring(toolkit_registry=toolkit_registry):
+            if self.atom2.is_in_ring(toolkit_registry=toolkit_registry):
+                return True
+
         return False
 
     def __repr__(self):
@@ -3031,7 +3037,6 @@ class FrozenMolecule(Serializable):
 
         self._cached_smiles = None
         # TODO: Clear fractional bond orders
-        self._rings = None
 
     def to_networkx(self):
         """Generate a NetworkX undirected graph from the Molecule.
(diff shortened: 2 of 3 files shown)


mattwthompson:
The toolkit wrappers are disagreeing on the number of rings in napthlaene:

openff/toolkit/topology/molecule.py
```diff
@@ -2107,10 +2113,10 @@ def is_in_ring(self):
...
             )
 
-        for ring in self._molecule.rings:
-            if self.atom1.molecule_atom_index in ring:
-                if self.atom2.molecule_atom_index in ring:
-                    return True
+        if self.atom1.is_in_ring(toolkit_registry=toolkit_registry):
+            if self.atom2.is_in_ring(toolkit_registry=toolkit_registry):
+                return True
+
```
mattwthompson, line 2119: This logic does not ensure that each atom is in _the same_ ring, only that it is is _some_ ring. A diphenyl is a simple tests case for this; depending the criteria there are 12 or 13 bonds in rings. Which definition should be used here?

j-wags: Hmmm... I think the "12" definition should be used. So we may need to delegate to these calls:
* https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.html?highlight=bond%20ring#rdkit.Chem.rdchem.Bond.IsInRing
* https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEBondBase.html#OEChem::OEBondBase::IsInRing

openff/toolkit/topology/molecule.py
```diff
@@ -5827,30 +5847,11 @@ def get_bond_between(self, i, j):
 
         raise NotBondedError("No bond between atom {} and {}".format(i, j))
 
-    @property
-    def rings(self):
```
j-wags, line 5831: In its current state, this PR removes `Molecule.rings` without replacing it.

Now that I know more about the area, I don't know that we should have exposed `Molecule.rings` and `Molecule.n_rings` in the first place (because, as indicated in both OE and RDKit docs, there are multiple distinct valid behaviors)

If we keep both methods and let them use either the OE or RDK backends, the docstring should have a big warning that different backends will give different outputs. I'm kinda concerned that supporting both backends would invite further scope creep ("if we read into the options we may be able to make them more consistent") and violate the principle of "OpenFF isn't a cheminformatics toolkit". 

On the other hand, if we deprecate `Molecule.rings` and `n_rings`, we could still offer `Atom.is_in_ring` and `Bond.is_in_ring` (which have unambiguous behavior, thereby satisfying @SimonBoothroyd's needs), having each call out directly to a toolkit. In the release notes, we could say "hey, if you need rings, you'll need to do it directly through a cheminformatics toolkit now".

I'm kinda in favor of the "deprecate" route. Thoughts?

SimonBoothroyd: +1 of just deprecating then removing `rings` and `n_rings`. It would also be good to consider allowing `is_in_ring` to take an 'of_size' argument if it's being changed into a function. This would allow us to more easily encode 'is in ring of size' into a feature vector using the toolkit!

mattwthompson:
Looks like a test is failing due to a toolkit disagreement in how many rings a molecule has. I'll look into that tomorrow, probably just removing it since it's a deprecated API point - otherwise this is ready for review @j-wags

openff/toolkit/tests/test_molecule.py
```diff
@@ -3711,25 +3711,46 @@ def test_get_bond_between(self):
...
+    def test_is_in_ring(self):
+        """
+        Test basic behavior of Atom.is_in_ring and Bond.is_in_ring API.
+        More extensive behavior testing is done in test_toolkits.py
+        """
+        molecule = Molecule.from_smiles(
+            "[H:9][c:3]1[c:2]([c:1]([c:6]([c:5]([c:4]1[H:10])[H:11])[H:12])[H:7])[H:8]"
         )
+        assert molecule.atoms[1].is_in_ring()
+        assert molecule.bonds[1].is_in_ring()
```
j-wags, line 3726: (blocking) This is not appropriate - the SMILES above does not specify an order for the bonds, so it would be valid for a toolkit to say `bonds[1]` is a C-H bond.

openff/toolkit/tests/test_molecule.py
```diff
@@ -3711,25 +3711,46 @@ def test_get_bond_between(self):
...
-            n_rings == Molecule.from_smiles(smiles, allow_undefined_stereo=True).n_rings
+        mol = Molecule.from_smiles(smiles, allow_undefined_stereo=True)
+        assert n_rings == mol.get_n_rings()
+
+    def test_is_in_ring(self):
+        """
+        Test basic behavior of Atom.is_in_ring and Bond.is_in_ring API.
+        More extensive behavior testing is done in test_toolkits.py
+        """
+        molecule = Molecule.from_smiles(
```
j-wags, line 3722: (blocking) This is not appropriate - `from_mapped_smiles` should be used if the provided atom order is required.

openff/toolkit/tests/test_molecule.py
```diff
@@ -3711,25 +3711,46 @@ def test_get_bond_between(self):
...
-            ("Cc1ccc(cc1Nc2nccc(n2)c3cccnc3)NC(=O)c4ccc(cc4)CN5CCN(CC5)C", 30, 30),
+            ("c1ccc2ccccc2c1", 2),
+            ("c1ccc(cc1)c2ccccc2", 2),
+            ("Cc1ccc(cc1Nc2nccc(n2)c3cccnc3)NC(=O)c4ccc(cc4)CN5CCN(CC5)C", 5),
         ],
     )
-    @requires_rdkit
-    def test_is_in_ring(self, smiles, n_atom_rings, n_bond_rings):
-        """Test Atom.is_in_ring and Bond.is_in_ring"""
+    def test_n_rings(self, smiles, n_rings, toolkit_wrapper):
```
j-wags, line 3741: (blocking) This test should be removed because `n_rings` is being deleted from the API

j-wags (changes requested):
Sorry to be nitpicky here, but I can't meaningfully review this if we're planning to fully remove `Molecule.rings` and `n_rings` before merging (like I think we've decided to do, though maybe we understood different things from the discussion). Most of the lines in the test diffs will be substantially changed and so the review would be huge and it'd be hard to keep track of everything under discussion. 

I think we should immediately remove those API points with no deprecation warning. It's bad, but we don't have a release planned where anyone would see a deprecation warning anyway, and 0.11.0 will hopefully be our only major API break this year, so this is probably the most realistic way forward.

mattwthompson:
My read of the discussion was that we'd deprecate those points in this release with the intention of removing them later - probably since that's what we would typically do. I will update this to remove them now, though.

docs/releasehistory.md
```diff
@@ -13,6 +13,17 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
...
+  [`Atom.is_in_ring`](openff.toolkit.topology.Atom.is_in_ring) and
+  [`Bond.is_in_ring`](openff.toolkit.topology.Bond.is_in_ring) to use corresponding
+  functionality in OpenEye and RDKit wrappers.
+
+## New deprecations
+- [PR #855](https://github.com/openforcefield/openff-toolkit/pull/855): Deprecates
+  [`Molecule.rings`](openff.toolkit.topology.Molecule.rings). To find rings in a
+  molecule, directuly use a cheminformatics toolkit after using
+  [`Molecule.to_rdkit`](openff.toolkit.topology.Molecule.to_rdkit) or
+  [`Molecule.to_openey`](openff.toolkit.topology.Molecule.to_openeye).
```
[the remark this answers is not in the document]
mattwthompson: My macbook's keyboard is starting to crap out - typically linters cover me in source code. Yikes!

openff/toolkit/tests/test_molecule.py
mattwthompson: 👍 I hadn't considered that bond order is not so trustworthy. Since this is a test in the Molecule API, it's intentionally simple. I rewrote it to check that every C-C bond in benzene is "in a ring"

openff/toolkit/topology/molecule.py
```diff
@@ -2096,22 +2103,36 @@ def molecule_bond_index(self):
...
         """
         Return whether or not this bond is in a ring(s) (of any size)
 
+        Note: Bonds containing atoms that are only in separate rings, i.e. the central bond in a biphenyl,
+            are not considered to be bondned by this criteria.
+
+        Parameters
+        ----------
+        toolkit_registry: openff.toolkit.utils.toolkits.ToolkitRegistry or openff.toolkit.utils.toolkits.ToolkitWrapper, default=GLOBAL_TOOLKIT_REGISTRY
+            :class:`ToolkitRegistry` or :class:`ToolkitWrapper` to use to enumerate the tautomers.
```
j-wags, line 2116:
```suggestion
        toolkit_registry: openff.toolkit.utils.toolkits.ToolkitRegistry, default=GLOBAL_TOOLKIT_REGISTRY
            :class:`ToolkitRegistry` to use to enumerate the tautomers.
```

openff/toolkit/topology/molecule.py
```diff
@@ -445,18 +445,25 @@ def is_bonded_to(self, atom2):
...
         if self._molecule is None:
             raise NotAttachedToMoleculeError(
                 "This Atom does not belong to a Molecule object"
             )
 
-        return any([self.molecule_atom_index in ring for ring in self._molecule.rings])
+        _is_in_ring = toolkit_registry.call(
+            "atom_is_in_ring",
+            self._molecule,
+            self.molecule_atom_index,
```
j-wags, line 463: Here and for `bond.is_in_ring`, we can just pass the `Atom` or `Bond`, since those know their `Molecule`

```suggestion
            self
```

openff/toolkit/topology/molecule.py
```diff
@@ -2096,22 +2103,36 @@ def molecule_bond_index(self):
...
 
-        for ring in self._molecule.rings:
-            if self.atom1.molecule_atom_index in ring:
-                if self.atom2.molecule_atom_index in ring:
-                    return True
-        return False
+        _is_in_ring = toolkit_registry.call(
+            "bond_is_in_ring",
+            self._molecule,
+            self.molecule_bond_index,
```
j-wags, line 2132:
```suggestion
            self
```

openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -1266,6 +1266,47 @@ def to_openeye(molecule, aromaticity_model=DEFAULT_AROMATICITY_MODEL):
 
         return oemol
 
+    def atom_is_in_ring(self, molecule: "Molecule", atom_index: int) -> bool:
```
j-wags, line 1269:
```suggestion
    def atom_is_in_ring(self, atom: Atom) -> bool:
```

j-wags (commented):
This is starting to come together really well. I think we can simplify the toolkitwrapper method signatures and avoid a lot of weird edge cases in the process -- See suggested code changes.

j-wags (approved):
Thanks for all the iteration on this. I'm approving this PR, pending changes in the comment before this post.

docs/releasehistory.md
```diff
@@ -13,6 +13,18 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
   discussion about constraints to the FAQ
 - [PR #1173](https://github.com/openforcefield/openforcefield/pull/1173): Expand
   on the SMIRNOFF section of the toolkit docs
+- [PR #855](https://github.com/openforcefield/openff-toolkit/pull/855): Refactors
+  [`Atom.is_in_ring`](openff.toolkit.topology.Atom.is_in_ring) and
+  [`Bond.is_in_ring`](openff.toolkit.topology.Bond.is_in_ring) to use corresponding
+  functionality in OpenEye and RDKit wrappers.
```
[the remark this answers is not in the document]
mattwthompson: I've left this section in but also added this note to the entry in the breaking changes section. If this is obviously bad it should stick out like a sore thumb to whoever is responsible for tidying up release notes, who has final say in things like that.

PR #859, opened 2021-03-02
Set the maximum number of tautomers to be generated by RDKit.

SimonBoothroyd:
## Description

This PR explicitly sets the maximum number of tautomers that RDKit should generate, rather than just pruning after generating the max number.

## TODOs

## Status

- [x] Ready to go


mattwthompson (changes requested):
I notice that the OpenEye equivalent has `options.SetMaxCount(max_states + 1)` - I don't know which one would be preferred but I think they should be the same?

https://docs.eyesopen.com/toolkits/python/quacpactk/OEProtonClasses/OEFormalChargeOptions.html#OEProton::OEFormalChargeOptions::SetMaxCount

https://www.rdkit.org/docs/source/rdkit.Chem.MolStandardize.rdMolStandardize.html#rdkit.Chem.MolStandardize.rdMolStandardize.TautomerEnumerator.SetMaxTautomers

j-wags:
Simon notes that, in many cases, the toolkit's "generate a certain number of X" methods actually generate a _much larger_ number, and then cut down the outputs. this is super inefficient if the number requested is "1", but internally the method generates a much larger number.

j-wags (approved):
I agree that there's a possible off-by-one ambiguity here, but this PR is already such a clear improvement that I don't think it requires digging into the implementation details to check for differences.

PR #861, opened 2021-03-02
Update testing matrix

mattwthompson:
Ready to go now that 0.9.1 is released

The upcoming toolkit release 0.9.1 will be the last release to support Python 3.6. Many other scientific packages have already dropped it ([NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#detailed-description) schedule this change for last summer!) and we are currently unaware of any users that require 3.6 support.

We agreed on the following matrix
image

Note that OpenEye has not published Python 3.9 builds for their toolkits [link](
https://anaconda.org/OpenEye/OpenEye-toolkits/files), so RDKit-only is the only option at the moment for 3.9. The 3.8 builds are being trimmed in order to get CI running a little quicker; all Python versions, each combination of toolkits, and both Linux and macOS are still being tested, just not all combinations of each.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,38 +18,40 @@ on:
 jobs:
   test:
     if: (github.event_name == 'schedule' && github.repository == 'openforcefield/openff-toolkit') || (github.event_name != 'schedule')
-    name: Test on ${{ matrix.cfg.os }}, Python ${{ matrix.python-version }}, RDKit=${{ matrix.cfg.rdkit }}, OpenEye=${{ matrix.cfg.openeye }}
-    runs-on: ${{ matrix.cfg.os }}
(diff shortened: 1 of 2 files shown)


j-wags (approved):
Looks great. Before merging, could you add some representation of this table to `docs/developing.rst`? I'm dumb and don't know whether a markdown-formatted table can be put in a `.rst` file (maybe @Yoshanuikabundi does?), but I've used [this tool](https://www.tablesgenerator.com/markdown_tables#) to great success before.

mattwthompson:
Turns out rST has its own table syntax, which seems no more powerful than Markdown but significantly more finicky (surprise?). I managed to get an acceptable solution as per Pycharm and a local build (thanks for the instructions in #863, by the way).

image

j-wags:
Looks great. Thanks for digging into this, @mattwthompson. Merge whenever you're ready!

PR #862, opened 2021-03-03
Clarified that systems are OpenMM `System`s

### LINKED ISSUE #618: Clearly designate "OpenMM System" in docs, examples
In some docs, examples, code comments, etc. there are references made to `System`s that are currently only OpenMM `System` objects. While the [`OpenFF System`](https://github.com/openforcefield/openff-system) is still early in development, there will be a future in which both objects exist as desirable outputs of the toolkit. This will be a significant point of confusion for users. I would like to get ahead of this by making more clear in the toolkit when exports are to an OpenMM `System`.

In docs, examples, etc. no references to `System` alone should be used. It should always be clear that, at least for existing code, it is an OpenMM `System`. For example, docstrings should never things like "this function exports to a System ..." and instead should be "this function exports to an OpenMM System ..."

**Additional context**
Note that `openff-system`/`OpenFF System` may have a different name in the future, but I think this change would be worthwhile independent of that. Given that OpenMM `System` is an external class, it would be useful to make that explicitly clear (if perhaps redundantly at times).

I could probably sneak this into 0.7.0 to really get out ahead of this, but 0.7.1 would probably work just as well.

Yoshanuikabundi:
This PR explicitly describes `System`s generated by the toolkit as OpenMM `System`s to resolve future confusion with the forthcoming OpenFF `System` object. The codebase already did a good job of this, but some of the documentation and examples needed clarification. Resolves #618.

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -211,7 +211,7 @@ But when we do need to make tradeoffs, "assume the user is an expert" is our gui
 
 At the same time, we aim for "automagic" behavior whenever a decision will clearly go one way over another.
 System parametrization is an inherently complex topic, and the OFF toolkit would be nearly unusable if we required the user to explicitly approve every aspect of the process.
-For example, if a ``Topology`` has its ``box_vectors`` attribute defined, we assume that the resulting ``System`` should be periodic.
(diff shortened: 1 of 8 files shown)


j-wags (approved):
Looks great! Approved for merge now that 0.9.1 is cut.

PR #866, opened 2021-03-04
Clean up readthedocs.yml

Yoshanuikabundi:
readthedocs.yml is present both in the root of the project (where the RTD docs [say they look](https://docs.readthedocs.io/en/stable/config-file/index.html)) and also in the docs folder. The two files have different contents. In this PR (now that we have PRs building on RTD) I'll consolidate and update the configuration to remove deprecated features.


.readthedocs.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+# Build documentation in the docs/ directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
+build:
+  image: latest
+
+python:
+  version: 3.6
```
j-wags, line 16: We're dropping 3.6 support very soon, so let's be a bit proactive with this. 

```suggestion
  version: 3.8
```

j-wags (approved):
Looks great. I spot-checked the RTD build and nothing seems to be out of place, so this should be good to merge!

PR #871, opened 2021-03-10
Moving to MyST

Yoshanuikabundi:
This PR replaces the m2r2 markdown parser with [MyST](https://myst-parser.readthedocs.io/en/latest/index.html), which is a more active and featureful project. MyST works alongside the Sphinx RST interpreter, so converting the documentation can be down piecemeal.

MyST is a substantial improvement over m2r2. It removes 5 spurious doc build warnings, allows admonitions like notes and warnings to be written in pure markdown (this fixes several previously unnoticed rendering errors in the spec), and uses standard markdown syntax for things like links to API documentation.

Unfortunately, MyST only covers pure documentation files - docstrings must still be written in markdown.

- [x] Docs build without m2r2 dependency
- [x] Update SMIRNOFF spec and ensure rendering is correct
- [x] Remove m2r2 and pip dependencies from RTD environment
- [x] Add links between existing markdown files and API docs (this is now possible!)
- [x] Convert at least one full RST file to Markdown
- [x] See how many warnings I can fix (fixed 5, introduced none)


j-wags (commented):
All of these changes make sense to me and seem like a clear improvement. I've also spot-checked the draft build and it looks just fine. I think that this is in a merge-able state.

However, this is my first exposure to MyST. I know that @mattwthompson has used MyST for his projects, so I'll let him be the final approver on this when @Yoshanuikabundi says it's ready for review. It doesn't seem like anything dramatic has changes, but Matt may have thoughts on "best practices" or anything else regarding its usage.

mattwthompson:
I have not done tons with MyST but I would be happy to review this when it's ready!

Yoshanuikabundi:
I've updated the SMIRNOFF spec to link to the API docs for the `ForceField` class where its mentioned, converted an API doc file over to MD so we can see what it looks like, and have updated the release history file to markdown (since this is the part of the affected docs most devs will have the most interaction with). 

I haven't updated the entirety of the release history to markdown, but instead have just wrapped everything from 0.9.1 backward in an `eval-rst` directive. I figured this was less error prone (and faster!) than updating the entire document. Unfortunately it looks off - the RST header sizes aren't consistent with what they should be, and there doesn't seem to be a way to correct it. So we probably need to decide what to do there.

But I think this is ready for review! @mattwthompson

mattwthompson (approved):
Looks good to me - I'm thrilled that we can now write new docs in Markdown!

I have
* Built the docs locally as a static page using the updated `environment.yml`
* Built the docs locally using`sphinx-autobuild`
* Skimmed some of the API docs, poking around a small number of links
* Compared the SMIRNOFF and release notes side-by-side with the live docs
* Briefly skimmed the RTD build of this PR

and everything looked good to my eye.

Non-blocking: When I build with `sphinx-autobuild`, it worked fine, but from a fresh install it hit a loop that at first I thought was infinite. It seems to be scanning the path including `docs/api/generated` for changes, and some rST files are changed each time the docs are built. For some reason, the docs were re-built N times, where N is the number of files matching `docs/api/generated/*rst`. This resulted in one of my CPU cores being hammered for ~5-10 minutes ~~can a path be excluded from `sphinx-autobuild`?~~ When I add ` --ignore=docs/api/generated/` this startup loop doesn't happen.


Non-blocking: I found a few things in the installation page I didn't like. Are you working on this in your follow-up PRs or can I safely take a swing at that without risking merge conflicts?


 > I haven't updated the entirety of the release history to markdown ...

I probably wouldn't have noticed the formatting differences if you didn't point them out. Ideally these are fixed at some point but I'm fine leaving that for future work. I don't know of any solution easier than converting rST to Markdown by hand. The fact that MyST does a pretty good job of parsing rST side-by-side with Markdown is impressive enough.

Yoshanuikabundi:
> Non-blocking: When I build with `sphinx-autobuild`, it worked fine, but from a fresh install it hit a loop that at first I thought was infinite. It seems to be scanning the path including `docs/api/generated` for changes, and some rST files are changed each time the docs are built. For some reason, the docs were re-built N times, where N is the number of files matching `docs/api/generated/*rst`. This resulted in one of my CPU cores being hammered for ~5-10 minutes ~can a path be excluded from `sphinx-autobuild`?~ When I add ` --ignore=docs/api/generated/` this startup loop doesn't happen.

Interesting. I observed the same loop and tried the same fix but it didn't work for me. I've tried it again just now and it still doesn't work, even after updating `sphinx-autobuild` or reinstalling the whole environment. It works as expected if you build the docs up first before running `sphinx-autobuild`, as the generated API stubs don't change - can you confirm that your modified `sphinx-autobuild` command still prevents the startup loop after you do a `rm -r docs/api/generated/ docs/_build` to clean out the generated stuff? That's why the instructions say to do a `make html` before using `sphinx-autobuild`, as this works around this startup loop. Should that be more explicit?

> Non-blocking: I found a few things in the installation page I didn't like. Are you working on this in your follow-up PRs or can I safely take a swing at that without risking merge conflicts?

I haven't touched installation.rst yet, but I'm planning to eventually work through all of the docs. Feel free to take a swing if you like, and I'll wait to work on that particular page, or, if you prefer, open an issue with an outline of the changes you want and I'll incorporate them when I get to it!

> > I haven't updated the entirety of the release history to markdown ...
> 
> I probably wouldn't have noticed the formatting differences if you didn't point them out. Ideally these are fixed at some point but I'm fine leaving that for future work. I don't know of any solution easier than converting rST to Markdown by hand. The fact that MyST does a pretty good job of parsing rST side-by-side with Markdown is impressive enough.

Did some more investigation and its treating the whole eval-rst directive as being under the last heading. So the first rST heading is one level lower than that. I tried a few tricks but couldn't come up with anything that (a) looks better than the status quo and (b) will still work once we do another release.

Yay!

mattwthompson:
> can you confirm that your modified command still prevents the startup loop ... ?

Nope, I was doing things out of order and building the docs prior to running `sphinx-autobuild` works just as described. (`--ignore` flag does not change behavior, startup loop happens only if docs are not built prior.) I think the instructions are fine as is, I just didn't read them.

PR #873, opened 2021-03-12
Update README for Python 3.6 drop, developer docs for other updates

mattwthompson:
While working on #861 I noticed some outdated sections of the developer docs (`Currently, we use a service called Travis CI` 😬 ). I'm opening this as a separate PR so that ideas here don't block progress there.

I also set out to change anywhere an officially-looking statement of Python 3.6 support is found, and I could only find it in the README.

- [x] Supported Python versions updated in README
- [x]  OFF -> OpenFF (this is perhaps opinionated)_
- [x]  Travis changes
- [x] OpenEye license changes
- [ ] Tag issue being addressed
- [ ] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/tests)
- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable
- [ ] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [ ] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/master/docs/releasehistory.rst)

PR #876, opened 2021-03-12
Update non-CI actions for Python 3.6 drop

mattwthompson:
Another small update for the Python 3.6 drops. I want to keep this separate from #873 in order to keep that one focused on documentation (and if it ends up being closed due to duplication of effort). This removed the Python 3.6 timebomb and updates the installer (it is even used?) and cron jobs to only use our supported versions.

I also skimmed `/devtools/` for anything Python 3.6 sneaking in but found no conflicts.


.github/workflows/installer.yml
```diff
@@ -13,19 +13,14 @@ jobs:
...
       CI_OS: ${{ matrix.os }}
       PYVER: ${{ matrix.python-version }}
 
     steps:
     - uses: actions/checkout@v2
-      
-    - name: Wait for 2 hours if triggered by a release
-      if: (github.event_name == 'release') || (github.event_name == 'prerelease')
-      run:
-        sleep 7200
```
j-wags, line 28: Thanks for getting rid of this. That was always fun to run down!

j-wags (approved):
Looks great :-)

.github/workflows/installer.yml
mattwthompson: And to think we put that in with such good intentions 😢

Issue #771, opened 2020-11-17
Custom exceptions


mattwthompson:
**Is your feature request related to a problem? Please describe.**
The toolkit, and our software in general, should raise descriptive, custom exceptions instead of re-using built-in exceptions. 

**Describe the solution you'd like**
We should have our own exceptions that are fairly specific and provide detailed feedback to the user as to what went wrong and why, and possibly hints at how to resolve the error. For example, this behavior (currently in the code) checks all the boxes:

```python3
In [1]: from openforcefield.topology.molecule import Molecule
Warning: Unable to load toolkit 'AmberTools'.

In [2]: from openforcefield.utils.toolkits import BuiltInToolkitWrapper

In [3]: mol = Molecule.from_smiles('CCO')

In [4]: BuiltInToolkitWrapper().assign_partial_charges(molecule=mol, partial_charge_method='am1bcc')
---------------------------------------------------------------------------
ChargeMethodUnavailableError              Traceback (most recent call last)
<ipython-input-4-5e6d5f4e9e1e> in <module>
----> 1 BuiltInToolkitWrapper().assign_partial_charges(molecule=mol, partial_charge_method='am1bcc')

~/software/openforcefield/openforcefield/utils/toolkits.py in assign_partial_charges(self, molecule, partial_charge_method, use_conformers, strict_n_conformers)
    466         if partial_charge_method not in PARTIAL_CHARGE_METHODS:
    467             raise ChargeMethodUnavailableError(
--> 468                 f'Partial charge method "{partial_charge_method}"" is not supported by '
    469                 f"the Built-in toolkit. Available charge methods are "
    470                 f"{list(PARTIAL_CHARGE_METHODS.keys())}"

ChargeMethodUnavailableError: Partial charge method "am1bcc"" is not supported by the Built-in toolkit. Available charge methods are ['zeros', 'formal_charge']
```

Today, I went through most of the codebase and recorded when we raise built-in exceptions to get a better picture of how much would be changed by implementing this (nearly) everywhere in the codebase. Some of these (maybe a little less than a half?) probably fit well with existing exceptions, but many will require new exceptions. I have yet to fill out the last column.

| Context | Existing exception | Proposed change |
|---|---|---|
| In many places: | 
| * Method not implemented | `NotImplementedError`
| In `topology/topology.py`: | 
| * Aromaticity model not in list of known, allowed atomaticity models | `ValueError` |
| * Charge model not in ... | `ValueError` |
[...]

PR #878, opened 2021-03-15
Fix some small to-dos

lilyminium:
Hello, I was looking through this repo to get more familiar with the OpenFF infrastructure and saw some old, small todos / questions / typos I thought I could address quickly. Each is in a separate, easily revertable commit in case things have changed and they should no longer be done. They are small enough that I didn't think they were worth raising issues or updating the changelog. I hope this is helpful.

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -117,8 +117,8 @@ class Particle(Serializable):
         Set the particle's molecule pointer. Note that this will only work if the particle currently
         doesn't have a molecule
         """
-        # TODO: Add informative exception here
-        assert self._molecule is None
+        err = f"{type(self).__name__} already has an associated molecule"
+        assert self._molecule is None, err
         self._molecule = molecule
 
     @property
--- a/openff/toolkit/utils/toolkits.py
+++ b/openff/toolkit/utils/toolkits.py
@@ -3353,8 +3353,8 @@ class RDKitToolkitWrapper(ToolkitWrapper):
                 # set it back to zero
                 atom.SetAtomMapNum(0)
 
-        # TODO: I think UpdatePropertyCache(strict=True) is called anyway in Chem.SanitizeMol().
-        rdmol.UpdatePropertyCache(strict=False)
+        # Chem.SanitizeMol calls updatePropertyCache so we don't need to call it ourselves
+        # https://www.rdkit.org/docs/cppapi/namespaceRDKit_1_1MolOps.html#a8d831787aaf2d65d9920c37b25b476f5
         Chem.SanitizeMol(
             rdmol,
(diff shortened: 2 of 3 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -117,8 +117,8 @@ def molecule(self, molecule):
         Set the particle's molecule pointer. Note that this will only work if the particle currently
         doesn't have a molecule
         """
-        # TODO: Add informative exception here
-        assert self._molecule is None
+        err = f"{type(self).__name__} already has an associated molecule"
+        assert self._molecule is None, err
```
mattwthompson, line 121: Nit: This is an improvement (at least this `AssertionEror` now provides informative context) but I think the goal is to have both a custom message and exception i.e. `MoleculeAlreadyAssociatedWithSomethingError` #771

lilyminium: In that case I will remove it from this PR. That is a very specific error type, though. Does it need its own type?

mattwthompson: IMO the change can stay  (it's an improvement) but the TODO should remain as part of the pending exception rework moving away from built-in excepitons.

I think the specificity of each exception is somewhat a matter of taste and we haven't converged on how specific or general we want exceptions to be. Something more general that covers a few different errors would probably be better. I now see these lines are in the `Particle` class whereas I previously assumed this was a `TopologyMolecule`/`ReferenceMolecule` bookkeeping error, so that exception doesn't make sense anyway.

openff/toolkit/utils/toolkits.py
```diff
@@ -3353,8 +3353,8 @@ def from_smiles(
                 # set it back to zero
                 atom.SetAtomMapNum(0)
 
-        # TODO: I think UpdatePropertyCache(strict=True) is called anyway in Chem.SanitizeMol().
-        rdmol.UpdatePropertyCache(strict=False)
+        # Chem.SanitizeMol calls updatePropertyCache so we don't need to call it ourselves
+        # https://www.rdkit.org/docs/cppapi/namespaceRDKit_1_1MolOps.html#a8d831787aaf2d65d9920c37b25b476f5
```
j-wags, line 3357: Ah, molecule sanitization. Any method that touches sanitization is a huge tangle of complexity so I try not to touch it too often. So this change made me a bit nervous, but I looked down the code paths in RDKit and agree that running `updatePropertyCache(strict=True)` (which happens inside `Chem.SanitizeMol` ) will not be affected by previous results arising from running with `strict=False`. I also reran our slow tests and this seems to be fine. So I think this change is good to go.

j-wags (approved):
Looks good. I got a bit nervous about the part that touched molecule sanitization, but the change seems to be fine. We need to hack through the sanitization tangle eventually, sothis is a good start.

j-wags:
Merged. Thanks @lilyminium!

PR #879, opened 2021-03-16
Fix numpydoc warning previously missed

Yoshanuikabundi:
This fixes an incorrectly titled section of the `FrozenMolecule.from_qcschema()` method docstring from "Example" to "Examples". This would raise a warning in Numpydoc, but only when the API was generated, so I missed it previously. As a result of this oversight, the examples for the method were not rendered.


j-wags (approved):
LGTM!

PR #880, opened 2021-03-16
Dev doc review

Yoshanuikabundi:
Developer docs need some love. This is a work in progress and is based off PRs #871 and #873, so they should be merged first.


docs/developing.md
```diff
@@ -0,0 +1,309 @@
...
+      The default implementation of this function should suffice for most developers.
+    - `check_handler_compatibility`: Checks whether this ParameterHandler is "compatible" with another.
+      This function is used when a ForceField is attempted to be constructed from *multiple* SMIRNOFF data sources, and it is necessary to check that two sections with the same tagname can be combined in a sane way.
+      For example, if the user instructed two `vdW` sections to be read, but the sections defined different vdW potentials, then this function should raise an Exception indicating that there is no safe way to combine the parameters.
+      The default implementation of this function should suffice for most developers.
+    - `postprocess_system`: operates identically to `create_force`, but is run after each ParameterHandlers' `create_force` has already been called.
+      The default implementation of this method simply does nothing, and should suffice for most developers.
+
+[`ParameterAttribute`](openff.toolkit.typing.engines.smirnoff.parameters.ParameterAttribute)
+: A single parameter value with units and runtime validation.
```
j-wags, line 113:
```suggestion
: A single value with runtime validation. This validation is highly customizable, and may do things like allowing only certain values for a string or enforcing the correct units or dimensionality on the value (validation can be defined using arbitrary code)
```

docs/developing.md
```diff
@@ -0,0 +1,309 @@
...
+
+[`IndexedParameterAttribute`](openff.toolkit.typing.engines.smirnoff.parameters.IndexedParameterAttribute)
+: A variable number of parameter values with units and runtime validation.
+For example, dihedral torsions are often parameterized as the sum of several sine wave terms.
+Each of the parameters of the sine wave `k`, `periodicity`, and `phase` are implemented as `IndexedParameterAttribute`s.
```
j-wags, line 118:
```suggestion
[`IndexedParameterAttribute`](openff.toolkit.typing.engines.smirnoff.parameters.IndexedParameterAttribute)
: A ParameterAttribute that may appear with multiple integer values. The exact name of an IndexedParameterAttribute is NOT expected to appear verbatim in a OFFXML file, but instead should appear with a numerical integer suffix (for example the IndexedParameterAttribute `k` should only appear as, `k1`, `k2`... in an OFFXML). The current implementation requires this indexing to start at 1 and subsequent values be contiguous (no skipping numbers), but does not enforce an upper limit on the integer. 

Old version:
A variable number of parameter values with units and runtime validation.
For example, dihedral torsions are often parameterized as the sum of several sine wave terms.
Each of the parameters of the sine wave `k`, `periodicity`, and `phase` are implemented as `IndexedParameterAttribute`s.
```

docs/developing.md
```diff
@@ -0,0 +1,309 @@
...
+
+[`ParameterType`](openff.toolkit.typing.engines.smirnoff.parameters.ParameterType)
+: A base class for the parameters of a `ParameterHandler`.
+  Extend this alongside `ParameterHandler` to define and validate the parameters of a new force.
+  For example, the Lennard-Jones potential can be parameterized through either the length `ParameterAttribute` `sigma` or `r_min`, alongside the energy `ParameterAttribute` `epsilon`. Both options are handled through the [`vdWType`](openff.toolkit.typing.engines.smirnoff.parameters.vdWHandler.vdWType) class, a subclass of `ParameterType`.
+  This is distinct from the [`vdWHandler`](openff.toolkit.typing.engines.smirnoff.parameters.vdWHandler) class, which includes other `ParameterAttribute`s that apply to all atoms in the system.
```
j-wags, line 124:
```suggestion
[`ParameterType`](openff.toolkit.typing.engines.smirnoff.parameters.ParameterType)
: A base class for the SMIRKS-based parameters of a `ParameterHandler`.
  Extend this alongside `ParameterHandler` to define and validate the SMIRKS-based (/ForceField) parameters of a new force.
  This is analagous to ParmEd's XType classes, like [BondType](https://parmed.github.io/ParmEd/html/api/parmed/parmed.html?highlight=bondtype#parmed.BondType). 
  
  For example, the Lennard-Jones potential can be parameterized through either the size `ParameterAttribute` `sigma` or `r_min`, alongside the energy `ParameterAttribute` `epsilon`. Both options are handled through the [`vdWType`](openff.toolkit.typing.engines.smirnoff.parameters.vdWHandler.vdWType) class, a subclass of `ParameterType`.
  
  
  
  ~This is distinct from the [`vdWHandler`](openff.toolkit.typing.engines.smirnoff.parameters.vdWHandler) class, which includes other `ParameterAttribute`s that apply to all atoms in the system.~ Somewhere we should make the statement that BOTH ParameterTypes AND ParameterHandlers should be defining ParameterAttributes
```

docs/developing.md
Yoshanuikabundi: Incorporated these suggestions in dbcc59a

docs/developing.md
Yoshanuikabundi: Incorporated these suggestions in dbcc59a

docs/developing.md
Yoshanuikabundi: Incorporated these suggestions in dbcc59a

docs/developing.md
```diff
@@ -0,0 +1,347 @@
...
+A single value that can be validated at runtime.
+
+`ParameterAttribute`s can be instantiated as Python class or instance members to define the kinds of value that a particular parameter can take.
+They are used in both `ParameterHandler`s and `ParameterType`s.
+The sorts of values a ParameterAttribute can take on are restricted by runtime validation.
+This validation is highly customizable, and may do things like allowing only certain values for a string or enforcing the correct units or array dimensions on the value; in fact, the validation can be defined using arbitrary code.
+The name of a ParameterAttribute should correspond exactly to the corresponding attribute in an OFFXML file.
+
+#### [`IndexedParameterAttribute`](openff.toolkit.typing.engines.smirnoff.parameters.IndexedParameterAttribute)
+A ParameterAttribute with a sequence of values, rather than just one. Each value in the sequence is indexed by an integer.
```
j-wags, line 123: I forgot that I could click on these and go to a fuller description in the API docs. Maybe let's also put an inline link to the API docs, so this reads "An `IndexedParameterAttribute`(link) is a ParameterAttribute..."

j-wags (approved):
Looks great. Thank @Yoshanuikabundi, and apologies that I didn't get a change to fill out the sections below (eg `Molecule.to/from_X`). I can do that later, no need to hold up this PR for that!

Yoshanuikabundi:
In our last meeting, @j-wags and I discussed the problem of pluralising type names in code spans. For instance, in this PR, the plural form of `ParameterAttribute` was previously rendered `ParameterAttributes` or `ParameterAttribute`s. Across the codebase, both these forms are used.

The former (`ParameterAttributes`) is undesirable because it is not the verbatim name of anything, but it is presented in a code span. The latter (`ParameterAttribute`s) seems more correct, especially in raw text form(`` `ParameterAttribute`s ``), but is difficult to read and ugly when rendered and so is also undesirable.

Commit 54f4502 splits this arrow by removing plurals of verbatim text altogether. Instead, it chooses phrasings that remove the pluralisation altogether without changing the meaning, or that place the plural on a different word in the sentence. Several different examples of this are in commit 54f4502, and the general principle is taken from [this StackOverflow answer](https://stackoverflow.com/a/41388152). This method tends to read more formally and less conversationally than the alternative, but avoids the downsides of the above methods.

docs/developing.md
Yoshanuikabundi: Fixed in 64c283d

PR #881, opened 2021-03-25
[WIP] Improve protein parameterization runtime

j-wags:
Adds `cachetools` as a requirement

- [x] (temporary, should be replaced by a more general solution) Adds a lightweight, index-sensitive hash method for molecules (`FrozenMolecule.ordered_connection_table_hash`)
- [x] Adds caching for `to_openeye` and `to_rdkit` using `cachelib`
- [x] Adds caching for `Atom.molecule_atom_index`, since the data structure makes self-index lookup an O(n_atoms_in_molecule) operation. `Molecule._invalidate_cached_properties` now recursively deletes this cached property on the molecule's Atoms. 
- [x] Removes `@classmethod` designation from `OpenEyeToolkitWrapper.to_openeye` and `RDKitToolkitWrapper.to_rdkit` (I don't see a need for them to be classmethods, this was probably from before when I knew what I was doing)
- [x] Makes `Molecule.to_openeye` and `to_rdkit` accept a toolkit registry kwarg. It's unclear why they didn't before (a user could reasonably want to divert `Molecule.to_rdkit` through a custom ToolkitWrapper)
    - [ ] appropriate cached molecule hash invalidation (eg when an atom/bond is added)
    - [ ] appropriate cached atom index invalidation (necessary? Atoms can only be appended at present)
    - [ ] cache for `to_openeye` / `to_rdkit` actually detects when a aromaticity model is requested
    - [ ] test that operations on cached outputs don't mutate the object in the cache

[protein_param_test.tar.gz](https://github.com/openforcefield/openff-toolkit/files/6201573/protein_param_test.tar.gz)

`master` branch runtime on the above is ~40 minutes for both OE and RDKit.

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -38,6 +38,7 @@ import warnings
 from collections import OrderedDict
 from copy import deepcopy
 from typing import Optional, Union
+from cached_property import cached_property
 
 import networkx as nx
 import numpy as np
@@ -459,7 +460,7 @@ class Atom(Particle):
         # for vsite in self._vsites:
         #    yield vsite
 
-    @property
+    @cached_property
     def molecule_atom_index(self):
         """
         The index of this Atom within the the list of atoms in ``Molecules``.
@@ -2040,6 +2041,20 @@ class FrozenMolecule(Serializable):
         """
         return hash(self.to_smiles())
 
+    #@cached_property
+    def ordered_connection_table_hash(self):
+        if self._ordered_connection_table_hash is not None:
+            return self._ordered_connection_table_hash
+        
+        id = ''
+        for atom in self.atoms:
+            id += f'{atom.element.symbol}_{atom.formal_charge}_{atom.stereochemistry}__'
+        for bond in self.bonds:
+            id += f'{bond.bond_order}_{bond.stereochemistry}_{bond.atom1_index}_{bond.atom2_index}__'
+        #return hash(id)
+        self._ordered_connection_table_hash = hash(id)
+        return self._ordered_connection_table_hash
+
     @classmethod
     def from_dict(cls, molecule_dict):
         """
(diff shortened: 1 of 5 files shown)


j-wags:
As of this commit, runtime is about 45 seconds with rdkit (26 of which are system creation) and 220 seconds with OE (197 of which are system creation). I'll focus back on OE improvements when I get back to this.

openff/toolkit/topology/molecule.py
```diff
@@ -2040,6 +2041,20 @@ def __hash__(self):
...
+            return self._ordered_connection_table_hash
+        
+        id = ''
+        for atom in self.atoms:
+            id += f'{atom.element.symbol}_{atom.formal_charge}_{atom.stereochemistry}__'
+        for bond in self.bonds:
+            id += f'{bond.bond_order}_{bond.stereochemistry}_{bond.atom1_index}_{bond.atom2_index}__'
+        #return hash(id)
+        self._ordered_connection_table_hash = hash(id)
+        return self._ordered_connection_table_hash
```
j-wags, line 2056: I need to think more about if this is totally suitable for caching outputs of `to_openeye` and `to_rdkit`. It captures the important attributes of the _connection table_ (so it's safe for SMARTS matching), but it will (incorrectly) return a false positive for molecules with the same connection table but different conformers, partial charges, or names. 

I could potentially either
* Separate the connection table conversion lines in `to_openeye` (ie the atom/bond addition) into a private method, and only apply the caching to that. 
* Expand the cache key to include more information (like molecule name and partial charges), but at that point this will be about as slow as `to_dict`
* Avoid caching altogether by making a multi-SMARTS matching method (instead of running find_smarts_matches repeatedly on one SMARTS each time)

j-wags:
#950 migrates these changes into the `topology-biopolymer-refactor` branch. This implementation is incomplete at this time, so the to-do list from the title post will also be migrated.

PR #888, opened 2021-03-29
Examples helper script

Yoshanuikabundi:
In this PR I'm developing a simple CLI app, installed with the toolkit, to aid in installing and running examples

- [x] Make available as executable at install time with setuptools
- [x] List available examples at command line
    - [ ] Provide blurb for each example
- [x] CLI to copy example files to a convenient user-writeable directory
    - [x] Identify notebook file
    - [x] Manage dependencies for notebooks
        - At this point I think I'll assume the user has conda installed and use that.
        - I can create a new environment for each example (in the example folder), switch to it, run the notebook, and then switch back when the user closes the notebook. 
        - Maybe one day we can use [`__pypackages__`](https://www.python.org/dev/peps/pep-0582/).
        - I think the simplest way to specify dependencies will be to create a general examples `enviroment.yaml`, and then allow environments to specify additional packages in their own `environment.yaml` files.
- [x] Write docstrings
- [x] Update [documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs)

To run the script, either do a `python /path/to/toolkit/repo/utilities/examples_helper/cli.py` or run setuptools again with the `pip install -e .` command described in the developer docs and then use `openff-examples`. The script is self-documenting using the Click library (which I think will need to become a dependency...?).

New dependencies are just Click.


Yoshanuikabundi:
@j-wags Today in our meeting we talked about moving this into the openff/toolkits/utils directory. Unfortunately, the `utils` module is quite noisy to import, which happens implicitly on `import openff.toolkit.utils.example_helper.bla`, and doing this gives this spiel at the start of every helper run if an OpenEye license is unavailable:

```
LICENSE: Could not open license file "oe_license.txt" in local directory
LICENSE: N.B. OE_LICENSE environment variable is not set
LICENSE: N.B. OE_DIR environment variable is not set
LICENSE: No product keys!
LICENSE: No product keys!
LICENSE: No product keys!
LICENSE: No product keys!
Warning: Unable to load toolkit 'OpenEye Toolkit'. The Open Force Field Toolkit does not require the OpenEye Toolkits, and can use RDKit/AmberTools instead. However, if you have a valid license for the OpenEye Toolkits, consider installing them for faster performance and additional file format support: https://docs.eyesopen.com/toolkits/python/quickstart-python/linuxosx.html OpenEye offers free Toolkit licenses for academics: https://www.eyesopen.com/academic-licensing
```

This is annoying at best, and at worst will encourage users to ignore all the (hopefully useful!) output the helper gives. So I've moved the examples_helper folder into openff/toolkit directly for the time being, which solves the issue as the `openff.toolkit.utils.toolkits` module is not imported when the helper is run. Maybe we can discuss some way to reduce the impact of these warnings at some point, as I'm guessing most users will not have, need, or want an OpenEye license. It's easy to move the helper around so we can pick a new spot at your leisure.

I guess there might also be a way to tell Python not to import parent modules, but this is probably just as ick as the previous runtime editing of `sys.path`.

Yoshanuikabundi:
In f59a791 I moved the non-deprecated notebooks from using comments to control `nbval` to using tags, which are a lot less intrusive - you need to have the tags toolbar on in notebook, and its off by default. See the [docs for nbval](https://nbval.readthedocs.io/en/latest/#Using-tags-instead-of-comments). The relevant cells are skipped in the tests, as they were previously.

I think, with this capacity to send instructions to `nbval` without disturbing the user experience, we should look at switching `nbval-lax` to `nbval` and using CI to make sure that notebook output doesn't change. I can do this in future PRs as I go through the examples to spruce them up.

I'm also a little uneasy about using the same CI to test the notebooks and the notebook example thingy. With the new examples `environment.yml` it would be easy to test notebooks separately to `offtk-examples`, and I can see a bug being introduced to `offtk-examples` that breaks the tests without causing them to fail. I think ideally the examples would be tested independently of the tool, and I know how to test the examples in this case, but not sure how to test the tool (I am very bad at testing)

.github/workflows/examples.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Run example scripts without OE
+        shell: bash -l {0}
+        run: |
+          if [[ "$RDKIT" == false ]]; then
+            PYTEST_ARGS+=" --ignore=examples/check_dataset_parameter_coverage"
+            PYTEST_ARGS+=" --ignore=examples/QCArchive_interface"
+          fi
+          pytest $PYTEST_ARGS openff/toolkit/tests/test_examples.py
+
+      - name: Run example notebooks
```
j-wags, line 88: Per our discussion today, let's add some sort of test for `openff-examples list`, maybe comparing to the output of `ls examples/ --ignore="deprecated"`

devtools/conda-envs/examples.yaml
```diff
@@ -0,0 +1,26 @@
+name: offtk-examples
```
j-wags, line 1: If possible, let's see if we can piggyback off of `rdkit.yaml` and `openeye.yaml`, since it's easy for yamls in this directory to fall out of date in a fail-deadly way, so it's good to have fewer of them.

.github/workflows/examples.yml
```diff
@@ -0,0 +1,106 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+        cfg:
+          - python-version: 3.7
+
```
j-wags, line 31: Per our discussion today, let's expand this testing matrix to test {rdkit-only, openeye-only}. The way we do that in the standard CI is [here](https://github.com/openforcefield/openff-toolkit/blob/7c3dfbaed9cda911b70065626414c7e382bf9ff9/.github/workflows/CI.yml#L73-L88)

openff/toolkit/examples_helper/example.py
```diff
@@ -0,0 +1,88 @@
...
+            return Example(EXAMPLES_DIR / value)
+        except ValueError:
+            pass
+
+        return self.fail(
+            f"Could not find example {value}. It should be a folder "
+            f"containing exactly one Jupyter notebook.",
+            param,
+            ctx,
+        )
```
j-wags, line 88: May want to double check that this is doing what we think it should -- [This example](https://github.com/openforcefield/openff-toolkit/tree/master/examples/using_smirnoff_with_amber_protein_forcefield) has two notebooks

examples/visualization/visualization.ipynb
```diff
@@ -261,7 +267,11 @@
   {
    "cell_type": "code",
    "execution_count": 9,
-   "metadata": {},
+   "metadata": {
+    "tags": [
+     "nbval-skip"
+    ]
+   },
```
j-wags, line 274: Maybe in this PR, maybe in a different one, could you add some sort of early "example standards" to the developer docs that links to the full docs for invisible nbval tags, and recommends when to use them?

.github/workflows/examples.yml
Yoshanuikabundi: Done in a4ac60b

openff/toolkit/examples_helper/example.py
Yoshanuikabundi: Fixed in 6a70698

Yoshanuikabundi:
OK.

I think I've had every single possible problem so far on this PR.

While trying to get the `openff-examples` entry point to run reliably in CI, I tried installing installing it as a source distribution, rather than in development mode, and discovered that the examples were not installed. This is a big problem as it is the way most users will access the toolkit. Through a long process of reading many out of date Stack Overflow questions, I eventually found the [appropriate documentation](https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html):

> Instead, the PyPA recommends that any data files you wish to be accessible at run time be included in the package.

So I followed this recommendation, which as far as I understand means moving the `examples` folder into the `openff` folder. This seems... less than ideal. I also had to copy the Conda environment file `rdkit.yaml` into the examples folder so it is accessible to the script.

Things I discovered doing this:
 - The `isort` and `black` configurations we had were incompatible. It was possible (one of the deprecated examples managed it) to get into a loop where formatting with one would break the other. This is fixed in ccabc9e.
 - While `ls` on Linux has the wonderful `--ignore` switch, on OSX it does not, and there is no equivalent switch. I've hacked together a `find` command that works (hopefully, test hasn't finished yet)
 - I hate `find`
 - `setuptools` is both confusing and not the same thing as `distutils`
 - Expanding the matrix to include OpenEye was easy. 
 - Expanding the matrix to include OSX was hard.
 - We'd left the examples tests in the original CI (fixed in 72c91b3)

This PR seems to basically work at this point, but is not really in an acceptable state in terms of what I've had to finagle to get it there. To proceed, I think we should...
- Open and merge a new PR that does not include the `openff-examples` tool. This PR would just be about getting the examples all running with the proper dependencies on CI. It would not require moving the examples directory. Most of this work is already present in this PR so I should be able to do this in a few hours.
- Plan out how the `openff-examples` tool should get access to the examples and implement it in a new PR. Most of the code can come from this PR, but I think there are plenty reasons not to want the examples in the package. Ideas I've had are:
    - Download the requested examples on demand from GitHub, taking package version into account. This would add complexity to the `openff-examples` tool
    - Add a new package, optionally installed with the toolkit, that includes the examples and the `openff-examples` tool. This package could even depend on the example dependencies. So there would then be three packages on Conda Forge: `openff-toolkit-base`, `openff-toolkit`, and `openff-toolkit-examples`. I think this is the cleanest approach, and it would let us take all the package management code out of the tool while still providing a super easy approach to having the example's dependencies installed. The toolkit and its examples can be kept in sync by having a new examples release alongside each toolkit release, and depending on the exact correct toolkit version. I think this can even be done in the existing Forge feedstock.
- Finally, change the CI to use the `openff-examples` tool internally (if we want - this would mostly serve as a test of the `openff-toolkit-examples package`

j-wags:
Based on our check-in today, let's split things into three stages of PRs:

* The easily-fixable stuff we found in this PR (examples that aren't being tested, CI running unit tests and example tests in serial, and isort/black conflicts) -- Let's fix these ASAP
* The examples packaging/structure issue -- Let's see about getting a [dedicated data package just for examples](https://github.com/conda-forge/openff-toolkit-feedstock/pull/9), and if that doesn't work, let's take a break on the "machinery" side of example deployment. 
* IF the dedicated examples data package works, let's revisit strategy for example-deployment machinery, since a dedicated conda package would gracefully handle dependency updates, so we don't need to roll that ourselves :-)

Full meeting notes here: https://openforcefield.atlassian.net/wiki/spaces/IN/pages/1615790099/2021-04-21+Mitchell+Wagner+Check+in

PR #891, opened 2021-03-30
Fix #890

### LINKED ISSUE #890: from_openeye mutates input molecule
**Describe the bug**
The `OpenEyeToolkitWrapper` mutates the input molecule rather than operating on a copy. This can lead to subtle errors when the input molecule is used by the user for other purposes after calling this function.

**To Reproduce**

The issue can be seen be inspecting the `from_openeye` method [here](https://github.com/openforcefield/openff-toolkit/blob/3c05a06d81ec0ccf36c82df966db90b6caea1559/openff/toolkit/utils/toolkits.py#L1251-L1306).

**Output**

 - Operating system

**Additional context**

SimonBoothroyd:
Fixes #890

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -10,7 +10,7 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ### Bugfixes
 
-- [PR #891](https://github.com/openforcefield/openforcefield/pull/891): Calls to `Molecule.from_openeye` no longer mutates the input OE molecule.
+- [PR #891](https://github.com/openforcefield/openforcefield/pull/891): Calls to `Molecule.from_openeye` no longer mutate the input OE molecule.
 
 ### Improved documentation and warnings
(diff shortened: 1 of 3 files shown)


docs/releasehistory.md
```diff
@@ -8,6 +8,10 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ## Changes since last release
 
+### Bugfixes
+
+- [PR #891](https://github.com/openforcefield/openforcefield/pull/891): Calls to `Molecule.from_openeye` no longer mutates the input OE molecule.
```
mattwthompson, line 13:
```suggestion
- [PR #891](https://github.com/openforcefield/openforcefield/pull/891): Calls to `Molecule.from_openeye` no longer mutate the input OE molecule.
```

mattwthompson (approved):
LGTM; I checked that the test fails on the current release. I assume that the `OEMol()` constructor safely creates a copy and there are no gotchas to be worried about?

j-wags (approved):
Also agree. This looks great. Thanks, Simon!

> I assume that the OEMol() constructor safely creates a copy and there are no gotchas to be worried about?

I believe that's correct, or at least that's how I've used it dozens of times in this codebase :-)

PR #895, opened 2021-03-30
Updates for RDKit 2021.03

j-wags:


j-wags:
The to/from inchi test failures seem to be complaining about nitrogen stereochemistry. Neither of the two failing cases have tetravalent nitrogen, which is confusing, since generally RDKit would only care about tetravalent N stereo (it doesn't care about trivalent nitrogen). 

Molecule 228 fails because of atom center stereo. Maybe something's special and stereogenic about a nitrogen somewhere in the fused rings in this mol?

image

Molecule 272 fails because of nitrogen double bond stereo. This is fine, I see a few N double bonds in this molecule and the confusion over how those are constrained by rings makes me think that RDKit is reasonable in flagging these. 
image

PR #909, opened 2021-04-20
Compute virtual site positions

### LINKED ISSUE #910: Compute virtual site positions from within TK
Currently the TK relies on an OpenMM system to calculate the spatial positions of virtual sites. It would be nice to have an API point that allows calculation these positions entirely within the toolkit, so downstream would not need to depend on OpenMM for this simple operation. For example, `openff-recharge`, in order to use virtual sites, needs to construct a design matrix of each BCC, which depends on particle pairwise distances, so having the ability to query a virtual sites position based on a conformer (or any set of given positions) would be great to have.

Currently proposed implemented in https://github.com/openforcefield/openff-toolkit/pull/909

Any discussion regarding API points, how to implement, etc, is welcome.

@SimonBoothroyd

trevorgokey:

This PR will allow the ability to compute the positions of virtual site particles within the toolkit, i.e. without going through OpenMM.

The main functionality implemented in this PR is
- `FrozenMolecule.compute_virtual_site_positions_from_conformer(index: int) -> numpy.ndarray * [Angstrom]`
- `VirtualSite.compute_virtual_site_positions_from_conformer(index: int) -> numpy.ndarray * [Angstrom]`
- `VirtualParticle.compute_position_from_conformer(index: int) -> numpy.ndarray * [Angstrom]`
- `FrozenMolecule.compute_virtual_site_positions_from_atom_positions(positions: numpy.ndarray) -> numpy.ndarray * [Angstrom]`
- `VirtualSite.compute_virtual_site_positions_from_atom_positions(positions: numpy.ndarray) -> numpy.ndarray * [Angstrom]`
- `VirtualParticle.compute_position_from_atom_positions(positions: numpy.ndarray) -> numpy.ndarray * [Angstrom]`

where positions is assumed to have shape `(N, 3)` where `N` is the number of atoms. The return shape is `(M, 3)` where `M` is the number of virtual particles.

This PR is intended to facilitate virtual site support in `openff-recharge` (https://github.com/openforcefield/openff-recharge/pull/63). Fixes #910

#### Update:
This PR also introduces a few side effects aimed at improving virtual site functionality and fixing bugs. Since this PR depends on these fixes, I figured to just add them here rather than create a new PR and have this code depend on it.

**Bugfix 1**: When adding a virtual site via the `Molecule` API, a call to `create_openmm_system` calls the `vdW` and `NonBonded` handlers, which adds empty particles to the system/force. Since VirtualSites manage particle creation internally, the particle count becomes inconsistent. The change implemented here will cause only atom particles to be added in the base `ParameterHandler.create_force`, and delegates adding virtual particles to the system/force to `VirtualSiteParameterHandler`. I suppose this was not an issue when creating virtual sites directly from the XML, since the topology only had atoms when the vdW and NonBonded handlers were called (Vsites have them in their dependencies).

**Bugfix 2**: When adding a virtual site via the `Molecule` API, the parameter terms (sigma, epsilon) do not use the validation of the parameter handlers, and default to None, causing issue downstream when accessing e.g. the OpenMM API. This changes the defaults to be 0 with appropriate units, such that the default settings for virtual sites produce particles that exert no force and do not change atomic partial charges. It appears that there is some desire to handle epsilon=0 as a special case, but since the OpenMM addParticle API point requires these, I am lead to believe None is not a good default.

#### Non-obvious behavior
Creating an OpenMM system from virtual sites created solely from the `Molecule` API is a bit troublesome, since there are certain patterns and switches that affect the behavior that only the parameter handler knows about (exclusions in particular). Additionally, adding a virtual site via the `Molecule` API with and then creating an OpenMM system using a force field without a `VirtualSites` tag gives rise to a state where the virtual particles are never created since the VS parameter handler does not exist. This could have been alleviated by allowing the vdW handler to create particles for both atoms and virtual sites (before adding bugfix 1), but then the vdW handler would need to then relay the particles off to some virtual site machinery in order to process the virtual sites correctly (constructing the appropriate OpenMM objects, etc). Therefore, **when creating an OpenMM system with virtual sites added via the `Molecule` API, one must ensure that applied force field has an empty VirtualSites tag in order for the parameter handler to be called and construct the virtual particles.** This behavior is already documented in `virtualsites.md` as a somewhat innocent bullet point. Since this behavior was already predicted and documented, bugfix 1 makes sense.

@SimonBoothroyd


openff/toolkit/tests/test_molecule.py
```diff
@@ -2870,6 +2870,192 @@ def test_add_trivalent_lone_pair_virtual_site(self, molecule):
...
+        atom3 = molecule.atoms[2]
+        distance = 0.2 * unit.angstrom
+        out_of_plane_angle = 90 * unit.degree
+        vsite1_index = molecule.add_divalent_lone_pair_virtual_site(
+            [atom1, atom2, atom3],
+            distance,
+            out_of_plane_angle=out_of_plane_angle,
+            symmetric=False,
+        )
+        expected = np.array([[0.0, 0.0, -0.2]])
```
j-wags, line 2999: All of the tests only have one orientation / particle. We should add/modify a test that expects multiple particles to ensure that the positions are returned correctly, and document whether a particular order is expected.

openff/toolkit/topology/molecule.py
```diff
@@ -3279,7 +3795,12 @@ def _add_bond_charge_virtual_site(self, atoms, distance, **kwargs):
         vsite = BondChargeVirtualSite(atoms, distance, **kwargs)
 
         self._add_virtual_site(vsite, replace=replace)
-        self._invalidate_cached_properties()
+
+        # If we invalidate, we can never generate virtual sites after
+        # conformers are generated. This seems off. Since virtual sites
+        # do not seem to affect anything cached, we keep it.
+        # self._invalidate_cached_properties()
+
```
j-wags, line 3803: Agreed. And I'm moving toward thinking that "cached properties" are strictly things that affect the connectivity table, so this change is fine and we can remove this comment. (same goes for the copies of this comment elsewhere)

j-wags (commented):
I don't have time to finish this review, but I'm adding my partial progress in the subcomments on this post. 

Also,

> when creating an OpenMM system with virtual sites added via the Molecule API, one must ensure that applied force field has an empty VirtualSites tag in order for the parameter handler to be called and construct the virtual particles.

VirtualSites are by far our hardest feature to support, extend, and (in this case) review because of this "users can either define them in the force field **OR** by using the Molecule API" approach. 

I don't see much of a benefit to allowing them to be created using the Molecule API, and the cost of supporting for them would become much cheaper if we just stated a plan to deprecate things like `Molecule.add_X_virtual_site`. 

@jthorton @SimonBoothroyd @trevorgokey -- Do you have any current or future need for defining vsites using direct calls to the `Molecule` class? Or would it be OK to announce deprecation of vsites from the Molecule API and only support adding them using force fields?

SimonBoothroyd:
> @j-wags I think removing this feature would make a lot of sense, from my view vsites are part of the forcefield and so should be treated in the same way as the other forcefield parameters and encoded in the forcefield.

+ 1. Having only the FF / v-site handler API expose a way to get out the v-sites sounds good with me.

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1302,8 +1302,12 @@ def create_openmm_system(self, topology, **kwargs):
...
 
-        # Add particles (both atoms and virtual sites) with appropriate masses
-        for atom in topology.topology_particles:
+        # Add atom particles with appropriate masses
+        # Virtual site particle creation is handled in the parameter handler
+        # create_force call
+        # This means that even though virtual sites may have been created via
+        # the molecule API, an empty VirtualSites tag must exist in the FF
+        for atom in topology.topology_atoms:
             system.addParticle(atom.atom.mass)
```
j-wags, line 1311: >  I suppose this was not an issue when creating virtual sites directly from the XML, since the topology only had atoms when the vdW and NonBonded handlers were called (Vsites have them in their dependencies).

Ugggh this took me so many read-throughs to understand, but now it makes perfect sense.

openff/toolkit/topology/molecule.py
```diff
@@ -3402,7 +3933,12 @@ def _add_trivalent_lone_pair_virtual_site(self, atoms, distance, **kwargs):
         vsite = TrivalentLonePairVirtualSite(atoms, distance, **kwargs)
 
         self._add_virtual_site(vsite, replace=replace)
-        self._invalidate_cached_properties()
+
+        # If we invalidate, we can never generate virtual sites after
+        # conformers are generated. This seems off. Since virtual sites
+        # do not seem to affect anything cached, we keep it.
+        # self._invalidate_cached_properties()
+
```
j-wags, line 3941:
```suggestion
```

openff/toolkit/topology/molecule.py
```diff
@@ -3369,7 +3898,12 @@ def _add_divalent_lone_pair_virtual_site(
         )
 
         self._add_virtual_site(vsite, replace=replace)
-        self._invalidate_cached_properties()
+
+        # If we invalidate, we can never generate virtual sites after
+        # conformers are generated. This seems off. Since virtual sites
+        # do not seem to affect anything cached, we keep it.
+        # self._invalidate_cached_properties()
+
```
j-wags, line 3906:
```suggestion
```

openff/toolkit/topology/molecule.py
```diff
@@ -3324,7 +3848,12 @@ def _add_monovalent_lone_pair_virtual_site(
...
```
j-wags, line 3856:
```suggestion
```

openff/toolkit/topology/molecule.py
j-wags:
```suggestion
```

j-wags (changes requested):
Alright -- This looks to be in pretty good shape @trevorgokey! Thanks for your patience on this. 

To get this merge-ready, could you: 
- [ ] Modify a test to handle two-particle sites
- [ ] Accept the cleanups suggested in the code change comments
- [ ] Delete the LonePairVSite class

I'll approve afterwards, and we can move towards the next release!

trevorgokey:
Thanks for looking this over @j-wags. I have gone through and made the changes. Let me know if there is anything else to do.

trevorgokey:
Going through this, when writing tests, made me realize that there might be value in keeping the molecule vsite API now that computing the positions are available. I agree that parameterizing these particles should be strictly left to the FF, but I somewhat like the idea of using the molecule API to define a vsite and the have the ability to calculate the positions. I can see cases where this could be useful to define dummy atoms, which have positions in space but no parameters. I could also see it useful for someone who uses the TK to load molecules, and just wants to add the particles with no parameters to save to a file with the extra particles e.g. xyz, sdf, or pdb.

In other words, it is definitely more cumbersome to try to define dummy atoms by using a FF handler. It is robust but not user friendly. Using the molecule API, like is done in the tests, is a really nice way to quickly get a dummy atom position quickly. We could have the case where all virtual particles are wiped if an FF is applied. Rewriting the tests without the molecule API would hike up the complexity quite a bit.

Just my 2 cents. Using vsites as a way to define dummy atoms is not the intended use, but is nice and an unintended feature enabled by our nice architecture.

j-wags (approved):
Awesome work. Thanks @trevorgokey!

> our nice architecture.

I wouldn't go that far :-) 

> there might be value in keeping the molecule vsite API now that computing the positions are available

Yeah, the decision to remove the API points isn't final. There are also intermediate steps that we could consider, like changing the method names to be private.

Anyway, this looks great. Merging now -- Thanks again!

PR #911, opened 2021-04-22
Add parallel CI for examples, including their own deps

Yoshanuikabundi:
In #888, I explored ways to improve the user experience of using examples. So began a journey across workflows, through CLI development, and into the depths of package management. Much was discovered, and this PR applies some of the lessons thusly learned.

Previously, examples were tested as a part of the general testing CI. This elongated execution times of CI. In addition, examples were tested with only the dependencies of the toolkit (and the testing infrastructure), meaning that notebooks with external dependencies were skipped.

This PR adds a new CI workflow that tests all of the examples, in parallel with other tests, with their dependencies. It also makes the dependencies of the examples available to users as a Conda `environment.yaml` file in the examples directory. In addition, it converts several comments that are used to control testing of individual notebook cells into metadata tags. This is described in the [`nbval` docs](https://nbval.readthedocs.io/en/latest/#Using-tags-instead-of-comments).

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -112,7 +112,7 @@ jobs:
         run: |
           EXAMPLES="$(find examples -mindepth 1 -maxdepth 1 -type d ! -name deprecated)"
           for example in $EXAMPLES; do
-            pytest $NB_ARGS $examples
+            pytest $NB_ARGS $example
           done
 
 
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -117,6 +117,11 @@ jobs:
       - name: Run example notebooks
         shell: bash -l {0}
         run: |
+          if [[ "$RDKIT" == false ]]; then
+            NB_ARGS+=" --ignore=examples/QCArchive_interface"
+            NB_ARGS+=" --ignore=examples/check_dataset_parameter_coverage"
+            NB_ARGS+=" --ignore=examples/conformer_energies"
+          fi
           pytest $NB_ARGS examples
 
       - name: Codecov
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -110,7 +110,10 @@ jobs:
       - name: Run example notebooks
         shell: bash -l {0}
         run: |
-          pytest $NB_ARGS examples/
+          EXAMPLES="$(find examples -mindepth 1 -maxdepth 1 -type d ! -name deprecated)"
+          for example in $EXAMPLES; do
+            pytest $NB_ARGS $examples
+          done
 
 
       # - name: Check openff-examples list is complete and correct
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -71,7 +71,7 @@ jobs:
       - name: Install additional example dependencies
         shell: bash -l {0}
         run: |
-          conda env update --file examples/environment.yaml
+          conda env update --file examples/environment.yaml --name test
 
       - name: Additional info about the build
         shell: bash -l {0}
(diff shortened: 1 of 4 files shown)


Yoshanuikabundi:
@j-wags, the only problem that I can see so far with this PR is that RDkit ends up installed even in the OpenEye runs. This is because it is a dependency of, believe it or not, `openff-toolkit`, which is installed from conda-forge as a dependency of `openmmforcefields`, which is required for one of the notebooks. I spotted this when I noticed that one of the notebooks directly calls RDkit, and should fail if its not present. You can see the solved dependency tree in the [OSX tests](https://github.com/openforcefield/openff-toolkit/pull/911/checks?check_run_id=2410748494) of e15e352. See the "Conda tree information" job toward the end of the log.

We only need `openmmforcefields` for the actual force field XML files, not the code. The example that uses it is [using_smirnoff_with_amber_protein_forcefield](https://github.com/openforcefield/openff-toolkit/blob/master/examples/using_smirnoff_with_amber_protein_forcefield/BRD4_inhibitor_benchmark.ipynb). Options might be:
 - Remove the `openmmforcefields` dependency and copy the relevant force field into the examples folder (I would do this without consulting you, only the `openmmforcefields` package is commented out all over the place and I'm not sure of the history behind it and I'm worried there might be a good reason we don't have the amber force field XMLs in our repo, even as an example)
 - Manually force-remove the `rdkit` dependency after conda has finished installing everything
 - Ignore it (and probably explicitly put `rdkit` in the examples `environment.yaml` to prevent breakage in the future)

If we remove the `rdkit` dependency, we need to decide what to do with the example(s) that will be broken by that change. Simplest fix would be just to ignore it for the OpenEye only run, as we did previously. Better fix might be to see if we can remove the requirement in the example, but this could be punted on til later.

Other than that, this PR is ready for review.

.github/workflows/examples.yml
```diff
@@ -0,0 +1,128 @@
...
+        run: |
+          pytest $NB_ARGS examples
+
+      - name: Conda tree information
+        shell: bash -l {0}
+        run: |
+          conda install -c conda-forge conda-tree
+          conda tree deptree
+          conda tree leaves
+          conda tree whoneeds rdkit
```
j-wags, line 121: This is slick. I'm going to use `conda tree` for debugging from now on!

.github/workflows/examples.yml
```diff
@@ -0,0 +1,128 @@
...
+      PYTEST_ARGS: -r fE --tb=short --cov=openff --cov-config=setup.cfg --cov-append --cov-report=xml
+      NB_ARGS: -v --nbval-lax --ignore=examples/deprecated
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - uses: conda-incubator/setup-miniconda@v2
+        name: Install only RDKit
```
j-wags, line 55:
```suggestion
        name: Install with only RDKit
```

.github/workflows/examples.yml
```diff
@@ -0,0 +1,128 @@
...
+        run: |
+          python setup.py develop --no-deps
+
+      - name: Environment Information
+        shell: bash -l {0}
+        run: |
+          conda info
+          conda list
+
+      - name: Run example scripts without OE
```
j-wags, line 101: I can't wrap my head around what this means. 

```suggestion
      - name: Run example scripts
```

.github/workflows/examples.yml
```diff
@@ -0,0 +1,128 @@
...
+        shell: bash
+        env:
+          OE_LICENSE_TEXT: ${{ secrets.OE_LICENSE }}
+        run: |
+          echo "${OE_LICENSE_TEXT}" > ${OE_LICENSE}
+
+      - name: Install package
+        shell: bash -l {0}
+        run: |
+          python setup.py develop --no-deps
```
j-wags, line 93: Because openmmforcefields would have pulled in the latest _packaged_ version of openff-toolkit, let's delete it here, to ensure that only the current codebase is being tested. 

```suggestion
          conda remove --force openff-toolkit
          python setup.py develop --no-deps
```

j-wags (commented):
**TL;DR:**
Looks pretty good! I'd ask for a few changes

* force remove RDKit in the OE-only tests (reasoning below)
* remove the `conda tree` stuff, since linux CI doesn't like it for some reason (you were probably going to do this, just wanted to make sure)
* some wording changes in the yaml that had me super confused (in inline suggestions)
* force-deleting the stable release of `openff-toolkit` installed by `openmmforcefields`
* "pushing the big red button" 🔴  and removing examples tests from the old CI jobs


**Full version:**

> Remove the openmmforcefields dependency and copy the relevant force field into the examples folder

I have a weird feeling around copying data files like that. A lot of evil has occurred in our field as a results of copying data files without provenance, and having future people interpret them without context.

> Manually force-remove the rdkit dependency after conda has finished installing everything
> If we remove the rdkit dependency, we need to decide what to do with the example(s) that will be broken by that change. Simplest fix would be just to ignore it for the OpenEye only run, as we did previously. 

Let's do this. It's a bit brute-force, but it makes sure that the tests do exactly what we're asking. 

> Better fix might be to see if we can remove the requirement in the example, but this could be punted on til later.

That would be nice, but we've got forward progress already, so lets wrap up what we have, and we can open a separate issue for this :-)

.github/workflows/examples.yml
Yoshanuikabundi: I only wish it could run the solver and give the tree for a hypothetical environment...

Yoshanuikabundi: Also it sucks that it seem to be broken on Linux...

.github/workflows/examples.yml
Yoshanuikabundi: Done in 0ac1b27

Yoshanuikabundi:
@j-wags Done :)

j-wags (approved):
Wonderful. Thanks, @Yoshanuikabundi. 

As an aside, I noticed that the CI time savings wasn't that huge (~1-2 minutes). But the REALLY helpful thing is that this will let us put in much more complex+longer-running examples without slowing down CI on PRs 🕐

Issue #286, opened 2019-04-23
Standard release process


j-wags:
### 2: Create a new release on GitHub, following this template:
 
https://github.com/openforcefield/openff-toolkit/releases/new

#### Title = `X.Y.Z [Descriptive Title]`

```
# 0.10.1 Minor feature and bugfix release

[**For the complete release notes, please see our documentation**](https://open-forcefield-toolkit.readthedocs.io/en/stable/releasehistory.html)

For help getting the OpenFF Toolkit running, see our [installation instructions](https://open-forcefield-toolkit.readthedocs.io/en/stable/installation.html).

Please report bugs, request features, or ask questions through our [issue tracker](https://github.com/openforcefield/openff-toolkit/issues).

**Please note that there may still be some changes to the API prior to a stable 1.0.0 release.**

```

Add the "See our install instructions" block verbatim from above

### 2.5: Update the stable tag to point to the newest release/latest commit

Just run [this action](https://github.com/openforcefield/openff-toolkit/actions/workflows/release.yml)

### 3: Trigger a new build on conda-forge.  

**Get a fork of the feedstock repo**

If you don't have a personal fork of https://github.com/conda-forge/openff-toolkit-feedstock, then make one.

If you do have an old personal fork of https://github.com/conda-forge/openff-toolkit-feedstock, then update it with the most recent changes by running (replacing `j-wags` with your GH username)

```
git pull --set-upstream git@github.com:conda-forge/openff-toolkit-feedstock.git
git push --set-upstream git@github.com:j-wags/openff-toolkit-feedstock.git
```

**Update the build info**

in `recipe/meta.yaml`, update
* The version string at the top
* The build number to `0` if that's not already set
* Update the SHA with the value from the latest release. Get this using `curl -sL https://github.com/openforcefield/openff-toolkit/archive/X.Y.Z.tar.gz | openssl sha256`
* Add any new dependencies since the last release (Looking at recent changes to the [test env](https://github.com/openforcefield/openff-toolkit/blame/main/devtools/conda-envs/test_env.yaml) may help catch these)

**Open a PR from your fork to conda-forge/openff-toolkit-feedstock**

Follow instructions, including rerendeing and such. Once tests pass, go ahead and merge.

Test the package
* Run the [conda testing workflow](https://github.com/openforcefield/openforcefield/actions?query=workflow%3AConda)

### 4: Update repo text

[...]

PR #913, opened 2021-04-26
Release workflow

Yoshanuikabundi:
As we discussed in #286, this PR adds a release workflow that uses a floating tag called `latest` to point to the latest release. In addition, it caches the release on DockerHub (addressing $416) and tells Binder to look there. Finally, it updates all links to Binder in the repository to point to the `latest` tag.

I think the workflow can be manually tested on an existing release?


j-wags (commented):
Update: This is blocked by me. `secrets.DOCKER_USERNAME` and `secrets.DOCKER_PASSWORD` are not currently populated, so this PR won't work until I get those.

j-wags:
@Yoshanuikabundi: I have access to our dockerhub username and password now. However, I started reviewing this PR, and I'm bothered by the use of high-value secrets in this repo (both because of the recent CodeCov leak, and in general because it prevents us from letting people submit PRs from forks). So I don't think we should add more secrets (like docker credentials) to this repo -- In fact, if we can get this repo down to zero secrets, it can be open to PRs from forks!

So, how hard would it be to move this automation to https://github.com/openforcefield/status/? That repo is more internal and there's no expectations of outside contributions to it.

j-wags (approved):
This looks great, but I'd prefer to keep the dockerhub secrets out of this repo. So let's:
* Remove the dockerhub build+upload from this PR
* Keep everything else in this PR
* Let's not worry about making/caching the docker images for now

Yoshanuikabundi:
@j-wags I was a little worried about this PR so I haven't merged it yet. I made a [test repo](https://github.com/Yoshanuikabundi/latest_tag_test) to see how it works. I did the following:

1. Create the repo in GitHub with a readme
2. Create a release `v0.1.0` from the initial commit through GitHub
3. Clone locally
4. `git tag latest && git push --tags`
5. Added and committed the Actions workflow
6. Made a small change to the readme, committed it, and pushed
7. Published a new release `v0.2.0` through GitHub

 As I feared, the `latest` tag shows up as a [release](https://github.com/Yoshanuikabundi/latest_tag_test/releases). But the workflow does work at least, and within about a minute after the `v0.2.0` release was published the tag had updated.

We could get around that minor ugliness by using a branch rather than a tag with [this action](https://github.com/Embraser01/update-git-branch-action). That would have the downside that it would be possible for someone to accidentally check out the `latest` branch and commit to it, which can't happen with a tag.

Just wanted to bring that up before I merge this PR and we're stuck with a weird pseudo-release forever. Which do you prefer? Sorry, I should've brought this up at our check in but it must've slipped my mind.

.github/workflows/release.yml
```diff
@@ -0,0 +1,15 @@
+name: Build Notebook Container
+on:
+  release:
+    types: [published]
```
j-wags, line 4: Let's keep this manual -- This way we won't get surprised if we some day make an alpha/non-stable release.

```suggestion
  workflow_dispatch:
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,15 @@
+name: Build Notebook Container
```
j-wags, line 1:
```suggestion
name: Update latest tag
```

j-wags (approved):
Thanks for checking into this, @Yoshanuikabundi. I think it's fine to have a `latest` tag, since it doesn't move to the top of the Releases list, even when it gets updated.

However, I'd like to keep this manual for now, so I've proposed that change. Let's also make sure to add this to the [standard release process](https://github.com/openforcefield/openff-toolkit/issues/286) when we merge.

j-wags:
Sorry for the flaky tests -- Merging!

PR #915, opened 2021-04-27
Update for black 21.4b0

mattwthompson:
Black 21.4b0 came out over the weekend. It's the first update since 20.8b1, so that's 8 months for some really tiny chances.

This should get CI closer to passing.


openff/toolkit/tests/test_forcefield.py
```diff
@@ -3868,7 +3868,7 @@ def test_freesolv_gbsa_energies(
         )
 
     def test_tip5p_dimer_energy(self):
-        """"""
+        """ """
```
j-wags, line 3871: Monumental improvements :-)

j-wags (approved):
Wunderbar!

Issue #914, opened 2021-04-27
Have create_openmm_system generate CustomNonbondedForce objects


mattwthompson:
**Is your feature request related to a problem? Please describe.**
OpenMM provides `NonbondedForce` as an way to handle non-bonded forces, both vdW and electrostatics, in the same object. This is great for vanilla setups, i.e. those with cutoff LJ and PME electrostatics. This is, however, not consistent with the design of the SMIRNOFF spec in which vdW and electrostatic forces are generally kept separate. For example, one should be able to design a SMIRNOFF-style force field that has different non-bonded methods or cutoffs for vdW and electrostatics, which gets hard/impossible to store in a single `NonbondedForce`.

**Describe the solution you'd like**
@j-wags recently learned that a single `NonbondedForce` can be split up into several `CustomNonbondedForce` objects with no or minimal performance losses (https://github.com/openforcefield/openff-toolkit/pull/833#issuecomment-821722615). I don't know if **all** instances of `NonbondedForce` should be taken out.

**Describe alternatives you've considered**
This is in general a tangly portion fo the codebase - it's been a daunting task to document, it takes about 100 LOC to encode in the unit tests, and we still run into cases in which a single `NonbondedForce` is incompatible with the SMIRNOFF spec. Things have been working fine for the main use case (our mainline force fields  that use cutoff vdW and PME electrostatics) but I doubt that a single `NonbondedForce` is just not extensible enough long-term - given the option of separate `CustomNonbondedForce`s!

**Additional context**
Some issues that have cropped up: #380 #191 #734 #882 https://github.com/openforcefield/openff-toolkit/issues/219

There might be impacts on downstream developers if people expect to be able to run `.getForces()` and extract data from the single `NonbondedForce` that is found. I would think most users would not be hugely impacted by a change that keeps simulations running similarly to as before.

j-wags:
Thinking about the design of a fix -- if at all possible, I'd advocate supporting the previous behavior as default, and splitting into two `CustomNonbondedForces` as a optional override. This would preserve people's old workflows that relied on getting the `NonbondedForce` from the resulting system. 

So the kwarg could be `forcefield.create_openmm_system(topology, try_to_combine_nonbonded_forces=True`). There would be three possibilities for its behavior:

[...]

Issue #886, opened 2021-03-26
GBSA test failures on FreeSolv dataset


mattwthompson:
**Describe the bug**
Some of the GBSA tests are reporting a different energy on a small (?) subset of the FreeSolv dataset. The differences look small to my eye but I'm not familiar enough with these tests to say they're not important.

These tests **only run** if OpenEye Toolkits are installed and "slow" tests are turned on via `--runslow`. These are automated on our nightly CI runs, but not the ones that run on each push and PR. These were previously missed because of #874 but not observed until last night's nighly because we do not run "slow" tests before merging. Presumably because of how long #874 lasted, there were several upstream updates, so maybe it's reasonable that some of these energies differ.

**To Reproduce**

Something like
```
pytest -v --runslow openff/toolkit/tests/test_forcefield.py::TestForceFieldParameterAssignment::test_freesolv_gbsa_energies
```
**Output**

Trimmed section of [log](https://pipelines.actions.githubusercontent.com/u2Wf9sTwFK9NZj8rRqcrkPO1UF3tPeGIdu4NysSHujaRyWbUAY/_apis/pipelines/1/runs/3036/signedlogcontent/5?urlExpires=2021-03-26T15%3A07%3A28.2666139Z&urlSigningMethod=HMACV1&urlSignature=NHb2pnozLBO4F6U5t0NdLryRLUjIAgtenMvLOdgPEeU%3D) 
```
2021-03-26T05:12:48.3733770Z =================================== FAILURES ===================================
2021-03-26T05:12:48.3736390Z _ TestForceFieldParameterAssignment.test_freesolv_gbsa_energies[1328936-0_0_2-False-OBC1-True] _
2021-03-26T05:12:48.3738320Z openff/toolkit/tests/test_forcefield.py:3867: in test_freesolv_gbsa_energies
2021-03-26T05:12:48.3739680Z     off_omm_system, amber_omm_system, positions, by_force_type=False
2021-03-26T05:12:48.3741270Z openff/toolkit/tests/utils.py:677: in compare_system_energies
2021-03-26T05:12:48.3742500Z     atol=atol,
2021-03-26T05:12:48.3743720Z openff/toolkit/tests/utils.py:503: in compare_context_energies
2021-03-26T05:12:48.3745230Z     format_args=(potential_energy1, potential_energy2),
2021-03-26T05:12:48.3746450Z openff/toolkit/tests/utils.py:492: in raise_assert
[...]

PR #917, opened 2021-04-28
Ensure NonbondedForce cutoff is set to vdW handler's cutoff

### LINKED ISSUE #734: VdW Cutoff Ignored when Electrostatic Handler Present
**Describe the bug**
The non-bonded cutoff is hardcoded in several places (the [`vdWHandler`](https://github.com/openforcefield/openforcefield/blob/7091de5daf03abfa487e7c3619eafcfcbff91d1d/openforcefield/typing/engines/smirnoff/parameters.py#L3226) and the [`ElectrostaticsHandler`](https://github.com/openforcefield/openforcefield/blob/7091de5daf03abfa487e7c3619eafcfcbff91d1d/openforcefield/typing/engines/smirnoff/parameters.py#L3524)). This hardcoded value overwrites the value set by both the `vdWHandler` and the `ElectrostaticsHandler`.

**To Reproduce**

```
# box vectors.
topology.box_vectors = numpy.eye(3) * unit.nanometers

# Create the FF
force_field = ForceField()

# Add only a VdW handler.
vdw_handler.method = "cutoff"
vdw_handler.cutoff = 6.0 * unit.angstrom
vdw_handler.scale14 = 1.0

vdw_handler.add_parameter(
    {
        "smirks": "[#18:1]",
        "sigma": 1.0 * unit.angstrom
    }
)

assert numpy.isclose(
    system.getForce(0).getCutoffDistance().value_in_unit(unit.angstrom),
    6.0
)

# Add an electrostatics handler.
electrostatics_handler = ElectrostaticsHandler(version=0.3)
electrostatics_handler.cutoff = 6.0 * unit.angstrom
electrostatics_handler.method = "PME"
force_field.register_parameter_handler(electrostatics_handler)

library_charges = LibraryChargeHandler(version=0.3)
library_charges.add_parameter(
    {
        "smirks": "[#18:1]",
        "charge1": 0.0 * unit.elementary_charge,
    }
)
force_field.register_parameter_handler(library_charges)

assert numpy.isclose(
    system.getForce(0).getCutoffDistance().value_in_unit(unit.angstrom),
    6.0
)
```

**Output**

```AssertionError``` on the final assert.

 - `openforcefield = 0.7.2`

mattwthompson:
Fixes #734 

I was working on bigger changes but wanted to make this (relatively) small patch first. I think the best approach is to forcefully fix this in the final steps of `create_openmm_system`.  Until we finish #914,  cutoffs can't be mismatched at all,, except for the case of cutoff vdW and PME electrostatics in which the PME "cutoff" isn't particularly meaningful anyway.


mattwthompson:
No clue why everything is passing given #886 is still open

j-wags (commented):
Do we have an authoritative reference for the statement "If the NonbondedForce is using the `PME` method, then the cutoff set on the NonbondedForce is used for vdW, but a different cutoff is used for the electrostatics PME short-range/long range transition"? 

At [one point in the docs](http://docs.openmm.org/latest/api-python/generated/simtk.openmm.openmm.NonbondedForce.html#simtk.openmm.openmm.NonbondedForce.setEwaldErrorTolerance), it's implied that the ewald error tolerance is used to determine the cutoff between the regions, but I'd like to make sure that this is the case. 

Otherwise this PR looks good, I just remember scratching my head about nonbonded methods about a year ago, and I think writing this stuff down now can save us time in the future.

mattwthompson:
> At one point in the docs, it's implied that the ewald error tolerance is used to determine the cutoff between the regions, but I'd like to make sure that this is the case.

I agree with your assessment here and actually came across the same text while half-reading your comment at first. ~~Still, I couldn't find a definitive answer myself, although some threads in the OpenMM issue tracker imply that either some PME tuning happens by default or the PME "cutoff" is not the same as `.getCutoffDistance().~~ However, I did find an ancient thread that points to a thoroughly-commented script, which includes this comment that somewhat specifically answers our question.

> The nonbonded cutoff does not affect the accuracy of the Coulomb interaction with PME.  You can
    freely vary the cutoff distance, and OpenMM will automatically select internal parameters to give
    whatever accuracy has been selected with the ewaldErrorTolerance parameter.  (The cutoff does affect
    other nonbonded interactions, such as Lennard-Jones, so this generally places a lower limit on the
    cutoffs you consider acceptable.)


https://github.com/openmm/openmm/issues/55

https://simtk.org/svn/pyopenmm/trunk/simtk/pyopenmm/extras/optimizepme.py

It _seems_ like there is no on-the-fly PME tuning that happens at runtime like what happens in GROMACS, but that's not the same as asking  "is the vdW cutoff the same distance used to switch electrostatics interactions from real space to reciprocal space?" For a few years I've been operating on the assumption that the PME "cutoff" is an implementation detail that can significantly impact performance but (if everything is set up correctly) shouldn't damage results.

I think it's reasonable to say that if we're both confused, this is an area of the OpenMM docs that could be improved.

j-wags:
So based on the discussion in the openMM issue, it seems like PME cutoff DOES change the boundary between direct-space and reciprocal-space electrostatics, but it doesn't matter, and PME cutoff shouldn't be part of the SMIRNOFF spec. 

I'll work on figuring out what that means for this PR. And also the spec, I guess.

j-wags (approved):
Thanks for all the thinking on this, @mattwthompson, and sorry it's taken me so long to get to it. These changes look good. 

This PR allows something that was previously unallowed, but I think that the SMIRNOFF spec will have to eventually change to be more specific about how the different cutoffs should be treated. At that time, I think there will be a different set of configuration options ADDED for PME electrostatics, and the cutoff will take on a different meaning than is used here. So I don't think that merging this now is likely to paint us into a reverse-incompatibility corner in the future. 

Could you update the releasenotes before you merge? Thanks!

PR #920, opened 2021-04-29
Optionally split out nonbonded force into multiple forces

mattwthompson:
Implements #914 

My current approach is to store all coulombic interactions in a `NonbondedForce` since it's easier to get PME working with a couple lines of code. (It may be best to later split this out into its own `CustomNonbondedForce`, I'm not sure.) The vdW interactions  are split out into their own `CustomNonbondedForce` - however, this class does not easily allow for scaled 1-4 interactions, so the current hack is just to add those back to the `NonbondedForce`, since it's easy, but that produces a really strange state. The vdW 1-4 interactions (but no intermolecular interactions) are stored with the coulombic interactions, whereas the other vdW interactions are all split out into another force. Yuck. Need to fix that.

This is already a pretty big task to swing at, so my goal for this PR is just to get things working with "vanilla" behavior (cutoff vdW, PME electrostatics, 12-6 LJ) when `combine_nonbonded_forces=False`, with the intention to add in complex behavior later on.

- [x] Fix setting `NonbondedForce.NoCutoff` for all non-periodic cases
- [x] Throw error in case of un-combinable nonbonded methods
- [ ] Fix 1-4 scaling factor hack
- [ ] Ensure 1-4 scaling factors are applied
- [ ] Match energies in cases where `combine_nonbonded_forces` should have no impact
- [ ] Ensure non-default, non-matching 1-4 scaling factors can be used
- [ ] Update non-bonded method section of dev docs


mattwthompson:
Aha! It took an embarassing amount of time reading through old OpenMM issues, but I eventually noticed that the 1-4 exceptions are meant to be added back as a [`CustomBondForce`]. My lazy reading read over [this](https://github.com/openmm/openmm/issues/1901#issuecomment-329258376) key comment several times, but read `CustomBondForce` as `CustomNonbondedForce`, which does not seem to actually offer a way to do that.

mattwthompson:
I've smooshed two important changes into one PR, which is not good. I'm closing this and I will cherry-pick into two or three new PRs aimed at a feature branch.

Issue #922, opened 2021-05-04
Generator for 1-4 pairs from a Topology


mattwthompson:
**Is your feature request related to a problem? Please describe.**
To interface with MD engines, I often need to access a list of the 1-4 pairs, defined as atoms separated by a shortest path of precisely 3 bonds

**Describe the solution you'd like**
I would like an API point like `Topology.pairs_14` that is a generator over tuples of atom indices of each 1-4 pair in a topology.

A more general solution might be to have a more general function that returns all pairs separated by a minimum distance, specified as an argument to the function `n`. This should behave identically to `.pairs_14` in the case of `n=3`. There are some force fields that scale 1-3 or 1-5 interactions (and even some [crazy people](https://github.com/openmm/openmm/issues/1926) that try to include 1-2 interactions!) so there would be some utility of making this behavior more general. It may be worth limiting the value of `n` to avoid computational explosions (I can't wrap my head around calling a function like this with i.e. `n` > 10).

I expect this to be feasible to implement with some intelligent looping and/or `networkx` calls. There should be tests that cover basic behavior in small molecules, "gotchas" in rings, and maybe a slow test that tests the performance of a system of a few thousand small molecules.

**Describe alternatives you've considered**
There are some workarounds, like one-off code I've written and [similar functionality](http://docs.openmm.org/latest/api-python/generated/simtk.openmm.openmm.NonbondedForce.html#simtk.openmm.openmm.NonbondedForce.createExceptionsFromBonds) in OpenMM. Just iterating over proper torsions (`Topology.propers`) and tracking atoms #1 and #4 does not work because some pairs in rings can be pairs by multiple paths, i.e. two atoms in a 5-membered ring can simultaneously be 1-4 and 1-3 pairs, and in typical MM force fields this would be treated as a 1-3 interaction and completely excluded.

Keeping behavior like this in one place (the toolkit) would help code duplication and buggy behavior resulting from people (like me) doing bad one-off implementations.

**Additional context**

cc @ijpulidos - Please add this to the requirements doc - though I'd expect any solution that can fit into the current API to be portable to the refactor.

j-wags:
I agree with this feature request -- The SMIRNOFF spec also hints at support for custom 1-3 and 1-5 scaling factors, so this will be handy in general. 

The specific spot for this logic would be in the `Molecule` class, with an additional access method in the `Topology`, which simply loops over all the molecules it contains. 

> Please add this to the requirements doc

[...]

PR #925, opened 2021-05-05
Update non-bonded methods to better match SMIRNOFF spec

mattwthompson:
This PR targets a new branch aimed at consolidating several changes to non-bonded treatments into one release (possibly 0.10.0). This hopefully condenses disruptions to users to one point in time an avoids confusing stopgaps for partially-implemented behavior that is then implemented a month later (this was an issue [here](https://github.com/openforcefield/openff-toolkit/pull/833#issuecomment-819956402)). It might be useful to consider staging some other PRs here (#896, #917).

The biggest problem here is that, without using `NonbondedForce` and no `CustomNonbondedForce`/`CustomBondedForce` setup, it's not possible to strictly adhere to the SMIRNOFF spec and use OpenMM to compute "gas-phase" single-molecule energies, which is a pretty common use case.
* PME cannot be applied to non-periodic topologies/systems
* The simplest way to get direct coulombic interactions is to just use  `openmm.NonbondedMethod.NoCutoff`
* No vdW method in the SMIRNOFF spec corresponds to `openmm.NonbondedMethod.NoCutoff`

This probably doesn't make a meaningful impact on conformer energies (especially compared to numerous other sources of error) as the main/only error is intramolecular vdW interactions past the 9 Angstrom cutoff. But there's still a huge mismatch between specified inputs (cutoff, with a switching function, and PME) and resulting methods actually used in a calculation (no cutoff and no cutoff).

This PR takes the approach of more strictly adhering to the spec, resulting in many more errors being raised and breaking tons of tests that rely on the `NoCutoff` behavior. My hope is that the follow-up PR implementing #922 can handle such cases while both adhering to the spec and working within the constraints of OpenMM's implementations (which of course differ from those in other engines). 


docs/developing.md
```diff
@@ -219,10 +219,10 @@ This table describes which [`NonbondedMethod`](http://docs.openmm.org/latest/use
 
 | vdw_method | electrostatics_method | periodic | OpenMM Nonbonded method or exception | Common case |
 |------------|-----------------------|----------|--------------------------------------|---|
-| cutoff     | Coulomb               | True     | raises `IncompatibleParameterError`
-| cutoff     | Coulomb               | False    | `openmm.NonbondedForce.NoCutoff`
-| cutoff     | reaction-field        | True     | raises `SMIRNOFFSpecUnimplementedError`
-| cutoff     | reaction-field        | False    | raises `SMIRNOFFSpecError`
+| cutoff     | Coulomb               | True     | raises `.SMIRNOFFSpecUnimplementedError`
+| cutoff     | Coulomb               | False    | raises `.SMIRNOFFSpecUnimplementedError`
```
j-wags, line 223: Let's have this return the same thing as before, and print a DeprecationWarning that this will change (or return something different)

docs/developing.md
```diff
@@ -219,10 +219,10 @@ This table describes which [`NonbondedMethod`](http://docs.openmm.org/latest/use
...
+| cutoff     | reaction-field        | True     | `openmm.NonbondedForce.CutoffPeriodic`
+| cutoff     | reaction-field        | False    | `openmm.NonbondedForce.CutoffNonPeriodic`
```
j-wags, line 225: There's some additional context about reaction-field and OpenMM that I need to dig up -- It's somewhere in the issues/PRs. I'll let you know when I find it.

mattwthompson:
Stalled, closing in favor of Interchange's implementations.

PR #938, opened 2021-05-14
[DNM] Add ForceField._create_openff_system

### LINKED ISSUE #310: Container class for parameterized systems

mattwthompson:
Resolves #310 

This PR begins the process of adding **private** API point(s) to the toolkit that construct OpenFF System objects.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -114,6 +114,7 @@ jobs:
 
       - name: Install experimental packages and dependencies
         shell: bash -l {0}
+        run: |
           pip install git+git://github.com/openforcefield/openff-system.git
           pip install git+git://github.com/mattwthompson/openff-units.git
           pip install git+git://github.com/mattwthompson/openff-utilities.git
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -112,6 +112,14 @@ jobs:
         run: |
           python setup.py develop --no-deps
 
+      - name: Install experimental packages and dependencies
+        shell: bash -l {0}
+          pip install git+git://github.com/openforcefield/openff-system.git
+          pip install git+git://github.com/mattwthompson/openff-units.git
+          pip install git+git://github.com/mattwthompson/openff-utilities.git
+
+          conda install ele mdtraj -c conda-forge -yq
+
       - name: Check installed toolkits
         shell: bash -l {0}
         run: |
--- a/openff/toolkit/tests/test_system.py
+++ b/openff/toolkit/tests/test_system.py
@@ -1,0 +1,15 @@
+from openff.system.stubs import ForceField
+
+from openff.toolkit.topology import Molecule, Topology
+
+
+def test_basic_construction():
+    top = Molecule.from_smiles("C").to_topology()
+    parsley = ForceField("openff-1.0.0.offxml")
+    off_sys = parsley.create_openff_system(top)
+
+    assert off_sys.topology is not None
+    assert off_sys.topology.mdtop is not None
+    assert off_sys.positions is None
+    for key in ["vdW", "Bonds", "Electrostatics"]:
+        assert key in off_sys.handlers
--- a/openff/toolkit/typing/engines/smirnoff/forcefield.py
+++ b/openff/toolkit/typing/engines/smirnoff/forcefield.py
@@ -50,6 +50,7 @@ from openff.toolkit.utils.utils import (
     convert_0_2_smirnoff_to_0_3,
     convert_all_quantities_to_string,
     convert_all_strings_to_quantity,
+    requires_package,
 )
 
 # =============================================================================================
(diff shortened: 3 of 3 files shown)


mattwthompson:
Closing these changes in favor of a simpler solution that relies on `System.from_smirnoff`. See  #191 for the implementation and some  context, and justification.

j-wags:
I think that moving this to be a `System` classmethod is a good call. Once `System` is stable + public we can easily add it as an optional import, and have `ForceField.create_openff_system` redirect straight to it.

PR #948, opened 2021-05-25
Add `normalize_partial_charges` kwarg to `assign_partial_charges`

### LINKED ISSUE #456: Round partial charges to enforce correct integral charge on molecule

* Some partial charge assignment methods only report three or four decimal places of precision
* Some System/Topology file formats only store three or four decimals of precision

Both of these issues can lead to rounding errors that become significant in larger molecules. 

[ParmEd implemented this approach](https://github.com/ParmEd/ParmEd/blob/5b9837dff455b7803f6785df24f68d0dd5372236/parmed/modeller/residue.py#L366-L423), which rounds existing charges to the desired precision, and then adds the sum of the rounding errors to the first particle to achieve the correct net charge on each residue. This may be tricky for OFFTK, since it has no concept of "Residues", thus a large biopolymer could require a substantial correction, which should **not** be dumped on a first atom (for example, if 1000 residues each contributed 0.001e of rounding error, then the first atom might end up getting +1 added to its charge).  

It seems likely that we could apply the same solution, but just with a slightly more complex algorithm to spread out the final charge correction over several atoms. It's not clear to me that this is a question with an "automagic" solution -- pathological cases could lead us to **introducing** errors with this functionality (for example, accumulating a large error given a biopolymer, or adding a dipole to a symmetric system). 

We'd want to figure out how this would interface with our current functionality. Basically, we should allow
* Users disabling the rounding altogether
* Users rounding to a custom precision
* Users applying this rounding only to certain molecules 

Further, we should figure out where this would live in our current API. Some possibilities include
* kwarg to create_openmm_system: `do_not_round_charges=[molecules]` or `custom_charge_rounding={molecule:precision}`
* Tag-level element in the ForceField itself -- likely in `ElectrostaticsHandler` (but this prevents us from handling different molecules individually). So, for example, `  <Electrostatics version="0.3" method="PME" scale12="0.0" scale13="0.0" scale14="0.833333" cutoff="9.0 * angstrom" rounding_precision="4"/>`
* In our eventual OFF `System` class, which would know which simulation format it is exporting to, and thus which precision to use. 

I'm most in favor of the first and third options, since the second would require a change to the SMIRNOFF spec and disallow treating molecules differently. The first option is the only one that is immediately available.

**Additional context**
In the LibraryCharges PR that was already merged (#433), we added a check that the sum of a molecule's partial charges must be within 0.01 of the molecule's formal charge. This can be overridden by a kwarg (`allow_nonintegral_charges=True`).

j-wags:
- [x] Resolves #456 
    - [x] Update existing tests to expect higher precision
    - [x] Add test that DOESN'T use round_partial_charges and ensure the charges are different
    - [x] Test supplying `round_partial_charges` to Molecule API as well as ToolkitWrapper API


openff/toolkit/tests/test_toolkits.py
```diff
@@ -2834,7 +2836,31 @@ def test_assign_partial_charges_am1bcc(self):
         for pc in molecule._partial_charges:
             charge_sum += pc
             abs_charge_sum += abs(pc)
-        assert abs(charge_sum) < 0.001 * unit.elementary_charge
+        assert abs(charge_sum) < 1e-8 * unit.elementary_charge
```
mattwthompson, line 2839: Not blocking: I wonder if this molecule (acetate) should have a smaller error than 1e-8?

mattwthompson (changes requested):
I found a few details in tests that I think should be looked at, but overall the implementation seems great. 👍 

If you went into this thinking that the toolkit often reports non-integer charges on small molecules, keep reading to have your confirmation bias totally validated.

----

Blocking:

1. Does OpenEye never return a molecule with this charge rounding issue? Even at a much lower rounding error than AmberTools?
3.  Independent of above, there are also a few tests in `TestOpenEyeToolkitWrapper` that I think could be updated with the assumption that the difference between the total charge and the sum of formal charges should be ~1e-8, not ~0.005. Searching for `_sum` hits many tests named `test_assign_partial_charges_.*`. I think that we can and should trust partial charge assignment to not give rounding integers of that magnitude, right?
4. Don't forget to update ` openff/toolkit/data/reference_energies/README.md` since you generated new files.


-------
Not blocking:

The energy differences are quite a bit larger than I expected. My first two thoughts are
* I have faith in how we set up the generation of these reference energies
* Maybe the energies before were "wrong," since those molecules plausibly had some charge rounding errors? It looks like we forced RDKit/AmberTools in `test_energies.py`.

But this not a change I'd want to sweep under the rug, so I tinkered with the script (https://github.com/openforcefield/openff-toolkit/commit/aa18faff9f0d93ed5bcded1861335c048ddc94b5) Here's the output of inspecting the `openmm.System` generated by each molecule:

```
Molecule with name '*****' and SMILES '[H][O][C]([H])([H])[C]([H])([H])[H]' 	 True 	 1.000000000070389e-06 e

Molecule with name '*****' and SMILES '[H][O][C]([H])([H])[C]([H])([H])[H]' 	 False 	 1.000000000070389e-06 e

[Molecule with name '' and SMILES '[H][C]([H])([H])[H]', Molecule with name '' and SMILES '[H][C]([H])([H])[H]'] 	 True 	 -0.003999999999999976 e

[Molecule with name '' and SMILES '[H][C]([H])([H])[H]', Molecule with name '' and SMILES '[H][C]([H])([H])[H]'] 	 False 	 -0.003999999999999976 e

Molecule with name '' and SMILES '[H][c]1[n][c]([H])[c]([H])[c](-[c]2[c]([H])[c]([H])[c]([O-])[c]([H])[c]2[H])[c]1[H]' 	 True 	 -0.999 e

Molecule with name '' and SMILES '[H][c]1[n][c]([H])[c]([H])[c](-[c]2[c]([H])[c]([H])[c]([O-])[c]([H])[c]2[H])[c]1[H]' 	 False 	 -0.999 e

```

OOF. No wonder there are some energy differences. Just to double-check, I cherry-picked the differences in that script on this branch and got:


```
Molecule with name '*****' and SMILES '[H][O][C]([H])([H])[C]([H])([H])[H]' 	 True 	 1.3877787807814457e-17 e

Molecule with name '*****' and SMILES '[H][O][C]([H])([H])[C]([H])([H])[H]' 	 False 	 1.3877787807814457e-17 e

[Molecule with name '' and SMILES '[H][C]([H])([H])[H]', Molecule with name '' and SMILES '[H][C]([H])([H])[H]'] 	 True 	 0.0 e

[Molecule with name '' and SMILES '[H][C]([H])([H])[H]', Molecule with name '' and SMILES '[H][C]([H])([H])[H]'] 	 False 	 0.0 e

Molecule with name '' and SMILES '[H][c]1[n][c]([H])[c]([H])[c](-[c]2[c]([H])[c]([H])[c]([O-])[c]([H])[c]2[H])[c]1[H]' 	 True 	 -1.0000000000000004 e

Molecule with name '' and SMILES '[H][c]1[n][c]([H])[c]([H])[c](-[c]2[c]([H])[c]([H])[c]([O-])[c]([H])[c]2[H])[c]1[H]' 	 False 	 -1.0000000000000004 e

```


So I'm relatively confident that the updated reference energies are more correct, this PR does what is intended, and the energies are not blocking.


I thought a little bit about if this behavior should be default, then realized it was [made clear](https://github.com/openforcefield/openff-toolkit/issues/456#issuecomment-779432146) in the pre-implementation discussions.

I don't know if it's better to call `molecule._normalize_partial_charges()` from within the toolkit wrapper calls or just at the end of `Molecule.assign_partial_charges()`. I think my intuition was to do the latter, but I can't think of anything wrong with this approach. And there are probably cases in which having that behavior in the toolkit wrappers is desirable, I just can't think of one off the top of my head.

Some of https://github.com/openforcefield/openff-toolkit/issues/456 centered around this being toggle-able by the user, proposed as a(nother) keyword argument to `ForceField.create_openmm_system`. I think it's okay to leave that switch out of this PR; implementing it in the future will not be a behavior break (as long as the default remains to do the charge-rounding). Another list of ideas from that thread was:

1. Users disabling the rounding altogether
1. Users rounding to a custom precision
1. Users applying this rounding only to certain molecules

I think (1) is partially-implemented; it wouldn't be easy to do but it seems possible. (2) is not implemented but I'm not sure it should be. (3) is also not implemented but arguably out of scope here. (Maybe in the future this is something that gets hidden away in the complexity of `ToolkitAM1BCCHandler`, and maybe it needs to be re-thought through in the context of biopolymer infrastructure.) So I don't think any of these are blockers.

openff/toolkit/tests/test_toolkits.py
j-wags: I've tightened all instances of this down to 1e-10, though if we want to be really strict, full python float precision touches 1e-15. I think 1e-10 is fine for now, though, since too tight a tolerance will win us lots of false negatives in the tests.

j-wags:
Thanks for the attentive review! The first three quotes below are responses to your blocking comments.

>    (blocking) Does OpenEye never return a molecule with this charge rounding issue? Even at a much lower rounding error than AmberTools?

The python-float-normalization gets the error down to 1e-16, while the original OE errors are on the order of 1e-8. So I've added tests to ensure that the outputs are in those ranges. I've also added tests for the new `normalize_partial_charges` keyword in all four `assign_partial_charges` providers - OE, AmberTools, RDKit, BuiltIn - though the latter two already returned errors on the order of 1e-16, even before this PR. Those additional tests will be handy to ensure that they continue accepting those keywords and complying with the high-level API.

>    (blocking) Independent of above, there are also a few tests in TestOpenEyeToolkitWrapper that I think could be updated with the assumption that the difference between the total charge and the sum of formal charges should be ~1e-8, not ~0.005. Searching for _sum hits many tests named test_assign_partial_charges_.*. I think that we can and should trust partial charge assignment to not give rounding integers of that magnitude, right?

Good call -- I've tightened the `abs(molecule.total_charge - sum(partial_charges))` tolerance down to 1e-10 on every test that I could. We could go tighter, but we'll start getting false positives at some point, so 1e-10 seems like plenty to me.

>    (blocking) Don't forget to update openff/toolkit/data/reference_energies/README.md since you generated new files.

Thanks for catching this. Done!

>  (comment) I don't know if it's better to call molecule._normalize_partial_charges() from within the toolkit wrapper calls or just at the end of Molecule.assign_partial_charges().

Since direct calls to the ToolkitWrappers (like `RDKitToolkitWrapper.assign_partial_charge`) are in the public API, I wanted the normalization to be accessible to users of that as well. But since the `_normalize_partial_charges(offmol)` method didn't really seem to have a natural place to go in `toolkits.py`, and someone with a weird molecule subclass may want to change what their partial charges mean or how they're normalized, I figured that it would be most logical to put it in `FrozenMolecule`. 

> (comment) Some of #456 centered around this being toggle-able by the user, proposed as a(nother) keyword argument to ForceField.create_openmm_system.

Yeah, I was thinking about this. It seems like a weird use case, but I agree that someone might one day want to do that. When that day comes, we can thread it through to the top-level `create_openmm_system` call. But generally, the kwargs on that method are a real tangle of complexity, and I don't want to add another if we don't need to. 

**I think this should be good for you to take another look when you have time!** The diff is a bit messier now that I've tightened tolerances all over the file, and added tests of the new kwarg for other toolkitwrappers. 

Also, there's a pretty large merge conflict with #993, so if you approve this, I'll close this PR, then make a new branch from master and port the changes into that (I'll merge those changes directly and you won't need to re-review)

mattwthompson (approved):
I agree with everything you've written - LGTM (given passing tests and a way to resolve merge conflcits)!

j-wags:
Thanks @mattwthompson :-) I'll close this and start a new PR, and will merge with CI turns green on that.

PR #949, opened 2021-05-25
Relocate the SMIRNOFF spec to standards

SimonBoothroyd:
Update the SMIRNOFF document to state it has been relocated to the standards repository. This should not be merged until https://github.com/openforcefield/standards/pull/1 is merged.


mattwthompson (approved):
I'm in favor of this change, and

> This should not be merged until openforcefield/standards#1 is merged.

has been merged!

j-wags:
I'm going to be out for a week, but I'm good for this to move forward. However, one request: If you do merge, please make sure that the bookkeeping stays clear - other links in our repos and websites that point to the SMIRNOFF spec need to point to the authoritative `standards` version. Otherwise every link leading to this perma-redirect page will be a sloppy user experience, and it will be just as convenient for people to pull up an old release's SMIRNOFF spec on RTD without realizing that it may be incomplete/out of date.

Also, the redirect should point to the SMIRNOFF spec page itself, not the top level of the `standards` repo.

mattwthompson:
I gave a reasonable attempt at fixing everything in this repo. There weren't too many, but they were scattered across comments, docstrings, example notebooks, and the user guide. Here are the two patterns I searched:


```
(openff-dev) [openforcefield] grep -r "smirnoff.html" .
./docs/releasehistory.md:More details and examples of this change are available in `The ChargeIncrementModel tag in the SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#chargeincrementmodel-small-molecule-and-fragment-charges>`_
./docs/releasehistory.md:`The ChargeIncrementModel tag in the SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#chargeincrementmodel-small-molecule-and-fragment-charges>`_
./docs/releasehistory.md:`the ProperTorsions section of the SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#fractional-torsion-bond-orders>`_.
./docs/releasehistory.md:  ``ChargeIncrementModel`` tag in the `SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#chargeincrementmodel-small-molecule-and-fragment-charges>`_.
./docs/releasehistory.md:`Library Charges <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#librarycharges-library-charges-for-polymeric-residues-and-special-solvent-models>`_.
./docs/releasehistory.md:  `LibraryCharges tag in the SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#librarycharges-library-charges-for-polymeric-residues-and-special-solvent-models>`_
./docs/releasehistory.md:`GBSA tag in the SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/0.5.0/smirnoff.html#gbsa>`_.
./docs/releasehistory.md:  `GBSA tag in the SMIRNOFF specification <https://open-forcefield-toolkit.readthedocs.io/en/0.5.0/smirnoff.html#gbsa>`_.
./docs/releasehistory.md:  `SMIRNOFF specification page <https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#supported-generalized-born-gb-models>`_.
./docs/releasehistory.md:This update also introduces the `SMIRNOFF 0.3 specification <https://open-forcefield-toolkit.readthedocs.io/en/0.4.0/smirnoff.html>`_.
./docs/releasehistory.md:`SMIRNOFF 0.3 specification updates <https://open-forcefield-toolkit.readthedocs.io/en/0.4.0/smirnoff.html>`_
./docs/releasehistory.md:* Major overhaul, resulting in the creation of the `SMIRNOFF 0.2 specification <https://open-forcefield-toolkit.readthedocs.io/en/master/smirnoff.html>`_ and its XML representation
(openff-dev) [openforcefield] grep -r "The-SMIR" .
./docs/users/smirnoff.md::::{include} ../../The-SMIRNOFF-force-field-format.md
```

I'm not sure we should be changing the release history, so I left those be. If the links in the docs all work, I think that's about all that can be done here.

Yoshanuikabundi:
I would be in favour of updating the links in the changelog. There's precedent; IIRC we've changed the changelog to keep links up-to-date, to migrate information to the main docs, to consistently use the spacing "force field", and to migrate to Markdown. And maybe more importantly, the purpose of the changelog is not to be a fixed-in-stone record of the state of the changelog at release time, but to document the release history. We should be as eager to keep it up-to-date as any other part of the docs.

We also need to make a decision about how to refer to the standard in the documentation. Having a chapter of the docs just say "this chapter lives over here now" isn't great. I think we should keep the root-directory Markdown file as-is, because there will be lots of links to it on GitHub that will appreciate the redirect, but in the [docs themselves](https://open-forcefield-toolkit--949.org.readthedocs.build/en/949/users/smirnoff.html) we could:

1. Remove the page from the table of contents (make it an orphan) and otherwise leave it as is - this would prevent breaking any existing links to the spec in the docs, but its a random markdown file that has to live there forever, and we would need to find somewhere else in the documentation to link people to the spec (probably the top level page).
2. Load the specification from the standards repository when we build the docs to render it in the documentation. This would create a link between versions of the Toolkit and of the spec, which may or may not be what we want. I would suggest making this link explicit - ie, we could have a disclaimer at the top of the SMIRNOFF page saying "*This version of the Toolkit implements version x.x.x of the SMIRNOFF Spec, reproduced below*"
3. Leave the spec page as a stub redirecting to the standards repo. If we do this, we might want to consider adding the version of the spec that a particular version of the toolkit implements to the redirect page.

mattwthompson:
> I would be in favour of updating the links in the changelog.

No objection here - sounds good to me.

> And maybe more importantly, the purpose of the changelog is not to be a fixed-in-stone record of the state of the changelog at release time, but to document the release history. We should be as eager to keep it up-to-date as any other part of the docs.

I think I take the other perspective in general, partially because updating the changelog makes it harder to understand the chnages in a release as they were made in that specific release - 

> We should be as eager to keep it up-to-date as any other part of the docs.

- and also because of how the overhead scales with number of releases/size of release notes.

Part of why I'm fine updating it here is that moving where the spec is hosted shouldn't change its content. I don't know if this is always the case. Not sure of a good example to highlight this, maybe something like a link to an example that was later removed or heavily modified. This is all in a grey area I haven't thought of much and I'd be curious to hear your thoughts.

j-wags:
Ok, these are really good questions and now my head hurts.

(per live discussion in our 2021_09_29 meeting) The following decisions:

We'll
* Still have a SMIRNOFF spec page on the top-level of the docs. This page will: 
    * have content added to effective be a page on "the relationship between the toolkit and the spec"
    * Prominently link to the current spec in the first screen of the page (for people who just want to click through)
    * mention "hey, the spec used to live here, but now it doesn't because X"
    * mention "the toolkit will try to be reverse-compatible, but can't be forward compatible. If you try to use an old version of the toolkit to load a new version of the spec you'll get an error"
* Either do or don't retroactively change the releasenotes. I'll accept this PR in either state -- This is a big question for marginal gain so I think we should do whatever's easiest.

@Yoshanuikabundi Could you 
* get this PR to a merge-able state and ping me for review/merging
* open a separate PR to populate the page content described above

j-wags (approved):
Looks great. Thanks for taking this on, @Yoshanuikabundi.

Issue #536, opened 2020-03-04
Adding  lur_cache to toolkits


jthorton:
**Is your feature request related to a problem? Please describe.**
We recently discussed on slack the inconsistent method decorators in backend toolkits which varied between `@staticmethod` and `@classmethod`. I think if we are able to switch to using instance methods where possible this should allow for the use of the `lur_cache` allowing the toolkit wrappers to quickly process molecules they have seen before with the same settings which could help improve performance.

This does slightly complicate the choice of hash for the molecule as this should not be a function of a method that can be in a cache as we will hit an infinite recursion loop I think (if someone knows how to get around this that would be great). My thinking is that if we have the current hash which is based on smiles and wrap the `to_smiles` method with a `lur_cache` then the cache will need to hash the molecule to check if it has seen it before which will call the `to_smiles` method again and so on. Ideally the hash could be a function of something that does not depend on the toolkits, possibly a hash of the networkx graph which should be unique however they will be order dependent for example:

```python
from openforcefield.tests.test_forcefield import create_ethanol, create_reversed_ethanol
ethanol = create_ethanol()
ethanol_reversed  = create_reversed_ethanol()
hash(ethanol.to_networkx())
```
> 8752032379945
```python
hash(ethanol_reversed.to_networkx())
```
> 8752032605205

InChIKey could be a good use for this as it will not depend on the order and will be consistent between which toolkit calculated it but that will have the same problems as with smiles. 

**Describe the solution you'd like**
It would be great to see all toolkit methods become instance ones which also would remove the small cost of having to init the wrappers repeatedly when running some operations and to add a cache to most if not all backend toolkit methods.

Issue #135, opened 2018-11-06
Preserve Molecule.properties through RDMol/OEMol conversions


j-wags:
Make sure that all Molecule properties (partial charges, Wiberg orders, conformers, etc) survive conversion to and from OEMol and RDMol.

Determine whether the same properties survive filo I/O in supported file formats.

Determine whether it's possible to preserve arbitrary Molecule.properties (anything serialize-able the user puts in the molecule.properties dictionary) through the above conversions.

j-wags:
Also, test specifically that charges are read from MOL2's

j-wags:
### Atom Core Property Mapping Sets
(Work in progress)

|   |   OFFAtom | OEAtom | RDAtom |
|--|------------|---------|---------|
| No name | Atom.name = "" | oeatom.SetData("name","") | rdatom.SetData("name","") ?|
| element | 
| formal charge | 
| is_aromatic | 
| stereochemistry |

j-wags:
### Bond core property mapping spec
| | OFFBond | OEBond | RDBond|
|-|-|-|-|
| No atom1 | Not permitted | N/A | N/A |
| No atom2 | Not permitted | N/A | N/A |
| No bond_order | Not permitted | N/A | N/A |
| No fractional_bond_order | None | oebond.HasData('fractional_bond_order') returns False | rdbond.HasProp("fractional_bond_order") returns 0 |
| is_aromatic | 
| No stereochemistry | None | oebond.HasStereoSpecified() returns False **OR** oechem.OECIPBondStereo_NotStereo **(This will require further inspection)** |

j-wags:
Meeting notes, also posted to Slack:

2018_11_07 Jeff Caitlin Chat
- There's a jupyter notebook documenting a OEmol --> RDmol --> OEmol conversion in the SMIRNOFF channel
- Relevant conversation in SMIRNOFF: https://openforcefieldgroup.slack.com/archives/C4VHEFXS5/p1530049256000370
- Relevant GIST of a jupyter notebook https://gist.github.com/bannanc/810ccc4636b930a4522636baab1965a6
- Caitlin is 95% sure that the GIST implementation is robust in all cases, but it would be good if I did some deep digging to ensure that we have it implemented the same way, and that the implementation is watertight.

j-wags:
[...]

PR #951, opened 2021-05-25
[WIP] Feature branch: Topology Biopolymer Refactor 

### LINKED ISSUE #946: Topologies cannot be instantiated from other topologies
**Describe the bug**

Trying to create a `Topology` from another `Topology` instance yields an excpetion due to a non-existant method

**To Reproduce**

```
from openff.toolkit.topology import Molecule, Topology
Topology(other=Topology.from_molecules(Molecule.from_smiles("C")))
```

**Output**

```
  File "/XXX/openff-toolkit/scratch/at-am1/run.py", line 160, in <module>
    main()
  File "/XXX/openff-toolkit/scratch/at-am1/run.py", line 110, in main
    Topology(other=Topology.from_molecules(Molecule.from_smiles("C")))
  File "/XXX/openff-toolkit/openff/toolkit/topology/topology.py", line 1321, in __init__
    self.copy_initializer(other)
AttributeError: 'Topology' object has no attribute 'copy_initializer'
```

 - Operating system

**Additional context**

j-wags:
Feature branch plans: https://openforcefield.atlassian.net/wiki/spaces/IN/pages/1579122724/2021+Topology+Refactor

# Do not edit this branch directly -- All changes should go through PR review!

## Notes and to-dos inherited from #881 

Adds `cachetools` as a requirement

- [x] Addresses #536 
- [x] (temporary, should be replaced by a more general solution) Adds a lightweight, index-sensitive hash method for molecules (`FrozenMolecule.ordered_connection_table_hash`)
- [x] Adds caching for `to_openeye` and `to_rdkit` using `cachelib`
- [x] Adds caching for `Atom.molecule_atom_index`, since the data structure makes self-index lookup an O(n_atoms_in_molecule) operation. `Molecule._invalidate_cached_properties` now recursively deletes this cached property on the molecule's Atoms. 
- [x] Removes `@classmethod` designation from `OpenEyeToolkitWrapper.to_openeye` and `RDKitToolkitWrapper.to_rdkit` (I don't see a need for them to be classmethods, this was probably from before when I knew what I was doing)
- [x] Makes `Molecule.to_openeye` and `to_rdkit` accept a toolkit registry kwarg. It's unclear why they didn't before (a user could reasonably want to divert `Molecule.to_rdkit` through a custom ToolkitWrapper)
- [x] Incorporate suggestions from #992 
    - [ ] appropriate cached molecule hash invalidation (eg when an atom/bond is added)
    - [ ] appropriate cached atom index invalidation (necessary? Atoms can only be appended at present)
    - [ ] cache for `to_openeye` / `to_rdkit` actually detects when a aromaticity model is requested
    - [ ] test that operations on cached outputs don't mutate the object in the cache

## Notes and to-dos inherited from #973 

Remaining things that this doesn't do are:
* [ ] Standard caps
    * Partial support via #1072 . We need a way to programatically extract the information from CCD or something similar.
* [x] uncapped C terminal 
* [x] Main-chain prolines
    * [ ] (possibly related) ~Figure out what's up with the proline substructure that gets an `N-` in it~ I'm pretty sure this is a problem with `components.cif`

## Notes and to-dos inherited from #982 
- [ ] Add `pycifrw` as dependency to some recipes
- [ ] Add tests for substructure dictionary creation

## Notes and to-dos from #1072 
- [ ] ~Better way to deal with strict chirality. Check [code dev notes](https://openforcefield.atlassian.net/wiki/spaces/IN/pages/2123956225/2021-09-20+Core+Developers+meeting+notes)~ We won't do this - A partnering company has asked us to [remove stereomarks from protein FFs.](https://github.com/openforcefield/amber-ff-porting/issues/38)

## Notes and to-dos from #1085
- [x] Implement `Molecule`-level `HierarchySchemes` and `HierarchyElements`, providing ways to generate a static index where atoms are grouped according to their common metadata.
- [ ] In our original API design/plan and there is an option where we want to initialize a Molecule with some hierarchy flavor.
- [ ] Add an example that both details how the default behavior is implemented and also implements custom hierarchy scheme.
- [x] Refactor to make Topology objects consist of explicit copies of `Molecule`s, deprecating `TopologyMolecule`, `TopologyAtom`, `TopologyVirtualSite`, etc
- [x] Implement access to `HierarchySchemes` from the Topology level.
- [x] Implement `to`/`from_dict` methods for Topology
[protein_param_test.tar.gz](https://github.com/openforcefield/openff-toolkit/files/6201573/protein_param_test.tar.gz)

## Notes and to-dos from #1097 
- [x] Refactor to make Topology objects consist of explicit copies of `Molecule`s, removing `TopologyMolecule`, `TopologyAtom`, `TopologyVirtualSite`, etc
- [x] Refactor `test_topology.py` to use basic pytest features
- [x] Implement access to `HierarchySchemes` from the Topology level, using `offtop.hierarchy_iterator` (I'm happy to change the name of this in a future PR)
- [x] Implement `to`/`from_dict` methods for Topology
- [x] Add `Molecule.particle_index(particle)`, `atom(index)`, `atom_index(atom)`, `virtual_site(index)`, `virtual_site_index(vsite)`, `virtual_site_particle_start_index(vsite)`, and `bond(index)`. 
- [x] Add `Topology.atoms`, `atom_index(atom)`, `particle_index(particle)`, `virtual_site_particle_start_index(virtual_site)`, `molecule_index(molecule)`, `molecule_atom_start_index(molecule)`, `molecule_virtual_particle_start_index(molecule)`
- [x] Fixes #946 (but SHOULDN'T close it until the refactor is merged to `master`)
- [x] Implement duplicate-molecule-detection to save time in `ToolkitAM1BCCHandler` (done in #1140)
- [ ] Implement duplicate-molecule-detection to save time during WBO assignment
- [ ] Test speed of atom/particle indexing operations on new `Topology` class  and implement heuristics to recover original performance
- [x] Have all deleted API points now either redirect to a equivalent new API point or raise a descriptive error
- [ ] Decide whether a topology should have an aromaticity model, or if it's redundant with the Molecule and `ForceField` aromaticity models
- [x] Ensure that all API points for permutations of `{Molecule,Topology}.{atom,particle,virtual_site}{,s,_index}` and counters like `n_atoms`are exposed and tested.
- [ ] Wire up residue and chain information to play nice with `Topology.{to,from}_openmm`

## Notes and to-dos from #1105 
- [x] Prototype implementation of reading a molecule from a PDB consisting of entirely "standard" components and assigning bond orders and formal charges from `aa_variants.cif` (and some manually-added caps)
- [ ] Fix issue where our peptide bond SMARTS could mislabel PRO ring-closing carbon as CA (maybe go over metadata assignment substructures in some sort of order that puts peptide+disulfide bonds last, and make earlier matches take precedence)
- [x] Code cleanup
- [x] More extensive testing
- [x] Ensure that `perceive_residues` still works with newly-generated substructure files
- [x] Resolve whether the same file can be used for BOTH metadata assignment (`perceive_residues`) AND bond order/formal charge assignment (`from_pdb`)
    - It can't -- The atom-naming dict needs to remove imidazole+guanidinium bond orders and formal charges to catch resonance states. The chemistry-assigning dict needs to know that exact information.
- [x] Make stereo warnings quiet (maybe don't do from_rdkit at all and just make offmol directly in OFFTK API?)
    - Loading process now ingests coordinates and perceives stereo from that.
- [x] Remove direct rdkit dependence from `molecule.py`
- [ ] Add extensive tests for PDB loading (take PDBs from PLBenchmarks, COVID moonshot?, other sources?)
- [ ] Add logic to gracefully error for unrecognized residues in PDB (like TPO)
- [ ] Add API points to support adding additional residues to the substructure dict
- [ ] ~Handle loading PDBs with multiple components (even if they're a mix of polymer and small molecules) (do we really want to support this? It's easy to split PDBs)~ This is out-of-scope for the 0.11.0 release
- [ ] ~Offer API point for residue substructure perception when loading OpenMM topologies~ This is out of scope for the 0.11.0 release
- [x] Fill `residue_name`, `residue_number` metadata
- [x] Fill `atom_name` metadata
- [x] Add tests for metadata loading
- [ ] Test roundtrips to/from PDB (Guarantee that we keep atom name, atom order, atom number(maybe not this one, if original indexing doesn't start at 0?), atom name, residue name, residue number, chain name, element, coordinates)
- [ ] Make sure `perceive_residues` assigns appropriate residue numbers at disulfide bridges (numbering should increment at each peptide bond, NOT jump disulfide bonds)
- [ ] Schedule meeting with RCSB to talk about ways to improve `aa-variants` file

## Notes and to-dos from #1140 
- [x] Add functionality to efficiently group identical molecules
- [x] Use new functionality to recover SMARTS matching performance
- [ ] Polish `Topology.identical_molecule_groups` docstring so actual humans can read it
- [x] Use new functionality to deduplicate molecules in special charge assignment cases (like ToolkitAM1BCC and friends)

## Notes and to-dos from #1195
- [x] Refactor PDB loading into toolkit wrappers
- [x] Refactor (`to`/`from`) (`openmm`,`rdkit`,`openeye`) to handle hierarchy metadata
    - [x] Document spec for hierarchy info transfer, edge cases, and roundtrip reliability
    - [ ] Document spec for other info transfer (basically get #135 out of issue tracker and into docs)
    - [x] Update docstrings to give basic info about hierarchy info transfer spec
    - [ ] Make public API point for loading custom substructure dicts
    - [ ] (maybe?) Don't assign info to non-tagged atoms in substructure SMARTS

## Notes and to-dos from #1285
- [x] Cherry pick vsite refactor
- [ ] (blocking) Ensure that tests from `test_parameters` highlighted at [this link](https://github.com/openforcefield/openff-toolkit/blob/9b319c0d4ae931dbecce8038a35c95766c77868b/openff/toolkit/tests/test_parameters.py#L2445-L2772) get ported to Interchange
- [ ] (blocking) Re-enable vsite_showcase notebook and get it to a working state


openff/toolkit/utils/toolkits.py
```diff
@@ -168,6 +169,10 @@ class AntechamberNotFoundError(MessageException):
 # UTILITY FUNCTIONS
 # =============================================================================================
 
+def mol_to_ctab_and_aro_key(self, molecule, aromaticity_model=DEFAULT_AROMATICITY_MODEL):
+    return f"{molecule.ordered_connection_table_hash()}-{aromaticity_model}"
```
j-wags, line 173: From @connordavel: Could add a kwarg here for `safe_caching` which would default to `True`. If it's true, this would use `to_dict` as a slow, but safe molecule hash. Otherwise it would use this janky connection table hash for speed (but it loses metadata, so most people won't want it)

openff/toolkit/topology/molecule.py
```diff
@@ -36,21 +36,21 @@
...
+from collections import OrderedDict, UserDict, defaultdict
 from copy import deepcopy
 from typing import Optional, Union
+from itertools import chain
 
+
+import json
 import networkx as nx
 import numpy as np
+from cached_property import cached_property
```
mattwthompson, line 48:
```suggestion
from functools import cached_property
```
?

j-wags: This may be related to `cached_property` only being available in 3.8

mattwthompson: Looks like you're usiing https://pypi.org/project/cached-property/ I'm not sure if https://pypi.org/project/backports.cached-property/ would be better, and the answer comes from how each works with 3.8+; a solution in the standard library should be used over a third-party backport. I have not looked into this but it should be a quick check.

mattwthompson:
With the T4 lysozyme, I'm getting a single atom with missing metadata, causing issues with export:


In:
```python3
import re
from collections import Counter

from openff.toolkit.topology import Molecule, Topology
from openff.units import unit
from simtk import unit as simtk_unit
from simtk.openmm.app import PDBFile

from openff.interchange.components.interchange import Interchange
from openff.interchange.drivers import get_gromacs_energies, get_openmm_energies
from openff.interchange.utils import get_test_file_path


def _fix_openff_impropers(force_field: ForceField):
    impropers = force_field.get_parameter_handler("ImproperTorsions")
    for imp in impropers.parameters:
        imp.smirks = re.sub("(:[23])(.+)(:[23])", r"\3\2\1", imp.smirks)

    return force_field


protein = Molecule.from_file(get_data_file_path("proteins/T4-protein.sdf"))
protein.perceive_residues()

top = protein.to_topology()

print("Counter(Length of metadata dict : number of atoms)")
print(Counter([len(a.atom.metadata) for a in top.topology_atoms]))

print("The atom with no metadata:")
for top_atom in top.topology_atoms:
    if len(top_atom.atom.metadata) == 0:
        print(top_atom)

```

Out:

```
Counter(Length of metadata dict : number of atoms)
Counter({3: 2633, 0: 1})
The atom with no metadata:
TopologyAtom 2633 with reference atom <Atom name='' atomic number='8'> and parent TopologyMolecule <openff.toolkit.topology.topology.TopologyMolecule object at 0x11993baf0>

**Comment by j-wags** [comment]
> With the T4 lysozyme, I'm getting a single atom with missing metadata, causing issues with export:

That's unfortunately expected -- It's an issue with the substructure dictionary missing coverage for the protein's C-terminal that we're working to fix.

**Comment by mattwthompson** [comment]
I've been using `Topology.atom_index(atom)` and `Molecule.atom_index(atom)` often, but I've noticed it's much slower to reach back to get the index of a known atom than just keep an index while iterating over the atoms. Minimal reproduction, using IPython magic that can be pasted into Jupyter:

```
In [1]: from openff.toolkit.topology import Topology, Molecule

In [2]: top = Topology.from_molecules([*100 *[Molecule.from_smiles(100 * "CC")]])
   ...:
   ...: def func1(topology):
   ...:     return [(atom, topology.atom_index(atom)) for atom in topology.atoms]
   ...:
   ...: def func2(topology):
   ...:     return [(atom, i) for i, atom in enumerate(topology.atoms)]
   ...:

In [3]: %%timeit
   ...: func1(top)
   ...:
   ...:
1.04 s ± 194 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: %%timeit
   ...: func2(top)
   ...:
   ...:
10.9 ms ± 2.42 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

1 second here and there doesn't matter, but I'm sometimes calling this many times while writing out files and this can compound to be most of the call time. Is there anything inherently dangerous about `func2`? Assume that I'm not adding atoms or molecules during iteration.

**Comment by j-wags** [comment]
Hm, `Molecule.atom_index` caches, so it should be near-instantaneous. `Topology.atom_index` doesn't, so it should indeed be slower. There's a subtle-yet-significant change we could make to fix this (have a Molecule know which Topology it's part of), which would improve performance but potentially add a lot of complexity.

Separately - The `enumerate(offtop.atoms)` line should work for the forseeable future. I think in some places we literally declare "the index is defined by the order in which things come out of the iterator", so we'll be VERY loud if the day ever comes that that isn't true.

**Comment by mattwthompson** [comment]
See #1142 for a number of changes now brought into this branch

**Comment by mattwthompson** [comment]
Minor detail, I noticed that `Atom.mass` is now unitless:

```
(master) $ python
Python 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 20:33:18)
[Clang 11.1.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from openff.toolkit.topology import *
>>> Molecule.from_smiles("CCO").atoms[0].mass
Quantity(value=12.01078, unit=dalton)
(master) [openff-system] conda deactivate
(base) [openff-system] conda activate interchange-topology
(interchange-topology) [openff-system] python
Python 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 20:33:18)
[Clang 11.1.0 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from openff.toolkit.topology import *
>>> Molecule.from_smiles("CCO").atoms[0].mass
12.011
```

**Comment by mattwthompson** [comment]
I'm finding a performance regression in `Topology.from_openmm`, though I haven't had time to profile it yet.


```python3
import time

import numpy as np
from openff.toolkit.topology import Molecule, Topology

for n_molecules in np.logspace(0, 2.5, num=10, dtype=int):
    topology = Topology.from_molecules(n_molecules * [molecule])
    openmm_topology = topology.to_openmm()
    start_time = time.time()
    Topology.from_openmm(openmm_topology, unique_molecules=[molecule])
    print(n_molecules, time.time() - start_time)
```

```
0.6627719402313232
1 0.698512077331543
3 3.0817441940307617
6 4.673622131347656
12 8.500097036361694
24 22.013742923736572
46 47.690475940704346
87 73.42918682098389
166 154.51772904396057
316 233.0623791217804
```

Which looks roughly linear in time, so at least it's not obviously scaling poorly.

image


I use this method a lot for parsing existing files and general tomfoolery. I could switch to using a different library for file parsing, but I'd need to make a new interface to get things into our object models. This is also a heavily-trafficked path, so I think it's justified on its own. I may get to this myself, but probably not until January.

jchodera:
@mattwthompson I'd like to put in a general request to *always label your axes in plots* to improve clarity.

I think this is "execution time (seconds)" vs "number of residues". That does indeed take a lot of time!
If you have a chance to profile just the largest system case, it would help us identify what is going on!

mattwthompson:
I noticed that one of the data structures in `Topology.identical_molecule_groups` seems to be a list that's always length 2, and looks a _lot_ like a dictionary; its first element is always an integer (key) and its second element is a dictionary (value). Am I missing a reason this should be a list, or could we make a dictionary? What I'm suggesting is that [this](https://github.com/openforcefield/openff-toolkit/blob/3a81f9efc0997d8661da534589405a52fdd3c19e/openff/toolkit/topology/topology.py#L1220) docstring should go from


```python3
{
    0: [
        [0, {0: 0, 1: 1, 2: 2}],
        [1, {0: 1, 1: 0, 2: 2}],
    ]
}
```

```python3
{
    0: [
        {0, {0: 0, 1: 1, 2: 2}},
        {1, {0: 1, 1: 0, 2: 2}},
    ]
}
```

j-wags:
Oh, I agree with you that we could make that a dict, assuming you meant to write:
```
{
    0: {
        0: {0: 0, 1: 1, 2: 2},
        1: {0: 1, 1: 0, 2: 2}},
    }
}
```

Feel free to open a PR!

I'm curious about the `Topology.from_openmm` slowdown. Probably something to do with `atom.topology_atom_index` caching if I had to guess... I'll look deeper into this after my next PR.

mattwthompson:
I ran a non-exhaustive, not-optimally-rigorous benchmark of loading some PDB files, using only those in our [protein-ligand benchmarking](https://github.com/openforcefield/protein-ligand-benchmark) repo. The script I used and results are below. The summary is that 18/20 files were loaded and most took roughly 1 minute (ranging from 16 to 116 seconds). I used the most recent commit with debugging print statements removed. This is not rigorous, but indicates that
* Given a dataset of well-formed PDB files, the toolkit will load most of them. I'm not defining "well-formed" nor are we able to pin down that number to something more quantitative, but we can have some confidence that the number is more likely to be 90% than 10%.
* Performance is something on the order of 5-10 residues per second

https://gist.github.com/mattwthompson/678e26f13cef12328d14680fc985fd16

mattwthompson:
This can probably be closed now that all of these changes are in the default branch.

mattwthompson:
I think we're forced to "close" this but it has been implicitly merged in - the default branch is now `main`, which was seeded from `topology-biopolymer-refactor`. (Updating the base branch to `main` would effectively move us back a month or two in time.) Changes that would have gone into this PR are now going straight into `main`, just as planned, and landing in beta/RC builds of version 0.11.0.

Of course, please re-open if I'm wrong! But I don't see any other ending here.

j-wags:
You served us well, giant PR! 

image

PR #955, opened 2021-05-26
Force field modification example refresh

Yoshanuikabundi:
Working through examples to refresh the suite. This refresh is an adaptation of the 2nd workflow of the @j-wags original Toolkit Showcase notebook (#763, ee506308 [2nd workflow](https://github.com/openforcefield/openff-toolkit/blob/ee506308c390a2fae3f1289652f47ad0717e9779/examples/toolkit_showcase/toolkit_showcase.ipynb)), spruced up with some more narrative and a bit more motivation and explanation.

[📒 Rendered on Binder](https://mybinder.org/v2/gh/openforcefield/openff-toolkit/ffmod-example-refresh?filepath=examples%2Fforcefield_modification%2Fforcefield_modification.ipynb)


j-wags:
SMARTS spec that we use (would be good to link in the text): https://www.daylight.com/dayhtml/doc/theory/theory.smarts.html

j-wags:
* For the cyclic nitrogen angle smirks, let's define it as a variable before we reuse it in the lower code block. 
* We can drop the part from the old example, since nonbonded cutoffs are less interesting.
* Re: The TODO about SMIRKS uniquely identifying a molecule -- It doesn't in theory uniquely ID a molecule, because the undefined formal charges and bond orders could allow another mol to match. For example, ammonia's SMARTS could be `[N:1]([H:2])([H:3])([H:4])`, but this would also match ammonium, because the `N:1` part will also catch a positively charged N.
* `Convince yourself that this SMIRKS code applies to all of the H-X-H angles in our molecule.` -- While this is true in this case, in general _many_ angle SMARTS could match a particular group of three bonded atoms, and the order that they're defined in the force field would define which one of them gets used in the System. So from looking at the SMARTS by itself we can't be confident that it would be assigned, since another term further down in the FF could match instead. 
* Can we find a more interesting molecule to work with, and/or more interesting changes to make? If we want a minimal one with a bunch of unique parameters, we could do something like:
```
mol = Molecule.from_smiles('[N-]=[N+]=N[C@H]1NCOC(F)=C1')
mol.generate_conformers()
mol.visualize(backend='nglview')
```

(might also be worth trying zidovudine `CC1=CN(C(=O)NC1=O)[C@H]2C[C@@H]([C@H](O2)CO)N=[N+]=[N-]` since that has a nitrile we can play with)

I'm happy to help with figuring out which FF terms to mess with if you do want to change the molecule. I'm also fine if we don't want to change the molecule and move on to other more urgent work. It'll also be good to have this example show creating a molecule from SMILES, since the toolkit showcase shows loading from SDF.

Yoshanuikabundi:
I added some more hydroxyl groups so the accidental linearisation of the hydroxyls jumps out more, and changed the first step to mess with the fluorine-carbon bond rather than the hydroxyl angle. I feel like I could spend an unlimited amount of time choosing a more perfect molecule and not actually improve the example much, so I haven't put much energy into it. It would be good if we could get some real user testing and feedback!

j-wags (approved):
I think this is just about ready to go. Thanks for all the thought you've put into this @Yoshanuikabundi! 

Before you merge, could you
- [ ] Remove the SDF file
- [ ] Update the README to reflect the new content (and include the toolkit showcase)

To fix the OpenEye tests, we should avoid using fixed atom numbering when referring to molecules built from SMILES. Instead we can replace those instances using snippets like:

```
fluorine = [atom for atom in ligand.atoms if atom.element.symbol == 'F'][0]
carbon = [neighbor_atom for neighbor_atom in fluorine.bonded_atoms][0]
print(ff_applied_parameters[0]['Bonds'][(fluorine.molecule_atom_index, carbon.molecule_atom_index)])
```


```
# Label the three atoms that should be a part of our angle
COC_motifs = ligand.chemical_environment_matches('[C:1][O:2][C:3]')
print(COC_motifs)

ligand.properties['atom_map'] = {COC_motifs[0][0]:1, 
                                 COC_motifs[0][1]:2, 
                                 COC_motifs[0][2]:3}
# Generate the SMIRKS code (any SMILES is a SMIRKS for a specific molecule)
smirks = ligand.to_smiles(mapped=True)
smirks
```


I agree that these changes may remove the need to use the `ligand.to_smiles(mapped=True)` trick. The API around atom maps is unstable and will have reverse-incompatible changes soon, so if we need to drop that bit to get OE tests passing, it's probably for the best.

j-wags:
Looks like that fixed the failing tests - Merging!

PR #956, opened 2021-05-26
Add `ForceField.get_partial_charges()` and `Molecule.assign_partial_charges()` methods

### LINKED ISSUE #945: Add ForceField.assign_partial_charges

Users often want to see which partial charges would be assigned to a `Topology` or `Molecule` by a given `ForceField` object. However, this isn't directly accessible. Instead, users have to run:

```
ff = ForceField('openff-1.3.0.offxml')
sys, ret_top = ff.create_openmm_system(mol.to_topology(), return_topology=True)
[*ret_top.reference_molecules][0].partial_charges
```

We could make a convenience method that just runs the above and returns the partial charges. 

```
class ForceField:
    ...
    def assign_partial_charges(molecule, toolkit_registry=None)
```

We could also modify the existing `Molecule.assign_partial_charges` to accept a `ForceField` as an argument (which would be exclusive with accepting a `partial_charge_method` or `use_conformers` kwarg)

The name might be easy to confuse with `Molecule.assign_partial_charges`. I don't know whether this is good or bad. Also, we may think about how we can change the name to keep people from thinking this is a necessary prerequisite to running `create_openmm_system`.

One tricky thing with the above implementation is that it will ALSO run through valence parameter assignment, and if any of those are missing, it will raise an error instead of returning the partial charges. This could by circumvented by knowing _ahead of time_ which handlers can assign partial charges, and only running those. However this conflicts with the current plugin architecture, since a user could define a new charge method and we wouldn't know about it in time to include it in this list.

Yoshanuikabundi:
In refreshing the Force Field Modification example (#955) I found that a convenient method to get the charges a force field would assign a molecule would be useful. These are (will be) those methods. Closes #945.

- [x] `ForceField.get_partial_charges()`
    - [x] Implementation
    - [x] Tests
    - [x] Docstrings
- [x] `Molecule.assign_partial_charges()`
    - [x] Implementation
    - [x] Tests
    - [x] Docstrings


openff/toolkit/tests/test_forcefield.py
```diff
@@ -4907,6 +4907,32 @@ def test_read_smirnoff_spec_freesolv(
         )
 
 
+class TestForceFieldGetPartialCharges:
+    def get_partial_charges(self, mol, ff):
```
j-wags, line 4911: Took me a sec to understand the use of this method. Since it's not in the public API we can make the name really big. 

```suggestion
    def get_partial_charges_from_create_openmm_system(self, mol, ff):
```

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1505,6 +1505,34 @@ def _get_parameter_handler_class(self, tagname):
...
+        Returns
+        -------
+        charges : simtk.unit.Quantity with shape (n_atoms,) with units compatible with elemental charge
+            The partial charges of the provided molecule in this force field.
+
+        """
+        _, top_with_charges = self.create_openmm_system(
+            molecule.to_topology(), return_topology=True, **kwargs
+        )
+        charges = [*top_with_charges.reference_molecules][0].partial_charges
```
j-wags, line 1533: We should do a check for whether this molecule has vsites `offmol.n_virtual_particles`. If it DOES, I'd be ok with raising an error, since the vsites may steal charge from the atoms, and it won't be clear to the user whether they're getting charges BEFORE or AFTER this adjustment.

j-wags: If you add the following OFFXML to the `test_forcefield`, you can make a test that will assign a virtualsite to ethanol

https://github.com/openforcefield/openff-toolkit/blob/18ea825c8598d02268f5607452116adb63ca6a26/openff/toolkit/tests/test_forcefield.py#L1976-L1995

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1505,6 +1505,40 @@ def _get_parameter_handler_class(self, tagname):
...
+        charges : simtk.unit.Quantity with shape (n_atoms,) with units compatible with elemental charge
+            The partial charges of the provided molecule in this force field.
+
+        """
+        _, top_with_charges = self.create_openmm_system(
+            molecule.to_topology(), return_topology=True, **kwargs
+        )
+
+        if top_with_charges.n_topology_virtual_sites != 0:
+            raise ValueError(
```
j-wags, line 1535: I'd like us to move away from re-using built-in exceptions, so let's replace this with a custom exception. You can find a template for doing this at the top of the file.

openff/toolkit/topology/molecule.py
```diff
@@ -2909,6 +2909,7 @@ def assign_partial_charges(
         partial_charge_method,
         strict_n_conformers=False,
         use_conformers=None,
+        force_field=None,
```
j-wags, line 2912: Thanks for looking into this, but I think this would introduce a lot of complexity/user surprises. So I think we should remove this in favor of only allowing users to get charges using `forcefield.get_partial_charges`.

j-wags: Note from @Yoshanuikabundi  during meeting: Add a note in the docstring pointing people to `get_partial_charges` if they do want to use a ff here.

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1505,6 +1505,40 @@ def _get_parameter_handler_class(self, tagname):
...
         return ph_class
 
+    def get_partial_charges(self, molecule, **kwargs):
+        """Generate the partial charges for the given molecule in this force field.
+
+        As any `ParameterHandler` may in principle modify charges, the entire
+        force field must be applied to the molecule to produce the charges.
+        Calls to this method from incorrectly or incompletely specified `ForceField`
+        objects thus may raise an exception.
+
```
j-wags, line 1515: from @Yoshanuikabundi Add a docstring example of using this to ASSIGN partial charges to a mol.

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1505,6 +1505,40 @@ def _get_parameter_handler_class(self, tagname):
...
+        ----------
+        molecule : openff.toolkit.topology.Molecule
+            The ``Molecule`` corresponding to the system to be parameterized
+        toolkit_registry : openff.toolkit.utils.toolkits.ToolkitRegistry, optional. default=GLOBAL_TOOLKIT_REGISTRY
+            The toolkit registry to use for operations like conformer generation and
+            partial charge assignment.
+
+        Returns
+        -------
+        charges : simtk.unit.Quantity with shape (n_atoms,) with units compatible with elemental charge
```
j-wags, line 1526:
```suggestion
        charges : simtk.unit.Quantity with shape (n_atoms,) with units compatible with elementary charge
```

j-wags (approved):
Looks great! Thanks @Yoshanuikabundi. This is really useful functionality. The only hard blocker is the removal of forcefield kwarg (and associated logic) in `assign_partial_charges` before you merge.

openff/toolkit/typing/engines/smirnoff/forcefield.py
Yoshanuikabundi: Done in 43a27a6

openff/toolkit/topology/molecule.py
Yoshanuikabundi: Done in 29fb897 and 43a27a6

openff/toolkit/typing/engines/smirnoff/forcefield.py
Yoshanuikabundi: Done in 43a27a6

openff/toolkit/typing/engines/smirnoff/forcefield.py
Yoshanuikabundi: Done in 43a27a6

PR #965, opened 2021-06-01
Implement topology.TagSortedDict for charge increment find_matches

trevorgokey:

This PR allows ChargeIncrementHandler to keep the tagged order of atoms when calling `find_matches`. This is allowed by implementing a `TagSortedDict`, which is similar to `SortedDict` but keeps the last seen permutation of a tuple of atoms. Below is the docstring for further explanation: 

```
class TagSortedDict(_TransformedDict):
    """
    A dictionary where keys are kept unsorted, but only allows one permutation of a key
    to exist. Certain situations require that atom indices are not transformed in any
    way, such as when the tagged order of a match is needed downstream. For example a
    parameter using charge increments needs the ordering of the tagged match, and so
    transforming the atom indices in any way will cause that information to be lost.

    Because deduplication is needed, we still must avoid the expected situation
    where we must not allow two permutations of the same atoms to coexist. For example,
    a parameter may have matched the indices (2, 0, 1), however a parameter with higher
    priority also matches the same indices, but the tagged order is (1, 0, 2). We need
    to make sure both keys don't exist, or else two parameters will apply to
    the same atoms. We allow the ability to query using either permutation and get
    identical behavior. The defining feature here, then, is that the stored indices are
    in tagged order, but one can supply any permutation and it will resolve to the
    stored value/parameter.

    As a subtle behavior, one must be careful if an external key is used that was not
    supplied from the TagSortedDict object itself. For example:

        >>> x = TagSortedDict({(2, 5, 0): 100})
        >>> y = x[(5, 0, 2)]

    The variable y will be 100, but this does not mean the tagged order is (5, 0, 2)
    as it was supplied from the external tuple. One should either use keys only from
    __iter__ (e.g.  `for k in x`) or one must transform the key:

        >>> key = (5, 0, 2)
        >>> y = x[key]
        >>> key = x.key_transform(key)

    Where `key` will now be `(2, 5, 0)`, as it is the key stored. One can overwrite
    this key with the new one as expected:

        >>> key = (5, 0, 2)
        >>> x[key] = 50

    Now the key `(2, 5, 0)` will no longer exist, as it was replaced with `(5, 0, 2)`.
    """
```

Implementation is done, but need to add tests. Here is an example after applying `label_molecules` on H2O with several toy parameters:
```
{'Angles': <openff.toolkit.topology.topology.ValenceDict object at 0x7fdf260863a0>,
 'Bonds': <openff.toolkit.topology.topology.ValenceDict object at 0x7fdf260864f0>,
 'ChargeIncrementModel': <openff.toolkit.topology.topology.TagSortedDict object at 0x7fdf26085df0>,
 'Electrostatics': <openff.toolkit.topology.topology.ValenceDict object at 0x7fdf26085f40>,
 'ImproperTorsions': <openff.toolkit.topology.topology.ImproperDict object at 0x7fdf26085b80>,
 'ProperTorsions': <openff.toolkit.topology.topology.ValenceDict object at 0x7fdf26085d60>,
 'ToolkitAM1BCC': <openff.toolkit.topology.topology.ValenceDict object at 0x7fdf26085bb0>,
 'VirtualSites': <openff.toolkit.topology.topology.UnsortedDict object at 0x7fdf26085eb0>,
 'vdW': <openff.toolkit.topology.topology.ValenceDict object at 0x7fdf26085cd0>}
OrderedDict([((1, 0),
              <ChargeIncrementType with smirks: [#8:2][#1:1]  charge_increment1: -0.6 e  charge_increment2: 0.6 e  >),
             ((1, 0, 2),
              <ChargeIncrementType with smirks: [#1:3][#8:2][#1:1]  charge_increment1: -0.6 e  charge_increment2: 0.6 e  charge_increment3: 0.6 e  >),
             ((2, 0),
              <ChargeIncrementType with smirks: [#8:2][#1:1]  charge_increment1: -0.6 e  charge_increment2: 0.6 e  >)])
OrderedDict([((1, 0),
              [<VirtualSiteBondChargeType with smirks: [#8:2][#1:1]  epsilon: 0.05 kcal/mol  distance: 0.7 A  charge_increment1: 0.2 e  charge_increment2: 0.0 e  type: BondCharge  sigma: 0.1 A  name: EP  match: once  >]),
             ((1, 0, 2),
              [<VirtualSiteDivalentLonePairType with smirks: [#1:1][#8:2][#1:3]  epsilon: 0.05 kcal/mol  distance: 0.5 A  charge_increment1: 0.2 e  charge_increment2: 0.0 e  charge_increment3: 0.0 e  type: DivalentLonePair  outOfPlaneAngle: 70.0 deg  sigma: 0.1 A  name: EP  match: all_permutations  >,
               <VirtualSiteMonovalentLonePairType with smirks: [#1:3][#8:2][#1:1]  epsilon: 0.05 kcal/mol  distance: 0.5 A  charge_increment1: 0.2 e  charge_increment2: 0.0 e  charge_increment3: 0.0 e  type: MonovalentLonePair  outOfPlaneAngle: 90.0 deg  inPlaneAngle: 180.0 deg  sigma: 0.1 A  name: E  match: all_permutations  >]),
             ((2, 0),
              [<VirtualSiteBondChargeType with smirks: [#8:2][#1:1]  epsilon: 0.05 kcal/mol  distance: 0.7 A  charge_increment1: 0.2 e  charge_increment2: 0.0 e  type: BondCharge  sigma: 0.1 A  name: EP  match: once  >]),
             ((2, 0, 1),
              [<VirtualSiteDivalentLonePairType with smirks: [#1:1][#8:2][#1:3]  epsilon: 0.05 kcal/mol  distance: 0.5 A  charge_increment1: 0.2 e  charge_increment2: 0.0 e  charge_increment3: 0.0 e  type: DivalentLonePair  outOfPlaneAngle: 70.0 deg  sigma: 0.1 A  name: EP  match: all_permutations  >,
               <VirtualSiteMonovalentLonePairType with smirks: [#1:3][#8:2][#1:1]  epsilon: 0.05 kcal/mol  distance: 0.5 A  charge_increment1: 0.2 e  charge_increment2: 0.0 e  charge_increment3: 0.0 e  type: MonovalentLonePair  outOfPlaneAngle: 90.0 deg  inPlaneAngle: 180.0 deg  sigma: 0.1 A  name: E  match: all_permutations  >])])
```

where I also show virtual sites, which unfortunately cannot make use of this new dictionary. Instead, because we see that Monovalent and Divalent parameters can be applied to the same atoms but different permutations, and we cannot collapse them into a single permutation. In this PR, the dictionary in use was changed from `dict` to `UnsortedDict`, which was technically an API bug. This modification to use `UnsortedDict` allows virtual sites to use the same API as the other handlers when calling `find_matches`.


openff/toolkit/topology/topology.py
```diff
@@ -214,6 +214,107 @@ def __keytransform__(self, key):
...
 
 
+class UnsortedDict(_TransformedDict):
+    ...
+
+
+class TagSortedDict(_TransformedDict):
+    """
+    A dictionary where keys are kept unsorted, but only allows one permutation of a key
+    to exist. Certain situations require that atom indices are not transformed in any
```
mattwthompson, line 224: The keys here are tuples of atom indices, correct? I first read this as if the topic was whether or not the keys in the dictionary were sorted, which confused me.

mattwthompson (commented):
I _think_ the design here is good, including storing the sorted keys inside as psuedo-duplicates. Please add some tests for the simple behavior, i.e. what's already in the docstring. (Is `_TransformedDict` explicitly tested anywhere? I couldn't find anything. Might be useful to add.)

openff/toolkit/topology/topology.py
trevorgokey: Yes, these keys are atom indices. Happy to accept modification if something isn't clear.

trevorgokey:
Let me know if it is desired to introduce an issue for this PR to tag.

openff/toolkit/topology/topology.py
mattwthompson: It probably makes sense to anybody who was already in the headspace of these dictionaries, but I didn't immediately remember what these keys store.

openff/toolkit/topology/topology.py
```diff
@@ -214,6 +214,107 @@ def __keytransform__(self, key):
         return key
 
 
+class UnsortedDict(_TransformedDict):
+    ...
+
+
+class TagSortedDict(_TransformedDict):
+    """
+    A dictionary where keys are kept unsorted, but only allows one permutation of a key
```
mattwthompson, line 223:
```suggestion
    A dictionary where keys, consisting of tuples of atom indices, are kept unsorted, but only allows one permutation of a key
```

Something like this would be useful here, or just moved up to `_TransformedDict`

openff/toolkit/topology/topology.py
```diff
@@ -214,6 +214,107 @@ def __keytransform__(self, key):
...
+    def key_transform(self, key):
+        key = tuple(key)
+        return self.__keytransform__(key)
+
+    def clear(self):
+        """
+        Clear the contents
+        """
+        self.store.clear()
+        self._stored.clear()
```
mattwthompson, line 315:
```suggestion
        self._sorted.clear()
```

mattwthompson:
No need to create an issue since you've explained it here

mattwthompson (approved):
I think this is in a good place to be merged - feel free to either squash + merge into `master` or into your other feature branch and keep working from there.

This class is "public" so I think it warrants a line in the release notes as well.

trevorgokey:
Excellent, thanks @mattwthompson! I think I am ready to merge. There is one CI test that didn't pass. If you think it is not due to this PR, I'll go ahead and merge.

mattwthompson:
> FAILED openff/toolkit/tests/test_forcefield.py::TestForceFieldParameterAssignment::test_overwrite_bond_orders

Yeah, this is a flaky one I've seen regularly over the past few months. Unrelated and not blocking - feel free to merge!

j-wags:
This should be good to merge. Is there additional context/a reason we should keep this open @trevorgokey?

trevorgokey:
I think so! I was waiting for tests to pass, then dropped the ball on merging. Looks like everything passed. I'll go ahead and merge master again, and close after the new checks pass.

PR #972, opened 2021-06-05
Theme refresh

Yoshanuikabundi:
The theme we use for the documentation could use an update.

Specific issues we want to fix:
 - Navigation is awkward, especially within pages - the only table of contents is in a dropdown from the navbar, and I'm basically certain that very few people will find it on their own. I think we'd be better off with a TOC in a sidebar, like in most Sphinx themes.
 - Having links and references and code in red all over the place is very jarring.

Opportunities:
 - Updating the theme would be an opportunity to define a unified look for documentation across OpenFF projects. I think just using the fonts and colours from openforcefield.org would result in a very snazzy, uniform look

Constraints:
 - We don't want to add any maintenance burden to the project
     - No JavaScript so that we don't open the docs up to security holes
     - Should we keep a fork of a theme in a repo for easy updates? see MolSSI: https://github.com/MolSSI/qcarchive-sphinx-theme

With this in mind, I've adopted the [sphinx-material](https://github.com/bashtage/sphinx-material) theme. I looked at a lot of themes but of the ones that I like the look of Material was by far the easiest to customise. I've made quite a number of changes - per my discussion with @j-wags, I've kept them in the toolkit repo, though they can easily be broken out into a new fork if we decide to go that way.

I think keeping a fork in a repo is the way to go. We won't ever need to worry about upstream changes and can make our own changes at our leisure. The small changes between projects will amount to a few changes in docs/conf.py to change accent colours and a logo. And we won't be missing out on anything upstream, because there's no JavaScript to worry about security holes with and we've already made all the changes we'd want bugfixes for.

## Changes in this PR

This PR is pretty maximalist. A minimalist version would just include `palette.css` and a cut down `theme.css`, and would be the default material theme with new fonts and colours. CSS is a pretty safe change, because its designed to cascade over changes upstream.

The most complex change in this PR were getting the OpenFF logos to appear correctly, which requires `header.html` and `sidebar.html` and parts of `theme.css`. Most of the complexity comes from the fact that I got both the logo for the toolkit and for the initiative to appear in the header on big screens and in the pull out draw on small screens. Changes to html templates are sketchier than CSS because they often override upstream theme changes. 

`layout.html` is purely a performance optimisation. Finally, `relbar.html` makes the current page appear in the breadcrumb top bar. All these changes will override changes upstream, and are part of the [documented configuration options](https://bashtage.github.io/sphinx-material/customization.html) of the theme - the only risk is if they do something like change the CSS framework they're using. Which would be rude.

`apiref.css` adds rules to make the API docs simpler and easier to read. They're based on the [Book theme](https://sphinx-themes.org/sample-sites/sphinx-book-theme/kitchen-sink/api/), which helpfully separates out its sphinx theming from its general web app CSS logic. This was essential as the material theme alone is extremely bare bones in the API department.

I also discovered and fixed a major bug in the documentation generation - the list of methods was printed twice, before and after the description of the __init__ method, for every class. Now its only printed once.

## Previewing the change

The PR should render correctly on ReadTheDocs, so check the statuses at the bottom of this PR for an up to date preview.

- [ ] Decide where to keep theme changes


j-wags:
I think this PR has been superseded by #979, which has the same effect but hosts the new theme in a different repo that all the OpenFF packages can share.

PR #978, opened 2021-06-09
Catch dummy atom KeyErrors in are_isomorphic

### LINKED ISSUE #977: Cannot get Molecule equality with dummy atoms
**Describe the bug**

Checking molecule equality for molecules with dummy atoms raises an error because the equality check first checks the hill_formula, which requires elements. A dummy atom (atomic number 0) does not have an element and therefore raises a KeyError.

**To Reproduce**

```python
mol = Molecule.from_smiles("[*]H")
mol == mol
```

**Output**

```
KeyError                                  Traceback (most recent call last)
<ipython-input-107-3735b8ff570b> in <module>
      1 mol = Molecule.from_smiles("[*]H")
----> 2 mol == mol

openff/toolkit/topology/molecule.py in __eq__(self, other)
   2226         # updated to use the new isomorphic checking method, with full matching
   2227         # TODO the doc string did not match the previous function what matching should this method do?
-> 2228         return Molecule.are_isomorphic(self, other, return_atom_map=False)[0]
   2229 
   2230     def to_smiles(

openff/toolkit/topology/molecule.py in are_isomorphic(mol1, mol2, return_atom_map, aromatic_matching, formal_charge_matching, bond_order_matching, atom_stereochemistry_matching, bond_stereochemistry_matching, strip_pyrimidal_n_atom_stereo, toolkit_registry)
   2576 
   2577         # Do a quick hill formula check first
-> 2578         if Molecule.to_hill_formula(mol1) != Molecule.to_hill_formula(mol2):
   2579             return False, None
   2580 

openff/toolkit/topology/molecule.py in to_hill_formula(molecule)
   3890         from collections import Counter
   3891 
-> 3892         atom_symbol_counts = Counter(
   3893             Element.getByAtomicNumber(atom_num).symbol for atom_num in atom_nums
   3894         )

~/anaconda3/envs/polymetrizer-viz/lib/python3.8/collections/__init__.py in __init__(self, iterable, **kwds)
    550         '''
    551         super(Counter, self).__init__()
--> 552         self.update(iterable, **kwds)
    553 
    554     def __missing__(self, key):

~/anaconda3/envs/polymetrizer-viz/lib/python3.8/collections/__init__.py in update(self, iterable, **kwds)
    635                     super(Counter, self).update(iterable) # fast path when counter is empty
    636             else:
--> 637                 _count_elements(self, iterable)
    638         if kwds:
    639             self.update(kwds)

openff/toolkit/topology/molecule.py in <genexpr>(.0)
   3891 
   3892         atom_symbol_counts = Counter(
-> 3893             Element.getByAtomicNumber(atom_num).symbol for atom_num in atom_nums
   3894         )
   3895 

simtk/openmm/app/element.py in getByAtomicNumber(atomic_number)
    105     @staticmethod
    106     def getByAtomicNumber(atomic_number):
--> 107         return Element._elements_by_atomic_number[atomic_number]
    108 
    109     @staticmethod

KeyError: 0

```

 - Operating system

**Additional context**

In general the KeyError leads to some unexpected behaviour. Could the toolkit return an element of `None` for all invalid atomic numbers?

lilyminium:
Fixes #977 

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -12,6 +12,8 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ### Bugfixes
 
+- [PR #978](https://github.com/openforcefield/openff-toolkit/pull/978) Adds support for
+  dummy atoms in Molecule equality checking
 - [PR #965](https://github.com/openforcefield/openforcefield/pull/965): Adds
   [`TagSortedDict`](openff.toolkit.topology.TagSortedDict),
   which is used primarily when atom indices need to be stored in a specific order, such
--- a/openff/toolkit/tests/test_molecule.py
+++ b/openff/toolkit/tests/test_molecule.py
@@ -3708,7 +3708,7 @@ class TestMolecule:
         assert isinstance(mol.visualize(backend="openeye"), IPython.core.display.Image)
 
     def test_equality_with_dummy_atoms(self):
-        mol = Molecule.from_smiles("[*]H")
+        mol = Molecule.from_smiles("[*][H]")
         assert mol == mol
 
 
--- a/openff/toolkit/tests/test_molecule.py
+++ b/openff/toolkit/tests/test_molecule.py
@@ -3707,6 +3707,10 @@ class TestMolecule:
 
         assert isinstance(mol.visualize(backend="openeye"), IPython.core.display.Image)
 
+    def test_equality_with_dummy_atoms(self):
+        mol = Molecule.from_smiles("[*]H")
+        assert mol == mol
+
 
 class MyMol(FrozenMolecule):
     """
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -3012,8 +3012,14 @@ class FrozenMolecule(Serializable):
         """
 
         # Do a quick hill formula check first
-        if Molecule.to_hill_formula(mol1) != Molecule.to_hill_formula(mol2):
-            return False, None
+        try:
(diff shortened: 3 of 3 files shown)


adalke:
This fixes the immediate problem - getting a fast-rejection path for `are_isomorphic()` to work. The underlying problem is that the underlying SMILES and SDF inputs can have atoms mapped to an atom with element 0. For SMILES, this is the wildcard atom `*`. For SD input these are often R-groups, like in https://www.ebi.ac.uk/chebi/searchId.do?chebiId=CHEBI:4986 .

OpenFF uses SimTK for this, specifically, the list in `simtk.openmm.app.element`, which has no definition for element 0.

I do not know enough about OpenFF's long-term goals. I assume it's outside of scope, because it's not a real atom. If so, then the I/O routines should have a validity check, for example, so that `from_smiles("[*][H]")` raises an exception, rather than downstream operations. With this approach, the proposed patch would not need to be accepted.

FWIW, I ran into the same underlying issue today. I was checking for structures with different atom counts between OEChem and RDKit. I used `Molecule.to_hill_formula()` to summarize the structures, only to have it fail, and have to patch around it.

That's why I think there should be a more systematic solution.

P.S. The simtk data set is out of date. Elements 112-116 use the placeholder names Uu*.  These all now have names. Eg, 112 is copernicium/Cn, adopted 2010. All elements up to 118 (oganesson/Og) are named. Support for these is almost certainly out of scope. ;)

j-wags:
Thanks for opening this @lilyminium, and I agree with you @adalke -- We _don't_ want to support dummy atoms, but the error should be raised at the time that a dummy atom is initialized instead of here. So I'm going to close this PR.

PR #985, opened 2021-06-14
Fix 984

### LINKED ISSUE #984: The default developer docs should be updated to include installing black and isort

mattwthompson:
- [x] Fix #984 
- [ ] ~Add  back `openmmforcefields`~ From Jeff -- I've undone this as we DON'T want to do this, see https://github.com/openforcefield/openff-toolkit/pull/985#discussion_r654676981
  - https://github.com/openmm/openmmforcefields/releases/tag/0.9.0
- [x] Re-organize into "base" and optional/test-only/dev-only/examples-only dependencies
  - https://github.com/conda-forge/openff-toolkit-feedstock/blob/c08b5a84c677166ba845d2c433ba989b89b70668/recipe/meta.yaml

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -26,7 +26,6 @@ dependencies:
   - openeye-toolkits
   - black
   - isort
-  - openmmforcefields
   - pyyaml
   - toml
   - bson
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -20,6 +20,9 @@ dependencies:
   - rdkit
   - openeye-toolkits
   - packaging
+  - black
+  - isort
+
     # Removed until a new release which targets openff-toolkit is made.
 #  - openmmforcefields
   - openmm
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -7,33 +7,30 @@ dependencies:
     # Base depends
   - python
   - pip
-
-    # Testing
+  - packaging
+  - openmm
+  - openff-forcefields
+  - smirnoff99Frosst
+  - numpy
+  - networkx
+  - ambertools
+  - rdkit
+  - mdtraj
+  - xmltodict
+    # Test-only/optional/dev
   - pytest
   - pytest-cov
   - nbval
   - codecov
   - coverage
-  - numpy
-  - networkx
-  - ambertools
-  - rdkit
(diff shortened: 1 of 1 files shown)


mattwthompson:
Getting these failures, which both are new (to me) and should be unrelated:

```
FAILED openff/toolkit/tests/test_links.py::test_readme_links[https://mybinder.org/badge_logo.svg]
FAILED openff/toolkit/tests/test_links.py::test_readme_links[https://mybinder.org/v2/gh/openforcefield/openff-toolkit/latest0]
FAILED openff/toolkit/tests/test_links.py::test_readme_links[https://mybinder.org/v2/gh/openforcefield/openff-toolkit/latest1]
=================== 3 failed, 46 passed, 3 skipped in 10.38s ===================
Error: Process completed with exit code 1.

File: devtools/conda-envs/test_env.yaml [hunk 1]
```diff
@@ -7,30 +7,30 @@ dependencies:
...
   - openeye-toolkits
-  - packaging
-    # Removed until a new release which targets openff-toolkit is made.
-#  - openmmforcefields
-  - openmm
-  - openff-forcefields
-  - smirnoff99Frosst
+  - black
+  - isort
+  - openmmforcefields
```
j-wags, line 29: `openmmforcefields` requires `openff-toolkit`, so this pulls in a stable `openff-toolkit` package. But the tests on `master` require new code since the last release, so the unit tests break. 

So we COULD `conda remove --force openff-toolkit-base`, but we don't even need to do _that_, since none of our unit tests require `openmmforcefields`. Only the example tests do., and they [do run the `conda remove...` line](https://github.com/openforcefield/openff-toolkit/blob/1ba11ebbafa67f7953b13f6c0af6183a5801f33e/.github/workflows/examples.yml#L99) to break the dependency loop. 
```suggestion
```

j-wags (approved):
Thanks fgor getting this started. I think this is good to go (the failing tests really DO seem to be spurious now). If you don't object to the change I made, please go ahead and merge, @mattwthompson :-)

mattwthompson:
Ugh, `openff/toolkit/tests/test_toolkits.py::TestAmberToolsToolkitWrapper::test_assign_fractional_bond_orders_openeye_installed` fails maybe 1% of runs, effectively as a false positive. It's probably a bad sign that we can't reliably get the same numbers out of what calls happen in that test, but we probably don't have the bandwidth to dig deeper into it.

Thanks for looking at this - merging now!

Issue #988, opened 2021-06-15
move each toolkit wrapper implementation into its own module


adalke:
The toolkit wrappers are in openff/toolkit/utils/toolkits.py I find that file very large (nearly 6,000 lines!) and a bit hard to understand and navigate because there are implementations for RDKit, OpenEye, and AmberTools all in the same file.

Not only is file is large, functions have internal imports (to defer importing modules that might not exist), there are classmethod and staticmethod solely to get around the stick-everything-into-a-class approach, and I get confused sometimes looking at a function like "to_object()" which is defined three times; I have to pay close attention to which one I'm looking at.

**Describe the solution you'd like**

I propose splitting the functions for the three wrapper implementation each into their own modules, like "openeye_wrapper.py", "rdkit_wrapper.py", and "ambertools_wrapper.py". There will still be an OpenEyeToolkitWrapper, RDKitToolkitWrapper, and AmberToolsToolkitWrapper class. On instantiation they will import the corresponding module, and load the appropriate functions.

I have a prototype of how that might work, at https://github.com/openforcefield/openff-toolkit/tree/toolkit-impl . It has a new ModuleToolkitWrapper(ToolkitWrapper) class that does the appropriate forwarding.

There's a new `publish()` decorator, which indicates that a given module's functions are to be public via the ToolkitWrapper class. This helps make a stronger distinction between the private API/internals and the public API.

The test_toolkits.py tests all pass with this change. I had to modify a couple of cases:

1) there is no longer a _check_license() test for OpenEye via mocks as that requires dynamic testing, while the public API to OpenEyeToolkitWrapper only has does the check at instantiation time. The new code only does that check at initial import time

2) I had to instantiate a few RDKitWrapperToolkit instances because the tests assumed a staticmethod to access some internal test functions. I left the tests as RDKitWrapperToolkit()._method_call and used "publish" to make those calls be available. A better solution would be to unit test the underlying (private) functions directly, in rdkit_wrapper.py.

I believe this layout is easier to understand and use, but I can be wrong. Take a look for yourself. 

**Describe alternatives you've considered**

Could leave it as-in.

A few months ago I explored a two-module solution, where you would import the public-facing module, and the toolkit-specific code was in a private implementation module. My thought was that the "import rdkit", "import openeye" could be pushed into the implementation modules, which are only imported if the public module said it was okay. In practice this was clumsy and hard-to-understand.

[...]

PR #993, opened 2021-06-22
Separate toolkit modules

### LINKED ISSUE #834: Split Monolithic Files

There are several files in the project which are over 5000 lines long, including `toolkits.py` and `molecule.py`. Having such large files make a) IDEs which perform static code analysis run slowly when editing the files, especially in terms of code completion and linting, b) makes it hard to find what you're looking for without scrolling over 1000s of lines, c) makes it daunting for potential new contributors.

It would be good to split these up into smaller sub-files. The module imports can be retained by exposing classes in the split files via the root `__init__.py` of a module so as not to break the API.

As a start, split the following would be a big plus:

```
- utils
    - toolkits
        - __init__.py         <- Expose all classes through here to not break API.
            | GLOBAL_TOOLKIT_REGISTRY
            | OPENEYE_AVAILABLE         <- Consider removing these as I'm not sure what they actually do?
            | RDKIT_AVAILABLE
            | AMBERTOOLS_AVAILABLE
            | BASIC_CHEMINFORMATICS_TOOLKITS  <- Remove the warning as it's never triggered?
        - exceptions.py         <- Maybe add a module for the general custom exceptions?
            | MissingPackageError         <- Delete as it is unused?
            | ToolkitUnavailableException
            | LicenseError
            | InvalidToolkitError
            | InvalidToolkitRegistryError
            | UndefinedStereochemistryError
            | ChargeMethodUnavailableError
            | IncorrectNumConformersError
            | IncorrectNumConformersWarning
            | ChargeCalculationError
            | InvalidIUPACNameError
        - toolkits.py
            | ToolkitWrapper
            | BuiltInToolkitWrapper
            | ToolkitRegistry
        - ambertools.py
            | GAFFAtomTypeWarning
            | AntechamberNotFoundError
        - rdkit.py
            | RDKitToolkitWrapper
        - openeye.py
            | OpenEyeToolkitWrapper
```

similar attention should likely be given to `molecule.py` and potentially `topology.py`.

adalke:
This PR breaks up the monolithic "toolkits.py" into several submodules:
```
ambertools_wrapper.py
base_wrapper.py
builtin_wrapper.py
constants.py
exceptions.py
openeye_wrapper.py
rdkit_wrapper.py
toolkit_registry.py
```

 "Split Monolithic Files" https://github.com/openforcefield/openff-toolkit/issues/834 . See also related discussion in "move each toolkit wrapper implementation into its own module"   https://github.com/openforcefield/openff-toolkit/issues/988

- [x] Resolves #834

No tests added - this is a straight re-factoring.

A couple of docstrings were chaned to use r"aw" strings because they contained unsupported \escapes and generated warnings during test.

Black and isort both pass. I added some relative imports, so relabelled "=== GLOBAL IMPORTS ===" to "=== IMPORTS ==="

Is this needed?


adalke:
I moved all of the exceptions into its own module. Some of the existing code refers to openff.topology.toolkit.util.utils like:
```
./toolkit/topology/molecule.py:
from openff.toolkit.utils.utils import MissingDependencyError, requires_package

./toolkit/tests/test_utils.py:
from openff.toolkit.utils.utils import MissingDependencyError, requires_package

./toolkit/typing/engines/smirnoff/parameters.py: 
from openff.toolkit.utils.utils import (
    IncompatibleUnitError,
    MessageException,
     ...
```

I have not changed those places. I assume they should be using `openff.toolkit.utils` instead of `openff.toolkit.utils.utils` and that the current location is not meant as part of the public API so can be changed.

mattwthompson:
@j-wags would this be an appropriate time to reconsider the module structure? Personal preference, but I'm not a huge fan of core functionality tucked away into `something.utils.something`. Something like `from openff.toolkit.wrappers import RDKitToolkitWrapper` feels better to me, and presumably things could be wired up to point in the right places in order to maintain backwards API compatibility. (This is probably most suitable for a follow-up PR to this one). Totally fine if this would be considered too much of a change.

adalke:
I agree with @mattwthompson both on the layout change, and the proposal that it be in another PR.

openff/toolkit/utils/toolkits.py
```diff
@@ -5930,3 +174,50 @@ def __repr__(self):
                 wrapper._toolkit_name, wrapper._toolkit_installation_instructions
             )
     print(msg)
+
+
+# Consistency check that the __all__ is correct.
+def _check_all():
```
j-wags, line 180: (blocking) I'm not sure if this was accidentally committed or not.  Right now it makes it so that, if a developer added a new Exception type to `exceptions.py` and updated that file's `__all__`, they'd get an error unless they also added the new Exception to _this_ file's `__all__`. Is that the intent? 

If it's unintentional, I'm ok with this check being deleted

If it's intentional:

* Is it worth having this check at all / do we WANT to force this file's `__all__` to be a superset of all the imported files `__all__`s? 
* Would be be better (or possible) to put this in `test_toolkits` instead of here?
* If this is the pattern we want, would this be generally satisfiable by using `from exceptions import *`?

j-wags (changes requested):
I think this looks pretty good @adalke! There's an added code block in here that I thought about for a while, but either I don't understand it or it's unintentional, so let's resolve that before merging. 

One neat formatting feature -- In the description text for this PR, you can add a line like `- [x] Resolves #834`, and github will automatically close the issue when this PR gets merged, leaving a link to the PR itself for future records.

> A couple of docstrings were changed to use r"aw" strings because they contained unsupported \escapes and generated warnings during test.

Excellent. Thanks! 

> >     Update changelog
> Is this needed?

You're right -- It's not.

> I assume they should be using openff.toolkit.utils instead of openff.toolkit.utils.utils and that the current location is not meant as part of the public API so can be changed.

Hm, I haven't really thought about whether the longer import paths shouldn't be considered part of the public API. Stylistically I agree that importing from `X.utils.utils` in our code looks bad, and that we should use the shorter paths throughout the codebase. But generally, is there a situation where we'd benefit by considering the longer paths to be outside the public API?

> would this be an appropriate time to reconsider the module structure?

The big motivator to do that in this PR would be if that would change a lot of lines again. I'd like to minimize the huge-number-of-changed-lines PRs, since those require a lot of my time to meaningfully review. But as I understand it, switching the modules to a different folder would be file _moves_, not file _edits_, so the line diff shouldn't be huge (just some changed imports). So those can be a different PR

Also, the changes in that move may be overridden by a refactor from `ToolkitWrappers` to `Toolbox`es depending on how we choose to proceed with @adalke's proposals. So let's hold off on that until we have the other moving pieces settled.

adalke:
> There's an added code block in here that I thought about for a while, but either I don't understand it or it's unintentional, so let's resolve that before merging.

It's a mistake. It was there to ensure my __all__ list was correct. I left it there because of habits from being primarily a sole developer - I know that it's potentially useful, but potentially not useful, and if in the future it proves to be not useful then I can remove it.

However, in a group projects like OpenFF, that context isn't visible, and the way I did it looks like it's trying to enforce a long-term policy - which wasn't my intent. So I removed it.

 
>  you can add a line like `- [x] Resolves #834`, and github will automatically close the issue when this PR gets merged, leaving a link to the PR itself for future records.

Updated.

> But generally, is there a situation where we'd benefit by considering the longer paths to be outside the public API?

Refactors like this.

The internal layout of a package should be the one that's easiest to maintain. This might not be the same as the logical layout. Unfortunately, Python makes it hard to create a module in place X but make it's repr(), etc. follow the logical layout.

If you want to refactor the internals in the future, you'll break code which reaches into the deeper structure. For example, reaching into toolbox.utils.utils for an exception, when now there's a toolbox.exceptions.

j-wags (approved):
> It's a mistake. It was there to ensure my all list was correct.

Awesome. Yeah, @mattwthompson and @jchodera have caught a lot of my unintentional commits and I'm really grateful for their time - The system works!

>>    But generally, is there a situation where we'd benefit by considering the longer paths to be outside the public API?
> Refactors like this.

Oh, of course! Thanks for clarifying that. A lot of how we've been discussing the API just "clicked" in my head. 

This should be good to merge!

PR #995, opened 2021-06-23
Molecular gastronomy

Yoshanuikabundi:
This is (will be) a cookbook for all the different ways to create a `Molecule`. Submitting draft PR now for discussion at my check-in with @j-wags tomorrow.

See https://openforcefield.atlassian.net/wiki/spaces/IN/pages/1893171258/2021-06-16+Wagner+Mitchell+Check-in

Stuff I've done:
 - Switched from `nbsphinx` to  [`myst-nb`](https://myst-nb.readthedocs.io/en/latest/use/markdown.html) to process notebooks into documentation. This lets us use exactly the same syntax for .md and .ipynb files, saving us from having three different markup languages in the docs
 - Moved the "Using the toolkit" documents to their own subfolder - the `docs` folder was getting a little crowded
 - Set up Intersphinx with OpenMM so we can link to the OpenMM System docs directly
 - Started a notebook that will eventually go through every way to make a `Molecule`

- [x] Document the other 9001 ways to make a molecule
- [x] Make sure CI complains if documentation notebooks break


j-wags (approved):
From our meeting today: I think this is in great shape. The only things that might be worth adding would be:

* Links to the actual data files that are being loaded, so that users can see examples of them.
* Disable the `Molecule.from_iupac` block from being executed (this leaves an error because the docs are built in a OE-less environment)
* (NOT IN THIS PR) When we move to GH Actions and can use OpenEye, it'd be great to also show loading from mol2 and re-enabling the iupac code block
* Remove remaining TODO annotations (but please ping me if there's something that I didn't catch in #1018)

PR #997, opened 2021-06-24
Move XML snippets in test_forcefield.py to module level

mattwthompson:
I am building tests off of the existing test suite and have found that the location of XML snippets is at best not consistent and at worst not import-able. Those defined at the module level are importable, as are those in classes (albeit with an extra step). Those defined inside of functions cannot be imported, as far as I know.

I moved everything to the module level, which decreases readability but adds uniformity and import-ability and the ability to keep things in sync across projects. (On that basis, there might be a better place to put this, like some test-facing data repo. Dumping everything here is fine, I think.). The major benefit is that when I write a test depending on i.e. a TIP5P XML snippet, I don't have to copy code, so there are no duplicates to keep in sync.

Everything passes locally and I also ran slow tests. This isn't really an API break since the test module isn't a part of the public API; I'm probably the only person that would be importing these things now. The diff and git history are/will be ugly, but squash-merging should make the history traversable enough to dig into the past.

Another approach might be to move each snippet into its own file, but that would be something like 20-30 files to track around. The readability is not much different either way, I don't think; it's content being in a different file or in a different location in the file.

-  [ ] Tag issue being addressed

### DIFF
--- a/openff/toolkit/tests/test_forcefield.py
+++ b/openff/toolkit/tests/test_forcefield.py
@@ -75,11 +75,6 @@ from openff.toolkit.utils.toolkits import (
 # GLOBAL CONSTANTS
 # ======================================================================
 
-# File paths.
-TIP3P_SDF_FILE_PATH = get_data_file_path(
-    os.path.join("systems", "monomers", "water.sdf")
-)
-
 XML_FF_GENERICS = """<?xml version='1.0' encoding='ASCII'?>
 <SMIRNOFF version="0.3" aromaticity_model="OEAroModel_MDL">
   <Bonds version="0.3">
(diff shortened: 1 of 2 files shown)


openff/toolkit/tests/test_forcefield.py
```diff
@@ -75,11 +75,6 @@
 # GLOBAL CONSTANTS
 # ======================================================================
 
-# File paths.
-TIP3P_SDF_FILE_PATH = get_data_file_path(
-    os.path.join("systems", "monomers", "water.sdf")
-)
```
j-wags, line 81: (not blocking) Oh, interesting. I guess this was never used?

j-wags (approved):
LGTM! Thanks @mattwthompson. As a courtesy, you might add a note to a "Tests Updated" section in the release history in case anyone _was_ importing these.

openff/toolkit/tests/test_forcefield.py
mattwthompson: It looks like the file is used in a few places, but this shortcut to the file is not used at the moment. Maybe it was in the past - not sure, thought it better to remove it anyway.

PR #998, opened 2021-06-24
Lazy load networkx

mattwthompson:
Whenever `Molecule` is imported , `import networkx as nx` is also called, whether or not it is needed for functionality. It's one of the slower imports, so I lazy-loading it can save ~0.15 seconds of import time in what I presume is a significant number of our users' workflows.

Before:
image

After:
image

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -3058,6 +3058,8 @@ class FrozenMolecule(Serializable):
         # Here we should work out what data type we have, also deal with lists?
         def to_networkx(data):
             """For the given data type, return the networkx graph"""
+            import networkx as nx
+
             from openff.toolkit.topology import TopologyMolecule
 
             if strip_pyrimidal_n_atom_stereo:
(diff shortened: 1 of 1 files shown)


adalke:
I'm all for deferred imports to improve initial import cost and skip importing when not needed.

In this case, I think your code path is deferring a bit too much. Every other code path will end up importing networkx, eg, by calling a `obj.to_networkx()` method.

The only performance advantage you'll get is if the input isn't one of the support types, and doesn't have a `__class__.__name__` matching "Graph". Which isn't going to be often enough to optimize.

My suggestion is to move the:
```
import networkx as nx
```
to the top of 
```
def to_networkx(data):
```

This also means the error message won't be different for the "Graph-object-but-not-a-networkx-Graph" case, but I don't think that's something to have.

mattwthompson:
That does seem quite a bit better; the diff here is simpler than I thought  it was going to be and I don't remember why this change wasn't already in place. I did some import profiling a few months back but maybe I didn't see that idea to completion ...

Lnaden:
Not something I have direct experience with but have been thinking of adding to a few projects is the built in `importlib.util.LazyLoader` which has been in since 3.5: 

https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader

That way you don't have to have `import` statements in a bunch of functions. Its a bit tricky to setup and use though: https://stackoverflow.com/questions/42703908/how-do-i-use-importlib-lazyloader

Not sure how helpful this is, but something to consider.

adalke (approved):
Looks good.

j-wags (approved):
Agreed! This looks good to me too. Thanks, @mattwthompson!

mattwthompson:
Thanks!

Issue #958, opened 2021-05-27
reduce use of test_toolkits.get_mini_drug_bank()


adalke:
I will be working on the toolkit wrapper API. I want to reduce the time needed to run the tests to make it easier to run the test suite after changes. Currently (with RDKit but not OEChem available) it takes about 11 seconds to import openff.toolkit.tests.toolkits. Of that, about 8 seconds is in test_toolkits.py:get_mini_drug_bank(). I think this can be reduced significantly.

That function is used only in a pytest decorator, like:
```
    @pytest.mark.parametrize("molecule", get_mini_drug_bank(OpenEyeToolkitWrapper))
    def test_to_inchi(self, molecule):
```

This parses molecules/MiniDrugBank.sdf at each call. There are three for each toolkit.

This could be reduced by caching the molecule loading, and/or moving the load into the test function rather than use pytest.mark.parameterize. The former would speed up all tests. The latter would help someone who wants to run only some of the tests in the module.

A further analysis suggests that most of those test cases are not needed. The test for toolkit_wrapper's to_inchi() is mostly forwarding the call to the underlying toolkit. There seems to be no benefit for OpenFF to augment the QA process for the toolkit developers.

If there is a benefit, then the test cases should be chosen to be more diverse, for example, auranofin with a gold atom, which gets disconnected by InChI.

Along these lines, InChI is not meant as an input structure format. How much does OpenFF need to support this option? (Perhaps best left for another issue.)

I will be working on this topic and adding notes/comments to this issue.

j-wags:
> A further analysis suggests that most of those test cases are not needed. The test for toolkit_wrapper's to_inchi() is mostly forwarding the call to the underlying toolkit. There seems to be no benefit for OpenFF to augment the QA process for the toolkit developers.

This is largely due to our ignorance -- We want this test to fail if a code change in `openff.toolkit.topology.Molecule.to_rdkit` would cause it to output some RDMols that are unable to make InChIs. But we don't know of a more "first principles" way to guard for this this without just throwing a bunch of molecules through the process. 

Do you know if there's a more general-purpose way to ensure that `Molecule.to_rdkit` is doing the right thing? 

> If there is a benefit, then the test cases should be chosen to be more diverse, for example, auranofin with a gold atom, which gets disconnected by InChI.

[...]

PR #999, opened 2021-06-25
Coverage based fixture reduction

adalke:
Reduce the number of fixtures in the test suite if they don't contribute to test coverage.

- [x] Addresses issue #1004 

- [x] Closes issue #958 

## Goal

Many test methods in `test_molecule.py` and `test_toolkits.py` use the 371 records in the mini-DrugBank data set (in respectively mol2 and sdf formats). These were used to verify a test can handle a variety of molecules.

## Adding test cases isn't always beneficial

While that resulted in a large number of tests, most of these tests did not add additional test assurance, that is, they didn't offer an improvement in the ability to catch possible failures. 

For example, the toolkits' `to_inchi()` method converts the Molecule into the toolkit-specific molecule object, then calls the toolkit-specific function to generate the requested InChI type. Other tests already covered the ability to turn the molecules into toolkit-specific molecule objects, so all that is needed is a check that the handful of InChI-specific conversion code is tested. This only needs one test case, because we can assume (until proven otherwise) that the vendor - OpenEye or RDKit - has tested their InChI generation code.

Similarly, OpenFF molecules can be converted to several serialization formats, including YAML, BSON, and msgpack. These all first use to_dict() to create a Python dictionary, then use a library function to serialize that dictionary. If we have a test that to_dict() is correct, then the test for those serialization functions again need only test one molecule.

## off_coverage

Some test cases require more extensive analysis to reduce the number of needless tests.

For that I used off_coverage's "func-pairs" tracer. This tracks the line number execution order for each function call in the openff package. More precisely, it tracks each pairs of executed line numbers, which captures not just line number coverage but also an (incomplete) sense of branching. In addition, if the program exits with a return or exception, then it also records that line number as part of the pairs.

This was used via a context manager set up like this:
```
coverage = openff.open_call_coverage("test.feats")
   ...
    @requires_openeye
    @pytest.mark.parametrize("mol", get_mini_drug_bank())
    def test_torsions(self, mol):
        def T():
          assert set(molecule.torsions) == set(molecule.propers) | set(molecule.impropers)   
          common_torsions = molecule.propers & molecule.impropers
          if len(common_torsions) > 0:
              for torsion in common_torsions:
                  assert is_three_membered_ring_torsion(torsion)

        with coverage.record(mol.name):
            T()
```

This saves the output to "test.feats" with one line per test case, named by the compound name, and with a feature for each unique code path. Note that I introduced the "T()" function because Python's trace mechanism is only triggered when entering a new scope, typically a function call. This let me record branch coverage even in the test function.

I then used off_coverage's `minimize` command to find a minimal set of compound needed to exercise the observed code paths. For test_torsions() this reported that  "DrugBank_5373", "DrugBank_6739", "DrugBank_2077", and "DrugBank_2642" were sufficient.

## differential off_coverage

In some cases I needed to look at the difference between two coverage pathways, for example, given that `to_dict` is tested, what test cases are needed for the additional code paths needed to test `to_json`.

For this I developed a new `subtract` method to off_coverage. I used the above recording mechanism to record the baseline for `to_dict`, then record the path coverage for `to_json`, then subtracted the features for the first from the second, to get a set of deltas, which I minimized to find the subset of tests needed to for the delta coverage.

## Loading text records instead of parsed molecules at import time

The `test_molecule.py` and `test_toolkits.py` methods loaded MiniDrugBank into openff Molecule objects during import time, to take advantage of `pytest.mark.parametrize()`. Unfortunately, loading 371 molecules takes time, and these were often loaded multiple times, once for each tested function.

This meant that if you use `pytest -k` to focus on some other function, you still had the import overhead of processing several thousand compounds at import time.

I changed the loader so it reads the SDF/mol2 records at import time, which are used to parameterize the test cases, but instead of returning a `Molecule`, it returns a `DrugBankRecord`. For test_toolkits.py, the `get_molecule(toolkit)` method uses the appropriate toolkit to parse the record and return a molecule. For test_molecule.py the `get_molecule()` method always uses OpenEyeToolkitWrapper, and caches the parsed molecule. This preserves and simplifies the previous behavior.

## Load selected text records

I also changed the `get_mini_drug_bank()` function with one that takes a list of selected identifies and returns the corresponding DrugBankRecord instances. Putting everthing together, the `test_torsions` function from earlier now looks like:

```
    @requires_openeye
    @pytest.mark.parametrize(
        "record",
        get_mini_drug_bank(
            ["DrugBank_5373", "DrugBank_6739", "DrugBank_2077", "DrugBank_2642"]
        ),
    )
    def test_torsions(self, record):
        """Test torsions property"""
        molecule = record.get_molecule()
          ...
```

This uses 1) the reduced set of test cases, 2) loaded as records during import time, rather than as molecules, and 3) converted to molecules in the function call (in this case, by returning a copy of the cached molecule).

## Removed xfail_mols and pdb 'wip'

RDKit could not parse some of the MiniDrugBank records. The previous version keep track of these records and marked them `pytest.mark.xfail`. These were for reasons like 'raises warning "Unable to make OFFMol from OEMol: OEMol has unspecified stereochemistry."' which were not part of the test code, but were meant to mark why an input molecule was not being used.

The new test design goal is that molecules aren't used if they don't add test assurance. These xfail non-tests didn't add test assurance, so I removed them. 

I also removed the '"Read from pdb has not been implemented properly yet"' test for PDB output format, marked "wip" to pytest, because that test didn't do anything, many times.


j-wags:
This is ready for review, so I'll take a look at this, and will be responsible for approving+merging.

openff/toolkit/tests/test_molecule.py
```diff
@@ -159,69 +160,80 @@ def is_three_membered_ring_torsion(torsion):
...
+        self._molecule = None  # Cache
+
+    def get_molecule(self):
+        if self._molecule is None:
+            with NamedTemporaryFile(mode="w+b", suffix=".mol2") as tmpfile:
+                tmpfile.write(self.record)
+                tmpfile.flush()
+                molecules = openeye_toolkit.from_file(
+                    tmpfile.name,
+                    "sdf",
```
adalke, line 192: That should be "mol2" not "sdf". It's a copy&paste error from similar code in test_toolkits.py. The error was hidden because prior to PR# 1006 the file_format parameter wasn't converted to the appropriate OEFormat_* and passed to `ifs.SetFormat()`. This told OEChem to use the default file type detection, based on the filename extension. The fix in that PR is at https://github.com/openforcefield/openff-toolkit/pull/1006/files#diff-e2e1869da0d73e84cb7e83cc8c464ef6bfe93f5d1b075de6ed3a21e212853a70R294 .

j-wags:
Closing as I never had time to get to this so it's gone stale and `master` branch has gone out of date.

Issue #898, opened 2021-03-31
Creating a ForceField can fail if a child directory is deleted


SimonBoothroyd:
**Describe the bug**

Creating a force field can fail if a child directory of a child of the current working directory is deleted while a force field is [searching for a file with the specified name](https://github.com/openforcefield/openff-toolkit/blob/e54052afcf0ee401fb65d84e995d53f1f77e3240/openff/toolkit/typing/engines/smirnoff/forcefield.py#L1097-L1103) due to `glob` trying to look for the next child in the directory it's currently looking in.

The current behaviour of that function does actually look slightly suspect but for different reasons than those that cause the issue - if I try to create a `ForceField` from `my-force-field.offxml` I expect the code to try and load the path as I've specified it, not to hunt for that file in all possible sub-directories and possibly finding a version of the file which I didn't want to load.

**To Reproduce**

The issue is tricky to reproduce without monkey patching internal functions, but I ran into this issue when I was trying to create a force field from a string representation, and the script that was doing that was based in a directory which contained thousands of child directories which were slowly being deleted.

**Output**

```
  File "/XXX/site-packages/openff/toolkit/typing/engines/smirnoff/forcefield.py", line 357, in __init__
    self.parse_sources(sources, allow_cosmetic_attributes=allow_cosmetic_attributes)
  File "/XXX/site-packages/openff/toolkit/typing/engines/smirnoff/forcefield.py", line 901, in parse_sources
    smirnoff_data = self.parse_smirnoff_from_source(source)
  File "/XXX/site-packages/openff/toolkit/typing/engines/smirnoff/forcefield.py", line 1099, in parse_smirnoff_from_source
    for file_path in dir_path.glob("**/*.offxml"):
  File "/XXX/pathlib.py", line 1125, in glob
    for p in selector.select_from(self):
  File "/XXX/pathlib.py", line 578, in _select_from
    for p in successor_select(starting_point, is_dir, exists, scandir):
  File "/XXX/pathlib.py", line 525, in _select_from
    with scandir(parent_path) as scandir_it:
FileNotFoundError: [Errno 2] No such file or directory: 'a/child/directory'
```

**Computing environment (please complete the following information):**
 - Operating system: OSX
 - Output of running `conda list`: 

```
openff-toolkit            0.9.1              pyhd8ed1ab_1    conda-forge
openff-toolkit-base       0.9.1              pyhd8ed1ab_1    conda-forge
[...]

PR #1002, opened 2021-06-29
Do not look for XML files in child directories

### LINKED ISSUE #899: Force field loaded from nested child directory
**Describe the bug**

It seems like #816 introduces a bug whereby a `ForceField` can inadvertently be loaded from a nested child of the current working directory. This can presumably lead to subtle issues whereby the user may think they are loading a force field in their current working directory, but if they forgot to create / copy a file there, load a force field from a child directory with the same name but not be the intended force field.

This issue is related to #898 

**To Reproduce**

```
import os

nested_directory = os.path.join("a", "b", "c")
os.makedirs(nested_directory, exist_ok=True)

# Create a FF in a nested directory
ForceField("openff-1.3.0.offxml").to_file(
    os.path.join(nested_directory, "force-field.offxml")
)

# Check that the file does not exist in the current working directory.
assert not os.path.isfile("force-field.offxml")

print(ForceField("force-field.offxml"))
```

**Output**

```
<openff.toolkit.typing.engines.smirnoff.forcefield.ForceField object at 0x7fa44f45acd0>
```

 - Operating system: OSX

```
openff-toolkit            0.9.1              pyhd8ed1ab_1    conda-forge
openff-toolkit-base       0.9.1              pyhd8ed1ab_1    conda-forge
python                    3.9.2           h2502468_0_cpython    conda-forge
```

**Additional context**

mattwthompson:
This PR involves
1. Taking the concise MWE from https://github.com/openforcefield/openff-toolkit/issues/899#issue-848178967 and making it a unit test
2. Reverting the extremely aggressive `.glob("**/*.offxml")`, which looks _recursively into all subdirectories_ for anything with an `.offxml` extenstion. https://github.com/openforcefield/openff-toolkit/issues/898 also explains why the current implementation (mine) was not the right approach.

This should directly fix  #899. #898 should be fixed by extension, although it would be tricky to test; I'd expect a file in a deleted subdirectory to not be looked at if files in subdirectories are not looked at in the first place.

Note that the check mark from ed00e65 is misleading, since tests didn't run on that commit. The point of committing in that order was to show the broken test and then how it's fixed, but that didn't work.


openff/toolkit/tests/test_forcefield.py
```diff
@@ -931,6 +931,25 @@ def test_force_field_case(self):
 
         assert hash(default_case) == hash(lower_case)
 
+    def test_do_not_load_in_child_dir(self):
```
j-wags, line 934: Could we do this test in a temp dir? Right now it leaves the directories around.
```suggestion
    def test_do_not_load_in_child_dir(self, tmp_path):
```

j-wags (approved):
Looks great. Thanks for putting this together, @mattwthompson! Small suggestion, also please update the releasehistory before you merge :-)

openff/toolkit/tests/test_forcefield.py
mattwthompson: Good catch - I wonder how many other times our tests leave temporary directories/files/souvenirs behind?

mattwthompson: Hmm, seems like that didn't quite do it:

```
$ # git clone https://github.com/openforcefield/openff-toolkit
$ # cd openff-toolkit
$ # conda activate openff-dev
$ # python -m pytest -v -r fE --tb=short --ignore=openff/toolkit/tests/test_examples.py  --ignore=openff/toolkit/tests/test_links.py --runslow openff/toolkit/tests
$ git status
git status                                                                                                                                         master  ✱
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	a/
	openff/toolkit/data/molecules/AlkEthOH_tripos/
	openff/toolkit/data/molecules/FreeSolv/
```

j-wags: Hm, I don't remember the exact convention with this, but we may need to have a context manager `temporary_cd` into that directory as well.

lilyminium: Yeah, just creating the tmp_path doesn't go to it. I use this construction a lot:

```python
def testy_test(tmpdir):
	with tmpdir.as_cwd():
		print(os.cwd())
```

Issue #1005, opened 2021-07-01
Improve toolkit file I/O consistency


adalke:
**Describe the bug**

There are a number of inconsistencies in the toolkit wrapper I/O routines. This is a record of the ones I've come across while putting together a PR:

- format_name should always be normalized to upper-case
- RDKit's from_file_obj() should do something with allow_undefined_stereo
- RDKit supports "MOL" as an alias for "SDF"; add an equivalent alias to OEChem
- always raise a ValueError if a format isn't supported (currently this may return [] or may raise an AttributeError)
- implement RDKit.from_file_obj(file_format="SMI") using SmilesMolSupplier() instead of a half-complete SMILES file reader
- all from_file_obj() implementations should support binary file objects
- OpenEye from_file() should implement file_format
- RDKit's from_file() and from_file_obj() should handle molecule processing the same way (in particular, the SDF reader from a file_obj should add hydrogens and implement allow_undefined_stereo!)
- use a consistent exception type and error message when a format isn't supported
- check for from_smiles() parse error
- add a new ParseError exception to throw when SMILES parsing fails
- have a consistent to_file(file_format="SMI") which
-- generates the same SMILES as to_smiles(), and
--  doesn't include a header line
- RDKit's SMILES reader should skip lines that cannot be parsed (matching the SDF reader behavior, and the OEChem reader)
- check OEChem oemolistream(filename) and oemolostream(filename) IsValid() and raise OSError if the file doesn't exist/cannot be opened
- don't use RDKit SupplierFromFilename() because we know (via `file_format`) that the file is in SD format and that function requires a supported filename extension. Instead, use ForwardSDMolSupplier directly

adalke:
Implemented in PR #1006

Issue #1004, opened 2021-07-01
remove test cases that don't contribute to test assurance


adalke:
Many of the unittest in `test_molecule.py` and `test_toolkits.py` use all 371 structures from MiniDrugBank as test parameters.

In many cases, the final 370 structures add no test assurance over the initial structure. For example, if all 371 molecules can be converted `to_dict` then only one test is needed to be assured that `to_json` works correctly.

Some of the test cases do require more than one input to exercise the different code paths.

I used `off_coverage` to help identify a large number of tests that can be reduced from those two test modules with a pretty high confidence that no important test cases were omitted.

The overall test performance is now significantly faster.

This work is available in PR #999 in the branch labeled 'coverage-based-fixture-reduction'.

adalke:
Grr, this is essentially a duplicate of #958 . Which I forgot was there.

PR #1006, opened 2021-07-02
Standardize toolkit I/O

adalke:
- [x] Addresses issue #1005

There are a number of inconsistencies in the RDKit and OpenEye toolkit wrapper interfaces. In particular, the cross-comparison code I developed expects from_file() and from_file_obj() to have the same behavior, which wasn't true for reading SD files with RDKit. See https://github.com/openforcefield/openff-toolkit/issues/1005 for the full list.

This adds a new test module, `test_toolkit_api.py` with top-level base classes to handle the test implementation. Each base class contains a derived class for OpenEyeToolkitWrapper and one for RDKitToolkitWrapper.

This ensures that both wrappers are tested with exactly the same parameters and checked for the same behaviors.

There's also a new ParseError exception, which inherits from MessageException and ValueError. This is thrown when a SMILES string cannot be parsed. (The previous implementation did not check for parse failures.)

I pulled in the (small number of) SMILES test cases in `test_toolkits.py` but did not change them as I didn't want to risk merge errrors with PR #1004 .

I think InChI parsing should also be moved to this new `test_toolkits_io.py`.

Note: I only summarized these as bug fixes. Some of them might be considered as "New features and behaviors changed".

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -46,6 +46,14 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
     set to 0 with appropriate units, rather than `None`
 - [PR #956](https://github.com/openforcefield/openforcefield/pull/956): Added `ForceField.get_partial_charges()` 
   to more easily compute the partial charges assigned by a force field for a molecule.
+- [PR  #1006](https://github.com/openforcefield/openff-toolkit/pull/1006):
+  Changed the SMILES file output in the `to_file()` RDKit and OpenEye
+  wrappers so the outputted SMILES is the one from `to_smiles()`, with
+  explicit hydrogens, rather than the toolkit's default of implicit hydrogens.
+- [PR  #1006](https://github.com/openforcefield/openff-toolkit/pull/1006):
+  Changed the SMILES output format for the RDKit wrapper's `to_file()`
+  so it does not include a header line. This improves the consistency
+  between the OpenEye and RDKit output formats.
 
 
 ### Bugfixes
@@ -58,17 +66,15 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
   was not set to the value specified by the vdW and Electrostatics handlers.
 - [PR #909](https://github.com/openforcefield/openforcefield/pull/909): It is now possible to create an
   OpenMM system with virtual sites created via the `Molecule` virtual site API
-- [PR  #1006](https://github.com/openforcefield/openff-toolkit/pull/1006)
+- [PR  #1006](https://github.com/openforcefield/openff-toolkit/pull/1006):
   Many small fixes to the toolkit wrapper I/O for better error
   handling, improved consistency between reading from a file vs. file
   object, and improved consistency between the RDKit and OEChem
   toolkit wrappers. For the full list see
   [Issue #1005](https://github.com/openforcefield/openff-toolkit/issues/1005). Some
-  of the more significant changes are:
+  of the more significant fixes are:
   - RDKitToolkitWrapper's `from_file_obj()` now uses the same
     structure normaliation as `from_file()`.
-  - Writing to a SMILES file now uses the same SMILES as
-  `to_smiles()`, and (for RDKit) does not include a header line.
   - `from_smiles()` now raises an `openff.toolkit.utils.ParseError` if
   the SMILES could not be parsed.
   - OEChem input and output files now raise an OSError if the file
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -47,13 +47,12 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 - [PR #956](https://github.com/openforcefield/openforcefield/pull/956): Added `ForceField.get_partial_charges()` 
   to more easily compute the partial charges assigned by a force field for a molecule.
 - [PR  #1006](https://github.com/openforcefield/openff-toolkit/pull/1006):
-  Changed the SMILES file output in the `to_file()` RDKit and OpenEye
-  wrappers so the outputted SMILES is the one from `to_smiles()`, with
-  explicit hydrogens, rather than the toolkit's default of implicit hydrogens.
-- [PR  #1006](https://github.com/openforcefield/openff-toolkit/pull/1006):
-  Changed the SMILES output format for the RDKit wrapper's `to_file()`
-  so it does not include a header line. This improves the consistency
-  between the OpenEye and RDKit output formats.
+  Two behavior changes in the SMILES output for `to_file()` and `to_file_obj()`:
+  - The RDKit and OpenEye wrappers now output the same SMILES as `to_smiles()`.
+   This uses explicit hydrogens rather than the toolkit's default of implicit hydrogens.
+  - The RDKit wrapper no longer includes a header line. This improves
+  the consistency between the OpenEye and RDKit outputs.
+
 
 
 ### Bugfixes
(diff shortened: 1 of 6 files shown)


adalke:
Still contains extra data records which aren't used, doesn't check if RDKit/OEChem aren't installed, and need tests for loading two SDF/SMILES molecules, and for loading a file with an error in it,.

openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -40,6 +42,21 @@
...
+
+
+def _require_text_file_obj(file_obj):
+    try:
+        file_obj.write("")
+    except TypeError:
+        # Switch to a ValueError and use a more informative exception
+        # message to match RDKit's toolkit writer.
+        raise ValueError(
+            "Need a text mode file object like StringIO or a file opened with mode 't'"
```
j-wags, line 56: Should this be mode `'r'`?

openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -339,21 +377,38 @@ def from_file_obj(
...
-                mol = self.from_smiles(line, _cls=_cls)
-                mols.append(mol)
+        elif file_format == "SMI":
+            # There's no good way to create a SmilesMolSuppler from a string
+            # other than to use a temporary file.
+            with tempfile.NamedTemporaryFile(suffix=".smi") as tmpfile:
+                content = file_obj.read()
+                if isinstance(content, str):
+                    # Support the older API, whihc required Unicode strings
+                    tmpfile.write(content.encode("utf8"))
```
j-wags, line 402: Nifty! It's cool that you know this.

openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -1488,7 +1543,8 @@ def from_smiles(
         from openeye import oechem
 
         oemol = oechem.OEGraphMol()
-        oechem.OESmilesToMol(oemol, smiles)
+        if not oechem.OESmilesToMol(oemol, smiles):
+            raise ParseError("Unable to parse the SMILES string")
```
j-wags, line 1547: Nice. I've seen that OE uses this pattern a lot in their example code, so I'm glad that you know where to add it in here.

docs/releasehistory.md
```diff
@@ -58,6 +58,22 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
...
+  Many small fixes to the toolkit wrapper I/O for better error
+  handling, improved consistency between reading from a file vs. file
+  object, and improved consistency between the RDKit and OEChem
+  toolkit wrappers. For the full list see
+  [Issue #1005](https://github.com/openforcefield/openff-toolkit/issues/1005). Some
+  of the more significant changes are:
+  - RDKitToolkitWrapper's `from_file_obj()` now uses the same
+    structure normaliation as `from_file()`.
+  - Writing to a SMILES file now uses the same SMILES as
+  `to_smiles()`, and (for RDKit) does not include a header line.
```
j-wags, line 71: (blocking) This header line thing has bugged me for a long time! Thanks for catching it. I know of a few people who have worked around this, and so I think they could get burned if they have an application that's been discarding the first line by default. So let's list this one as a behavior change.

openff/toolkit/tests/test_toolkit_io.py
```diff
@@ -0,0 +1,1200 @@
...
+
+    @pytest.mark.parametrize(
+        "name,file_format", [("caffeine_2d_sdf", "SDF"), ("caffeine_smi", "SMI")]
+    )
+    def test_from_file_handles_cls(self, name, file_format):
+        filename = getattr(file_manager, name)
+        mol = self.toolkit_wrapper.from_file(
+            filename, file_format, _cls=SingingMolecule
+        )[0]
+        mol.sing()
```
j-wags, line 953: Brilliant! These molecule subclass tests are wonderful.

openff/toolkit/utils/rdkit_wrapper.py
adalke: "r" is for "read". "rt" is to read in text mode. "rb" is to read in binary mode. If unspecified, the default is "t", so you rarely see it written explicitly. https://docs.python.org/3/library/functions.html?highlight=text+mode#open

openff/toolkit/utils/rdkit_wrapper.py
adalke: To clarify, that's OpenFF's "older" API. I know this because when I passed in a ByteIO() it failed. There's also a test for this backwards compatibility support:
```
    def test_from_file_obj_smi_supports_stringio(self):
        # Backwards compability. Older versions of OpenFF supported
        # string-based file objects, but not file-based ones.
        with open(file_manager.caffeine_smi) as file_obj:
            mol = self.toolkit_wrapper.from_file_obj(file_obj, "SMI")[0]
        assert mol.name == "CHEMBL113"
```

Grr. I just realized that "file-based" should be "binary-based".  Based on your comment, I see that code comment for the test should instead be in the `rdkit_wrapper` code you highlighted. Just committed and pushed.

docs/releasehistory.md
adalke: Added two bullets to the "New features and behaviors changed", one for the change from implicit to explicit [H] in both toolkits and the other about the removal of the title line in the RDKit wrapper.

adalke:
I, umm, caused the documentation build to fail with "Command killed due to excessive memory consumption"??!?

It appears to be in the 'conda env create --quiet --name 1006 --file docs/environment.yml' step, so I don't think it's due to any change I made.

lilyminium:
It doesn't look like OpenFF runs its own doc check?

Maybe give it a kick by pushing another commit or closing/reopening this PR. Otherwise, MDAnalysis ran into a similar issue. When the kind people at RTD troubleshooted for us, they found that enabling `CONDA_USES_MAMBA` solved it in the meantime. These flags are only configurable on their end so OpenFF may have to contact them.

adalke:
Seemingly a transient RTD hiccup. Though I did update releasehistory, there's no way that could have affected things. ... right? ;)

j-wags (approved):
Thanks for putting this together, @adalke! I think it's in good shape -- There are just two blocking comments. Please feel free to merge once they're resolved!

openff/toolkit/tests/test_toolkit_io.py
```diff
@@ -0,0 +1,1200 @@
...
+        assert mol.n_bonds == 4
+        assert molecule.partial_charges is None
+
+    def test_parse_methane_with_explicit_Hs(self):
+        mol = self.toolkit_wrapper.from_smiles(
+            "[C]([H])([H])([H])([H])", hydrogens_are_explicit=True
+        )
+        # add hydrogens
+        assert mol.n_atoms == 5
+        assert mol.n_bonds == 4
```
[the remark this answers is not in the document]
adalke: Well spotted. I used code coverage and found two other similarly shadowed methods, now fixed.

openff/toolkit/tests/test_toolkit_io.py
```diff
@@ -0,0 +1,1200 @@
...
+        file_obj = getattr(file_obj_manager, name)
+        mol = self.toolkit_wrapper.from_file_obj(
+            file_obj, file_format, _cls=SingingMolecule
+        )[0]
+        mol.sing()
+
+
+@pytest.fixture(scope="class")
+def init_toolkit(request):
+    request.cls.toolkit_wrapper = request.cls.toolkit_wrapper_class()
```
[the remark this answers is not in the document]
adalke: Comment added. It was to defer instantiating OpenEyeToolkitWrapper/RDKitToolkitWrapper until the @requires decorators did their availability checks.

PR #1007, opened 2021-07-03
port changes from #948 to new branch instead of resolving merge conflict

### LINKED ISSUE #456: Round partial charges to enforce correct integral charge on molecule

j-wags:
- [x] Resolves #456 
- [x] Superseded #948
    - [x] Update existing tests to expect higher precision
    - [x] Add test that DOESN'T use round_partial_charges and ensure the charges are different
    - [x] Test supplying `round_partial_charges` to Molecule API as well as ToolkitWrapper API


openff/toolkit/topology/molecule.py
```diff
@@ -3463,6 +3470,21 @@ def assign_partial_charges(
...
+    def _normalize_partial_charges(self):
+        """
+        Add offsets to each partial charge to ensure that they sum to the formal charge of the molecule,
+        to the limit of a python float's precision. Modifies the partial charges in-place.
+        """
+        expected_charge = self.total_charge
+
+        current_charge = 0.0 * unit.elementary_charge
+        for pc in self.partial_charges:
+            current_charge += pc
```
adalke, line 3482: Suggestion: Could write this using `sum`:
```
current_charge = sum(self.partial_charges, 0.0 * unit.elementary_charge)
```

lilyminium: What's wrong with `np.sum(self.partial_charges)`? It doesn't start with the 0 base case so no unit errors.

openff/toolkit/tests/test_toolkits.py
```diff
@@ -2834,7 +2859,32 @@ def test_assign_partial_charges_am1bcc(self):
...
+        molecule.assign_partial_charges(
+            partial_charge_method="am1bcc",
+            toolkit_registry=toolkit_registry,
+            normalize_partial_charges=False,
+        )
+        charge_sum = 0 * unit.elementary_charge
+        abs_charge_sum = 0 * unit.elementary_charge
+        for pc in molecule._partial_charges:
+            charge_sum += pc
+            abs_charge_sum += abs(pc)
```
adalke, line 2885: Are `charge_sum` and `abs_charge_sum` expected to be the same for both toolkits? That is, could the test here us a known good number rather than computing it? I ask because as testing philosophy, this test seems to verify that two implementations of the same algorithm, when run twice, give the same answer. I usually like to have known "golden" value to compare against, or a different algorithm, with a strong preference for the former.

openff/toolkit/topology/molecule.py
adalke: Indeed! My cheminformatics work so rarely uses numpy that I've got a blind spot.

openff/toolkit/tests/test_toolkits.py
j-wags: For methods that require 3D geometries, like AM1(-BCC) and PM3, we can't expect the same charges from all providers. This is ultimately due to deep implementation details (Both OE and AmberTools AM1 calculations do a geometry optimization of the input conformers, but with different settings, so they can reasonably provide two different-yet-correct sets of charges). 

On the other hand, for graph-based charge methods (like MMFF94 and Gasteiger), we could do a strict test for "gold standard" outputs. I'd consider that outside the scope of this PR, and somewhat low priority, since 99% of our users will be using AM1.

openff/toolkit/topology/molecule.py
j-wags: This is awesome! I've switched the manual partial charge sums throughout test_toolkits to this one-liner in b71702d.

lilyminium: Out of curiosity, what's the problem with numpy?

j-wags: Sorry about that @lilyminium - I saw your message in my email but GitHub didn't show it when I loaded the page, and I figured you had deleted it for some reason. 

I don't have a really strong preference between the two. The additional argument in @adalke's approach is a small overhead, but so is using numpy when a builtin function could work. I suppose the builtin `sum` may be a little more future-proof if we switch to a unit package that doesn't play well with numpy? Though, numpy compatibility seems like the most important thing we'd look for in a units package. Anyway, both are way better than the manual sum that we were using before.

j-wags:
Thanks for the feedback! I've got a matter decompressor to attend to today, but I'm treating this PR as having inherited Matt's approval from #948, so I'll be taking one more proofread of this on Tuesday before merging and cutting the long-overdue 0.9.3 release.

openff/toolkit/topology/molecule.py
lilyminium: I'd probably think of numpy as more transferrable.

Regardless, I've just realised you can do `self.partial_charges.sum()`, which might be the best solution overall. I think that both behaviours are compatible with pint as well, which IIRC was one of the major contenders for unit replacement?

adalke: > Out of curiosity, what's the problem with numpy?

No so much a problem, rather that I rarely need it in my code. Take a look at https://github.com/rdkit/mmpdb/ - not an array of numerical data in sight, nor use of anything else in NumPy. Lack of experience with NumPy means I have to think harder to consider NumPy solutions over built-in Python ones.

Issue #830, opened 2021-02-09
Look for antechamber in $AMBERHOME if not found in $AMBERHOME/bin


mattwthompson:
**Is your feature request related to a problem? Please describe.**
With some Amber installs, `antechamber` isn't properly added to `$PATH`.

**Describe the solution you'd like**
If `antechamber` isn't in the path, check to see if `$AMBERHOME` has been set and search within `${AMBERHOME}/bin`? 

**Describe alternatives you've considered**

**Additional context**
https://github.com/michellab/BioSimSpace/issues/175#issuecomment-775877860
https://github.com/openforcefield/openff-toolkit/blob/de8a4a545351301adfe424dff0d879b2dd13bc0b/openff/toolkit/utils/toolkits.py#L4442
https://github.com/openforcefield/openff-toolkit/blob/de8a4a545351301adfe424dff0d879b2dd13bc0b/openff/toolkit/utils/toolkits.py#L4458-L4461

This will probably be hard to reproduce in CI, since the `AmberTools` package on `conda-forge` puts `antechamber` in `$PATH`. On my machine ` $AMBERHOME` is set to `$CONDA_PREFIX`, which we don't have to remove.

I've never actually built AmberTools myself, so I could be overlooking something more obvious.

lohedges:
In your test script you could monkey-patch `distutils.spawn.find_executable`, i.e. bind it to another function that returns `None`, which is what it returns when it fails to find `antechamber` in the `PATH`.

mattwthompson:
This implementation should also check that the AmberTools version is newer than some minimum version (I think 2020 or newer). I think we have previously enforced this version check only at the packaging level - this is not currently a suitable solution because `openff-toolkit-base` does not enforce this (it could be updated to do so).

Related https://github.com/openforcefield/openff-toolkit/issues/571 and probably other issues

lohedges:
We now enforce our own check within BioSimSpace by running `antechamber -v` and checking that the version in the welcome message is >= 20. (If there is no output, or the output doesn't match the desired format, then we also disable OpenFF support.)

mattwthompson:
This hasn't cropped up as a problem in a couple of years - please re-open if it is.

PR #1017, opened 2021-07-14
Remove ambertools when we remove rdkit in CI

j-wags:
- [x] This PR is to test for a failure that @ocmadin and @j-wags resolved on a remote host, but can not reproduce now. Either way, it makes the testing marginally more robust by ensuring that openeye-only CI builds don't have RDKit and AmberTools. 


mattwthompson:
I checked the logs and this isn't actually removing AmberTools as intended. I took at stab at fixing it, may or may not remember to check back and verify.

mattwthompson:
If AmberTools is truly not intended to be installed in the OpenEye-only builds, this line could also be dropped: https://github.com/openforcefield/openff-toolkit/blob/bf19ccbd470c2bc15af8916a4f15c4678100765c/devtools/conda-envs/openeye.yaml#L19

mattwthompson:
Looks like this is working as expected:

image


No AmberTools.

j-wags:
Hm, so the tests are still passing. I feel like I'm crazy. 

@ocmadin, do you remember how this came up a few weeks ago? As I recall, we had a problem where the Toolkit would raise [an error about not finding antechamber's version](https://github.com/openforcefield/openff-toolkit/blob/master/openff/toolkit/utils/ambertools_wrapper.py#L59-L61), when 

* The root problem was that some path wasn't being set or some conda env wasn't being activated in an autogenerated openff-evaluator queue submission script
* I was confused because, in that case, the Toolkit should have said "ambertools is unavailable" instead of crashing on the version check

We solved the problem by finding the queue submission script template and adding `conda activate owens-openff-env`. 

I'm unable to reproduce this locally (that is, any way I try to mangle an environment, I get "ambertools in unavaialble" instead of a version check crash).

I guess my leading two hypotheses are 
* The root cause is #830, where an env _has_ ambertools, but AMBERPATH isn't in PATH
* `distutils.spawn.find_executable('antechamber')` runs under different conditions than `subprocess.check_output(["antechamber", "-L"])`, where the former does find antechamber, but the latter can't.

j-wags:
Ok, this PR doesn't need to be dragged out. Let's consider the review+merge of this PR separately from jeffs fantastical ghost hunt for the underlying issues. 

@mattwthompson, could you review this in its current state? I think the changes are an improvement to the CI.

I'll keep thinking about what may have been happening when @ocmadin and I worked on this and will open a new issue if I can get it to reproduce.

mattwthompson (approved):
Not sure why docs are failing (didn't click through) but the code changes are consistent with what I understand the motivation to be.

Manual checks of some logs:

image

image


When RDKit is flipped on, 2000 more tests pass, 1000 fewer are skipped (?). I think elsewhere I confused myself in forgetting that many of our skipped tests are slow tests, so 9000-10000 skipped tests is not "bad" in the scope of PR CI. 

It also looks like the RDKit=False cases have both RDKit and AmberTools not installed, which is I think one of the motivations here.

mattwthompson:
> Hm, so the tests are still passing. I feel like I'm crazy.

Sorry, what should not be passing? As I understand it, the intent is that the RDKit=False, OpenEye=True builds should test the behavior with no RDKit or AmberTools installed, which would involve a lot of skips but no failures (if the source code and test code are written as intended).

> ... will open a new issue if I can get it to reproduce.

+1; think changes like https://github.com/openforcefield/openff-toolkit/pull/1017/files#diff-8906835152921ef903f55779586f3a092362f65fca98df94d71801cf974ec95bR78 are inherently useful, even if the environments intentionally missing RDKit are a little obtuse.

I don't totally follow the case that originally motivated this PR, but I think where it's at right now is a clear improvement. If something funky is happening with $AMBERHOME, then it should crop up again; if something with a non-standard install is the issue, then it's not under our purview. And until it's reproducible, there isn't much to act on.

j-wags:
Thanks for the review, @mattwthompson. I've updated the releasehistory and will go ahead and merge!

> what should not be passing? 

Sorry, this PR was super vague and opened in the middle of a debugging session for a problem that it turns out I can't reproduce. Thanks for reviewing this change in a context-less vacuum.

PR #1021, opened 2021-07-20
Exception rework part 1

mattwthompson:
A step toward resolving #771, specifically https://github.com/openforcefield/openff-toolkit/issues/771#issuecomment-882672838

The process was
* search through the big files and find custom exceptions that were defined alongside functional code; move them to `openff/toolkit/utils/exceptions.py`
* `%s/MessageException/OpenFFToolkitException`
* Fix imports in tests
* Fix some definitions in `__all__`s
* Wrestled with multiple `ParseError` definitions

1. API isn't broken (with a small caveat and exception below) since any submodule that uses an exception that was moved from it to `exceptions.py` should be importing it anyway.
2. Actually implementing a deprecation warning is tricky; putting it in the body of the class will cause it to be triggered when anything from the module is imported (unworkable IMO - running the README example will trigger this despite not importing it) but tucking it away in the `__init__` means it will be triggered only if `MessageException` is _raised_, but _not when it's imported_. I went with the second option since I couldn't find a way to throw the warning when it's imported but not when other exceptions are.
2. It turns out there are two `ParseError` in the code! One is for failing to parse SMIRNOFF source, the other is for when a wrapped toolkit fails to parse a SMILES string. I've renamed each to be more descriptive, deprecated the SMIRNOFF one in a way that shouldn't break existing code, and removed the SMILES one as I think it's less likely that somebody is capturing that specific exception.
2. A subtlety with the imports is that, as a side effect of using star imports in `openff/toolkit/typing/engines/smirnoff/__init__.py`, the exceptions imported into `parameters.py` and `forcefield.py` *must be defined in the* `__all__` *of those files* for i.e. `from openff.toolkit.typing.engines.smirnoff import UnassignedValenceParameterException` to work, since that import path goes through the star import (`from .parameters import *`). `from openff.toolkit.typing.engines.smirnoff.parameters import UnassignedValenceParameterException` works no matter what's in the `__all__`, since that was imported into that module. This is a side effect of using star imports.


lilyminium:
> Actually implementing a deprecation warning is tricky; putting it in the body of the class will cause it to be triggered when anything from the module is imported (unworkable IMO - running the README example will trigger this despite not importing it) but tucking it away in the __init__ means it will be triggered only if MessageException is raised, but not when it's imported. I went with the second option since I couldn't find a way to throw the warning when it's imported but not when other exceptions are.

Which Python versions are you aiming for? The return of the module `__getattr__` is super handy for this, and an example is even given in the rationale. It's 3.7+

https://www.python.org/dev/peps/pep-0562/#id8

mattwthompson:
Wow - that's exactly what isneeded here. Thanks @lilyminium!


Now I get pretty much the behavior I want - a relevant warning emitted when a deprecated exception is imported, but nothing when not:
```
$ python -c "from openff.toolkit.utils.exceptions import AntechamberNotFoundError"  # No warnings!
$ python -c "from openff.toolkit.utils.exceptions import MessageException" 
/Users/mwt/software/openforcefield/openff/toolkit/utils/exceptions.py:16: DeprecationWarning: MessageException is DEPRECATED and will be removed in a future release of the OpenFF Toolkit. All custom exceptions now inherit from OpenFFToolkitException, which should be used as a replacement for MessageException. Import it via `from openff.toolkit.utils.exceptions import OpenFFToolkitException`.
  warnings.warn(warning_msg, DeprecationWarning)
```

One small downside of "deprecating" the `ParseError` and not just removing it is that, because I inherit `SMIRNOFFParseError` from it, that gets spat out when `ForceField` is loaded. Not sure if that inheritance shouldn't happen - I did it so that existing code can catch `ParseError` in a deprecation period.
```
$ python -c "from openff.toolkit.utils.exceptions import ParseError" 
/Users/mwt/software/openforcefield/openff/toolkit/utils/exceptions.py:16: DeprecationWarning: ParseError is DEPRECATED and will be removed in a future release of the OpenFF Toolkit.
  warnings.warn(warning_msg, DeprecationWarning)
$ python -c "from openff.toolkit.typing.engines.smirnoff import ForceField"
/Users/mwt/software/openforcefield/openff/toolkit/utils/exceptions.py:16: DeprecationWarning: ParseError is DEPRECATED and will be removed in a future release of the OpenFF Toolkit.
  warnings.warn(warning_msg, DeprecationWarning)
```

mattwthompson:
>  It's 3.7+

Oh, and we're pretty much 3.7+ at this point. We'll likely support 3.7 for a while, so 3.8+ features that aren't backported by a third party are more or less off limits. Our 3.6 status is not officially supported but unofficially still exists (https://github.com/conda-forge/openff-toolkit-feedstock/issues/11).  Nothing's broken that we're aware of, and we're mostly not testing it anymore.

lilyminium:
Yay!

> One small downside of "deprecating" the ParseError and not just removing it is that, because I inherit SMIRNOFFParseError from it, that gets spat out when ForceField is loaded.

Yeah, that's a real downside of star imports -- I'm guessing you're still `import ParseError`ing because users expect it in all sorts of namespaces and import `__all__`? Otherwise you could try propagating the `__getattr__`, although I can't currently think of a way to do it non-repetitively. You can dynamically set `__getattr__` on module imports in `__init__` though.

Defined namespaces would make this much easier: `from openff.toolkit.utils import exceptions; raise exceptions.ParseError()`

> Oh, and we're pretty much 3.7+ at this point.

Thanks, good to know. That's pretty on-pace with numpy.

mattwthompson:
> I'm guessing you're still `import ParseError`ing because users expect it in all sorts of namespaces and import `__all__`?

Pretty much - i.e. [this](https://github.com/openforcefield/openff-toolkit/blob/24953a407c853411bee9584d29fa0fb953e59151/examples/check_dataset_parameter_coverage/check_parameter_coverage.ipynb) notebook expects it to be there

```python3
from openff.toolkit.typing.engines.smirnoff import (ForceField, 
UnassignedValenceParameterException, BondHandler, AngleHandler,
ProperTorsionHandler, ImproperTorsionHandler,
vdWHandler)
```

In a perfect world I wouldn't want code like this to exist after a change, but as things stand right now it's valuable to keep kinda-old code working when there's not a large feature associated with the API break.

adalke:
The ParseError for SMILES was my doing. It was added a few weeks ago in https://github.com/openforcefield/openff-toolkit/pull/1006 . Previously, parse errors were ignored. I didn't realize there was a SMIRNOFF ParseError and didn't think to check because it wasn't in exceptions.py! So, definitely needed. :)

In any case, there shouldn't be anyone using that SMILES exception yet.

openff/toolkit/utils/exceptions.py
```diff
@@ -1,48 +1,45 @@
...
+class OpenFFToolkitException(Exception):
+    """Base exception for custom exceptions raised by the OpenFF Toolkit"""
 
     def __init__(self, msg):
-        super().__init__(self, msg)
+        super().__init__(msg)
         self.msg = msg
 
     def __str__(self):
         return self.msg
```
lilyminium, line 33: I'm curious what additional functionality this adds to `Exception`? Is it having the `.msg` attribute?

```python
>>> exc = ValueError("my message")
>>> str(exc)
'my message'
>>> exc.args
('my message',)
```

openff/toolkit/utils/exceptions.py
```diff
@@ -71,45 +68,218 @@ class LicenseError(ToolkitUnavailableException):
...
+    This error is raised when the conformer added to the molecule
+    has a different connectivity to that already defined.
+    or anyother conformer related issues.
+    """
+
+
+class SmilesParsingError(Exception):
+    """
+    This error is rasied when parsing a smiles string results in an error.
+    """
```
lilyminium, line 126: > It turns out there are two ParseError in the code! One is for failing to parse SMIRNOFF source, the other is for when a wrapped toolkit fails to parse a SMILES string. I've renamed each to be more descriptive, deprecated the SMIRNOFF one in a way that shouldn't break existing code, and removed the SMILES one as I think it's less likely that somebody is capturing that specific exception.

Were they the original `SMILESParseError` and `SmilesParsingError`? I still don't understand the difference even with looking at the docstrings -- I put my understanding of the errors below. The similar names are very confusing.

```suggestion
class SMILESParseError(ValueError):
    """The record could not be parsed into the given format"""


class InvalidConformerError(Exception):
    """
    This error is raised for any conformer related issues.
    """


class SmilesParsingError(Exception):
    """
    This error is raised when a toolkit package cannot parse a SMILES string.
    """
```

openff/toolkit/utils/exceptions.py
```diff
@@ -71,45 +68,218 @@ class LicenseError(ToolkitUnavailableException):
...
+
+    def __init__(self, msg):
+        super().__init__(msg)
+        self.msg = msg
+
+    def __str__(self):
+        return self.msg
+
+
+class SMIRNOFFParseError(OpenFFToolkitException, _DeprecatedParseError):
```
lilyminium, line 205: Could I suggest sticking to only one of XXParseError or XXParsingError so it's easier to remember? So far I've seen:

* SMILESParseError
* SmilesParsingError
* SMIRKSParsingError
* SMIRNOFFParseError

openff/toolkit/utils/exceptions.py
mattwthompson: Pretty much, yeah. I could go either way on this; I recall seeing/using some code that relies on `.msg` existing, but I couldn't find it with a quick. I might prefer the aesthetic of clearly having a message attribute over hoping that calling `str()` on it. I think that should always work, but I vaguely recall being bitten by cases in which that isn't true.

openff/toolkit/utils/exceptions.py
mattwthompson: This is just me being sloppy. They should be the same thing, and also inherit as you describe

openff/toolkit/utils/exceptions.py
mattwthompson: Totally - the SMILES duplicates were my mistake, and I have no memory of interacting with the SMIRKS error so it must be purely copy-pasted. Any suggestions which one?  I suppose `XXParsingError` rolls off the tongue better but I don't really care.

openff/toolkit/utils/exceptions.py
lilyminium: It'll give you functional differences if `msg` ever happens to not be a string, I'm not sure if that'd affect scripts and things. `OFFTKExc.msg` would just naively return 3, I assume.

```python
>>> exc = ValueError(3)
>>> str(exc)
'3'
>>> exc.args
(3,)
```

Normal exceptions also take in multiple arguments.

```python
>>> exc2 = ValueError("This is probably good for complex errors", "but I've never seen anyone use it", 3)
>>> str(exc2)
'(\'This is probably good for complex errors\', "but I\'ve never seen anyone use it", 3)'
>>> exc2.args
('This is probably good for complex errors', "but I've never seen anyone use it", 3)
>>> raise exc2
ValueError: ('This is probably good for complex errors', "but I've never seen anyone use it", 3)
```

openff/toolkit/utils/exceptions.py
lilyminium: Looking at the existing exceptions there's not a lot of Verb/ing ones to follow. Although ParsingError sounds nicer, I might lean the other way though and prefer XXParseError because that's what RDKit writes out when it has a parsing error.

mattwthompson:
> In any case, there shouldn't be anyone using that SMILES exception yet.

I opened #1023 to hopefully make that change before the next release, which may be v0.10.0. (The rest of this PR is aimed at a something later).

I realize this adds an extra moving piece, I think it's worthwhile.

openff/toolkit/utils/exceptions.py
mattwthompson: Hmm, I don't know why I dropped the `self` compared to the same code in `MessageException`; it looks out of place, but I probably did it at some point to get an ephemeral block of code to do what I wanted. The inside of `OpenFFToolkitException` is otherwise copied from `MessageException`, I didn't put care into thinking if that was wise.

What I'm gathering from this code is that ... overriding `__init__` and `__str__` is unnecessary? Whether writing tests or being a user of the toolkit, I'd be ok with patterns like

```python3
try:
    do_toolkit_thing(...)
except SomeError as e:
    if "foo" in str(e):
        ...
    elif "bar" in str(e):
        ...
```

which is functionally the same as `if "foo" in e.msg` but doesn't override anything while subclassing out of the stdlib

mattwthompson:
I've merged  #1023, which should resolve the duplicate `ParseError` issue.

There should be a flurry of upstream changes in the next couple days while 0.10.0 is prepped; I'm going to slow down or hold off on moving this PR forward until everything has settled down; hopefully in just a couple of days.

openff/toolkit/tests/test_utils.py
```diff
@@ -133,3 +134,13 @@ def test_sort_smirnoff_dict():
...
+
+
+def test_import_message_exception_raises_warning(caplog):
+    # TODO: Remove when removing MessageException
+    deprecation_msg = (
+        "DEPRECATED and will be removed in a future release of the OpenFF Toolkit"
+    )
+
+    with pytest.warns(DeprecationWarning, match=deprecation_msg):
+        from openff.toolkit.utils.exceptions import MessageException
```
lilyminium, line 146: You could also add a test that importing something else does *not* raise a warning.

```
with pytest.warns(None) as rec:
	from openff.toolkit.utils.exceptions import SMILESParseError
assert len(rec) == 0  # no warnings
```

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1127,15 +1097,15 @@ def parse_smirnoff_from_source(self, source):
...
                 exception_msg = e.msg
             except (FileNotFoundError, OSError):
                 # If this is not a file path or a file handle, attempt parsing as a string.
                 try:
                     smirnoff_data = parameter_io_handler.parse_string(source)
                     return smirnoff_data
-                except ParseError as e:
-                    exception_msg = e.msg
+                except SMIRNOFFParseError as e:
+                    exception_msg = e.args[0]
```
lilyminium, line 1108: This seems a bit inconsistent. (`e.msg` vs `e.args[0]`)

openff/toolkit/utils/exceptions.py
```diff
@@ -71,45 +68,218 @@ class LicenseError(ToolkitUnavailableException):
...
     """The requested partial_charge_method expects a different number of conformers than was provided"""
 
 
 class IncorrectNumConformersWarning(Warning):
     """The requested partial_charge_method expects a different number of conformers than was provided"""
 
 
-class ChargeCalculationError(MessageException):
+class ChargeCalculationError(OpenFFToolkitException):
     """An unhandled error occured in an external toolkit during charge calculation"""
```
lilyminium, line 100:
```suggestion
    """An unhandled error occurred in an external toolkit during charge calculation"""
```

openff/toolkit/utils/exceptions.py
```diff
@@ -71,45 +68,218 @@ class LicenseError(ToolkitUnavailableException):
...
 
 
-class AntechamberNotFoundError(MessageException):
+class AntechamberNotFoundError(OpenFFToolkitException):
     """The antechamber executable was not found"""
 
 
-class SMILESParseError(MessageException, ValueError):
+class SMILESParseError(ValueError):
     """The record couple not be parsed into the given format"""
```
lilyminium, line 112:
```suggestion
    """The record could not be parsed into the given format"""
```

openff/toolkit/utils/exceptions.py
```diff
@@ -71,45 +68,218 @@ class LicenseError(ToolkitUnavailableException):
...
+
+
+class InvalidConformerError(Exception):
+    """
+    This error is raised when the conformer added to the molecule
+    has a different connectivity to that already defined.
+    or anyother conformer related issues.
```
lilyminium, line 119:
```suggestion
    or any other conformer related issues.
```

openff/toolkit/utils/exceptions.py
```diff
@@ -71,45 +68,218 @@ class LicenseError(ToolkitUnavailableException):
...
+    """
+
+
+class SmilesParsingError(Exception):
+    """
+    This error is rasied when parsing a smiles string results in an error.
```
lilyminium, line 125:
```suggestion
    This error is raised when parsing a SMILES string results in an error.
```

openff/toolkit/utils/exceptions.py
```diff
@@ -71,45 +68,218 @@ class LicenseError(ToolkitUnavailableException):
...
+
+class InvalidPeriodicityError(OpenFFToolkitException):
+    """
+    Exception for setting invalid periodicity
+    """
+
+
+class MissingUniqueMoleculesError(OpenFFToolkitException):
+    """
+    Exception for a when unique_molecules is required but not found
```
lilyminium, line 159:
```suggestion
    Exception for when unique_molecules is required but not found
```

openff/toolkit/utils/exceptions.py
```diff
@@ -71,45 +68,218 @@ class LicenseError(ToolkitUnavailableException):
...
+
+class ParameterHandlerRegistrationError(OpenFFToolkitException):
+    """
+    Exception for errors in ParameterHandler registration
+    """
+
+
+class SMIRNOFFVersionError(OpenFFToolkitException):
+    """
+    Exception thrown when an incompatible SMIRNOFF version data structure in attempted to be read.
```
lilyminium, line 184:
```suggestion
    Exception thrown when an incompatible SMIRNOFF version data structure is attempted to be read.
```

openff/toolkit/utils/exceptions.py
```diff
@@ -1,48 +1,45 @@
...
+                "which should be used as a replacement for MessageException. Import it via "
+                "`from openff.toolkit.utils.exceptions import OpenFFToolkitException`."
+            )
+        warnings.warn(warning_msg, DeprecationWarning)
+        return globals()[f"_Deprecated{name}"]
+    raise AttributeError(f"module {__name__} has no attribute {name}")
+
+
+class _DeprecatedMessageException(Exception):
+    """DEPRECATED: A base class for exceptions that print out a string given in their constructor"""
```
lilyminium, line 22: IMO this should have the same functionality it used to, given that it's still supported albeit in deprecated status.

```suggestion
    """DEPRECATED: A base class for exceptions that print out a string given in their constructor"""
    
   def __init__(self, msg):
        super().__init__(self, msg)
        self.msg = msg

    def __str__(self):
        return self.msg
```

Then `_DeprecatedParseError` can inherit from `_DeprecatedMessageException`, with the added bonus that I *think* that'll get rid of the LGTM warning about `__init__`.

lilyminium (changes requested):
It feels so much neater having the exceptions in one file :-) Most of my comments are about typos. The important one is that _DeprecatedMessageException should still function the same as old MessageException, for those who were relying on `MessageException.msg`.

docs/releasehistory.md
```diff
@@ -8,6 +8,19 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ## Current Development
 
+### Behaviors changed
+
+- [PR #1021](https://github.com/openforcefield/openforcefield/pull/1021): Renames
+  [`openff.toolkit.utils.exceptions.ParseError`](openff.toolkit.utils.exceptions.ParseError) to
```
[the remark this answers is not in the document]
mattwthompson: In general I'm pretty sloppy with release notes between releases, since Jeff and/or Josh will clean them up before making an actual release.

But in this case it doesn't really make sense to try to link to an exception that no longer exists, does it?

openff/toolkit/tests/test_utils.py
```diff
@@ -9,16 +9,17 @@
...
 import os
 
 import pytest
 from simtk import unit
 
+# =============================================================================================
+# GLOBAL IMPORTS
+# =============================================================================================
+
+
```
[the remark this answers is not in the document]
mattwthompson: Probably automation - I'd  love to remove all of these blocks myself, I don't think they serve any purpose in legible code and can lead to confusing cases like this.

openff/toolkit/tests/test_utils.py
```diff
@@ -133,3 +134,13 @@ def test_sort_smirnoff_dict():
     assert smirnoff_dict == OrderedDict(
         sort_smirnoff_dict(forcefield._to_smirnoff_data())
     )
+
+
+def test_import_message_exception_raises_warning(caplog):
+    # TODO: Remove when removing MessageException
+    deprecation_msg = (
+        "DEPRECATED and will be removed in a future release of the OpenFF Toolkit"
+    )
```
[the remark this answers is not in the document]
mattwthompson: My guess is that this was longer at some point and I shortened it to where it's at now. I agree that looks odd

openff/toolkit/tests/test_toolkit_io.py
```diff
@@ -26,7 +26,11 @@
 from openff.toolkit.tests import create_molecules
 from openff.toolkit.tests.utils import requires_openeye, requires_rdkit
 from openff.toolkit.topology.molecule import Molecule
-from openff.toolkit.utils import OpenEyeToolkitWrapper, RDKitToolkitWrapper, exceptions
+from openff.toolkit.utils import OpenEyeToolkitWrapper, RDKitToolkitWrapper
+from openff.toolkit.utils.exceptions import (
```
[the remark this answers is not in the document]
mattwthompson: Preference, not performance, I think; I tend to prefer importing only things that are needed in a particular scope. This module should be pretty lightweight at import time so it should be a wash in terms of performance. Also, even though this can go either way, I have a marginal preference for keeping things consistent without a project and for whatever reason everywhere else uses this pattern. Then again, importing each exception one-by-one was the only way to import them before everything was collected into one namespace ...

mattwthompson:
Thanks for the review @lilyminium! Sorry about all the typos

openff/toolkit/utils/exceptions.py
```diff
@@ -71,45 +75,218 @@ class LicenseError(ToolkitUnavailableException):
...
+
+class _DeprecatedParseError(Exception):
+    """DEPRECATED: Error for when a SMIRNOFF data structure is not parseable by a ForceField"""
+
+    def __init__(self, msg):
+        super().__init__(msg)
+        self.msg = msg
+
+    def __str__(self):
+        return self.msg
```
lilyminium, line 209: I'd just inherit from _DeprecatedMessageException. As noted in https://github.com/openforcefield/openff-toolkit/pull/1021/files#r687269171, this will hopefully get rid of that LGTM alert too.

```suggestion
class _DeprecatedParseError(_DeprecatedMessageException):
    """DEPRECATED: Error for when a SMIRNOFF data structure is not parseable by a ForceField"""
```

lilyminium (changes requested):

openff/toolkit/tests/test_toolkit_io.py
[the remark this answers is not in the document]
lilyminium: Fair enough, consistency is a good design pattern!

mattwthompson:
Left to do:

 * Merge `master` in
 * `SMIRNOFFParsingError` -> `SMIRNOFFParseError` / Ensure consistency of above (particularly in docstrings)
 * Make all exceptions import from `OpenFFToolkitException`, or have a really good reason for not doing so
 * Make `OpenFFToolkitException` import from `_DeprecatedMessageException`, which allows for current code to emit a warning instead of breaking
 * Make notes about deprecation plan (done)
 * Do a once-over on API *breaks* before merging

j-wags (approved):
LGTM pending the to-dos in @mattwthompson's message above.

mattwthompson:
I noticed that I broke the API in one place that could cause issues for users who bend the rules. Around cb5dfc027fd1911b7486bd9daa80655b678a7cc2 the behavior is


```python3
>>> from openff.toolkit.utils.toolkits import OpenEyeToolkitWrapper
>>> from openff.toolkit.utils.toolkits import MessageException
ImportError: cannot import name 'MessageException' from 'openff.toolkit.utils.toolkits' (/Users/mwt/software/openforcefield/openff/toolkit/utils/toolkits.py)
```

which is not ideal.

I think 85a00bac53e49fa4986d60319bcf1c31aa83b362 fixes this, without emitting the warning for safer imports:

```python3
>>> from openff.toolkit.utils.toolkits import OpenEyeToolkitWrapper
>>> from openff.toolkit.utils.toolkits import MessageException
/Users/mwt/software/openforcefield/openff/toolkit/utils/exceptions.py:18: DeprecationWarning: MessageException is DEPRECATED and will be removed in a future release of the OpenFF Toolkit. All custom exceptions now inherit from OpenFFToolkitException, which should be used as a replacement for MessageException. Import it via `from openff.toolkit.utils.exceptions import OpenFFToolkitException`.
  warnings.warn(warning_msg, DeprecationWarning)

j-wags:
Good catch, and nice fix @mattwthompson. Thanks for the attention to detail!

PR #1022, opened 2021-07-20
Lint notebooks

mattwthompson:
- [ ] Lint notebooks
- [ ] Set up [automated `pre-commit` checking](https://pre-commit.ci/) on notebooks only

The codebase has autoformatters that partially apply a consistent style, but the notebooks do not. This PR uses [nbQA with `pre-commit`](https://github.com/nbQA-dev/nbQA#pre-commit) to apply auto-formatters. I have seen this used in [other projects](https://github.com/glotzerlab/hoomd-examples) to keep consistent style between the code and examples.

Here is the config I used (`pre-commit-config.yaml`):

```yaml
repos:
- repo: https://github.com/nbQA-dev/nbQA
  rev: 0.13.1
  hooks:
    - id: nbqa-pyupgrade
      args:
        - --nbqa-mutate
        - --py37-plus
    - id: nbqa-isort
      args:
        - --nbqa-mutate
    - id: nbqa-black
      args:
        - --nbqa-mutate
```

If we were to adopt pre-commit, this configuration (and future modifications of it) would be the de facto code style standard by committing. If not, this would at least provide a one-off fix of the style in the notebooks.

If desired I can re-do this using a bot, giving the bot "credit" for all the changes.


j-wags (approved):
Thanks @mattwthompson. These changes look good. TL;DR I like this and think we should add the pre-commit yaml and merge. I don't think we should add a bot.

More detailed version:

> If we were to adopt pre-commit, this configuration (and future modifications of it) would be the de facto code style standard by committing.

I think this is a good idea - The big thing that removed a lot of my doubts was reading up on pre-commits and seeing that _not_ having pre-commit installed won't have any negative effect on unaware developers (as long as we don't check for notebook linting in CI). So we can put "install pre-commit" in the developers guide (maybe in this PR, actually) but in the future we won't sweat it if an external contributor puts a notebook in a PR isn't linted.

> If desired I can re-do this using a bot, giving the bot "credit" for all the changes.

I'm becoming more open-minded about bot commits, largely because when I _do_ rely on the git history, I'm doing a `git blame`. In that situation my goal is usually to find the origin PR (which usually has lots of context, including the commit) instead of the commit itself (which frequently doesn't help me understand why we did something). But a bot commit would lead to the same PR as a human commit, so I'm thinking maybe bot commits aren't so bad. 

The downside is that my whole "git blame" workflow is fragile and dependent not on the _actual git tree_ but instead it relies on _github metadata_. So if someday we move to gitlab or another platform, losing this metadata in the process, then the transition costs will be lower with a clean git history. Also, there may be other git forensics (other than "blame") that require a clean git history that we end up using in the future.  

Also, notebooks are especially hairy, since it's possible to commit them in a "dirty" state, where cells are executed out of order or only partially (I see the [hoomd-example pre commits](https://github.com/glotzerlab/hoomd-examples/blob/155a864da12f60f1f096ad67f9c2af1942280fce/.pre-commit-config.yaml#L48-L55) have an extra formatting step to sanitize an already-run notebook). Since uploading a notebook in a "dirty" state is so easy, I'm starting to think that any bot-reliant solution for notebooks should _both_ lint _and_ re-execute (@Yoshanuikabundi set up notebook execution in docs rendering in #995, maybe there's something analogous we could do mainline examples?). I guess this isn't a blocker to having a bot linter in the meantime, but ultimately it'll be easier to justify bot commits in the git tree with something like a bot-notebook-executor.

So that's probably more than is useful to put here, but the short version is I'd like to hold off on bot commits for the time being.

mattwthompson:
Soft blocked by consensus re #1039

examples/toolkit_showcase/toolkit_showcase.ipynb
```diff
@@ -252,11 +253,11 @@
...
     "# Load protein and water force field parameters\n",
-    "omm_forcefield = openmm.app.ForceField(\n",
-    "    \"amber99sb.xml\", \"tip3p.xml\"\n",
-    ")\n",
+    "omm_forcefield = openmm.app.ForceField(\"amber99sb.xml\", \"tip3p.xml\")\n",
     "# Teach OpenMM about the ligand molecule and the Parsley force field\n",
-    "smirnoff = SMIRNOFFTemplateGenerator(forcefield=\"openff-1.3.0.offxml\", molecules=[ligand])\n",
+    "smirnoff = SMIRNOFFTemplateGenerator(\n",
+    "    forcefield=\"openff-1.3.0.offxml\", molecules=[ligand]\n",
+    ")\n",
```
Yoshanuikabundi, line 260: I think calls to functions should either be a single line:

```python
smirnoff = SMIRNOFFTemplateGenerator(forcefield="openff-1.3.0.offxml", molecules=[ligand])
```

or have each argument on its own line:

```python
smirnoff = SMIRNOFFTemplateGenerator(
    forcefield="openff-1.3.0.offxml", 
    molecules=[ligand],
)
```

But not have multiple arguments per line: 

```python
smirnoff = SMIRNOFFTemplateGenerator(
    forcefield="openff-1.3.0.offxml", molecules=[ligand]
)
```

I think the latter makes it easy to miss arguments other than the first - especially important in examples where users are likely to skim read code. Is it possible to configure black like this? And if that would mean changing the configuration used for the whole project, would that be acceptable?

Yoshanuikabundi:
Sorry for taking so long to look at this! I love lints and I think this is fantastic. I just have one comment about a formatting pattern I don't love, above.

Yoshanuikabundi:
Oh, and I absolutely agree with @j-wags about re-executing notebooks after linting them to sanitize. The toolkit showcase, for example, currently has an absolute file path to the gromacs installation on Jeff's machine in the output. There are probably potential privacy concerns there, and its definitely not as profesh as it could be.

examples/toolkit_showcase/toolkit_showcase.ipynb
mattwthompson: My preference is generally towards opinionated formatters (of which Black is one - "Black is opinionated so you don't have to be") even though I don't personally like all of its defaults (again, including black as a case). That being said, I generally agree with your suggestion here, if only because it makes diffs while adding/removing arguments more legible.

That's all too high-minded, though, since in this case `black`, I think, can be tricked into doing this:

```
$ cat t.py
smirnoff = SMIRNOFFTemplateGenerator(
    forcefield="openff-1.3.0.offxml", molecules=[ligand]
)

smirnoff = SMIRNOFFTemplateGenerator(
    forcefield="openff-1.3.0.offxml", molecules=[ligand],
)

smirnoff = SMIRNOFFTemplateGenerator(
    forcefield="openff-1.3.0.offxml",
    molecules=[ligand],
)
$ black t.py
$ cat t.py
smirnoff = SMIRNOFFTemplateGenerator(
    forcefield="openff-1.3.0.offxml", molecules=[ligand]
)

smirnoff = SMIRNOFFTemplateGenerator(
    forcefield="openff-1.3.0.offxml",
    molecules=[ligand],
)

smirnoff = SMIRNOFFTemplateGenerator(
    forcefield="openff-1.3.0.offxml",
    molecules=[ligand],
)
```

mattwthompson: Okay, I've gone through the notebooks and tried to find every case of this. I might have missed some, since there is so much to go through, but I think it's all cleaned up now.

mattwthompson:
>  re-executing notebooks after linting them to sanitize.

I'm kinda confused here - how can the notebooks be run without knowing the paths to the necessary executables? There are also i.e. warnings that show where on my computer I have things installed, like

```
examples/swap_amber_parameters/.ipynb_checkpoints/swap_existing_ligand_parameters_with_openmmforcefields-checkpoint.ipynb:      "/Users/mwt/miniconda3/envs/openff-dev/lib/python3.9/site-packages/parmed/amber/netcdffiles.py:409: UserWarning: Could not find netCDF4 module. Falling back on scipy implementation, which can significantly slow down simulations if used as a reporter\n",
```

I don't care much about people knowing that I installed miniconda in my home directory but I would like a cleaner solution that doesn't show that path at all. Is there something obvious I'm missing here?

mattwthompson:
Okay, to get the status checks above working, I authorized the [app](https://github.com/marketplace/pre-commit-ci) on this repo, which should be evident from the status checks of the last commit. However, that will screw up other PRs/commits/etc. until there is a `.pre-commit-config.yaml` file committed to the `master` branch, so I've uninstalled it. If/when this (or https://github.com/openforcefield/openff-toolkit/pull/1040) is merged, it should be re-authorized.

examples/toolkit_showcase/toolkit_showcase.ipynb
Yoshanuikabundi: Took me a second but trailing commas! Nice! Thanks for the tip.

j-wags:
> I'm kinda confused here - how can the notebooks be run without knowing the paths to the necessary executables?

I think @Yoshanuikabundi was basically saying what you said later -- The current arrangement shows the world that you have miniconda in your home directory. Not a big deal imho.

So the local file paths are weird but I think they're ok (the alternative is either bot commits or some opinionated sanitization automation that could have bugs of its own). 

I'm still thinking about bot commits, but this and #1040 can go ahead in the meantime. Thanks for pushing on this @mattwthompson!

mattwthompson:
Thanks!

PR #1023, opened 2021-07-21
Rename duplciate ParseError to SMILESParseError

mattwthompson:
See https://github.com/openforcefield/openff-toolkit/pull/1021#issuecomment-884258180 for context; in #1006 there was a `ParseError` added to `utils/exceptions.py` to handle failures in parsing SMILES strings in the toolkit wrappers. However, there's already a dedicated `ParseError` in the SMIRNOFF XML parsing code for handling failures there.

To fix this, in #1021 I moved them into their own exceptions, choosing to deprecating the SMIRNOFF `ParseError` and just remove the SMILES one. However, it would be much simpler to just prevent the SMILES one  from ever getting into a release, so I'm splitting this out from the other PR. **It would be great to sneak this into 0.10.0**, since it will prevent a soft API break from needing to take place in the following release. Again, since #1006 was merged after the previous release, this won't break any existing API.

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -78,7 +78,7 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
   of the more significant fixes are:
   - RDKitToolkitWrapper's `from_file_obj()` now uses the same
     structure normaliation as `from_file()`.
-  - `from_smiles()` now raises an `openff.toolkit.utils.ParseError` if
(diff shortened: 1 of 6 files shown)


j-wags (approved):
This looks great. Thanks for the sharp eye, @mattwthompson!

I'll also try to review #1021 but I agree that this will be good to put in sooner so that we can get it into the release.

mattwthompson:
I think reviewing #1021 should be lower in priority than anything that would go into the next release. Although it's in a good place, I'm pausing work on it while things move around upstream (including this one).

PR #1024, opened 2021-07-22
Fixes for OpenMM 7.6 beta

mattwthompson:


mattwthompson:
@j-wags looks like one important import will be broken by OpenMM 7.6: https://github.com/openforcefield/openff-toolkit/runs/3135095516 which I fixed in dbc1fcd. To keep this changeset small, I didn't do the try/except blocks that would suppress the deprecation warning.

I don't know the timeline for 7.6 to be released, but getting this into a released version of the toolkit should pre-empt that headache.

openff/toolkit/topology/topology.py
```diff
@@ -2327,7 +2327,11 @@ def to_openmm(self, ensure_unique_atom_names=True):
         from simtk.openmm.app import Aromatic, Double, Single
         from simtk.openmm.app import Topology as OMMTopology
         from simtk.openmm.app import Triple
-        from simtk.openmm.app.element import Element as OMMElement
+
+        try:
+            from simtk.openmm.app.element import Element as OMMElement
+        except ImportError:
+            from openmm.app.element import Element as OMMElement
```
mattwthompson, line 2334: Using the beta:

```python3
>>> from simtk.openmm.app.element import Element as OMMElement
Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.
ModuleNotFoundError: No module named 'simtk.openmm.app.element'
>>> from openmm.app.element import Element as OMMElement  # works fine
```

mattwthompson:
Also looks like anything using `pmd.openmm.load_topology` is failing, with the cause being a version check, not behavior (or at least not obviously): https://github.com/ParmEd/ParmEd/blob/1be8ca0f2f5b232db79384cb02454a9c53f37749/parmed/utils/decorators.py

I'll see if I can get that patched there, but I don't know the lead time for a ParmEd release to be bundled with AmberTools ...

mattwthompson:
https://github.com/ParmEd/ParmEd/issues/1185

j-wags (approved):
Thanks for digging into this., @mattwthompson - It looks perfect. I'll switch CI back and we can merge if it passes!

mattwthompson:
Looks like everything passed. As a minimal safeguard against the warning leaking out, I grabbed this patch and checked the README snippet on both versions of OpenMM. On 7.5.1 it runs quietly without surprise, but back on the beta is doesn't squelch the warning. Mimially reproducing (on 7.6beta):

```shell
$ python -c "from openff.toolkit.topology import Molecule"
Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.
```

which got me thinking I did the try/except in the wrong order. This snippet

```python3
>>> try:
...     from openmm.app.element import Element as OMMElement
... except ImportError:
...     from simtk.openmm.app.element import Element as OMMElement
```
runs on both versions and does not produce warning. This seems like a better fix, since it will only use the deprecated import path if it must.

j-wags:
Ah, got it. Let's go ahead and switch the the new way, then we should be good to merge!

Issue #971, opened 2021-06-04
LibraryCharge handler slow SMARTS matching


SimonBoothroyd:
**Describe the bug**

I've come across a couple of cases where the library change handler SMARTS matching code hangs when attempting to match a full molecule SMILES.

This is likely because there are tens (to hundered) of thousands of ways the SMILES pattern can be matched, all of which are enumerated and considered even though only one of the matches will eventually be used.

**To Reproduce**
Steps to reproduce the behavior. A minimal reproducing set of python commands is ideal.

```
import time

import numpy
from simtk import unit

from openff.toolkit.topology import Molecule
from openff.toolkit.typing.engines.smirnoff import LibraryChargeHandler
from openff.toolkit.utils import (
    GLOBAL_TOOLKIT_REGISTRY,
    OpenEyeToolkitWrapper,
    RDKitToolkitWrapper,
)

molecule: Molecule = Molecule.from_mapped_smiles(
    "[C:1]([N:2]1[C:3]([H:23])=[C:4]([H:24])[C:5]([H:25])=[C:6]1[C:7](=[S:8])"
    "[N:9]=[P:10]([N:11]([C:12]([H:26])([H:27])[H:28])[C:13]([H:29])([H:30])[H:31])"
    "([N:14]([C:15]([H:32])([H:33])[H:34])[C:16]([H:35])([H:36])[H:37])[N:17]"
    "([C:18]([H:38])([H:39])[H:40])[C:19]([H:41])([H:42])[H:43])([H:20])([H:21])"
    "[H:22]"
)

library_charge_handler = LibraryChargeHandler(version=0.3)
library_charge_handler.add_parameter(
    {
        "smirks": molecule.to_smiles(mapped=True),
        "charge": numpy.ones(molecule.n_atoms) * unit.elementary_charge
    }
)

for toolkit_wrapper in [OpenEyeToolkitWrapper, RDKitToolkitWrapper]:

    GLOBAL_TOOLKIT_REGISTRY._toolkits = [toolkit_wrapper()]

    start_time = time.perf_counter()

    matches = library_charge_handler.find_matches(molecule.to_topology())

    end_time = time.perf_counter()

    print(toolkit_wrapper.__name__, end_time - start_time, len(matches))
```

**Output**
The full error message (may be large, that's fine. Better to paste too much than too little.)

**Computing environment (please complete the following information):**
 - Operating system OSX
 - Output of running `conda list` 

**Additional context**

[...]

Issue #967, opened 2021-06-01
[Request] chemical_environment_matches keyword(unique=False, max_matches=None)


lilyminium:
**Is your feature request related to a problem? Please describe.**

Sometimes I want unique matches without having to sort through the non-unique matches myself.

**Describe the solution you'd like**

It would be easy and convenient to give `chemical_environment_matches` `unique` and `max_matches` keywords to pass to OpenEye or RDKit respectively. `unique=False` and `max_matches=None` would replicate current behaviour, where `max_matches is None` sets it to 0 for OpenEye and the largest possible int in RDKit.

**Describe alternatives you've considered**

**Additional context**

lilyminium:
Generally closed by #1036; max_matches isn't really important for sorting through non-unique matches.

j-wags:
Thanks for running the bookkeeping on this, @lilyminium. I'm super behind on everything as you can tell so I really do appreciate this.

PR #1036, opened 2021-07-30
SMARTS matching performance improvements

mattwthompson:
@lilyminium did the work (https://github.com/openforcefield/openff-toolkit/issues/971#issuecomment-859057461) - I'm opening it as a PR since I need this behavior added in order to trust my energy comparison tests.

Related #967 

- [ ] Toggle on slow tests with `True` arguments
- [ ] Ensure protein tests pass
- [ ] Add arguments to `.create_openmm_system` an `.create_force`
- [ ] Docstrings


jeff231li:
Is there a way to control the new arguments in `find_matches` (`unique` and `match_heavy_first`) when we call the method `ForceField.create_openmm_system`?

j-wags:
Hm, I think that both of these changes (`unique` and `match_heavy_first`), if they provide the same behavior, should default to `True` throughout the codebase. 

The two things I'd like to see to convince myself that they're behaving well are:
- [ ] The protein parameterization tests (both [di- and tri-peptides](https://github.com/openforcefield/amber-ff-porting) and the [T4 lysozyme example](https://github.com/openforcefield/openff-toolkit/pull/881)) running successfully with them defaulting to `True`
- [ ] The `--runslow` test suite passing with them defaulting to `True`

lilyminium:
> Hm, I think that both of these changes (unique and match_heavy_first), if they provide the same behavior, should default to True throughout the codebase.

`unique=False, match_heavy_first=False` is status quo. Shifting `unique=True` would result in different answers for `find_matches` and the like -- e.g. in the example molecule, 1296 vs 1 match. https://github.com/openforcefield/openff-toolkit/issues/971#issuecomment-859057461

`unique=True` toggles enumeration of every combination of the same atoms. So a SMARTS pattern matching methane CH4 would only give (C, H1, H2, H3, H4) for `unique=True`. However, `unique=False` would return `[(C, H1, H2, H3, H4), (C, H1, H2, H4, H3), (C, H1, H3, H2, H4), (C, H1, H3, H4, H2), (C, H1, H4, H2, H3), (C, H1, H4, H3, H2), (C, H2, H1, H3, H4), (C, H2, H1, H4, H3), (C, H2, H3, H1, H4), (C, H2, H3, H4, H1), (C, H2, H4, H1, H3), (C, H2, H4, H3, H1), (C, H3, H1, H2, H4), (C, H3, H1, H4, H2), (C, H3, H2, H1, H4), (C, H3, H2, H4, H1), (C, H3, H4, H1, H2), (C, H3, H4, H2, H1), (C, H4, H1, H2, H3), (C, H4, H1, H3, H2), (C, H4, H2, H1, H3), (C, H4, H2, H3, H1), (C, H4, H3, H1, H2), (C, H4, H3, H2, H1)]`. 

I *think* `match_heavy_first=True|False` should give identical behaviour (currently RDKit only, I couldn't work out OpenEye). For each heavy match, it pins heavy atoms and enumerates though the remaining Hs.

mattwthompson:
> Is there a way to control the new arguments in `find_matches` (`unique` and `match_heavy_first`) when we call the method `ForceField.create_openmm_system`?

@jeff231li No, but thanks for asking, since this question makes me realize how there would be some cost to wiring through another argument or two from `ForceField.create_openmm_system` to every one of the dozen or so parameter handlers.

mattwthompson:
We met on this and decided:
* `unique=True` should be implemented and made the default for `LibraryChargeHandler`
* `unique=False` should remain for all other handlers
* `unique` is not to be exposed as an argument in `create_openmm_system` but will be added as an argument to `find_matches` and `chemical_environment_matches`
* `match_heavy_first=True` is probably fine, and has been stress tested, but _might_ cause issues and therefore we should split it out

j-wags:
My profiling is showing about a 90% runtime reduction for whole-molecule librarycharge assignment using `uniquify=True`. Reproducing example here: https://github.com/openforcefield/openff-toolkit/issues/971#issuecomment-895602640

mattwthompson:
Looks like `max_matches` is exposed inconsistently - where do we want this to exist as an argument, and where don't we?

j-wags:
Limiting `max_matches` in this context is fairly dangerous, so I'd in favor of never exposing that as a parameter.

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -2222,13 +2222,15 @@ def __init__(self, parameter_type, environment_match):
...
-    def find_matches(self, entity):
+    def find_matches(self, entity, unique=False):
         """Find the elements of the topology/molecule matched by a parameter type.
 
         Parameters
         ----------
         entity : openff.toolkit.topology.Topology
             Topology to search.
+        unique : bool, default=False
+            If True, SMARTS matching will enumerate every possible combination of atoms.
```
j-wags, line 2233:
```suggestion
            If False, SMARTS matching will enumerate every valid permutation of matching atoms. 
            If True, only one order of each unique match will be returned.
```

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -2252,6 +2259,9 @@ def _find_matches(self, entity, transformed_dict_cls=ValenceDict):
             will determine how groups of atom indices are stored
             and accessed (e.g for angles indices should be 0-1-2
             and not 2-1-0).
+        unique : bool, default=False
+            If True, SMARTS matching will enumerate every possible combination of atoms.
```
j-wags, line 2263:
```suggestion
            If False, SMARTS matching will enumerate every valid permutation of matching atoms. 
            If True, only one order of each unique match will be returned.
```

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -2241,7 +2243,12 @@ def find_matches(self, entity):
         #  Should we reduce its scope and have a check here to make sure entity is a Topology?
         return self._find_matches(entity)
```
j-wags, line 2244: (blocking) Do we need to thread the `unique` kwarg through here? Ditto for all other instances of this pattern in the other ParameterHandlers.

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -3343,7 +3354,7 @@ def check_handler_compatibility(self, other_handler):
             tolerance_attrs=float_attrs_to_compare,
         )
 
-    def find_matches(self, entity):
+    def find_matches(self, entity, unique=False):
```
j-wags, line 3357: (blocking) (1) The `unique` kwags isn't used in this method. I think it needs to be added.

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -4298,7 +4315,7 @@ def check_handler_compatibility(
             identical_attrs=string_attrs_to_compare + int_attrs_to_compare,
         )
 
-    def find_matches(self, entity):
+    def find_matches(self, entity, unique=False):
```
j-wags, line 4318: ditto (1)

openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -1960,6 +1965,36 @@ def _find_smarts_matches(rdmol, smirks, aromaticity_model="OEAroModel_MDL"):
         """
         from rdkit import Chem
 
+        def _match_smarts_with_heavy_atoms_first(rdmol, qmol, match_kwargs):
```
j-wags, line 1968:
```suggestion
        # This code is part of a possible performance optimization that hasn't been validated 
        # for production use yet.
        def _match_smarts_with_heavy_atoms_first(rdmol, qmol, match_kwargs):
```

j-wags (changes requested):
Thanks -- This is looking really close to production-ready, though I've found a few places where I think a kwarg should be getting passed/possible typos. 

Also, could you add a pair of tests for `RDKTKW.find_smarts_matches` and `OETKW.find_smarts_matches` to ensure that matching `CCC` to propane returns two matches when `unique=False` and just one when `unique=True`?

openff/toolkit/typing/engines/smirnoff/parameters.py
mattwthompson: Yes, we do. I got wrapped up in out-of-scope details when thinking about how to wire it through. I think the argument is passed through each handler now; please verify that this is what you had in mind.

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -4017,7 +4032,9 @@ def from_molecule(cls, molecule):
     _INFOTYPE = LibraryChargeType  # info type to store
     _DEPENDENCIES = [vdWHandler, ElectrostaticsHandler]
 
-    def find_matches(self, entity):
+    _KWARGS = ["unique"]
```
j-wags, line 4035: (not blocking) I don't think this should be here -- It's a list of strings that tells `ff.create_openmm_system(args, kwargs)` which kwargs are consumed by `LibraryChargeHandler.create_force` and so should be forwarded on. As is coded now there isn't special handling for the `unique` kwarg in `LibraryChargeHandler.create_force` so this isn't doing anything. I think it should be removed but may have overlooked a detail.

j-wags (approved):
This looks great. Thanks for iterating on it @mattwthompson, and thanks @lilyminium for pushing the idea to begin with! I suggested one more cleanup, other than that, please update the releasehistory (maybe file this under _New Features and Behavior Changed_ just in case there's unintended side effects?) and this should be good to merge!

openff/toolkit/typing/engines/smirnoff/parameters.py
mattwthompson: That sounds correct, thanks for catching this. Lost to history is a period of time in which I confused myself about which kwargs are passed to `find_matches` and `create_force`, and even a brief attempt at refactoring how each handler accepts kwargs.

PR #1038, opened 2021-08-02
Update version checks in CI

j-wags:
The current "latest version finder" lines in our CI aren't sorting versions correctly (`0.10.0` appears next to `0.1.0` instead of after `0.9.2`)


mattwthompson:
Nice catch!

PR #1041, opened 2021-08-10
Raise more informative exception on post-typing connectivity failures

mattwthompson:
While debugging something else, @lilyminium noticed that when connectivity checks after parametrization fail, the exception is not so informative. Right now they're something like

```
NotBondedError: No bond between atom 4 and 19
```

Not only does this not specify which parameter caused the failure, it also fails to specify which _type_ of parameter cause it, i.e. this exception does not say if the failure was from a bond, angle, or dihedral. Now the bottom of the traceback gives some breadcrumbs:

```
NotBondedError: While processing improper with SMIRKS [H][C@@:1]([C])([N+])[C:3](=[O:4])[N:2]([H])[C]: No bond between atom 4 and 19
```

The same exception type is raised, so I wouldn't anticipate this breaking existing code.

If this is approved I intend to cherry-pick it into the topology feature branch.

### DIFF
--- a/openff/toolkit/typing/engines/smirnoff/parameters.py
+++ b/openff/toolkit/typing/engines/smirnoff/parameters.py
@@ -69,6 +69,7 @@ from openff.toolkit.topology import (
     ValenceDict,
 )
 from openff.toolkit.topology.molecule import Molecule
+from openff.toolkit.topology.topology import NotBondedError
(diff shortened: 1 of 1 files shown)


j-wags (approved):
LGTM!

mattwthompson:
The diff coverage is a little concerning, but probably due to other moving pieces. Might be worth looking into another time.

PR #1042, opened 2021-08-11
Add moldoc

mattwthompson:
It works locally :tm:

image

This was surprisingly quick and straightforward to spin up, so it's possible that we'd actually be able to roll this out if we saw a need that it fills. It should be as simple as adding blocks like this:

```
.. moldoc::

    molecule.generate_conformers(n_conformers=1)

    moldoc_display_molecule = molecule.to_moldoc()

```
`docmol` just looks for an object named `moldoc_display_molecule` and tries to display it if it's the right class. It has its own minimal molecule class, which  `Molecule.to_moldoc()` returns. IIRC this can be put into both rST files directly ("Book") or tucked away in docstrings ("Reference").

cc: @Yoshanuikabundi 

- [ ] Package onto `conda-forge`


j-wags:
@Yoshanuikabundi -- I'm generally against merging this PR because I'm out of bandwidth to maintain any more optional deps/extensions. But I'm also known for writing bad docs, so this PR doesn't have much of an upside for me. Could you review this and let me know if it would slot into a current need in our docs? If so we should merge it, but otherwise I'd prefer to avoid complexity.

Yoshanuikabundi:
@mattwthompson I think this is really cool! Unfortunately it doesn't work for me locally or on RTD, and I'm not sure it solves enough of a problem for the docs to be worth putting a whole lot of effort into it, especially if it's just going to break again after another month or so. Maybe we should close this, and if things change in the future and we need it we can reopen?

j-wags:
Sounds good. I'm in favor of keeping this "parked" in this branch so that we can come back if we find a cool spot to add this to the docs!

PR #1043, opened 2021-08-15
Use mamba for CI checks

Yoshanuikabundi:
We've had slow CI for a long time. Conda takes more than 8 minutes just to install dependencies in CI. With Mamba, entire documentation builds can complete in about 5 on RTD. Conda incubator supports Mamba, and I've been using this approach without any problems on the theme repo, so maybe it helps here.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -70,6 +70,9 @@ jobs:
           activate-environment: test
           environment-file: devtools/conda-envs/test_env.yaml
           auto-activate-base: false
+          miniforge-version: latest
+          miniforge-variant: Mambaforge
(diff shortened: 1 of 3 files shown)


Yoshanuikabundi:
It looks like dependency solving takes seconds rather than minutes, but downloading still takes time. Adds up to 3-5 minutes saved on each check, which is a third to a quarter of the total time.

mattwthompson:
I thought we'd been using mamba this entire time :pensive:

I've re-kicked CI, which should pass if QCArchive is back online

j-wags (approved):
Suh-weet! Thanks @Yoshanuikabundi!

PR #1059, opened 2021-08-27
Preparations for supporting type annotations

mattwthompson:
Typing infrastructure in the standard library is only recently starting to harden and third-party support is somewhere between zero and unreliable in most cases. Still, I think it's time to make this push. NumPy recently started shipping  with (partial) support and my hope is that it won't be too long until it becomes more common in scientific software in general. Actually implementing type annotations can be tricky, and because things are still in flux (i.e. different behavior on Python 3.7 versus 3.8+), there's a certain amount of creativity to the implementation. I'm sure there are cleaner ways to do what I've done here.

Mypy by default is not strict, so if something does not have type annotations, it will be ignored and mypy will not complain. This happens with most of the codebase. So, despite almost nothing here having annotations, this can happen:

```
$ mypy --show-error-codes --warn-unused-configs --warn-unused-ignores --namespace-packages -p "openff.toolkit"
Success: no issues found in 62 source files
```

There are a handful of other tiny things that it will complain about, however, and this PR fixes those.  These changes should have no effect on users, and until a `py.typed` file is added there is no downstream guarantee that stubs are shipped with packaged versions. See [PEP 561](https://www.python.org/dev/peps/pep-0561/).

This should be followed up with a more involved PR that aims to add [PEP 484](https://www.python.org/dev/peps/pep-0484/)-style annotations to some/most classes, functions, etc. and not break things; once something does have an annotation, mypy will complain instead of ignoring them. A long-term goal would be to run `mypy --strict ...`, but that's quite a ways off and arguably not worth the trouble.


openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -1812,26 +1812,27 @@ class ParameterHandler(_ParameterAttributeHandler):
...
+    _SMIRNOFF_VERSION_INTRODUCED = 0.0
+    _SMIRNOFF_VERSION_DEPRECATED = None
+    # if deprecated, the first SMIRNOFF version number it is no longer used
     _MIN_SUPPORTED_SECTION_VERSION = 0.3
     _MAX_SUPPORTED_SECTION_VERSION = 0.3
 
     version = ParameterAttribute()
 
-    @version.converter
+    @version.converter  # type: ignore[no-redef]
```
j-wags, line 1835: In general I like this PR. However these directives to make mypy happy kinda hurt the readability of the codebase - I'd prefer if there was some way to keep these out/minimize them so that first-time contributors aren't getting confused or getting a failing linter status.

openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -524,7 +524,7 @@ def enumerate_stereoisomers(
 
         """
         from rdkit import Chem
-        from rdkit.Chem.EnumerateStereoisomers import (
+        from rdkit.Chem.EnumerateStereoisomers import (  # type: ignore[import]
```
mattwthompson, line 527: A comment like this could be removed and instead moved into the config file. Neither are great solutions but putting it in the config file makes it re-usable if new additions to the toolkit also pull from this submodule.

j-wags (approved):
I think this is good to go. It'll just be good to make sure of two things:

1) Check whether type annotations could break compatibility with anything kinda odd that people may be using in conjunction with OFFTK (like, I'd like to know if this would break py3.5 compatibility, since this is a nice-to-have feature but not absolutely essential)
2) (not immediately actionable) Ensure that we don't take the FUTURE step of doing something functional that has a hard requirement on type annotations, and therefore could somehow break upstream/downstream/forward/reverse/plugin compatibility.

mattwthompson:
It seems like this [can't work](https://simondobson.org/blog/2020/12/09/backporting-python-types/) (without extensive hacks) on Python 3.5 since even relatively basic [variable annotation](https://stackoverflow.com/a/51789147/4248961) wasn't introduced until 3.6. It was a serious effort to even get an environment with 3.5 in it; I had to pin to old versions of several packages and also install from multiple non-standard labels, which also resulted in pulling down some old versions of other packages (OpenMM 7.1, pytest 4.5, NumPy 1.14!) which is not at all advisable.

I made some headway in testing against Python 3.6, but surprisingly found that we already broke it in #1021 with the use of PEP 562! Getting an environment working with 3.6 wasn't as difficult, but it's getting near its [EOL](https://www.python.org/dev/peps/pep-0494/#lifespan) upstream and has been dropped by [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table) long ago. (It also recommends dropping 3.7 this December.)

I'm still in favor of merging this, but I'll give an opening for objections before doing so.

lilyminium:
Just chiming in in general favour of dropping old Python versions. New features are fun.

> (like, I'd like to know if this would break py3.5 compatibility, since this is a nice-to-have feature but not absolutely essential)

3.5 especially reached end of life last year, and 3.6 will in December this year (as Matt mentioned).

PR #1061, opened 2021-08-30
Minor fixes for OpenMM 7.6 API changes

mattwthompson:
Until https://github.com/ParmEd/ParmEd/issues/1185, a significant portion of our tests will fail. This band-aids the issue if we need CI to pass.

The ideal solution involves everything being fixed upstream and this PR being closed without merging.


mattwthompson:
Related band-aid in package builds: https://github.com/conda-forge/openff-toolkit-feedstock/pull/12

mattwthompson:
There's now a patch that should result in OpenMM 7.5.1 being installed with ParmEd 3.4.2.


Later today, ParmEd 3.4.3 should be online and allow OpenMM 7.6 to be installed and work.

This PR is just used as a place to test these combinations - hopefully we can close this today.

mattwthompson:
Okay, 7.5.1 is being pulled in, as expected/hoped for.

Hilariously, tests are failing because _the API break resulted in a dead link to API docs._

mattwthompson:
I'm not sure why Python 3.7 gives different behavior - most things are still built with 3.6+ and I can't replicate the error locally.

If https://github.com/conda-forge/parmed-feedstock/pull/19 isn't merged by tomorrow morning, I'll have to merge this in so that we can make progress on other PRs.

mattwthompson:
The `Python 3.7, RDKit=true, OpenEye=true` builds are still pulling down the wrong versions. At some point it's downgrading to AmberTools  20.15 whereas one would expect 21.7 to be pulled down.

mattwthompson:
I can't reproduce this locally. I investigated the deptree and found nothing with a problematic pins around these packages. `conda env create ...` works fine but something unexpected is happening inside of `setup-miniconda`, which uses `mamba`, when it does `mamba env update ...`.

mattwthompson:
https://github.com/conda-forge/parmed-feedstock/pull/19 is merged and I'm pulling down 7.6/3.4.3 locally! Hopefully computers in an unspecified data center produce the same behavior ...

mattwthompson:
Everything is working for 7.6 and 7.5, at least in the core set of tests and examples. Merging.

j-wags:
Thanks, Matt!

PR #1063, opened 2021-09-02
Safely import OpenMM

mattwthompson:
This is a follow-on to #1061, which I was letting get bloated. While working on the notebooks there, I neglected to consider that the actual source code is responsible for emitting the `Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.`

This PR aims to fix that so that users aren't confused by a deprecation warnings they're not responsible for. This is a tedious fix, but I don't think there's any way around it. Simply dropping the old imports breaks anything that might happen to use `from simtk.openmm import ...` and requires we pin `openmm >=7.6` everywhere in our stack.

- [x] Make sure the warning doesn't show up in test logs


j-wags (approved):
This looks great @mattwthompson. Could you do a final sanity check and re-run CI with openmm=7.5 and 7.6, just to make sure that nothing slips between the cracks? Then our CI could test against 7.6 forever after. 

We won't plan on pinning to `openmm>7.6`, since it's such a core part of our ecosystem and that pin has the potential to break a whole ton of recipes. But new features may eventually break 7.5 support, and we'll wait for a user report on that day so we know when to add a pin.

mattwthompson:
The "slow" tests on macOS have been running for 3 hours, but everything else has worked as expected. I checked a log and the warning wasn't being emitted, so I'm hopeful it won't show up when our users import one of our classes.

PR #1064, opened 2021-09-07
Replace nbqa-black with black

mattwthompson:
The creator/maintainer of nbQA dropped by to say he'll soon deprecate the `nbqa-black` hook now that it is natively supported in `black`. https://github.com/openforcefield/openff-toolkit/pull/1060#issuecomment-912955601

The process was:
1. Edit `.pre-commit-config.yaml`
2. Run `pre-commit run --all-files` - this is how there is a diff in `examples/conformer_energies/conformer_energies.py`, which was previously missed because the tool only looked for notebooks in that path
3. `pre-commit autoupdate` - no changes

I one-off tested this and found that it would catch formatting errors in notebook files as one would expect.  The only "gotcha" is that I needed to do `pip install black[jupyter]`, which was clearly indicated in the error message if it isn't installed. Something was funky with the executable not being found from the `conda-forge` build, but I doubt that's a problem outside of my machine and not a huge issue given it's not typically necessary for developers to run it manually.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,8 @@ repos:
   - id: black
     files: ^openff
     args: [--check]
+  - id: black-jupyter
+    files: ^examples/((?!deprecated).)*$
 - repo: https://github.com/PyCQA/isort
   rev: 5.9.3
   hooks:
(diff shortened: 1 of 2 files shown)


j-wags (approved):
Excellent. Thanks @mattwthompson! It's especially convincing that the current maintainer of `nbqa-black` is recommending that people migrate. No need to update the docs or releasenotes, this should be good to merge :-)

PR #1065, opened 2021-09-08
Update Example notebooks to Sage force field

Yoshanuikabundi:


mattwthompson (approved):
* I've skimmed through each commit and the changes look fine, as far as I can tell. I have no doubt you made the one-liner updates accurately, and in the case of a typo the failures should crash the notebooks yet there are green check marks everywhere.
* Re-running the notebooks every time makes it difficult to tell what's actually being changed vs. the differences as a result of stochastic outputs. In the past I've tinkered with services (https://github.com/jupyter/nbdime ?) but I can't recall anything that solves this problem. Non-blocking/out of scope, but something to think about. Maybe raise an issue if this is a topic you've already done research on?
* It might be nice to have #1064 merged into this one, although the diffs imply you've manually brought those changes into here. Non-blocking.
* Adding each `.gitignore` like you did is super appreciated. At very least this makes my local copy of the toolkit cleaner, but it also safeguards us against aggressive `git add .`

examples/conformer_energies/conformer_energies.py
```diff
@@ -29,42 +30,59 @@ def compute_conformer_energies_from_file(filename):
 
     n_molecules = len(molecules)
     n_conformers = sum([mol.n_conformers for mol in molecules])
-    print(f'{n_molecules} unique molecule(s) loaded, with {n_conformers} total conformers')
+    print(
+        f"{n_molecules} unique molecule(s) loaded, with {n_conformers} total conformers"
+    )
 
     # Load the openff-1.1.0 force field appropriate for vacuum calculations (without constraints)
```
j-wags, line 37: (blocking) Thanks! @mattwthompson's approval stands, I just wanted to add this one to-do because this script is special case. This file is kinda weird since its neither library code nor a notebook example, and the way that it's structured will make this be a "silent behavior change" for the users. So in addition to this change, we should

- [ ] Update the README for this example to state that it now uses 2.0.0
- [ ] Record the change of force field in the `Behavior changed` section of the releasehistory. 
- [ ] (if possible) print which FF is used in the print statement at the end of this script

```suggestion
    # Load the openff-2.0.0 force field appropriate for vacuum calculations (without constraints)
```

PR #1072, opened 2021-09-09
[WIP] Perceive common caps

ijpulidos:

This now allows for the perception of some common caps when perceiving residues.


ijpulidos:
Caps are extracted from the CCD but since they are in the small molecule free form, thus it doesn't knows it is part of some protein chain, I had to manually flag one of the H's in ACE as leaving to account for this. Partially works so far, but we would probably want a more robust way of getting this directly from the CCD cif files.

j-wags (approved):
Awesome @ijpulidos! I see this is getting tangled up with #1081, so let's see if we can get this merged. I think this is good to merge in its current form, except that it'd be good to have some way to programmatically add the caps to the `aa_substructures...json` file (even if that's just a python script that contains the additional dict entries to add).

Feel free to merge without tests passing, just be sure to roll over any additional to-dos (like "find a way to get the caps directly from the CCD"  into the issue description in #951.

PR #1074, opened 2021-09-10
Add flake8

mattwthompson:
I iteratively ran `flake8 openff` and either fixed its warnings or turned them off (either via `setup.cfg` or `# noqa` comments).

This will allow us to comfortably remove LGTM, which isn't even a tool we can run ourselves.


mattwthompson:
We won't get this one merged in the current release line or the next one. I'd still love to adopt flake8 in this codebase at some point!

PR #1081, opened 2021-09-16
Strict chirality option for perceiving residues

ijpulidos:
Perceiving residues for molecule now has the option to specify if it uses strict chirality or not, by ignoring the stereomarks (`@`) in the SMARTS substructures matching. This is useful for cyclic peptides which commonly have different chiralities.

### DIFF
--- a/openff/toolkit/tests/test_molecule.py
+++ b/openff/toolkit/tests/test_molecule.py
@@ -3767,13 +3767,14 @@ class TestMolecule:
 
         assert isinstance(mol.visualize(backend="openeye"), IPython.core.display.Image)
 
-    def test_perceive_residues_natoms_nterminal_alanine(self):
+    @pytest.mark.parametrize("strict_chirality", (True, False))
+    def test_perceive_residues_natoms_nterminal_alanine(self, strict_chirality):
         """Test number of matches atoms in residue perception with NTerminal form of Alanine."""
         offmol = Molecule.from_file(
             get_data_file_path('proteins/NTerminal_ALA.sdf')
         )
         # Perceive residue substructures
-        offmol.perceive_residues()
+        offmol.perceive_residues(strict_chirality=strict_chirality)
         counter = 0  # matched atom counter
         for atom in offmol.atoms:
             if atom.metadata:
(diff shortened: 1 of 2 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -6468,14 +6468,40 @@ def visualize(self, backend="rdkit", width=500, height=300):
...
+            substructure_dictionary_no_chirality = deepcopy(substructure_dictionary)
+            # Update inner key (SMARTS) maintaining its value
+            for res_name, inner_dict in substructure_dictionary.items():
+                for smarts, atom_types in inner_dict.items():
+                    smarts_no_chirality = smarts.replace('@', '')  # remove @ in smarts
+                    substructure_dictionary_no_chirality[res_name][smarts_no_chirality] = \
+                        substructure_dictionary_no_chirality[res_name].pop(smarts)  # update key
+            # replace with the new substructure dictionary
+            substructure_dictionary = substructure_dictionary_no_chirality
+
```
lilyminium, line 6504: It doesn't look like this information will change over multiple function calls to me. Perhaps you could make and read another file?

openff/toolkit/tests/test_molecule.py
```diff
@@ -3887,14 +3894,15 @@ def test_perceive_residues_natoms_mainchain_histidine(self):
         assert unlabelled_counter == 12
         assert counter == offmol.n_atoms - 12
 
-    def test_perceive_residues_natoms_cyxteine(self):
+    @pytest.mark.parametrize("strict_chirality", (True, False))
```
lilyminium, line 3897: If these perceive residues tests were their own class (e.g. TestMoleculeResiduePerception), I think you could parametrize the whole class at once?

lilyminium: Also, it would be nice to have a test where changing `strict_chirality` affected the outcome of `perceive_residues` :)

openff/toolkit/topology/molecule.py
ijpulidos: As far as I see this, we might want sometimes to read with strict chirality and sometimes without it, so, I guess it is fine that this doesn't persist in multiple function calls. It is true that this will need to read the substructure dictionary every time but I think that's fine for now, that's fast and the memory footprint should be negligible. Maybe something to consider if we have a LOT of molecules, true.

In the long run we could be just fine by not using strict chirality at all, and making that the default. We would have to check this with detail.

openff/toolkit/tests/test_molecule.py
ijpulidos: Yes to both. Thanks! Will do that.

openff/toolkit/topology/molecule.py
j-wags: I agree with @ijpulidos -- We're still in "behavior seeking" mode right now with this new functionality, so it's probably best to keep this "messy" to make it easier to modify in later iterations.

ijpulidos:
I'm closing this PR in favor of #951 which already includes the same features/capabilities. Sorry for the noise, it seemed like a good idea at first, but I needed this in the same branch in the end.

PR #1085, opened 2021-09-20
Initial implementation of HierarchyElements and Schemes

j-wags:
- [x] Implement `Molecule`-level `HierarchySchemes` and `HierarchyElements`, providing ways to generate a static index where atoms are grouped according to their common metadata.
- [ ] Refactor to make Topology objects consist of explicit copies of `Molecule`s, deprecating `TopologyMolecule`, `TopologyAtom`, `TopologyVirtualSite`, etc
- [ ] Implement access to `HierarchySchemes` from the Topology level.
- [ ] Implement `to`/`from_dict` methods for Topology


openff/toolkit/tests/create_molecules.py
```diff
@@ -18,8 +18,11 @@
 
 import numpy as np
 from simtk import unit
+import pytest
```
mattwthompson, line 21: I wonder if the existing molecules should be fixtures as well?

j-wags: I think they should all become fixtures, and I had to stop myself from converting them all over so that this PR wouldn't get absolutely humongous.

mattwthompson: Sounds good - agree it's best to leave it be in this scope

openff/toolkit/utils/_cif_to_substructure_dict.py
```diff
@@ -6,13 +6,15 @@
 """
 
 import copy
+import json
```
j-wags, line 9: Skip reviewing this file -- Its only change was getting blacked and isorted

openff/toolkit/utils/base_wrapper.py
```diff
@@ -15,12 +15,14 @@
 )
 
 from .constants import DEFAULT_AROMATICITY_MODEL
+
```
j-wags, line 18: Skip reviewing this file -- Its only change was getting blacked and isorted

openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -14,10 +14,9 @@
 from collections import defaultdict
 from functools import wraps
 from typing import TYPE_CHECKING, List, Optional, Tuple
-from cachetools import LRUCache, cached
-
 
 import numpy as np
+from cachetools import LRUCache, cached
```
j-wags, line 19: Skip reviewing this file -- Its only change was getting blacked and isorted

openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -1615,8 +1615,8 @@ def from_rdkit(
             offmol.partial_charges = None
         return offmol
 
-
     to_rdkit_cache = LRUCache(maxsize=4096)
+
```
j-wags, line 1619: Skip reviewing this file -- Its only change was getting blacked and isorted

openff/toolkit/topology/molecule.py
```diff
@@ -70,22 +76,6 @@
...
-)
-
-class NotAttachedToMoleculeError(MessageException):
-    """Exception for when a component does not belong to a Molecule object, but is queried"""
-
-
-class InvalidAtomMetadataError(MessageException):
-    """The program attempted to set atom metadata to an invalid type"""
-
-
```
j-wags, line 88: These were moved to `exceptions.py`

j-wags:
@mattwthompson Could you take a glance at this PR in its current state? There's lots of linting nonsense, but the big changes I could use eyes on are the new classes in `molecule.py` and the new tests in `test_molecule.oy` 

The TopologyMolecule deprecation changes are getting really extensive, so I'll do them as a separate PR. So this one should enable manipulating `Molecules` using the hierarchy API, but hierarchy iterators can't be exposed on `Topology` until `TopologyMolecule` is deprecated.

openff/toolkit/topology/molecule.py
```diff
@@ -2726,6 +2735,73 @@ def __eq__(self, other):
...
+        self._hierarchy_schemes.pop(iter_name)
+        if hasattr(self, iter_name):
+            delattr(self, iter_name)
+
+    def perceive_hierarchy(self, iter_names=None):
+        """
+        Parameters
+        ----------
+        iter_names : Iterable of str, Optional
+            Only perceive hierarchy for HierarchySchemes that expose specific iterator names.
```
mattwthompson, line 2789:
```suggestion
            Only perceive hierarchy for HierarchySchemes that expose these iterator names.
```

openff/toolkit/tests/test_molecule.py
```diff
@@ -4027,3 +4020,78 @@ def test_molecule_subclass_from_dict(self):
         orig_mol = Molecule.from_smiles("CCO")
         mol = MyMol.from_dict(orig_mol.to_dict())
         assert isinstance(mol, MyMol)
+
+
+class TestHierarchies:
+    def test_nothing_perceived_dipeptide(self, dipeptide):
+        with pytest.raises(KeyError) as context:
```
mattwthompson, line 4027:
```suggestion
        with pytest.raises(KeyError):
```
FYI: When using this context manager, you don't need to store the return value if not using it. Flake8 will complain about this

openff/toolkit/topology/molecule.py
```diff
@@ -2726,6 +2735,73 @@ def __eq__(self, other):
...
+    ):
+        """
+        Parameters
+        ----------
+        uniqueness_criteria : tuple of str
+        iterator_name : str
+            Name of the iterator that will be exposed to access the HierarchyElements generated
+            by this scheme
+        sort_func : function
+            The function for deciding the iteration order of HierarchyElements generated by this scheme.
```
mattwthompson, line 2757: Probably drop this from the docstring but add it as a TODO, since it looks like it will be hard to implement on the first pass?

openff/toolkit/topology/molecule.py
```diff
@@ -2726,6 +2735,73 @@ def __eq__(self, other):
...
+
+        Returns
+        -------
+        new_hier_scheme : openff.toolkit.topology.HierarchyScheme
+            The newly created HierarchyScheme
+        """
+        assert iterator_name not in self._hierarchy_schemes
```
mattwthompson, line 2764:
```suggestion
        if iterator_name in self._hierarchy_schemes:
            raise IteratorNameAlreadyInExistingHierarchySchemesOrSomethingOrOtherError
```

I ran into this while calling `._add_default_hierarchy_schemes()` a second time on the same molecule, so I think it will be a relatively common exception to hit and worth ma king a custom exception

mattwthompson (commented):
Things look good, at least as deep as I can understand them. I was able to do a small amount of tinkering in an interpreter, as guided by the unit tests, but I think I'll be able to do more once this is all shipped up to the `Topology` level. I've added a few minor suggestions below, some of which you've probably already thought of.




A small improvement would be a getter for the current hierarchy schemes and/or iterator names, enabling this sort of thing without accessing private methods:


```python3
In [39]: dipeptide._hierarchy_schemes.keys()
Out[39]: dict_keys([])

In [40]: dipeptide._add_default_hierarchy_schemes()

In [41]: dipeptide._hierarchy_schemes.keys()
Out[41]: dict_keys(['residues', 'chains'])

In [42]: dipeptide.delete_hierarchy_scheme('chains')

In [43]: dipeptide._hierarchy_schemes.keys()
Out[43]: dict_keys(['residues'])
```

Another wishlist item (for some point in the future) would be an example that 
 both details how the default behavior is implemented and also implements custom hierarchy scheme.

openff/toolkit/topology/molecule.py
```diff
@@ -6629,3 +6722,149 @@ def _ipython_display_(self):
...
+        particle : openff.toolkit.topology.Particle
+        """
+        return self._particles[index]
+
+    @property
+    def parent(self):
+        return self.scheme.parent
+
+    def __str__(self):
+        return f"HierarchyElement {self.identifier} containing {len(self.particles)} particles"
```
ijpulidos, line 6867: Just a detail here, `self.particles` is a generator so applying `len` to it returns a `TypeError`. Now, I wonder if we should add tests for these things (maybe some tests that are specific to `HierarchyElement` and `HierarchyScheme` classes/objects), what do you think?

ijpulidos (changes requested):
This is awesome @j-wags, very nice to see this coming together!! I was reviewing our original API design/plan and there is an option where we want to initialize a `Molecule` with some hierarchy flavor. Do we want to have this in this iteration?

I suggested a small change. In any case this seems great!

openff/toolkit/topology/molecule.py
j-wags: Wow!  Thanks for the close review :-) Fixing that uncovered a ton of other bugs, and now I've fixed those and added more tests. 

Unfortunately, the way it's designed, we can't test `HierarchySchemes` or `HierarchyElements` individually, since we can't instantiate them without a parent `Molecule` (for schemes) or `HierarchyScheme` (for elements). 

(The OpenFF `Atom` class is a good example of what happens if we allow users to treat parents as optional. An `Atom` CAN be instantiated without a `Molecule` (it's allowed to be `None`), but about half of API points just throw errors if a user tries to access them, since they need information from the `self.molecule`. So that's a design decision I regret from earlier in OpenFF)

openff/toolkit/topology/molecule.py
j-wags: Fixed in fba36aa

openff/toolkit/topology/molecule.py
j-wags: Done in fba36aa

j-wags:
Thanks @ijpulidos and @mattwthompson for the great feedback! I've incorporated some of those suggestions tonight, and will do the others tomorrow.

Address or implement: 
- [ ] (From Matt's review) Add "a getter for the current hierarchy schemes and/or iterator names"
- [ ] (From Ivan's review) our original API design/plan and there is an option where we want to initialize a Molecule with some hierarchy flavor. Do we want to have this in this iteration?
- [ ] (From Matt's review)  an example that both details how the default behavior is implemented and also implements custom hierarchy scheme. (I don't plan to do that here, but will likely push this to-do up to the larger PR)

j-wags:
Regarding @ijpulidos question:

> our original API design/plan and there is an option where we want to initialize a Molecule with some hierarchy flavor. Do we want to have this in this iteration?

I _do_ want this eventually, but maybe not in this iteration. My attempts to do this found that it adds a lot of complexity (for example, if `Molecule.from_pdb` populates some `HierarchySchemes`, we have to get a bit clever about how we handle the double-addition errors if molecules also default to be initialized with those `HierarchySchemes`). So I'm going to bump this up to the larger PR.

j-wags:
I'm merging this and promoting the remaining to-dos to the larger topology refactor PR.

PR #1086, opened 2021-09-21
Add `Topology.to_smiles`

jthorton:
- [ ] This feature is to help with QM dataset generation for multi component systems see [here](https://github.com/openmm/qmdataset/issues/4#issuecomment-923369287) for more details.


j-wags (commented):
So I've been hesitating on this because I've been busy, and I'm concerned about the direct string manipulation of SMILES in this PR (the `'.'.join(smiles)` process). 

This morning @jthorton had mentioned an alternative: that we could convert the whole topology to an RDMol and have generate the SMILES from that. But thinking some more, that would add a lot of complexity to the RDKit interface. AND, if we did it that way, we'd want to also add equivalent functionality for OE, so the function can be run with OE-only builds. So upon further thought, that's a lot of work, and it doesn't seem like a better alternative.

I spent some time trying to think of a third way, and nothing suitable came to mind. 

So, I'll go ahead and move this PR forward, since it'll give us a more natural path for multi-component QC submissions than holding multiple interconnected graphs in a single OFFMol/RDMol. Let's make sure that this does something sensible in the future when we start making TypedMolecules, since they won't have the chemical info needed to make a SMILES.

j-wags:
@jthorton I think this is ready in its current form, but I do wonder - Will this be useless without a `Topology.from_mapped_smiles` method as well? Or will the mapped smiles be processed later through a different mechanism?

mattwthompson:
This has gone a bit stale since the fall - if people still want these changes, could this be rebased onto `main`?

jchodera:
This would indeed be super useful!

j-wags:
Just a weekend drive-by comment on this since it's gone so long without progress, and we're making the decision on whether to put more work into it - This set of functionality really worries me the more I think about it, for two reasons:

* To the best of my knowledge, there is no API point anywhere that will read a multi-molecule mapped SMILES back into an OFF Topology. Even worse, there are many API points that will read it without complaint and provide something surprising and wrong
* The manual manipulation of SMILES (a very complex format) as a strings smells funny to me

The first one in particular seems dangerous. Is this less severe than I think? Or is this functionality critical to Rosemary or another deliverable?

jchodera:
> To the best of my knowledge, there is no API point anywhere that will read a multi-molecule mapped SMILES back into an OFF Topology. 

Definitely agree that the pair of methods (Topology -> mapped SMILES -> Topology (+Molecules)) is probably best implemented together.

peastman:
What's the status of this?  I have a case where I really need this functionality.  I have a Topology composed of multiple Molecules, and I need to generate a single canonical ordered mapped SMILES string for it.  Is that possible now?

j-wags:
I think the code in this PR probably works fine in a pinch, but it's outside the scope of what the OpenFF Toolkit should do. Again - I don't know of any other tools that will ingest a mapped topology SMILES in a meaningful way, and I'm not enthusiastic about using it as a serialization format for OpenFF Topologies (especially given that we already support several serialization formats for them). 

Can you use the code from this PR in your own script? It may make more sense that way, since my major hesitation is that these ~20 lines of code add a new public API point that seems highly likely to cause user confusion.

jchodera:
Are we positive we don't want to provide this as an experimental method? We could mark the API as "experimental and subject to change". 

There are a number of use cases where this feature would be important:
* The current use case is ensuring that data generated with QCFractal is useful for force field parameterization or assessment when the system contains more than one molecule. OpenFF will definitely need to use this feature for assessing force fields, building force fields containing metals (an integral part of our OpenFF NIH proposal), and generating data in which interaction energetics may be necessary.
* When paired with a `.from_smiles()` method, indexed SMILES strings specifying more than one molecule provide a convenient way to construct unit tests. We use this frequently with single molecules, but this would be incredibly convenient for multiple molecules.
* Aside from indexed tagged SMILES, there is really no other portable way to represent molecular topologies as a string other than SDF files, which are very much not compact and inherently contain 3D information that may be unused.

j-wags:
I don't think the "experimental and subject to change" approach will work if this functionality will need to be stably supported to access production datasets.

> there is really no other portable way to represent molecular topologies as a string

`topology.to_json` should work within the same OFFTK version.

If I'm interpreting this correctly, the major use case here is to be able to submit/retrieve multi-molecule jobs on QCFractal. If that's the case, I'd be a bit more comfortable implementing this in a different form, possibly as `Topology.to/from_qcschema` or something equivalent either here or in QCSubmit. 

Would the creation of `Topology.to/from_qcschema` methods be a satisfactory solution here?

jchodera:
> I don't think the "experimental and subject to change" approach will work if this functionality will need to be stably supported to access production datasets.

The API could be experimental and subject to change, but the behavior is simple and well-defined. The API point could be migrated elsewhere, but the tagged SMILES strings would still be unambiguous and compatible with major cheminformatics toolkits such as RDKit and OpenEye.

> topology.to_json should work within the same OFFTK version.

I think we need this to be (1) portable (could easily be implemented in or interpreted by other codes) and (2) survive beyond a single OFFTK version. For (2), the tagged SMILES format is already unambiguous and directly compatible with major cheminformatics toolkits such as RDKit and OpenEye.

> Would the creation of Topology.to/from_qcschema methods be a satisfactory solution here?

The problem with `Topology.to/from_qcschema` is that the [QCSchema spec](https://molssi-qc-schema.readthedocs.io/en/latest/spec_components.html) includes much more information that is far out of scope for `Topology`, such as [`driver`](https://molssi-qc-schema.readthedocs.io/en/latest/spec_components.html#driver), [`model`](https://molssi-qc-schema.readthedocs.io/en/latest/spec_components.html#model), [`keywords`](https://molssi-qc-schema.readthedocs.io/en/latest/spec_components.html#keywords) and even [`conformation`](https://github.com/openforcefield/openff-toolkit/pull/1086) is optional. The standard [QCSchema spec](https://molssi-qc-schema.readthedocs.io/en/latest/spec_components.html#keywords) does not include information that the tagged SMILES string does---instead, [we encode the relevant information in an `extras` block](https://github.com/openforcefield/openff-toolkit/blob/dbd3e9e62afe7ee45dd3b55f0a12ffa2f469fe75/openff/toolkit/topology/molecule.py#L4434) which is not standardized. 

Considering there will be other storage formats for distributing the QCArchive data---including several HDF5 formats floating around---this doesn't seem like the best route if only the tagged SMILES is required.

peastman:
It does seem like something that ought to be built in.  The SMILES syntax for multiple molecules is well defined, and other programs have no trouble with it.  The thing that's different in OpenFF Toolkit is that its API has more levels: Molecule to represent a single molecule, and Topology to represent multiple interacting molecules.  RDKit doesn't have anything comparable to Topology.  It uses a single Mol class to represent both single molecules and multiple interacting molecules.  Since OpenFF has different classes for the two cases, I'd expect both of them to support SMILES conversion.

mattwthompson:
> Molecule to represent a single molecule, and Topology to represent multiple interacting molecules 

This is not currently the case, though I understand it to be the intended design. It would be great to enforce it (#1587) independently of this context, though it could safeguard against some footguns like `Molecule.from_smiles(my_topology.to_smiles())`

peastman:
It isn't strictly required that you use Topology, but trying to avoid it and use Molecule exclusively is very difficult.  For example, loading Molecules from PDB files doesn't work correctly (#1657).  And you can create a Topology from an OpenMM Topology, but not a Molecule.

An alternative would be to officially support representing multiple molecules with a single Molecule object, and change `Molecule.from_topology()` to remove the requirement that the Topology only contain a single Molecule.

PR #1079, opened 2021-09-14
Switch to `openff-units` registry

mattwthompson:
- [ ] Tag issue being addressed
- [ ] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/tests)
- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable
- [ ] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [ ] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/master/docs/releasehistory.rst)

PR #1088, opened 2021-09-23
Use Quantity.value_in_unit() instead of dividing quantities by units

mattwthompson:
This is a subset of #1079 that should be a pure upgrade with no API breaks, changed dependencies, etc.

OpenMM quantities can have their units stripped (and appropriate conversions applies) to float/array/etc. representations by dividing by a `Unit` object. This works fine, but  argue it's a little unsafe given that the same thing is accessible from the API.

```
>>> from openmm import unit
>>> x = 1.0 * unit.nanometer
>>> x
Quantity(value=1.0, unit=nanometer)
>>> x.value_in_unit(unit.nanometer)
1.0
>>> x.value_in_unit(unit.meter)
1e-09
>>> x / unit.nanometer
1.0
>>> x / unit.meter
1e-09
>>> x.value_in_unit(unit.nanometer) == x / x.unit
True
```

I scanned through the source code using regex patterns like `\/ unit\/`, `unitless = `, and probably missed a few cases. A tool to automate this would find a way to detect any time a `unit.Quantity` object is divided by a `unit.Unit` object, but I'm not quite sure how to build that.

### DIFF
--- a/openff/toolkit/tests/test_forcefield.py
+++ b/openff/toolkit/tests/test_forcefield.py
@@ -3242,7 +3242,10 @@ class TestForceFieldConstraints:
                 molecule.atoms[atom2_idx].element.symbol,
             }
             assert atom_elements == bond_elements
-            assert np.isclose(distance / unit.angstrom, bond_length / unit.angstrom)
+            assert np.isclose(
+                distance.value_in_unit(unit.angstrom),
+                bond_length.value_in_unit(unit.angstrom),
+            )
 
     def test_constraints_hbonds(self):
         """Test that hydrogen bonds constraints are applied correctly to a ethane molecule."""
@@ -3821,12 +3824,13 @@ class TestForceFieldParameterAssignment:
         ref_ene = 0.0011797690240 * unit.kilojoule_per_mole
 
         assert np.allclose(
-            off_crds / unit.angstrom, ref_crds_with_vsite / unit.angstrom
+            off_crds.value_in_unit(unit.angstrom),
+            ref_crds_with_vsite.value_in_unit(unit.angstrom),
         )
         # allow 1% error in energy difference (default is .001%)
         assert np.allclose(
-            off_ene / unit.kilocalorie_per_mole,
-            ref_ene / unit.kilocalorie_per_mole,
+            off_ene.value_in_unit(unit.kilocalorie_per_mole),
+            ref_ene.value_in_unit(unit.kilocalorie_per_mole),
             rtol=0.05,
         )
 
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -257,7 +257,9 @@ class Atom(Particle):
         # TODO
         atom_dict = OrderedDict()
         atom_dict["atomic_number"] = self._atomic_number
-        atom_dict["formal_charge"] = self._formal_charge / unit.elementary_charge
+        atom_dict["formal_charge"] = self._formal_charge.value_in_unit(
+            unit.elementary_charge
+        )
         atom_dict["is_aromatic"] = self._is_aromatic
         atom_dict["stereochemistry"] = self._stereochemistry
         # TODO: Should we let atoms have names?
@@ -2449,7 +2451,7 @@ class FrozenMolecule(Serializable):
                 "conformers_unit"
             ] = "angstrom"  # Have this defined as a class variable?
             for conf in self._conformers:
-                conf_unitless = conf / unit.angstrom
+                conf_unitless = conf.value_in_unit(unit.angstrom)
                 conf_serialized, conf_shape = serialize_numpy((conf_unitless))
                 molecule_dict["conformers"].append(conf_serialized)
         if self._partial_charges is None:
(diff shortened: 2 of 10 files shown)


openff/toolkit/tests/test_forcefield.py
```diff
@@ -3242,7 +3242,10 @@ def check_molecule_constraints(cls, molecule, system, bond_elements, bond_length
                 molecule.atoms[atom2_idx].element.symbol,
             }
             assert atom_elements == bond_elements
-            assert np.isclose(distance / unit.angstrom, bond_length / unit.angstrom)
+            assert np.isclose(
+                distance.value_in_unit(unit.angstrom),
+                bond_length.value_in_unit(unit.angstrom),
+            )
```
lilyminium, line 3248: Out of curiosity, does the below work? I know pint understands units.

```python
assert np.isclose(distance, bond_length)
```

**Pint**

```python
>>> import pint
>>> unit = pint.UnitRegistry()
>>> a = np.array([1.1]) * unit.angstrom
>>> b = np.array([0.11]) * unit.nanometer
>>> np.isclose(a, b)
array([ True])
>>> a = 1.1 * unit.angstrom
>>> b = 0.11 * unit.nanometer
>>> np.isclose(a, b)
True
```

lilyminium: No to both. :(

Perhaps something to keep in mind if/when the move to pint happens.

```python
>>> from openff.toolkit.topology.molecule import unit
>>> a = np.array([1.1]) * unit.angstrom
>>> b = np.array([0.11]) * unit.nanometer
>>> np.isclose(a, b)
array([False])
>>> a = 1.1 * unit.angstrom
>>> b = 0.11 * unit.nanometer
>>> np.isclose(a, b)
  File "<__array_function__ internals>", line 5, in isclose
  File "/Users/lily/anaconda3/envs/polymetrizer/lib/python3.9/site-packages/numpy/core/numeric.py", line 2355, in isclose
    xfin = isfinite(x)
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

lilyminium:
> probably missed a few cases

Not sure if you want to update the code in `test_parameters.py` with lines like the below?
https://github.com/openforcefield/openff-toolkit/blob/d39905b488850483c77de782db643af1a7ed1fb9/openff/toolkit/tests/test_parameters.py#L1398

There are a number of constructions in topology/molecule.py that use `_unit` instead of directly dividing by a unit, e.g. 

https://github.com/openforcefield/openff-toolkit/blob/d39905b488850483c77de782db643af1a7ed1fb9/openff/toolkit/topology/molecule.py#L599-L600

openff/toolkit/tests/test_forcefield.py
mattwthompson: This confusing behavior is one reason I use `pint` wherever I have the choice.

mattwthompson:
Perfect - thanks for catching those!

j-wags (commented):
I've read over this and it looks really good. Thanks for the detailed feedback @lilyminium, and for undertaking this effort @mattwthompson -- I really appreciate both of your work. 

This is in a state where I would be comfortable approving it, but could you do the final review, @lilyminium?

openff/toolkit/topology/molecule.py
```diff
@@ -596,8 +598,7 @@ def _position(self, atom_positions):
 
         originwt, xdir, ydir = self.virtual_site.local_frame_weights
         disp = self.virtual_site.local_frame_position
-        _unit = disp.unit
-        x, y, z = disp / _unit
+        x, y, z = disp.value_in_unit(disp.unit)
```
lilyminium, line 601: I'm not sure how you feel about using private attributes, but this could be written more concisely as `x, y, z = disp._value` (analogous to `pint.Quantity.magnitude`).

Alternatively, you could try `x, y, z = disp.value_in_unit(unit.angstrom)` or `x, y, z = disp.value_in_unit(atom_positions.unit)` and see if anything breaks. If I understand correctly -- @trevorgokey please correct me -- this whole chain of `local_frame_position` etc. seems to rely on the unit implicitly being `unit.angstrom`, since all the documentation refers to angstrom. For the most part it doesn't really affect the output of the function, except for this one, where the units of the input `atom_positions` really should match the units of `self.virtual_site.local_frame_position` for the output to make sense. But that might be outside the scope of this PR.

Either way LGTM!

lilyminium (approved):

openff/toolkit/topology/molecule.py
trevorgokey: Yes,  `atom_positions` is actually positions (L668), so the docstring is outdated. It appears that, if we don't assume anything about any units, that yes we should use `disp.value_in_unit(atom_positions.unit)` since it will later be multiplied by `*axis` (L626), which takes its units from the atoms (and `origin` which also takes its units from the atoms).

Overall, I do suppose `x, y, z = disp.value_in_unit(disp.unit)` is a bug if the unit is different from the atoms. Nice catch!

trevorgokey:
```suggestion
        x, y, z = disp.value_in_unit(atom_positions.unit)
```

openff/toolkit/topology/molecule.py
```diff
@@ -626,7 +627,7 @@ def _normalize(axis):
 
         position = origin + x * xaxis + y * yaxis + z * zaxis
 
-        return unit.Quantity(position, unit=_unit)
+        return unit.Quantity(position, unit=disp.unit)
```
trevorgokey, line 630:
```suggestion
        return unit.Quantity(position, unit=atom_positions.unit)
```

mattwthompson:
CI failure is a (new! and probably stochastic!) codecov failure.

Thanks for all the feedback - I'm pretty sure this is all tidied up so I'm going to merge.

PR #1070, opened 2021-09-09
Speed up rdkit distances

lilyminium:
Fixes #1069

The first commit is regarding A, B, and C in issue #1069. I get about a 8 us speed-up in general. This corresponds to 50% the time in the test-case in `test_elf_compute_electrostatic_energy`, or 87% in a bigger molecule.

```python
def old_coords(coordinates):
    distances = np.sqrt(
            np.sum(np.square(coordinates)[:, np.newaxis, :], axis=2)
            - 2 * coordinates.dot(coordinates.T)
            + np.sum(np.square(coordinates), axis=1)
        )
    return distances

def new_coords(coordinates):
    ab = coordinates.dot(coordinates.T)
    diag = np.einsum('ii->i', ab) # faster than np.diag
    ab += (ab - diag - diag[..., np.newaxis])
    diag[:] = -0.
    matrix = np.sqrt(-ab)
    return matrix

conformer = np.array(
            [
                [1.0, 0.0, 0.0],
                [0.0, 0.0, 0.0],
                [-1.0, 0.0, 0.0],
                [0.0, 1.0, 0.0],
                [0.0, -1.0, 0.0],
            ]
        )

>>> %timeit old_coords(conformer)
18.6 µs ± 229 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

>>> %timeit new_coords(conformer)
9.7 µs ± 213 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

from openff.toolkit.topology.molecule import Molecule, unit
mol = Molecule.from_smiles("CCCCCCCCCCCCCCCCCCCCCCCCCCCC")
mol.generate_conformers()
conformer = mol.conformers[0].value_in_unit(unit.angstrom)

>>> %timeit old_coords(conformer)
52.8 µs ± 1.38 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

>>> %timeit new_coords(conformer)
45.5 µs ± 439 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

```

The second (non-CI related) commit is related to D), and just takes advantage of numpy's tuple indexing. I didn't time this to see if it's faster.

- [x] Tag issue being addressed
- [ ] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/tests)
- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable
[...]

PR #1089, opened 2021-09-27
Improve speed of RDKit greedy diverse conformer selection

lilyminium:
This does not address any bug per se, but the code is faster and (IMO) much more readable, which makes future maintenance easier. I get a fairly consistent 2x+ speedup from my function, most likely from not working with numpy masked arrays.

**Original function**
```python
def greedy_select(rms_matrix, limit=10, rms_tolerance=0.05):
    # Apply the greedy selection process.
    closed_list = np.zeros(limit).astype(int)
    closed_mask = np.zeros(rms_matrix.shape[0], dtype=bool)

    n_selected = 1

    for i in range(limit - 1):

        distances = rms_matrix[closed_list[: i + 1], :].sum(axis=0)

        # Exclude already selected conformers or conformers which are too similar
        # to those already selected.
        closed_mask[
            np.any(
                rms_matrix[closed_list[: i + 1], :]
                < rms_tolerance,
                axis=0,
            )
        ] = True

        if np.all(closed_mask):
            # Stop of there are no more distinct conformers to select from.
            break

        distant_index = np.ma.array(distances, mask=closed_mask).argmax()
        closed_list[i + 1] = distant_index

        n_selected += 1

    return [i.item() for i in closed_list[:n_selected]]
```

**Newer function**
```python
def greedy_select2(rms_matrix, limit=10, rms_tolerance=0.05):
    selected_indices = [0]
    for i in range(limit - 1):
        selected_rms = rms_matrix[selected_indices]
        any_too_close = np.any(selected_rms < rms_tolerance, axis=0)
        if np.all(any_too_close):
            break
        
        rmsdist = np.where(any_too_close, -np.inf, selected_rms.sum(axis=0))
        selected_indices.append(rmsdist.argmax())
    return selected_indices
```

**Speed comparison**

```python
import timeit

for shape in [3, 10, 100, 1000]:
    arr = np.random.rand(shape, shape)
    arr += arr.T
    
    assert np.allclose(greedy_select(arr), greedy_select2(arr))
    original = timeit.timeit(lambda: greedy_select(arr), number=1000)
    new = timeit.timeit(lambda: greedy_select2(arr), number=1000)
    speedup = original / new
    shape_text = f"Shape ({shape}, {shape})"
    print(f"{shape_text:>20s}: {original:.3f} => {new:.3f} ({speedup:.1f}x speedup)")
```

```
        Shape (3, 3): 0.372 => 0.159 (2.3x speedup)
      Shape (10, 10): 0.370 => 0.158 (2.3x speedup)
    Shape (100, 100): 0.377 => 0.162 (2.3x speedup)
  Shape (1000, 1000): 0.522 => 0.243 (2.1x speedup)

```

### DIFF
--- a/openff/toolkit/utils/rdkit_wrapper.py
+++ b/openff/toolkit/utils/rdkit_wrapper.py
@@ -1250,35 +1250,22 @@ class RDKitToolkitWrapper(base_wrapper.ToolkitWrapper):
         rms_matrix = cls._elf_compute_rms_matrix(molecule)
 
         # Apply the greedy selection process.
-        closed_list = np.zeros(limit).astype(int)
-        closed_mask = np.zeros(rms_matrix.shape[0], dtype=bool)
-
-        n_selected = 1
-
-        for i in range(min(molecule.n_conformers, limit - 1)):
-
-            distances = rms_matrix[closed_list[: i + 1], :].sum(axis=0)
-
-            # Exclude already selected conformers or conformers which are too similar
-            # to those already selected.
-            closed_mask[
-                np.any(
-                    rms_matrix[closed_list[: i + 1], :]
-                    < rms_tolerance.value_in_unit(unit.angstrom),
-                    axis=0,
-                )
-            ] = True
-
-            if np.all(closed_mask):
-                # Stop of there are no more distinct conformers to select from.
+        selected_indices = [0]
+        angstrom_tol = rms_tolerance.value_in_unit(unit.angstrom)
+        for i in range(min(limit, molecule.n_conformers) - 1):
+            selected_rms = rms_matrix[selected_indices]
+            any_too_close = np.any(selected_rms < angstrom_tol, axis=0)
+            if np.all(any_too_close):
+                # stop if all conformers remaining are within RMS
+                # threshold of any selected conformer
                 break
 
-            distant_index = np.ma.array(distances, mask=closed_mask).argmax()
-            closed_list[i + 1] = distant_index
+            # add the next conformer with the largest summed RMS distance
+            # to current selected conformers
+            rmsdist = np.where(any_too_close, -np.inf, selected_rms.sum(axis=0))
+            selected_indices.append(rmsdist.argmax())
 
-            n_selected += 1
-
-        return [ranked_conformers[i.item()] for i in closed_list[:n_selected]]
+        return [ranked_conformers[i] for i in selected_indices]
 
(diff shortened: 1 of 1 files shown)


SimonBoothroyd:
> (IMO) much more readable

We can maybe agree to disagree here 😉 

> I get a fairly consistent 2x+ speedup from my function

Looking at the timing it seems you typically save about 0.2ms using the code changes assuming I understand the output of `timeit`, which isn't too much.

In general I'd advise caution in investing too much time in micro-optimisations to these functions (same with #1070 where the saving was only microsecond) - bear in mind these are typically coupled with function calls that take on the order of seconds to minutes and so their latency is almost always hidden, and the speed ups your getting would typically require you to run this code thousands of times before you'd notice the speed-up in a real sense.

lilyminium:
> Looking at the timing it seems you typically save about 0.2ms using the code changes assuming I understand the output of timeit, which isn't too much.

It does decrease but I do think the speed savings will be some factor of total time rather than an absolute difference, because the masked array creation / usage is looped over (masked arrays are notoriously slow). (below timings are per function call)
```
        Shape (3, 3): 0.312 => 0.152 msec (2.1x speedup over 1000 iterations)
      Shape (10, 10): 0.311 => 0.140 msec (2.2x speedup over 1000 iterations)
    Shape (100, 100): 0.324 => 0.146 msec (2.2x speedup over 1000 iterations)
  Shape (1000, 1000): 0.430 => 0.204 msec (2.1x speedup over 1000 iterations)
Shape (10000, 10000): 1.024 => 0.672 msec (1.5x speedup over 500 iterations)
Shape (20000, 20000): 2.115 => 1.335 msec (1.6x speedup over 100 iterations)
```

> In general I'd advise caution in investing too much time in micro-optimisations to these functions (same with #1070 where the saving was only microsecond) - bear in mind these are typically coupled with function calls that take on the order of seconds to minutes and so their latency is almost always hidden, and the speed ups your getting would typically require you to run this code thousands of times before you'd notice the speed-up in a real sense.

Normally I'd agree but these PRs are all code I modified for something else that I thought might be beneficial in the toolkit :-)

mattwthompson:
It seems this has gone a bit stale, but these changes are probably good and they're definitely small in scope. Could this be rebased onto `main` (previously `topology-biopolymer-refactor`)?

j-wags (approved):
This has been open for long enough that we should just give it a go/no-go. I think this is a go :-)

PR #1091, opened 2021-09-27
Index Conformers in RDKit instead of using Ids

### LINKED ISSUE #1090: Mixup with conformer indices and ids in RDKitToolkitWrapper._elf_compute_rms_matrix
**Describe the bug**

I think there was a bit of confusion as to whether conformer IDs were IDs or indices; they are IDs, i.e. labels. For the vast majority of the time, the IDs do correspond to indices; however, if the toolkit changes its implementation of `to_rdkit()` to not do this, the below code will abruptly stop working as intended.

https://github.com/openforcefield/openff-toolkit/blob/d39905b488850483c77de782db643af1a7ed1fb9/openff/toolkit/utils/rdkit_wrapper.py#L1178-L1191

**To Reproduce**

**Output**

 - Operating system

**Additional context**

lilyminium:
Fixes #1090


mattwthompson (changes requested):
Could you write a unit test that will fail before this patch, but not after? If I understand the problem, I think that the simplest one would be a molecule with one conformer labelled with an id > 0, probably resulting in a `KeyError`?

lilyminium:
> Could you write a unit test that will fail before this patch, but not after? If I understand the problem, I think that the simplest one would be a molecule with one conformer labelled with an id > 0, probably resulting in a KeyError?

That would be hard, because it doesn't currently fail -- `to_rdkit()` is called with `assignId=True` hardcoded in, which results in index-like IDs. I put in this change because I worried that someone making changes to `to_rdkit` in the future (e.g. making that a user-chosen kwarg) wouldn't realize possible downstream impacts like this one. Instead of making this change I could add a comment to that line in `to_rdkit`?

https://github.com/openforcefield/openff-toolkit/blob/d39905b488850483c77de782db643af1a7ed1fb9/openff/toolkit/utils/rdkit_wrapper.py#L1802

mattwthompson:
Oops, I didn't see that `_elf_compute_rms_matrix ` took in an OFFmol; I was thinking this test would modify an rdmol to have weird conformer IDs and then directly pass it to that function, but now I see that's not possible. I think documenting the behavior of `.to_rdkit()` here should be sufficient - probably the docstring is the right place to do this.

I'd think that eventually the various gotchas in the toolkit wrappers should probably have a different home, but that's out of scope here.

lilyminium:
> I think documenting the behavior of .to_rdkit() here should be sufficient - probably the docstring is the right place to do this.

Do you mean reverting this change and documenting the conformer ID-setting behaviour in the docstring of `to_rdkit()`? And/or adding some documentation to this function?

mattwthompson:
No, I was thinking that documenting the behavior of how`.to_rdkit()` assigns conformer IDs was the only thing to do in lieu of a unit test. But now I realize I was wrong, since the behavior actually isn't very weird. I mean, there would be no harm in documenting it, but not really any point since it's not a gotcha or something that a user would be surprised by.

Anyway, I think this good and stands alone. I just confused myself along the way.

mattwthompson (approved):
I don't think release notes need to be updated, since this is not a behavior or API change.

PR #1096, opened 2021-09-27
Make atom names more unique

### LINKED ISSUE #1095: Make atom names less likely to collide with conventional atom type names
The atom names generated by default in `Molecule. generate_unique_atom_names` can clash with atom names generated by downstream libraries (i.e. ParmEd loading an OpenMM system). This is because the basic algorithm of "element plus an incrementing index" is used by both; a hack to make ours more obscure would likely prevent some of these clashes.

```python3
>>> from openff.toolkit.topology import Molecule
>>> mol = Molecule.from_smiles("CCO")
>>> mol.generate_unique_atom_names()
>>> [a.name for a in mol.atoms]
['C1', 'C2', 'O1', 'H1', 'H2', 'H3', 'H4', 'H5', 'H6']
```
Lots of other things might have similar "atom names," even though we don't have atom types.

I double-checked that these would survive the conversion to OpenMM:
```python3
>>> for a in mol.atoms:
...     a.name = a.name + 'x'
...
>>> [a.name for a in mol.to_topology().to_openmm().atoms()]
['C1x', 'C2x', 'O1x', 'H1x', 'H2x', 'H3x', 'H4x', 'H5x', 'H6x']
```

**Additional context**

This would not be an API break, although it would be a behavior change. It could be triggered by an optional flag `Molecule. generate_unique_atom_names(make_them_weird=False)` but I'm not a fan of that idea.

mattwthompson:
May fix #1095

I'm not sure if something more than appending `'x'` is appropriate; something even longer would reduce the changes of collisions to zero but might cause other issues, like Amber files not being able to write them out at all due to the limitations of how their FORTRAn parsers are designed.


j-wags (approved):
Thanks @mattwthompson! This is a definite improvement and removes some mines from the system-conversion minefield. In the future, we may also consider letting users customize the suffix, so that multiple OpenFF ligands don't have atom type collisions. But this change is straightforward and should solve most of the cases where the old problem occured. 

Thanks again, and please update the releasenotes before merging!

PR #1098, opened 2021-09-29
Raise exception when dealing with radicals

### LINKED ISSUE #1075: Should the Molecule class support radicals?
**Describe the bug**
`Molecule.to_rdkit()` can produce a `Molecule` with a different number of atoms than the source RDKit molecule.

This appears to be because the `hydrogens_are_explicit` flag is set to `False` by default in [this line](https://github.com/openforcefield/openff-toolkit/blob/b3bcc885ba318f39d771aa7f58504ece3ab4fe31/openff/toolkit/topology/molecule.py#L5076), which means the molecule can be modified when converted from RDKit to Molecule. This is incredibly dangerous, since the point of being able to move between these representations is to guarantee we are always working with a one-to-one correspondence between atoms in all representations. This makes it easy to completely break this assumption. Worse, there is no way to override this, making the toolkit essentially useless for tasks that involve preserving atom indices and ordering---e.g. force field parameterization.

**To Reproduce**
We're working on a simple reproduction case [here](https://github.com/openmm/qmdataset/pull/5#issuecomment-917666335) with @peastman

**Output**
This is a related error with `generate_conformers()` for this `Molecule`.
```
  File "createDipeptides.py", line 175, in <module>
    createConformations(outputfile, forcefield, topology, positions, f'{name1}-{name2}', charges)
  File "createDipeptides.py", line 86, in createConformations
    mol.generate_conformers(n_conformers=10, rms_cutoff=0*unit.nanometers)
  File "/home/peastman/workspace/openff-toolkit/openff/toolkit/topology/molecule.py", line 3221, in generate_conformers
    return toolkit_registry.call(
  File "/home/peastman/workspace/openff-toolkit/openff/toolkit/utils/toolkit_registry.py", line 366, in call
    raise e
  File "/home/peastman/workspace/openff-toolkit/openff/toolkit/utils/toolkit_registry.py", line 362, in call
    return method(*args, **kwargs)
  File "/home/peastman/workspace/openff-toolkit/openff/toolkit/utils/rdkit_wrapper.py", line 899, in generate_conformers
    molecule._add_conformer(conformer)
  File "/home/peastman/workspace/openff-toolkit/openff/toolkit/topology/molecule.py", line 3997, in _add_conformer
    raise Exception(
Exception: molecule.add_conformer given input of the wrong shape: Given (42, 3), expected (40, 3)
```

(will have to fill in later)

**Additional context**
cc https://github.com/openmm/qmdataset/pull/5#issuecomment-917666335

mattwthompson:
Resolves #1075

I could not figure out how to detect radicals with `oechem`. Maybe I'm too novice, but [searching the docs](https://docs.eyesopen.com/toolkits/python/search.html?q=radical&check_keywords=yes&area=default) or skimming through the [API docs](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEAtomBase.html) did not get me anywhere, nor did inspecting an atom object in memory. I would have expected either to lead me somewhere, so either it's not offered or I'm missing something under my nose.

I have not thought about this affects parsing other files, but my guess is that enough paths run through `from_rdkit`/`from_openeye` that those serve as chokepoints for catching this.
 
This implementation does add a method call for each atom in `from_rdkit`; it's quick (< 1 us / atom) enough that I don't know if there'd be a better API point to use.

```
In [1]: from rdkit import Chem

In [2]: rdmol = Chem.MolFromSmiles(100 * 'CC')

In [3]: %timeit [atom for atom in rdmol.GetAtoms()]
141 µs ± 2.28 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [4]: %timeit [atom.GetNumRadicalElectrons() for atom in rdmol.GetAtoms()]
183 µs ± 2.11 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

- [ ] Implement for `OpenEyeToolkitWrapper`

### DIFF
--- a/openff/toolkit/tests/test_toolkits.py
+++ b/openff/toolkit/tests/test_toolkits.py
@@ -46,6 +46,7 @@ from openff.toolkit.utils.exceptions import (
     IncorrectNumConformersWarning,
     InvalidIUPACNameError,
     InvalidToolkitError,
+    RadicalsNotSupportedError,
     ToolkitUnavailableException,
     UndefinedStereochemistryError,
 )
@@ -1842,6 +1843,11 @@ class TestRDKitToolkitWrapper:
         smiles2 = molecule.to_smiles(toolkit_registry=toolkit_wrapper)
         assert smiles2 == expected_output_smiles
 
+    def test_rdkit_from_smiles_radical(self):
+        """Test that parsing an SMILES with a radical raises RadicalsNotSupportedError."""
+        with pytest.raises(RadicalsNotSupportedError):
+            RDKitToolkitWrapper().from_smiles("[CH3]")
+
     def test_rdkit_from_smiles_hydrogens_are_explicit(self):
         """
         Test to ensure that RDKitToolkitWrapper.from_smiles has the proper behavior with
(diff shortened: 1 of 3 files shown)


mattwthompson:
OpenEye got back to me (it took 6 months because of me; they responded in a business day and a half) and suggested I get the MDL valence and compare it to the valence that's found on the atom. I've incorporated those changes into a new PR targeting the active branch: #1236

PR #1111, opened 2021-10-25
Add automated tests against OpenMM/OpenEye RCs

mattwthompson:
This is copied/adapted/taken from https://github.com/openmm/openmm/issues/3232#issuecomment-909478595 

There is too much copy-pasted code here, but I couldn't immediately think of a solution that didn't involve making ~5 new conda YAMLs.

I don't think there's really a way to cleanly test this, but we could modify one of them on a trial basis to also check betas; OpenEye tends to release several betas per release so one should trigger before an RC.


ijpulidos:
I cannot seem to make sense of the RC label in the `OpenEye` conda channel, last main release is 2021.1.1 but I don't see rc for it in https://anaconda.org/OpenEye/OpenEye-toolkits/files?channel=rc, just for 2021.1.0. Is there something I'm missing here?

Even though, it seems that the latest rc is indeed related to the latest main, just have a minor/build version difference. Is that it?

mattwthompson:
My guess is that they don't publish RCs for the the releases that increment only the last digit. I don't know the versioning scheme they use, it looks like a `YYYY.MINOR.MICRO` style of [Calver](https://calver.org/), implying that bumping the last digit is bugfix-only and not worth building RCs for. Other projects I follow only build RCs for major versions, i.e. there are are RCs for Python 3.10.0 but none for 3.10.1 https://www.python.org/dev/peps/pep-0619/#release-schedule

As a separate matter, these scripts seem to be flexible enough that if we know how each project communicated the state of test-only builds, we can set different labels accordingly.

jchodera:
Suggest you [contact OpenEye support](https://www.eyesopen.com/support) if there are any questions. They usually respond within a few hours!

j-wags:
This is a great PR, and it should keep us from getting blindsided in the future. Thanks for opening it

I'd like to propose removing the checks for new release candidate packages, and instead unconditionally testing against RCs every 7 days. There are three reasons for this:
* Other package developers are free to go nuts before big releases. OE had like 7 release candidate packages in rapid succession for a recent release. That would be lots of static for us.
* The "is there a new RC?" action increases the complexity of the codebase, and uses actions from two small channels that could be hijacked and become security vulnerabilities.
* In essence, we want tests to alert us if 
    1) OFFTK stable and OETK stable don't work together ("high alert!", our daily "conda" action
    2) OFFTK dev and OETK stable don't work together (the current "CI" action, triggered daily and in PRs)
    3) OFFTK dev and OETK dev don't work together ("this will be a problem in a few weeks", this PR)
* We already have tests for `i` and `ii`, and this adds tests for `iii`. But this doesn't QUITE fit what we want for `iii`. Because what we want is to run tests against OETK dev whenever there's a change to OETK dev OR OFFTK dev. But it'd be even MORE complex to run this action on changes to either, and it would substantially increase our total CI time requirements. 

So, for the reasons above, I'd say that we should remove the scheduling scripts and have this be a weekly CI run.

devtools/conda-envs/beta_rc_env.yaml
```diff
@@ -0,0 +1,47 @@
+name: test
+channels:
+  - conda-forge/label/openmm_rc
+  - openeye/label/beta
+  - openeye/label/rc
```
mattwthompson, line 5: I'm not confident in this; if I have them in the other order, an old RC will be picked up instead of newer betas. This order ensures betas are picked up when they're newer, but I'm not sure what happens when an RC is newer than a beta.

mattwthompson:
Looks like this actually found a behavior change! I was able to reproduce the following locally:

```
__________ TestForceFieldParameterAssignment.test_tip5p_dimer_energy ___________
openff/toolkit/tests/test_forcefield.py:3804: in test_tip5p_dimer_energy
    [molecule1, molecule2], off_ff, minimize=False
openff/toolkit/tests/utils.py:1792: in evaluate_molecules_off
    top.to_openmm(), sys, atom_xyz, minimize=minimize
openff/toolkit/tests/utils.py:1628: in openmm_evaluate_vsites_and_energy
    sim = openmm.app.Simulation(omm_top, omm_sys, integ)
/usr/share/miniconda3/envs/test/lib/python3.7/site-packages/openmm/app/simulation.py:99: in __init__
    self.context = mm.Context(self.system, self.integrator)
/usr/share/miniconda3/envs/test/lib/python3.7/site-packages/openmm/openmm.py:6133: in __init__
    _openmm.Context_swiginit(self, _openmm.new_Context(*args))
E   openmm.OpenMMException: NonbondedForce: sigma for an exception must be positive
=============================== warnings summary ===============================
<<<SNIP>>>
=========================== short test summary info ============================
FAILED openff/toolkit/tests/test_forcefield.py::TestForceFieldParameterAssignment::test_tip5p_dimer_energy
= 1 failed, 26058 passed, 24 skipped, 82 xfailed, 314 xpassed, 4572 warnings in 9288.25s (2:34:48) =
```

The exception is thrown [here](https://github.com/openmm/openmm/blob/7.7.0beta/openmmapi/src/NonbondedForceImpl.cpp#L97), and the [change](https://github.com/openmm/openmm/pull/3286/files#diff-9000df78aa9160328f25b9a5867cd31ffc8bf8a8f07fdb7f699e05dc26fb40d0R97) was introduced in a [PR](https://github.com/openmm/openmm/pull/3286) that in part stemmed from [OpenFF's report](https://github.com/openmm/openmm/issues/3185) that OpenMM would casually accept many non-physical parameter values.

I think this will require a little more digging to fix (I'm not sure exceptions with zero sigma are that offensive, figuring the chances that they have zero epsilon are high). I'll note here that these tests were able to detect an issue in a way that should not require human effort if it happens again - and also that the issue doesn't seem to be a critical bug.

mattwthompson:
As one might predict, the sigmas are only zero when epsilons are also zero. Putting a debug trace in L1628 of `openff/toolkit/tests/utils.py` was fortunately quick and to the point:

```
ipdb> f = omm_sys.getForces()[1]
ipdb> for i in range(f.getNumExceptions()):
    params = f.getExceptionParameters(i)
    print(params[3], params[4])



0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
0.0 nm 0.0 kJ/mol
1.0 nm 0.0 kJ/mol
1.0 nm 0.0 kJ/mol
1.0 nm 0.0 kJ/mol
1.0 nm 0.0 kJ/mol
1.0 nm 0.0 kJ/mol
1.0 nm 0.0 kJ/mol
```

mattwthompson:
I think this good to go, or at least I think this does what it's meant to do in its scope; the remaining issues I see as non-blockers

1. I've raised an issue with OpenMM about the above: https://github.com/openmm/openmm/issues/3325
1. I have an email out to OpenEye support about whether or not there is a schedule for betas and RCs.

j-wags:
(non blocking) There are a few layers to the zero-sigma exclusion thing. 
1) Why is zero sigma a problem at all? (you covered this by opening the openmm issue, but maybe @jchodera or @davidlmobley can shortcut the answer. I recall there was some other reason that existing ffs always assigned nonzero sigmas to hydroxyl Hs, even when epsilons were zero, but I don't remember the specifics. Maybe something with free energy calcs?)
2) Why is a zero sigma appearing in tip5p water test and nowhere else? The [one place I can find a zero sigma in our test force fields](https://github.com/openforcefield/openff-toolkit/blob/2ff02fd1eaee019c33f76f8b670573876ce925c1/openff/toolkit/tests/test_forcefield.py#L676) ISN'T used in the failing tip5p test, and the [tip5p offxml string](https://github.com/openforcefield/openff-toolkit/blob/2ff02fd1eaee019c33f76f8b670573876ce925c1/openff/toolkit/tests/test_forcefield.py#L362-L380) has all nonzero sigmas.

devtools/conda-envs/beta_rc_env.yaml
j-wags: (non blocking) I saw you CC'ed me on an email to them. Let's stick with this ordering -- Maybe their response will let us reduce down to just one channel/label.

j-wags (approved):
Awesome. Thanks, @mattwthompson!

jchodera:
> Why is zero sigma a problem at all? 

In Lennard Jones energy expressions, you have ratios `(r/sigma)` everywhere. Unless great care is taken to avoid numerical instabilities or errors, `sigma=0` is not allowed in the domain for `sigma` because it will lead to NaNdyland.

j-wags:
Ha! Gotcha. Thanks for the quick reply!

mattwthompson:
> Why is a zero sigma appearing in tip5p water test and nowhere else? 

This confused me as well for the same reasons. No clue.

mattwthompson:
Hmm .... the checks that ran on 05ea2be4596e102d4a7780ce5fcbd7cdb66978ad pulled down OpenMM 7.6 instead of the 7.7 beta.

davidlmobley:
I was gonna answer, but John beat me to it.

I think some codes "notice" when epsilon is set to zero and explicitly zero out those interactions so this is a non-issue, whereas others very much need both to be set to zero for the reasons he noted.

mattwthompson:
Some loose ends, listed below, but this will never be finished if perfection is the standard. It's been approved, already added value, and I'm happy with it, so I'm merging when the CI & Example runs pass. The new workflows should NOT run during normal development PRs after this.

* Unclear exactly how OpenEye works through their beta process (ticket to support is open)
* Many runs picked OpenMM 7.6 over the 7.7 beta. I don't know why; my understanding of the solver led me to believe that shouldn't happen (newer versions take priority, even if beta) so I set channel priority to strict. This may be a problem in the case of a beta/RC being older than the most recent full release (i.e. the final 7.7 RC will be pulled over the full 7.7 release), but shouldn't be a HUGE problem in such a case because an "old" RC should be equivalent to a newer release.
* Confusing why the TIP5P test was failing; its XML content did not have zero sigma and some other snippet(s) did. Worth revisiting in a presumed next 7.7 beta that likely includes the patch https://github.com/openmm/openmm/pull/3326

jchodera:
> Unclear exactly how OpenEye works through their beta process (ticket to support is open)

Here's what they told us (edited slightly):

> Yes, generally speaking, when we are ready for the first release candidate (RC1), it is a mirror of the last beta version that was built. However, we cannot always guarantee that because on occasion, some very minor and well tested changes make their way into RC1. As far as the number of beta versions that we build prior to an RC1, this is always a moving target and something we can never really predict. In regards to nightly builds, we do not have snapshots that are available anywhere. The weekly beta builds are the closest thing we can offer in terms of pre-production development builds.

jchodera:
> Confusing why the TIP5P test was failing; its XML content did not have zero sigma and some other snippet(s) did. Worth revisiting in a presumed next 7.7 beta that likely includes the patch Removed more unnecessarily restrictive parameter checks openmm/openmm#3326

There is generally only one OpenMM beta for each release. If another one is critical, this would be important to request on the OpenMM issue tracker.

mattwthompson:
Good to know. Probably the next RC will be fine then, since it should include the changes since the beta. If that gives us some trouble I will request another one.

jchodera:
The OpenMM RC will indeed be identical to the released version of 7.7 (assuming no showstopper bugs!).

mattwthompson:
OpenEye Support got back to me. They were unable to provide any details about their release process, but they suggested we do not test against their betas - just the RCs - which seems fair. No reason to let their development builds cause broken builds here without an agreement (explicit or implicit) that those development builds are intended to be tested against.

PR #1115, opened 2021-10-28
Add 404 page to docs

Yoshanuikabundi:
While sorting out some housekeeping issues, I realised it might be possible to provide our own 404 page, which would be much better than using the RTD 404 page that doesn't even link to our docs. This PR is testing that.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -93,6 +93,24 @@ jupyter_execute_notebooks = "force"
 # List of notebooks NOT to execute (use output stored in notebook instead)
 execution_excludepatterns = []
 
+
+# sphinx-notfound-page
+# https://github.com/readthedocs/sphinx-notfound-page
+# Renders a 404 page with absolute links
+import importlib
+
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -93,6 +93,24 @@
...
+import importlib
+
+if importlib.util.find_spec("notfound"):
+    extensions.append("notfound.extension")
+
+    notfound_context = {
+        "title": "404: File Not Found",
+        "body": """
+    <h1>404: File Not Found</h1>
+    <p>Sorry, we couldn't find that page.</p>
```
j-wags, line 109: (not blocking) Could we add something like "This usually happens as a result of following an outdated link. Please ensure you're looking at the [most recent](https://open-forcefield-toolkit.readthedocs.io/en/) documentation unless you are deliberately trying to use an older version."

j-wags (approved):
Nice. One non-blocking suggestion. Thanks for thinking of this @Yoshanuikabundi!

PR #1120, opened 2021-11-02
Make FrozenMolecule.to_hill_formula a class method

### LINKED ISSUE #1110: Molecule.to_hill_formula
`Molecule.to_hill_formula` is actually a staticmethod, taking a `Molecule`, whereas `Molecule.hill_formula` is the property which actually gives a Hill formula.  This is sort of ugly compared to all the other `to_X` methods.  If you're stlil at the point where you can break things, this would be a good option...

mattwthompson:
Resolves #1110 by delegating most of the logic to a private function. If people *really* want a Hill formula from a NetworkX graph, there is another private method `from openff.toolkit.topology.molecule import _networkx_graph_to_hill_formula`); it could be made public but at the moment that path does not appear to be commonly used.


openff/toolkit/topology/molecule.py
```diff
@@ -3028,11 +3031,6 @@ def are_isomorphic(
             [Dict[int,int]] ordered by mol1 indexing {mol1_index: mol2_index}
             If molecules are not isomorphic given input arguments, will return None instead of dict.
         """
-
-        # Do a quick hill formula check first
-        if Molecule.to_hill_formula(mol1) != Molecule.to_hill_formula(mol2):
-            return False, None
-
```
mattwthompson, line 3035: This might introduce a performance regression. I do not have a strong intuition as to how often this short-circuit was triggered; I'd think rarely but I don't know.

openff/toolkit/topology/molecule.py
```diff
@@ -6470,3 +6410,57 @@ def _ipython_display_(self):
             return display(self.visualize(backend="openeye"))
         except ValueError:
             pass
+
+
+@requires_package("networkx")
```
j-wags, line 6415: (discussed in our meeting today) NetworkX is a hard dep of the toolkit, so we don't need to check for it here.

openff/toolkit/topology/molecule.py
j-wags: (discussed in our meeting today) Hm, this is a pretty important performance speedup (I think this is used for input validation/conformer handling in `openff-benchmark`), so I'd like to preserve this (either by having a check here for whether the input is a networkx graph or a molecule, or by making something like a private method for `Molecule._object_to_hill_formula(obj)`)

openff/toolkit/topology/molecule.py
mattwthompson: For context: it's lazy-loaded elsewhere in the module, which I confused with being optional

mattwthompson:
I've reworked everything to be backwards-compatible via private method(s), particularly `FrozenMolecule._object_to_hill_formula` which tries to figure out the type of the object and delegate out accordingly to either `_topologymolecule_to_hill_formula` or `_networkx_graph_to_hill_formula`. I'm opposed to making this a part of the public API since I want to move away from the squishiness of duck-typing inputs. The proposed changes clarify the API  and should not fundamentally change the existing behavior that's used by the benchmarking infrastructure.


Separately, it can still behave like a static method (despite not being one) if the object it's passed it sufficiently like a `FrozenMolecule`:

```
In [1]: from openff.toolkit.topology import *

In [2]: molecule = Molecule.from_smiles("CCO")

In [3]: topology = Topology.from_molecules(molecule)
   ...: topmol = TopologyMolecule(molecule, topology)

In [4]: Molecule.to_hill_formula(topmol)
Out[4]: 'C2H6O'

In [5]: Molecule.to_hill_formula(molecule)
Out[5]: 'C2H6O'
```

which feels weird but also how instance method seem to work in general:

```
In [6]: Molecule.to_smiles(molecule)
Out[6]: '[H]C([H])([H])C([H])([H])O[H]'

In [7]: Molecule.to_bson(molecule) == molecule.to_bson()
Out[7]: True
```

Okay, that's enough of me stumbling across things I maybe should have known for years.

j-wags (approved):
Looks great! Thanks, @mattwthompson.

PR #1125, opened 2021-11-10
Streamline CI github actions

lilyminium:
Github Actions can work off a job matrix, which might be more streamlined and easier to modify in the future than a list of jobs. The matrix can `include` or `exclude` jobs -- I've chosen to exclude the currently excluded configurations, as it seems neater to me to start off with the ideal matrix and specify which sections you don't want. For example, if trying to diagnose weird behaviour across a matrix of configurations, it's easier to comment the excluded portions than to add new `cfg` jobs.

This should hopefully change nothing about the current CI, although I'm not quite sure if true/false are boolean (yaml) or strings (bash) and how the comparisons will work out.

Alternate ways to write this include:

* having a cfg that includes all the environment variables, e.g. below. This is more similar to the current approach.
```
      matrix:
        os: [ubuntu-latest, macos-latest]
        cfg:
          - rdkit: true
            openeye: true
            envfile: test_env
            jobname: RDKit and OpenEye
            toolkit_checks: RDKIT OPENEYE
            packages_to_remove: ''
            python-version: 3.7
```


.github/workflows/CI.yml
```diff
@@ -111,58 +99,41 @@ jobs:
...
+              if [[ $(conda list | grep $cpkg) ]]; then conda remove --force $cpkg --yes ; fi
+            done
           fi
 
+      - name: Check installed toolkits
+        run: |
+          for tk in ${{ env.TOOLKIT_CHECKS }}; do
+            python -c "from openff.toolkit.utils.toolkits import ${tk}_AVAILABLE; assert ${tk}_AVAILABLE, '${tk} unavailable'"
+          done
+
```
mattwthompson, line 120: If I'm reading this right, it checks that the desired toolkits are installed (behavior that seems to be working) but does not explicitly check that that undesired toolkits are **not** installed? This may seem like an obtuse use case but for testing some OpenEye behavior, we want to make sure `OpenEyeToolkitWrapper` is being used.

.github/workflows/CI.yml
```diff
@@ -111,58 +99,41 @@ jobs:
...
-        if: ${{ matrix.cfg.rdkit == 'true' && matrix.cfg.openeye == 'TRUE' }}
-        run: |
-          pytest -r fE --tb=short openff/toolkit/tests/test_links.py
+        if: ${{ matrix.rdkit == true && matrix.openeye == true }}
+        run: pytest -r fE --tb=short openff/toolkit/tests/test_links.py
 
       - name: Run mypy
-        shell: bash -l {0}
         # Try to prevent failure from preventing later steps, but also display red X
         # https://stackoverflow.com/a/62112985/4248961
```
mattwthompson, line 132: This comment (which I think I wrote) seems to be wrong with the current code. Might as well remove it now.

mattwthompson (changes requested):
These changes look good to me, save one blocking discrepancy I ran into. I don't have a strong opinion on the "build a matrix and exclude some elements" idea, but it does seem to maintain the behavior and not limit flexibility when we want to change things up (it won't be long until I advocate for dropping Python 3.7, for example). Removing the `        shell: bash -l {0}` copy-paste and condensing the `setup-miniconda` down to a single call are wins IMO.

Blocking:
* I found RDKit installed in "Test on ubuntu-latest, Python 3.7, RDKit=false, OpenEye=true", all other rows in the matrix looked fine w/r/t installed toolkits.

.github/workflows/CI.yml
lilyminium: Is it still desired? Could you use `if: always()` to always run something, or `continue-on-error`?

.github/workflows/CI.yml
lilyminium: Added!

.github/workflows/CI.yml
mattwthompson: The thing I worry about, and the reason why it's here at all, is when a dumb little thing breaks `mypy` and therefore prevents unit tests from running, losing the ~5 mins of previous investment in the run getting conda and everything else set up. Probably the best way to do this, though, is just move this step to the very end?

mattwthompson (approved):
LGTM - I checked the log just to be sure that RDKit is *not* installed in the run it's not meant to be installed in. Thanks!

j-wags:
Awesome. Thanks @lilyminium and @mattwthompson!!

PR #1139, opened 2021-11-30
Update FAQ.md

j-wags:
- [x] Update the FAQ based on common user questions


FAQ.md
```diff
@@ -31,22 +30,34 @@ This is especially problematic for PDB files from X-ray crystallography which ty
...
+## What about starting from an XYZ file?
+
+XYZ files generally only contain elements and positions, and are therefore similar in content to PDB files. See the above section "What about starting from a PDB file?" for more information.
+
 ## What do you recommend as a starting point?
 
 For application of SMIRNOFF force fields, we recommend that you begin your work with formats which provide the chemical identity of your small molecule (including formal charge and bond order).
 This means we recommend one of the following or equivalent:
-- A `.mol2` file or files for the molecules comprising your system, with correct bond orders and formal charges. (Note: Do NOT generate this from a simulation package or tool which does not have access to bond order information; you may end up with a `.mol2` file, but the bond orders will be incorrect)
+- A `.mol`, `.sdf`, or `.mol2` file or files for the molecules comprising your system, with correct bond orders and formal charges. (Note: Do NOT generate this from a simulation package or tool which does not have access to bond order information; you may end up with a correct-seeming file, but the bond orders will be incorrect)
```
mattwthompson, line 45: This was a particularly glaring omission that I forgot to mention earlier - given that we seem to prefer SDF to MOL2!

FAQ.md
```diff
@@ -31,22 +30,34 @@ This is especially problematic for PDB files from X-ray crystallography which ty
...
 - Chemical Identity Registry numbers for the components of your system
 - IUPAC names for the components of your system
 
 Essentially, anything which provides the full identity of what you want to simulate (including stereochemistry) should work, though it may require more or less work to get it into an acceptable format.
 
+## I understand the risks and want to perform bond and formal charge inference anyway
+
+If you are unable to provide a molecule in the formats recommended above and want to attempt to infer the bond orders and atomic formal charges, there are tools available elsewhere that can provide guesses for this problem. These tools are not perfect, and the inference problem itself is poorly defined, so you should review each output closely. Some tools we know of include:
+
+- the OpenEye toolkits' `OEPerceiveBondOrders` functionality
```
mattwthompson, line 57:
```suggestion
- the OpenEye toolkits' [`OEPerceiveBondOrders`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEPerceiveBondOrders.html) functionality
```

FAQ.md
```diff
@@ -31,22 +30,34 @@ This is especially problematic for PDB files from X-ray crystallography which ty
...
+- [MDAnalysis' RDKit converter](https://docs.mdanalysis.org/stable/documentation_pages/converters/RDKit.html?highlight=rdkit#module-MDAnalysis.converters.RDKit)
```
mattwthompson, line 58: Just linking this leaves out a good bit of information

* That MDAnalysis is capable of gobbling up a ton of different formats
* That the inference happens during the MDAnalysis -> RDKit conversion

such that I'm not sure I'd be able to run with this if I didn't have Lily's code from earlier in my head.

In this case, however, I'm fine with leaving users just a few breadcrumbs instead of hand-holding through our API.

FAQ.md
```diff
@@ -31,22 +30,34 @@ This is especially problematic for PDB files from X-ray crystallography which ty
...
 - Isomeric SMILES strings for the components of your system
 - InCHI strings for the components of your system
 - Chemical Identity Registry numbers for the components of your system
 - IUPAC names for the components of your system
 
 Essentially, anything which provides the full identity of what you want to simulate (including stereochemistry) should work, though it may require more or less work to get it into an acceptable format.
 
+## I understand the risks and want to perform bond and formal charge inference anyway
+
+If you are unable to provide a molecule in the formats recommended above and want to attempt to infer the bond orders and atomic formal charges, there are tools available elsewhere that can provide guesses for this problem. These tools are not perfect, and the inference problem itself is poorly defined, so you should review each output closely. Some tools we know of include:
```
mattwthompson, line 55: What about stereochemistry? I often include stereochemistry when I describe the difference between representations. But I'm not sure if any of these tools do that?

mattwthompson (approved):
I'm emphatically in support of rapidly updating the FAQs as we run into common problems.

A couple comments but nothing blocking IMO.

FAQ.md
j-wags: Yeah, I haven't looked at the FAQ in a while it seems!

FAQ.md
j-wags: I left it out here for the sake of conciseness, but what I'd say with more words is "in an XYZ file you're totally missing the bond orders and formal charges - they're neither directly readable nor inferrable from the file. But you DO have stereo in there, since the coordinates have all the info needed to unambiguously assign it once the connectivity is perceived". So the user will still probably want to run `OE/RDK.PerceiveStereoFrom3D` to avoid stereo warnings, but that's all separate from the "scary guessing" topic for this point.

FAQ.md
j-wags:
```suggestion
- [MDAnalysis' RDKit converter](https://docs.mdanalysis.org/stable/documentation_pages/converters/RDKit.html?highlight=rdkit#module-MDAnalysis.converters.RDKit) , with an [example here](https://github.com/openforcefield/openff-toolkit/issues/1126#issuecomment-969712195)
```

FAQ.md
```diff
@@ -59,6 +70,16 @@ conda activate <my_new_env>
...
+## My conda installation of the toolkit STILL doesn't appear to work. 
+
+Many of our users encounter issues that are ultimately due to their terminal finding a different `conda` at higher priority in their `PATH` than the `conda` deployment where OpenFF is installed. To fix this, find the conda deployment where OpenFF is installed. Then, if that folder is something like `~/miniconda3`, run in the terminal:
+
+```shell
+source ~/miniconda3/etc/profile.d/conda.sh
+```
+
+and then try rerunning and/or reinstalling the Toolkit. 
+
```
Yoshanuikabundi, line 82: This makes me think that it might be best to describe how to check if you already have conda installed in the installation guide.

FAQ.md
Yoshanuikabundi: I think if we're saying "review each output closely", it might be nice to give a brief checklist of things to check - stereochemistry, bond order, partial charges, and whatever I've forgotten (which proves my point :wink:)

FAQ.md
Yoshanuikabundi:
```suggestion
- A `.sdf`, `.mol`, or `.mol2` file or files for the molecules comprising your system, with correct bond orders and formal charges. (Note: Do NOT generate this from a simulation package or tool which does not have access to bond order information; you may end up with a correct-seeming file, but the bond orders will be incorrect)
```

This is the nittiest nit, but elsewhere in the documentation for this project and others we're very consistent about recommending SDF, so it makes sense to put it first here.

Yoshanuikabundi (commented):
1 comment and two nits that I think we should fix. Otherwise this looks great! Much needed revision, thank you.

FAQ.md
j-wags: Conda is such a deep topic that I'm not actually sure we should to do this - The scope of that explanation could get way too wide ("here's what it means if your conda is in a system folder... it's not a showstopper but it is way more complicated in the following ways"), or have us suggest things that break someone else's computing environment (most of the bashrcs I find in the wild are clearly not understood by the user and have geologic layers of cruft). I like the `conda.sh` script because it's a "steamroll my current terminal session environment and just make it work" command that doesn't depend on the user's system details.

FAQ.md
j-wags: Good call. I've linked to the "Core concepts" page here, which starts with a discussion of what's in a Molecule

PR #1142, opened 2021-12-01
Use `openff-units` in topology refactor, make OpenMM and MDTraj optional dependencies

mattwthompson:
* Merge a relatively recent version of `master` into `biopolymer-topology-refacator` (!!!)
* Drop support for `simtk` namespace (formerly #1144)
* Make `openmm` optional
  * Functions that directly interact with OpenMM API/object are decorated with `@required_package("openmm")`
  * Packaging changes are TBD
* Use `mendeleev` for handling elements
* Make `mdtraj` an optional dependence
  * No core API points were using it
  * Packaging changes are TBD
* Use `openff-units` registry (formerly #1079)


j-wags (approved):
Approved pending the porting of all the relevant commits from #1079!

mattwthompson:
I've update the original comment in an attempt to clarify the many moving pieces at play here.

The bulk of this history has been approved once; in principle it would be useful to have this reviewed again, but I don't think there is any easy way to do that. Since this is not going to hit production for a few months, I'd advocate for it to be merged into the big refactor branch and whatever problems arise from this be dealt with these.

This is assuming I can get back to having **most** tests pass - might take a couple of iterations.

j-wags:
>  I'd advocate for it to be merged into the big refactor branch and whatever problems arise from this be dealt with these.
> This is assuming I can get back to having most tests pass - might take a couple of iterations.

Yup, agreed. It's fine if things aren't 100% perfect and we find some bugs later. Thanks again for doing so much legwork on this update :-)

Issue #1145, opened 2021-12-03
Should we add a Molecule.from_xyz() to allow easy perception of chemistry from dataset that lack chemical information?


jchodera:
**Is your feature request related to a problem? Please describe.**
I repeatedly hear from users that have erased the chemical information in their molecule (transforming into something like an XYZ or PDB file that only contains elements and 3D coordinates) that they would like a convenient way to create a `Molecule` to use OpenFF toolkit features, even if the heuristics for reconstructing chemical information are error-prone. It seems like we should support some path for doing this conveniently, but perhaps require an optional argument in the constructor or a special API point that requires the user to implicitly acknowledge this is a less desirable path to constructing a molecule.

**Describe the solution you'd like**
We could support either a constructor or static method that enables file formats that lack chemical information to create a `Molecule` object, provided a special API point or optional argument is provided that acknowledges this approach is likely error-prone.

* The OpenEye toolkit provides [a heuristic to perceive chemistry from positions](https://docs.eyesopen.com/toolkits/python/oechemtk/connectivity.html).
* RDKit provides a way to [construct a molecule from a PDB file block](https://www.rdkit.org/docs/source/rdkit.Chem.rdmolfiles.html#rdkit.Chem.rdmolfiles.MolFromPDBBlock)
* AmberTools provides a way to perceive connectivity and formal charges from a PDB/XYZ file and write an SDF file using a separate set of heuristics (with drawbacks described [here](https://jcheminf.biomedcentral.com/articles/10.1186/1758-2946-4-26
)).

**Describe alternatives you've considered**
This problem comes up frequently, so having an explicit API for dealing with it, together with explicit documentation on why this is not the preferred route, is a better practice than the current state of things.

j-wags:
Hm, this a big source of user confusion, so something like this for XYZ (and maybe some PDBs that have element info) could be nice. 

On the other hand, this adds to our scope, and could lead us toward being responsible for differences between guessing heuristics in different packages.

If we do implement this, I recommend the second API point you suggested. 

> `molecule = Molecule.from_xyz_unsafe(xyzfile)`

I'd be happy to lay out more of a spec for this if someone wants to take it on. For now, since workarounds are available (and the workarounds indicate to the user that we're not responsible for the correctness of the guesses), I'll tag this as medium priority.

davidlmobley:
[...]

Issue #1137, opened 2021-11-29
Allow either direct tuple of species and coordinates or temporary file to create Molecule object. 


khuddzu:
I am trying to implement your software in my project. Currently our data is in h5 format, and we have tensors of species and coordinates with dimensions [M, Na] and [M, Na, 3] respectively, where M is number of conformers and Na is number of atoms in conformation. 
If species and coordinates are s and c, then I have tried to create the Molecule object using (s, c) to see if this would work. It would be nice if it would, so you don't actually have to have a file you just need your data. 
I have also tried creating a temporary file with the data I have using the following code: 

for k,v in ds.items():                                                         # ds is my dataset
    species = ds[k]['species']
    coordinates = ds[k]['coordinates']
    for s, c in zip(species, coordinates):                                      #iterating through each conformation
        s = s.unsqueeze(0)
        c = c.unsqueeze(0)
        temp = tempfile.NamedTemporaryFile(mode='r+', suffix='.xyz')            #Create temp file 
        tensor_to_xyz(temp.name, (s, c), truncate_output_file=True)             #make xyz 
        mol = next(pybel.readfile('xyz', temp.name))                            #create mol object
        temp.close()                                                            #close temp file
        mol_raw = mol.write(format='mol')                                       #create mol file object
        mol = Molecule.from_file(mol_raw, file_format="XYZ")

I have also tried the above code but instead of the mol file object, I then created an rdkit Mol object and then tried to create your Molecule object, and this didn't work. 
Also, it should be noted that it is still reading a xyz file so the error I am getting is Bad File Input. with the temp file. this error might be solved with allowing XYZ as a file format for the inout of the Molecule object, but I am not sure. 

The most convenient option for me. however, is to allow for direct input of species and coordinates. 

Thank you. Please let me know if I can help in any way to resolve this issue.

mattwthompson:
Hi Kate,

[...]

PR #1148, opened 2021-12-07
Add specific exception for Molecule.from_file with XYZ files

mattwthompson:
Partial solution for confusing user experiences (#1137), possibly to be changed in the future pending decisions #1145

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -57,6 +57,7 @@ from openff.toolkit.utils.exceptions import (
     InvalidConformerError,
     NotAttachedToMoleculeError,
     SmilesParsingError,
+    UnsupportedFileTypeError,
 )
 from openff.toolkit.utils.serialization import Serializable
 from openff.toolkit.utils.toolkits import (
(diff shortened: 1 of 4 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -4743,6 +4744,14 @@ def from_file(
                 file_format = file_path.split(".")[-1]
         file_format = file_format.upper()
 
+        if file_format == "XYZ":
+            raise UnsupportedFileTypeError(
+                "Parsing `.xyz` files is not currently supported because they lack sufficient "
+                "chemical information to be used with SMIRNOFF force fields. For more information, "
+                "see https://open-forcefield-toolkit.readthedocs.io/en/latest/faq.html or to provide "
+                "feedback please visit https://github.com/openforcefield/openff-toolkit/issues/1145."
```
mattwthompson, line 4752: Directly linking to an issue is probably to aggressive, though I've seen some other projects do this in other contextx (i.e. issue pypa/pip issue 9617). It would be great to in general turn these "we don't support X" points into something more concrete, if only to get a sense of how commonly an unsupported behavior is reached.

openff/toolkit/topology/molecule.py
```diff
@@ -4743,6 +4744,14 @@ def from_file(
                 file_format = file_path.split(".")[-1]
         file_format = file_format.upper()
 
+        if file_format == "XYZ":
+            raise UnsupportedFileTypeError(
+                "Parsing `.xyz` files is not currently supported because they lack sufficient "
+                "chemical information to be used with SMIRNOFF force fields. For more information, "
+                "see https://open-forcefield-toolkit.readthedocs.io/en/latest/faq.html or to provide "
```
mattwthompson, line 4751: This redirects to the 0.10.1 build, which is fine, but will be slightly updated in the new release that includes #1139

openff/toolkit/topology/molecule.py
j-wags: I think that directly linking an issue is fine here. It's a living topic and, as you say, it's likely to change in the short term in a way that could use feedback.

j-wags (approved):
Excellent. Thanks @mattwthompson! Please update the releasenotes before merging :-)

PR #1153, opened 2021-12-08
Fix 1152

### LINKED ISSUE #1152: OpenEye 2D "courtesy conformer" conflicts with downstream omega use
**Describe the bug**

A user encountered a bug resulting from the interaction of two seemingly innocuous behaviors.

1. If a user makes an OFFMol from SMILES, then immediately writes it to SDF using OpenEyeToolkitWrapper, a 2D "courtesy conformer" of the molecule is automatically generated. 
2. If omega is passed an input molecule with both graph stereochemistry and 3D stereochemistry, it will "trust" the 3D stereochemistry. 

**To Reproduce**

The reproducing case below relies on (sigh) what I think is another bug where we trust the 3D stereochemistry from a loaded molecule, even if it contradicts the graph stereochemistry. But this can be used in this example to show that the conformer generated from a loaded 2D molecule has the wrong stereochemistry.

```python
# Make a simple chiral molecule from SMILES
smiles = "[C@H](F)(Cl)(Br)"
offmol = Molecule.from_smiles(smiles)
# Show that the first stereocenter is S and the mol has 0 conformers
print(offmol.atoms[0].stereochemistry, offmol.conformers)

# Save it to file without adding any conformers
offmol.to_file('xxx.sdf', file_format='sdf')
```
> S None

```python
# Load the molecule we just saved
offmol2 = Molecule.from_file('xxx.sdf', file_format='sdf')
# Show that the first stereocenter is still S and that the mol now has 1 conformer (the "courtesy conformer")
print(offmol2.atoms[0].stereochemistry, offmol2.conformers)
```

> S [Quantity(value=array([[ 0.,  0.,  0.],
>        [ 1.,  0.,  0.],
>        [-0., -1.,  0.],
>        [ 0.,  1.,  0.],
>        [-1.,  0.,  0.]]), unit=angstrom)]

```python
# Generate conformers and show that the GRAPH still thinks the stereo is S, and that 
# a new conformer has replaced the courtesy conformer.
offmol2.generate_conformers()
print(offmol2.atoms[0].stereochemistry, offmol2.conformers)
offmol2.to_file('yyy.sdf', file_format='sdf')

```

> S [Quantity(value=array([[ 1.20802656e-01, ...]]) , unit=angstrom)]

```python
# Load the newly generated conformer from file, triggering a perception of stereo 
# from the geometry (possibly another bug, since I think it should should trust 
# the graph stereo...)
offmol3 = Molecule.from_file('yyy.sdf')
# Show that the correct geometry was loaded, but that it's actually the R isomer
print(offmol3.atoms[0].stereochemistry, offmol3.conformers)
```
> R [Quantity(value=array([[ 1.20802656e-01, ...]]) , unit=angstrom)]

**Suggested resolution**

If I delete the courtesy conformer before running `offmol2.generate_conformers`, everything works fine:

```python
...
offmol2._conformers = None
offmol2.generate_conformers()
offmol2.to_file('yyy.sdf', file_format='sdf')

offmol3 = Molecule.from_file('yyy.sdf')
print(offmol3.atoms[0].stereochemistry)
```
> S

So I'd propose scrubbing all conformers before invoking omega for conformer generation, so it HAS to read the graph stereo. 

 - Mac OSX 12
 - OFFTK 0.10.1, OETK 2020.2.2

**Additional context**

There may be a less brute-force workaround by manipulating [`OEOmegaOptions.SetIncludeInput`](https://docs.eyesopen.com/toolkits/python/omegatk/OEConfGenClasses/OEOmegaOptions.html#OEConfGen::OEOmegaOptions::SetIncludeInput) and/or [`OEMolBuilderOptions.SetFromCT`](https://docs.eyesopen.com/toolkits/python/omegatk/OEConfGenClasses/OEMolBuilderOptions.html#OEConfGen::OEMolBuilderOptions::SetFromCT) but the former seems unlikely to help, and the default value for the latter should yield the behavior that we want, so I didn't bother trying to dig in to switch them.

j-wags:
- [x] Fixes #1152
- [ ] ~Add tests~ This behavior is too complex to readily test, I've left a link to #1152 by the code change

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -15,6 +15,11 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
   and no longer accepts input of NetworkX graphs.
 - [PR #1130](https://github.com/openforcefield/openforcefield/pull/1130): Running unit tests will
   no longer generate force field files in the local directory.
+- [PR #1153](https://github.com/openforcefield/openforcefield/pull/1153): Fixes
+  [Issue #1152](https://github.com/openforcefield/openff-toolkit/issues/1052) in which running 
+  [`Molecule.generate_conformers`](openff.toolkit.topology.Molecule.generate_conformers)
+  using the OpenEye backend would use the stereochemistry from an existing conformer instead 
+  of the stereochemistry from the molecular graph, leading to undefined behavior if the molecule had a 2D conformer. 
 
 ### Examples added
 
--- a/openff/toolkit/utils/openeye_wrapper.py
+++ b/openff/toolkit/utils/openeye_wrapper.py
@@ -1674,8 +1674,9 @@ class OpenEyeToolkitWrapper(base_wrapper.ToolkitWrapper):
         clear_existing : bool, default=True
             Whether to overwrite existing conformers for the molecule
         """
-        from openeye import oeomega
         import copy
+
+        from openeye import oeomega
 
         # Copy the molecule and scrub the conformers so that omega HAS to read stereo from graph mol
         # See https://github.com/openforcefield/openff-toolkit/issues/1152
--- a/openff/toolkit/utils/openeye_wrapper.py
+++ b/openff/toolkit/utils/openeye_wrapper.py
@@ -1675,8 +1675,14 @@ class OpenEyeToolkitWrapper(base_wrapper.ToolkitWrapper):
             Whether to overwrite existing conformers for the molecule
         """
         from openeye import oeomega
+        import copy
 
-        oemol = self.to_openeye(molecule)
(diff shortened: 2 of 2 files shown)


mattwthompson (approved):
As discussed, this might not be a perfect fix for all 2D/3D conflicts, but it seems to address the issue.

Left unresolved is the possibility - not investigated deeply, but seems likely - that `Molecule.from_openeye` uses the stereochemistry inferred from the coordinates instead of the graph. If this is the case, it is inconsistent with the behavior of `Molecule.to_openeye` and a general design point that's currently somewhat ambiguous elsewhere. This should be addressed in another issue - "What wins when 2D and 3D conflict?"

j-wags:
Thanks, @mattwthompson!

Issue #1154, opened 2021-12-09
Big endian not enforced in float arrays in de/serialisation


lilyminium:
**Describe the bug**

**Serialisation**
https://github.com/openforcefield/openff-toolkit/blob/ce84d821cd2097dc249abac43de2e61249bcf5d5/openff/toolkit/utils/utils.py#L664-L666

`newbyteorder` does not change the order of the bytes, but only the order that they are interpreted in. e.g.

```python
>>> arr = np.arange(3).astype(float)
>>> arr
array([0., 1., 2.])
>>> arr.newbyteorder(">")
array([0.00000e+000, 3.03865e-319, 3.16202e-322])
```

So this function writes exactly the same bytes in the same order that were in the array before.

**Deserialisation**

https://github.com/openforcefield/openff-toolkit/blob/ce84d821cd2097dc249abac43de2e61249bcf5d5/openff/toolkit/utils/utils.py#L692-L694

`newbyteorder` is not side-effecting, so this doesn't do anything. See https://github.com/openforcefield/openff-interchange/issues/345 for more.

**To Reproduce**

**Output**

**Computing environment (please complete the following information):**
 - Operating system
 - Output of running `conda list` 

**Additional context**

The reason that the endian enforcement has not shown up in tests is because both functions have no effect, so they default to the native endiannness of the system. This likely hasn't been a problem for users transferring files because by and large people work on x86, which is little-endian. I think one could test the functions individually by creating arrays with opposite endians and checking that they are serialized and deserialized properly. e.g. on my Mac

```
>>> import numpy as np
>>> from openff.toolkit.utils.utils import serialize_numpy, deserialize_numpy
>>> import sys
>>> sys.byteorder
'little'
>>> dt_bigendian = np.dtype(float).newbyteorder(">")
>>> arr = np.arange(3).astype(dt_bigendian)
>>> arr
array([0., 1., 2.])
>>> np.frombuffer(arr.tobytes())
array([0.00000e+000, 3.03865e-319, 3.16202e-322])
>>> actually_serialize_numpy = lambda x: (x.astype(dt_bigendian).tobytes(), x.shape)
>>> actually_deserialize_numpy = lambda x, y: np.reshape(np.frombuffer(x, dtype=dt_bigendian), y)
>>> np.frombuffer(actually_serialize_numpy(arr)[0])
array([0.00000e+000, 3.03865e-319, 3.16202e-322])
>>> little_arr = np.arange(3).astype(float)
[...]

PR #1155, opened 2021-12-09
Enforce and test big-endianness in NumPy serialization/deserialization

mattwthompson:
#1154


openff/toolkit/tests/test_utils_serialization.py
```diff
@@ -118,6 +120,40 @@ def test_pickle(self):
...
+    def test_serialize_endianness(self, dtype):
+        """Test that arrays are serialized as big-endian"""
+        # Force big- (">") or little-endianness ("<") , overriding architecture default
+        arr = np.arange(3).astype(float).astype(dtype)
+
+        # This bytestring should be a big-endian representation, generated by
+        # np.arange(3).astype(np.dtype(float).newbyteorder(">")).tobytes()
+        assert serialize_numpy(arr)[0] == (
+            b"\x00\x00\x00\x00\x00\x00\x00\x00?\xf0\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00"
+        )
```
lilyminium, line 140:
```suggestion
    @pytest.mark.parametrize("endian", [">", "<"])
    def test_serialize_endianness(self, endian):
        """Test that arrays are serialized as big-endian"""
        # Force big- (">") or little-endianness ("<") , overriding architecture default
        dtype = np.dtype(float).newbyteorder(endian)
        arr = np.arange(3).astype(dtype)

        # This bytestring should be a big-endian representation, generated by
        # np.arange(3).astype(np.dtype(float).newbyteorder(">")).tobytes()
        assert serialize_numpy(arr)[0] == (
            b"\x00\x00\x00\x00\x00\x00\x00\x00?\xf0\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x00"
        )
```

Maybe just passing the endian direction would save some vertical space?

lilyminium:
Ah, sorry about the pushes!! I was on the wrong branch

lilyminium:
I haven't really looked into this, but I wonder if OpenEye's C code just assumes things are little-endian -- in the failing tests, is the data passed to the C layer or through a Python interpretation? Switching to little endian fixed the tests: https://github.com/openforcefield/openff-toolkit/runs/4498459402?check_suite_focus=true

mattwthompson:
Thanks for having a look - I don't see any problems as the OpenEye interface, and I'm not sure if we ever pass bytes to it. I only see JSON roundtrips failing and AFAICT we do all of our JSON serialization with our own code. I did find a function that manually called [`np.frombuffer`](https://github.com/openforcefield/openff-toolkit/blob/444254a94fd832d3c0f3425cad72aff6919f9bdf/openff/toolkit/utils/serialization.py#L473-L491) without going through `deserialize_numpy`. This seems consistent with seeing your patch pass, since it forces the endianness to be the architecture default.

8e66076 / 7e095e5  fix the failing tests locally by ensuring `np.frombuffer` knows the bytes it's passed are supposed to be endian, we'll see if they fix them in CI.

It's not something I want fixed here, but the whole idea of [`_prep_numpy_data_for_json`](https://github.com/openforcefield/openff-toolkit/blob/444254a94fd832d3c0f3425cad72aff6919f9bdf/openff/toolkit/utils/serialization.py#L126-L130) is maybe a little inefficient, since conformer data passed to JSON will be thrown to bytes, then back into memory as an array, and then written out to a list, and then finally a string. I doubt this has a practical impact on performance since these operations should be pretty low-level and quick. I have done some profiling elsewhere and most of the time is spend doing unit-checking.

lilyminium:
Ah I see, I was thrown off by the tests passing for RDKit and not OpenEye -- thanks for actually looking at the code!

openff/toolkit/utils/utils.py
```diff
@@ -641,9 +639,9 @@ def detach_units(unit_bearing_dict, output_units=None):
...
 
 
-def serialize_numpy(np_array):
+def serialize_numpy(np_array) -> Tuple[bytes, Tuple[int]]:
     """
-    Serializes a numpy array into a JSON-compatible string. Leverages the numpy.save function,
+    Serializes a numpy array into a JSON-compatible big-endian bytestring. Leverages the numpy.save function,
     thereby preserving the shape of the input array
 
     from https://stackoverflow.com/questions/30698004/how-can-i-serialize-a-numpy-array-while-preserving-matrix-dimensions#30699208
```
lilyminium, line 647:
```suggestion
    Serializes a numpy array into a JSON-compatible big-endian bytestring.
```

Non-blocking -- perhaps I'm missing something, but this docstring no longer seems accurate; np.save isn't really used?

openff/toolkit/utils/utils.py
```diff
@@ -655,21 +653,24 @@ def serialize_numpy(np_array):
...
 
 
-def deserialize_numpy(serialized_np, shape):
+def deserialize_numpy(serialized_np: Union[bytes, List], shape: Tuple[int]):
     """
-    Deserializes a numpy array from a JSON-compatible string.
+    Deserializes a numpy array from a JSON-compatible bytestring. The input, if a bytestring, is
+    assumed to be in big-endian byte order.
 
     from https://stackoverflow.com/questions/30698004/how-can-i-serialize-a-numpy-array-while-preserving-matrix-dimensions#30699208
```
lilyminium, line 675: Again I'm not sure if this answer really applies here.

lilyminium (approved):
I'm not sure if this needs an entry in the releasehistory -- it's pretty base-level, but I also have an uneasy feeling that you're more likely to encounter potential future problems by choosing big- over little-endian, and it might be good to have a handy link in the docs just in case.

Otherwise LGTM -- I do agree that `_prep_numpy_for_json` might need future work, although not here. Apologies again for pushing to the branch and adding the xyz stuff.

openff/toolkit/utils/utils.py
mattwthompson: Correct, this either didn't ever use `np.save` or did so in a way that was lost to the sands of time/git history

PR #1158, opened 2021-12-20
Fix Molecule._ipython_display_ breaking without nglview

### LINKED ISSUE #1157: Molecule repr seems to raise a ModuleNotFoundError (without nglview)
**Describe the bug**

In a notebook, repr'ing a Molecule without nglview throws an error.  This seems a little aggressive / unwelcoming.  An alternative would be to fall back to a simpler repr on import failure?

**To Reproduce**

```
m = Molecule('...')
repr(m)
```

**Output**

```
ModuleNotFoundError                       Traceback (most recent call last)
~/miniconda3/envs/mda/lib/python3.8/site-packages/openff/toolkit/topology/molecule.py in visualize(self, backend, width, height, show_all_hydrogens)
   6365             try:
-> 6366                 import nglview as nv
   6367             except ImportError:

ModuleNotFoundError: No module named 'nglview'

During handling of the above exception, another exception occurred:

MissingDependencyError                    Traceback (most recent call last)
~/miniconda3/envs/mda/lib/python3.8/site-packages/IPython/core/formatters.py in __call__(self, obj)
    916             method = get_real_method(obj, self.print_method)
    917             if method is not None:
--> 918                 method()
    919                 return True
    920 

~/miniconda3/envs/mda/lib/python3.8/site-packages/openff/toolkit/topology/molecule.py in _ipython_display_(self)
   6457 
   6458         try:
-> 6459             return display(self.visualize(backend="nglview"))
   6460         except (ImportError, ValueError):
   6461             pass

~/miniconda3/envs/mda/lib/python3.8/site-packages/openff/toolkit/topology/molecule.py in visualize(self, backend, width, height, show_all_hydrogens)
   6366                 import nglview as nv
   6367             except ImportError:
-> 6368                 raise MissingDependencyError("nglview")
   6369 
   6370             if width is not None or height is not None:

MissingDependencyError: Missing dependency nglview. Try installing it with

$ conda install nglview -c conda-forge
```

 - Operating system

**Additional context**

mattwthompson:
Fixes #1157, which I _think_ was introduced in #1021 and a part of 0.10.1.

`nglview` is pulled in by all of our test environments, so the unit tests does not actually do anything on CI now. 

```
$ conda-tree whoneeds nglview                                 viz-no-nglview  ✈ ✱
['qcportal']
$ grep -r qcportal devtools                                   viz-no-nglview  ✈ ✱
devtools/conda-envs/rdkit.yaml:  - qcportal
devtools/conda-envs/openeye.yaml:  - qcportal
devtools/conda-envs/beta_rc_env.yaml:  - qcportal
devtools/conda-envs/test_env.yaml:  - qcportal
```

It fails locally before the fix (d37070b44af27e09174116b67f5d2c7b83549204); I guess you'll have to take me on my word there.

Reviewing this should include firing up a notebook and reproducing #1157 with and without these changes.

### DIFF
--- a/openff/toolkit/utils/exceptions.py
+++ b/openff/toolkit/utils/exceptions.py
@@ -51,7 +51,7 @@ class IncompatibleUnitError(OpenFFToolkitException):
     """
 
 
-class MissingDependencyError(OpenFFToolkitException):
+class MissingDependencyError(OpenFFToolkitException, ImportError):
     """
     Exception for when an optional dependency is needed but not installed
 
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -24,6 +24,9 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
   [`Molecule.generate_conformers`](openff.toolkit.topology.Molecule.generate_conformers)
   using the OpenEye backend would use the stereochemistry from an existing conformer instead 
   of the stereochemistry from the molecular graph, leading to undefined behavior if the molecule had a 2D conformer. 
+- [PR #1158](https://github.com/openforcefield/openff-toolkit/pull/1158): Fixes the default
+  representation of [`Molecule`](openff.toolkit.topology.Molecule) failing in Jupyter notebooks when
+  NGLview is not installed.
 - [PR #1151](https://github.com/openforcefield/openforcefield/pull/1151): Fixes 
   [Issue #1150](https://github.com/openforcefield/openff-toolkit/issues/1150), in which calling 
   [`Molecule.assign_fractional_bond_orders`](openff.toolkit.topology.Molecule.assign_fractional_bond_orders)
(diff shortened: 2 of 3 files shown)


openff/toolkit/utils/exceptions.py
```diff
@@ -51,7 +51,7 @@ class IncompatibleUnitError(OpenFFToolkitException):
     """
 
 
-class MissingDependencyError(OpenFFToolkitException):
+class MissingDependencyError(OpenFFToolkitException, ImportError):
```
j-wags, line 54: Is this diff supposed to be here?

j-wags (approved):
Tested this locally - Reproduced the error before the changes in this PR, and then that the PR resolves it. LGTM! One potentially spurious line, but feel free to leave it in if it's important.

openff/toolkit/utils/exceptions.py
mattwthompson: Yes, this is the change that changes the bug/undesired behavior. The other approach would be to catch this `MissingDependencyError` _instead of_ `ImportError` in `_ipython_display_` [here](https://github.com/openforcefield/openff-toolkit/blob/master/openff/toolkit/topology/molecule.py#L6436-L6439). I left this stand while #771 played out, and the discussion there seems to be in favor of this approach, so I left it in place.

mattwthompson: Thanks for the speedy review!

PR #1160, opened 2021-12-21
fix #1159 and add test

### LINKED ISSUE #1159: Incorrect v-site orientation
**Describe the bug**

The toolkit currently by default sorts the list of parent atom indices when creating a v-site which leads to different orientations for the v-site than were expected.

Presumably this will lead to v-sites being displaced in the wrong direction. 

**To Reproduce**

```
from openmm import unit

force_field = ForceField()

vsite_handler = force_field.get_parameter_handler("VirtualSites")
vsite_handler.add_parameter(
    parameter_kwargs={
        "smirks": "[#6:2]=[#8:1]",
        "name": "EP",
        "type": "BondCharge",
        "distance": 0.7 * unit.nanometers,
        "match": "once",
        "charge_increment1": 0.2 * unit.elementary_charge,
        "charge_increment2": 0.1 * unit.elementary_charge,
        "sigma": 1.0 * unit.angstrom,
        "epsilon": 2.0 / 4.184 * unit.kilocalorie_per_mole,
    }
)

molecule = Molecule.from_mapped_smiles("[O:2]=[C:1]=[O:3]")

from openmm import System
omm_system: System
omm_system, topology = force_field.create_openmm_system(
    molecule.to_topology(), return_topology=True
)

for molecule in topology.reference_molecules:
    for v_site in molecule.virtual_sites:
        for particle in v_site.particles:
            print(particle.orientation)

for i in range(omm_system.getNumParticles()):

    if not omm_system.isVirtualSite(i):
        continue

    omm_v_site = omm_system.getVirtualSite(i)

    print(
        tuple(omm_v_site.getParticle(j) for j in range(omm_v_site.getNumParticles()))
    )
```

**Output**

_Actual_

```
(0, 1)
(0, 2)
(0, 1)
(0, 2)
```

_Expected_

```
(1, 0)
(2, 0)
(1, 0)
(2, 0)
```

 - Operating system
 - Output of running `conda list`   `openff-toolkit = 0.10.1`

**Additional context**

j-wags:
- [x] Fix #1159 

Status

- [x] Ready for review

### DIFF
--- a/openff/toolkit/tests/test_forcefield.py
+++ b/openff/toolkit/tests/test_forcefield.py
@@ -2212,6 +2212,58 @@ class TestForceFieldVirtualSites:
 
         self._test_physical_parameters(toolkit_registry, *args.values())
 
+    def test_orientation_preserved_if_match_once(self):
+        """
+        Test that orientation-mangling bug from https://github.com/openforcefield/openff-toolkit/issues/1159
+        is resolved.
+        """
+        from openff.toolkit.typing.engines.smirnoff import ForceField
+        from openmm import unit
+
+        force_field = ForceField()
+
+        vsite_handler = force_field.get_parameter_handler("VirtualSites")
+        vsite_handler.add_parameter(
+            parameter_kwargs={
+                "smirks": "[#6:2]=[#8:1]",
+                "name": "EP",
+                "type": "BondCharge",
+                "distance": 0.7 * unit.nanometers,
+                "match": "once",
+                "charge_increment1": 0.2 * unit.elementary_charge,
+                "charge_increment2": 0.1 * unit.elementary_charge,
+                "sigma": 1.0 * unit.angstrom,
+                "epsilon": 2.0 / 4.184 * unit.kilocalorie_per_mole,
+            }
+        )
+
+        from openff.toolkit.topology import Molecule
+        molecule = Molecule.from_mapped_smiles("[O:2]=[C:1]=[O:3]")
+
+        from openmm import System
+        omm_system: System
+        omm_system, topology = force_field.create_openmm_system(
+            molecule.to_topology(), return_topology=True
+        )
+
+        for molecule in topology.reference_molecules:
+            for v_site in molecule.virtual_sites:
+                for particle in v_site.particles:
+                    print(particle.orientation)
+
+        omm_particle_tuples = []
+        for i in range(omm_system.getNumParticles()):
+
+            if not omm_system.isVirtualSite(i):
+                continue
+
+            omm_v_site = omm_system.getVirtualSite(i)
+
+            omm_particle_tuples.append(tuple(omm_v_site.getParticle(j) for j in range(omm_v_site.getNumParticles())))
+        assert (1,0) in omm_particle_tuples
+        assert (2,0) in omm_particle_tuples
+
+
 
 def generate_monatomic_ions():
     return (
--- a/openff/toolkit/tests/test_forcefield.py
+++ b/openff/toolkit/tests/test_forcefield.py
@@ -2238,9 +2238,11 @@ class TestForceFieldVirtualSites:
         )
 
         from openff.toolkit.topology import Molecule
+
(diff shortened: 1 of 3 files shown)


j-wags:
I'm gonna mark this as not-yet-ready-for-review, because I want to think a bit more about other bugs it may introduce. I'll take another look at this tomorrow morning and open it for review if it still makes sense!

j-wags:
@trevorgokey Ok -- I've thought about this a bit more, and I still can't fully wrap my head around it, but I think this PR is ready for review. My thinking is that this _might_ create a new bug somewhere else, but I'm unable to reason my way to what that bug could be. So as long as this solves a problem that _definitely_ exists now, while only _maybe_ creating a new problem, I think we should put it forward for the sake of timeliness.

I'm going to tag you for review, but just let me know if you're already out for the holidays and I'll find someone else :-)

openff/toolkit/tests/test_forcefield.py
```diff
@@ -2212,6 +2212,57 @@ def test_trivalent_charge_virtual_site(
...
+
+        omm_system: System
+        omm_system, topology = force_field.create_openmm_system(
+            molecule.to_topology(), return_topology=True
+        )
+
+        for molecule in topology.reference_molecules:
+            for v_site in molecule.virtual_sites:
+                for particle in v_site.particles:
+                    print(particle.orientation)
```
trevorgokey, line 2247:
```suggestion

```

openff/toolkit/tests/test_forcefield.py
```diff
@@ -2212,6 +2212,57 @@ def test_trivalent_charge_virtual_site(
...
+            omm_v_site = omm_system.getVirtualSite(i)
+
+            omm_particle_tuples.append(
+                tuple(
+                    omm_v_site.getParticle(j)
+                    for j in range(omm_v_site.getNumParticles())
+                )
+            )
+        assert (1, 0) in omm_particle_tuples
+        assert (2, 0) in omm_particle_tuples
```
trevorgokey, line 2264:
```suggestion
        for molecule in topology.reference_molecules:
            for v_site in molecule.virtual_sites:
                for particle in v_site.particles:
                    # correct particle.orientation is (1,0) and (2,0)
                    # and not (0, 1), (0, 2)
                    assert particle.orientation in omm_particle_tuples
```

trevorgokey (approved):
Looks good!

openff/toolkit/tests/test_forcefield.py
j-wags: Oh 🤦 good catch!

openff/toolkit/tests/test_forcefield.py
j-wags: This suggested diff replaces the original check with a test for another step in the process. This other check is also a good idea, but it's not a replacement for the original. So I'm going to modify it to do BOTH the original check, and also the OFF vs. OMM vsite particle index check.

j-wags, line 2264:
```suggestion
        assert (1, 0) in omm_particle_tuples
        assert (2, 0) in omm_particle_tuples
        for molecule in topology.reference_molecules:
            for v_site in molecule.virtual_sites:
                for particle in v_site.particles:
                    # correct particle.orientation is (1,0) and (2,0)
                    # and not (0, 1), (0, 2)
                    assert particle.orientation in omm_particle_tuples
```

mattwthompson:
The release notes say

> PR #\1160: Fixes a major bug identified in Issue #\1159, in which the order of atoms defining a BondChargeVirtualSite (and possibly other virtual sites types too)


I'm seeing only seeing a behavior change with a unit test with a `VirtualSiteMonovalentLonePairType` since my tests are extremely spotty, so FTR this bug probably affected other types as well.

j-wags:
Yeah, I could convince myself that TrivalentLonePairSite SHOULDN'T by this bug, and that BondChargeVirtualSites SHOULD, but I wasn't sure about the others. I could believe that Mono- and DivalentLonePair would be affected too.

PR #1162, opened 2022-01-03
Turn off `--warn-unused-ignores` to avoid false failures

mattwthompson:
I set up Mypy with [`--warn-unused-ignores`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-unused-ignores), which actually throws an error. This has been causing what I see as false negative test failures in CI for a few weeks now. I've fixed it locally but I think the option should just be turned off in CI. It's not a bad idea to use it locally when making bigger changes.


j-wags (approved):
Agreed. Thanks!

PR #1165, opened 2022-01-06
Add `use_interchange` argument to `ForceField.create_openmm_system`

mattwthompson:
* The existing `create_openmm_system` is moved to `_old_create_openmm_system`
* A new `create_openmm_system` delegates the creation of the `openmm.System` to Interchange
* A `use_interchange` argument is added, defaulting to `False` here.
* `ForceField.create_interchange` is now public

The default behavior is unchanged.


openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1377,11 +1399,24 @@ def create_openmm_system(self, topology, **kwargs):
             return system
 
     @requires_package("openff.interchange")
-    def _to_interchange(self, topology, box=None):
+    def to_interchange(self, topology: "Topology", box=None):
```
j-wags, line 1402: Since this doesn't convert the forcefield inTO an `Interchange` object, but rather is a method that adds a lot of information content and CREATEs an `Interchange` object, what would you think about calling this `create_interchange`?

mattwthompson: Sure, seems like a pure improvement to me

mattwthompson:
Since the `use_interchange` keyword argument only exists in this branch, It's currently difficult to use any other feature branches (i.e. #1182) while testing experimental features. I'm advocating for this to be merged with `use_interchange=False` **so that the argument can be used**. Otherwise I will continue to have duplicate copy+pasted code in many locations and potentially other feature branches may diverge over time.

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1377,11 +1405,24 @@ def create_openmm_system(self, topology, **kwargs):
...
-    def _to_interchange(self, topology, box=None):
+    def create_interchange(self, topology: "Topology", box=None):
         """
         Create an Interchange object from a ForceField, Topology, and (optionally) box vectors.
 
         WARNING: This API and functionality are experimental and not suitable for production.
+
+        Parameters
+        ----------
+        topology : openff.system.Topology
```
j-wags, line 1416:
```suggestion
        topology : openff.toolkit.topology.Topology
```

.github/workflows/CI.yml
```diff
@@ -101,7 +101,7 @@ jobs:
 
       - name: Install package
         run: |
-          conda remove --force openff-toolkit openff-toolkit-base
+          # conda remove --force openff-toolkit openff-toolkit-base
```
j-wags, line 104: (not blocking) I'm not sure if this is necessary, and it will be a headache if we forget about it later. Could you see if this can be reintroduced?

devtools/conda-envs/dev_env.yaml
```diff
@@ -0,0 +1,51 @@
+name: pr-1165-env
```
j-wags, line 1: (blocking) This file isn't used anywhere at this point, could it be removed?

j-wags (approved):
Excellent. Thanks @mattwthompson. I'd be OK to make `use_interchange=True` when you think it's ready, but to be safe we can keep it as `False` until it passes the regression testing laid out in the Interchange adoption project. 

One or two blocking requests, other than that this should be good to go.

.github/workflows/CI.yml
mattwthompson: Turns out it **is** necessary to have this commented out _while Interchange is being installed with `pip`_, since `openff-toolkit-base` is not there to be uninstalled:

```
Run conda remove --force openff-toolkit openff-toolkit-base

PackagesNotFoundError: The following packages are missing from the target environment:
  - openff-toolkit
  - openff-toolkit-base
```

I'll note this more clearly in hopes we notice it later. I expect it'll cause problems and get out attention that way, though.

mattwthompson:
The failing test is, I believe, unrelated

```
openff/toolkit/tests/test_molecule.py::TestMolecule::test_chemical_environment_matches_OE

PR #1167, opened 2022-01-10
Use dicts to map between Identical molecules and atom maps

mattwthompson:
This PR implements a small change to `Topology.identical_molecule_groups` that I request, see [here](https://github.com/openforcefield/openff-toolkit/pull/951#issuecomment-1006942523). Currently, the function returns a nested structure in which lists are uses to map between (topology) molecule indices and atom maps in a way that a dictionary is more naturally suited for.

To illustrate the behavior change, I split out the [docexample](https://github.com/openforcefield/openff-toolkit/blob/3a81f9efc0997d8661da534589405a52fdd3c19e/openff/toolkit/topology/topology.py#L1200) into a script:
```
$ python ex.py
{0: [[0, {0: 0, 1: 1, 2: 2}], [1, {0: 1, 1: 0, 2: 2}]]}
$ git checkout identical-molecules-dict
$ python ex.py
{0: [{0: {0: 0, 1: 1, 2: 2}}, {1: {0: 1, 1: 0, 2: 2}}]}
```

- [ ] Mypy passing locally
- [ ] Mypy passing in CI
- [ ] No (new) tests failing in CI

### DIFF
--- a/openff/toolkit/topology/topology.py
+++ b/openff/toolkit/topology/topology.py
@@ -16,12 +16,12 @@ Class definitions to represent a molecular system and its chemical components
    * Use `attrs <http://www.attrs.org/>`_ for object setter boilerplate?
 
 """
-
 import copy
 import itertools
 import warnings
 from collections import OrderedDict
 from collections.abc import MutableMapping
+from typing import Dict, List, Tuple
 
 import numpy as np
 from openff.units import unit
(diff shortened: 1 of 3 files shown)


mattwthompson:
CI is still running two hours later, so I killed it. I ran `test_topology.py` locally and found `openff/toolkit/tests/test_topology.py::TestTopology::test_from_openmm` stalling for several minutes - that's probably the issue, which has already been noted.

I've updated the unit test so everything passes locally (`python -m pytest -v openff/toolkit/tests/test_topology.py::TestTopology::test_group_chemically_identical_molecules -s`, the only one I could find using this API point). I'm not _completely_ confident that the behavior is unbroken and the same as what was originally desired, so it being hammered on would be useful.

mattwthompson:
Closing for now

PR #1169, opened 2022-01-11
Have conda CI install test plugins

j-wags:
- [x] Wrap up remaining to-do ffrom #1163 
- [x] Ensure that conda CI successfully runs with this change


mattwthompson (approved):
LGRM - Python 3.6 is EOL and it is expected that our upstreams will start to cease supporting it.

PR #1171, opened 2022-01-14
Fix trans COOh in Elf10 conformer selection

### LINKED ISSUE #428: OpenEye charging for trans-carboxylic acids will fail; incorporate workaround
Currently, OpenEye charging for (neutral) trans carboxylic acids will fail for reasons I'll explain below. OpenEye has come up with a fix/workaround which we should likely propagate into our charging code, see https://github.com/oess/openmm_orion/blob/gcalabro_data_record/MDOrion/LigPrep/ff_utils.py#L28-L103

(Tangentially, Antechamber charging for trans carboxylic acids will succeed but likely lead to incorrect/artifactual charges, which is a somewhat separate issue but worth noting.)

*Chemical explanation follows*:
Chris Bayly's ELF10 AM1-BCC implementation in the OpenEye toolkits tries to avoid strong internal electrostatic interactions when assigning charges, as these can lead to artifactual charges. So, if provided with a multi-conformer carboxylic acid with both cis and trans conformers, it will throw away the trans conformers when charging so as to avoid such interactions, and use only cis. This often works fine, but if for some reason no cis conformers are generated by Omega, charging will fail in such cases (I ran into one such molecule in a separate project). To fix this, they've implemented a workaround where such carboxylic acid proton torsions are converted to cis before charging so that some cis conformers will always be available. The above points to the code for this. 

The Antechamber AM1-BCC implementation will cheerfully assign charges in such cases, which, in Chris's view, will lead to artifactual charges.

Yoshanuikabundi:
OpenEye Omega has a bug where when it is asked to sample different hydrogen positions, it produces trans carboxylic acids, quite often taking precedence over the more physical cis carboxylic acids. This is especially problematic when the conformations are passed on to the `apply_elf_conformer_selection()` method, which rejects trans-COOH and produces no conformers. Previously, this resulted in an exception without any text. As @j-wags and I discussed in our [last checkin](https://openforcefield.atlassian.net/wiki/spaces/IN/pages/2223243280/2022+01+12+Mitchell+Wagner+Check+in), this PR:

- Adds documentation to the `apply_elf_conformer_selection` describing this problem
- Adds a more descriptive error when ELF conformer selection fails without a message
- Adds a private `_make_carboxylic_acids_cis()` method to `FrozenMolecule`. This method is implemented in Numpy with no Python loops and no reference to external chemical toolkits, so it should be fast even with lots of conformations
- Adds tests for the new method
- Fixes and adds a test for calling the RDKit `apply_elf_conformer_selection` method using the toolkit wrapper, which previously did not work
- Adds an argument, `make_carboxylic_acids_cis`, to the OpenEye `generate_conformers()` method. False by default, this argument runs  `_make_carboxylic_acids_cis()` on the output molecule if set to True
- Sets the `make_carboxylic_acids_cis` argument to True when `generate_conformers` is run via `assign_partial_charges` or `assign_fractional_bond_orders()` with an ELF method, which are the main external API points where users will touch this workflow.

Open questions:
- `_make_carboxylic_acids_cis()` defines convenience methods for computing dot products, norms, and dihedral angles in Numpy arrays. Should these methods be broken out, perhaps into the `utils` module for everyone to use?
- Should we make `make_carboxylic_acids_cis` True by default in the `generate_conformers()` method?
- Should we add an argument to `assign_partial_charges` and `assign_fractional_bond_orders` to not make carboxylic acids cis?

Fixes #428 (I think)


openff/toolkit/topology/molecule.py
```diff
@@ -3275,6 +3275,147 @@ def generate_conformers(
...
+            return
+
+        # Convert all conformers into one big array
+        conformers = np.asarray(
+            [q.value_in_unit(unit.angstrom) for q in self._conformers]
+        )
+
+        # Pull out the coordinates of all carboxylic acid groups into cooh_xyz
+        # cooh_xyz is an array with shape (n_cooh_groups, n_conformers, 4, 3)
+        cooh_indices = self.chemical_environment_matches("[C:2]([O:3][H:4])=[O:1]")
```
j-wags, line 3323: (not blocking) In other circumstances we'd want to pass in a toolkit registry here, and have the method take a toolkit registry as input. But since this is a private method I don't think it's necessary

j-wags (approved):
Great work - I'm super impressed at the linearization of correcting arbitrary numbers of conformers without any for loops. I'm approving this, pending the changes requested below and @SimonBoothroyd's approval.

@SimonBoothroyd This may have an effect on outputted partial charges, by making conformers that would otherwise be discarded from ELF (due to trans-COOH conformations) get corrected before being ELF conformer selection runs. We think this change should be minor, but could you let us know if this is acceptable? 

Two asks:
* This should add the `make_carboxylic_acids_cis` kwarg to `Molecule.generate_conformers` and `RDKitToolkitWrapper.generate_conformers`, to keep the method signatures consistent. 
* If possible, `_make_carboxylic_acids_cis` should take a ToolkitWrapper/Registry to dictate which backend is used for SMARTS matching. 

> * _make_carboxylic_acids_cis() defines convenience methods for computing dot products, norms, and dihedral angles in Numpy arrays. Should these methods be broken out, perhaps into the utils module for everyone to use?

Good question. The math is highly optimized and pretty complex, so I'm fine with it living inside of the method, since that where a developer would be looking if there were a problem.

> * Should we make make_carboxylic_acids_cis True by default in the generate_conformers() method?

I think "no", because this puts us in the position of being a cheminformatics toolkit, and that opens a dangerous door.

> * Should we add an argument to assign_partial_charges and assign_fractional_bond_orders to not make carboxylic acids cis?

I don't think this is necessary. If users want to see conformers get deleted they can provide them themselves, and you point out that this only is set to True for ELF-based charge methods.

openff/toolkit/tests/test_molecule.py
```diff
@@ -3615,6 +3619,110 @@ def test_apply_elf_conformer_selection(self, toolkit):
             initial_conformers[1].value_in_unit(unit.angstrom),
         )
 
+    def test_make_carboxylic_acids_cis(self):
+        # First, check that we get exactly the right coords for cis and trans formic acid
+        offmol = Molecule.from_mapped_smiles(
+            "[H:5][C:1]([H:6])([H:7])[C:2](=[O:4])[O:3][H:8]"
```
SimonBoothroyd, line 3625:
```suggestion
            "[H:1][C:2](=[O:3])[O:4][H:5]"
```

Maybe swap to formic acid for simplicity?

openff/toolkit/tests/test_molecule.py
```diff
@@ -3615,6 +3619,110 @@ def test_apply_elf_conformer_selection(self, toolkit):
...
+                    [-0.95982963, 1.56161928, 1.87139809],
+                    [-2.65173483, 0.05333299, 1.62961793],
+                    [-0.28333247, -0.44931144, -0.04270577],
+                    [-1.79993558, 0.06760395, -0.85680676],
+                    [-0.53051436, 1.2521683, -0.48778746],
+                    [-0.18957424, 1.88568068, 1.35755765],
+                ]
+            ),
+            unit=unit.angstrom,
+        )
```
SimonBoothroyd, line 3659: It's pretty to hard to tell from these noisy values that these truly are cis and trans conformers. For *unit* tests like these I'd recommend just using dummy values where it is immediately obvious if the expected behaviour is met!

```suggestion
        cis_xyz = np.array(
            [
                [-1.0, -1.0, 0.0],  # HC
                [+0.0, +0.0, 0.0],  # C
                [-1.0, +1.0, 0.0],  # =O
                [+1.0, +0.0, 0.0],  # -O
                [+2.0, +1.0, 0.0]   # HO
            ]
        ) * unit.angstrom
        trans_xyz = np.array(
            [
                [-1.0, -1.0, 0.0],  # HC
                [+0.0, +0.0, 0.0],  # C
                [-1.0, +1.0, 0.0],  # =O
                [+1.0, +0.0, 0.0],  # -O
                [+2.0, -1.0, 0.0]   # HO
            ]
        ) * unit.angstrom
```

SimonBoothroyd (changes requested):
Thanks for this @Yoshanuikabundi ! In general this looks good, I'd just want to see the tests tidied up and made a bit more obvious with less maintenance burden.

@davidlmobley may want to have a quick scan of this, [in particular these lines](https://github.com/openforcefield/openff-toolkit/blob/b52e4f930b1ac95a83657a7affdbaf9b6e92c51d/openff/toolkit/topology/molecule.py#L3288-L3317) to make sure this aligns with his understanding of the problem as well as he's had more conversations with Chris about this.

openff/toolkit/tests/test_molecule.py
Yoshanuikabundi: That's genius

openff/toolkit/tests/test_molecule.py
```diff
@@ -3615,6 +3619,110 @@ def test_apply_elf_conformer_selection(self, toolkit):
...
+
+        offmol._make_carboxylic_acids_cis()
+
+        assert np.all(
+            np.abs(np.asarray(offmol.conformers) - np.asarray(expected_conformers))
+            < 1e-5
+        )
+
+        # OK, we've checked a single conformer in great detail, now lets check
+        # a thousand quickly but less robustly
```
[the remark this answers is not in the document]
Yoshanuikabundi: I basically just wanted to throw a lot of different stuff at it and check that a few properties held: 

1. It was cis now
2. Only the H in the COOH groups had moved
3. OH bond length hadn't changed

Poor man's fuzzing I guess. Having a weird molecule that would generate conformers that were all over the place, and that had plenty of atoms for buggy code to accidentally mutate, seemed like an easy way to generate a lot data for it. The idea was to try and catch edge cases that I wouldn't have thought of testing. I guess that's not really a unit test though (is there somewhere else I should put something similar?)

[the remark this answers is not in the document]
Yoshanuikabundi: I definitely agree that the test is a mess, should I just get rid of it and write seperate tests for all the edge cases I can think of? Or am I just being paranoid.

openff/toolkit/tests/test_molecule.py
j-wags: Oh, that's awesome. Great call.

openff/toolkit/tests/test_molecule.py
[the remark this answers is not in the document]
j-wags: Yeah, it's a mixed bag in terms of code complexity and test thoroughness. I appreciate the points on both sides, but let's go ahead and delete the lines after this comment.

davidlmobley:
Replying to @SimonBoothroyd :
> @davidlmobley may want to have a quick scan of this, [in particular these lines](https://github.com/openforcefield/openff-toolkit/blob/b52e4f930b1ac95a83657a7affdbaf9b6e92c51d/openff/toolkit/topology/molecule.py#L3288-L3317) to make sure this aligns with his understanding of the problem as well as he's had more conversations with Chris about this.

This is consistent with my recollection. However, I have this quirky problem where I frequently get this backwards, so I am doublechecking with Christopher. However, for now let's assume that you have it right AND my memory is right. :) 

Will skim the rest of the issue and see if there's anything else I should chime in on, so if I don't respond again it means I agree with everything here. :) cc @j-wags

j-wags:
Thanks for the feedback @davidlmobley, you've resolved the blocking question that we had! 

Interestingly, it looks like we largely disagree on the specific questions originally raised by @Yoshanuikabundi. This can easily get bogged down into bikeshedding, so @Yoshanuikabundi  please resolve in the following order:

1) If you have a strong preference to take @davidlmobley's answer instead of mine, do so.
2) If you don't have a strong preference, take my answer.

I'll see if I can check in with @SimonBoothroyd  before he heads out about whether the his merge block should persist after we delete the extended test that he requested removing, and will report back if so.

Yoshanuikabundi:
OK. I have a strong preference for making `make_carboxylic_acids_cis` `True` by default in `generate_conformers()`, but otherwise don't have strong feelings, so I've made that change.

> Actually, maybe I'll reconsider -- "hard but possible" may be the right place to be, as I can see wanting to check how much assigned force field parameters would vary depending on COOH orientation when using WBO interpolation, and it would be substantially harder if the toolkit didn't allow it. So maybe there should be a toolkit-level override that comes with disclaimers/warnings.

The current code in `assign_partial_charges()` only makes everything cis if it generates conformers. If conformers are provided it'll use them directly. For instance, this code produces trans COOH:

```python3
from openff.toolkit.utils import OpenEyeToolkitWrapper

mol = Molecule.from_smiles("C(=O)OH", toolkit_registry=OpenEyeToolkitWrapper())
mol.generate_conformers(make_carboxylic_acids_cis=False, toolkit_registry=OpenEyeToolkitWrapper())
mol.assign_partial_charges("am1bccelf10", use_conformers=mol.conformers, toolkit_registry=OpenEyeToolkitWrapper())

mol.visualize("nglview")
```

j-wags (approved):
Great. Thanks for the back-and-forth, everyone. I've received permission from Simon to remove the block from merging this once the requested code changes were made, so this should be good to go now!

PR #1177, opened 2022-01-20
Faster element lookups in mendeleev

j-wags:
- [x] Used tips from https://github.com/lmmentel/mendeleev/issues/7 to speed up element/symbol lookups, solving the performance regression identified in https://github.com/openforcefield/openff-toolkit/pull/951#issuecomment-1000428092

Reproducing code from Matt's post above:

```python
import time

import numpy as np
from openff.toolkit.topology import Molecule, Topology

for n_molecules in np.logspace(0, 2.5, num=10, dtype=int):
    topology = Topology.from_molecules(n_molecules * [molecule])
    openmm_topology = topology.to_openmm()
    start_time = time.time()
    Topology.from_openmm(openmm_topology, unique_molecules=[molecule])
    print(n_molecules, time.time() - start_time)
```

Results before this PR:
```
0.6627719402313232
1 0.698512077331543
3 3.0817441940307617
6 4.673622131347656
12 8.500097036361694
24 22.013742923736572
46 47.690475940704346
87 73.42918682098389
166 154.51772904396057
316 233.0623791217804
```
results after this PR
```
1 0.002974987030029297
1 0.0026962757110595703
3 0.005962848663330078
6 0.014431953430175781
12 0.033457040786743164
24 0.050592899322509766
46 0.09189701080322266
87 0.16967487335205078
166 0.3632802963256836
316 0.8385882377624512
```

So, this gives a good performance improvement, however it makes a mendeleev table (`MENDELEEV_ELEMENTS_DATAFRAME`) at the module scope. I wonder if there's a cleaner way to do this?

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -38,7 +38,10 @@ from typing import TYPE_CHECKING, List, Optional, Union
 
 import networkx as nx
 import numpy as np
+import mendeleev
 from mendeleev import element
+from mendeleev.fetch import fetch_table
+
 from openff.units import unit
 from openff.units.openmm import to_openmm
 
@@ -80,6 +83,9 @@ from openff.toolkit.utils.utils import (
 # =============================================================================================
 # GLOBAL PARAMETERS
 # =============================================================================================
+
+MENDELEEV_ELEMENTS_DATAFRAME = fetch_table("elements")
+MENDELEEV_ELEMENTS_DATAFRAME.set_index("atomic_number")
 
 # TODO: Can we have the `ALLOWED_*_MODELS` list automatically appear in the docstrings below?
 # TODO: Should `ALLOWED_*_MODELS` be objects instead of strings?
@@ -413,7 +419,9 @@ class Atom(Particle):
         -------
         mendeleev.models.Element
         """
-        return element(self._atomic_number)
+        return getattr(
+            mendeleev, MENDELEEV_ELEMENTS_DATAFRAME.loc[self.atomic_number, "symbol"]
+        )
 
     @property
     def atomic_number(self):
@@ -5906,7 +5914,7 @@ class FrozenMolecule(Serializable):
         connectivity = [
             (bond.atom1_index, bond.atom2_index, bond.bond_order) for bond in self.bonds
         ]
-        symbols = [element(atom.atomic_number).symbol for atom in self.atoms]
+        symbols = [atom.element.symbol for atom in self.atoms]
         if extras is not None:
             extras[
                 "canonical_isomeric_explicit_hydrogen_mapped_smiles"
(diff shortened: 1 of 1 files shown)


lilyminium:
Would a dictionary be simpler and easier? 

e.g.
```python
ATOMIC_NUMBERS_TO_ELEMENTS = {
    row.atomic_number: getattr(mendeleev, row.symbol)
    for row in MENDELEEV_ELEMENTS_DATAFRAME.itertuples()
}
```

...
```python
    @property
    def element(self):
        """
        The element of this atom.

        Returns
        -------
        mendeleev.models.Element
        """
        return ATOMIC_NUMBERS_TO_ELEMENTS[self.atomic_number]
```
...

```python
    atom_symbol_counts = Counter(ATOMIC_NUMBERS_TO_ELEMENTS[atom_num].symbol
                                 for atom_num in atom_nums)
```

I also find that it's minorly faster for me, but on the order of 0.01s for 316 molecules.

richardjgowers:
A dataframe seems like overengineering, a dict (i.e. this is a hashtable problem) is fine?

mattwthompson (approved):
FTR the regression was because of L7253

```
Counter(element(atom_num).symbol for atom_num in atom_nums)
```

from yours truly, which did a literally wasteful conversion from atomic number to symbol for every atom.

These changes are pure improvements that I expect to fix a lot of poor performance that's been slowing down tests in Interchange. Based on our working session yesterday, I don't think it can get much quicker than this as long as the `element` property returns a `mendeleev` object.

I had some more thoughts on the approach here, which I've split out into https://github.com/openforcefield/openff-toolkit/issues/1178 so as to not permanently prevent this from being merged at all. (This patch is definitely a "good" in the "don't let the perfect be the enemy of the good" and merging it will make my 0-timestep system of 1000 argon molecules not take 10 minutes to run.)

j-wags:
> Would a dictionary be simpler and easier?

Yup! That's a good idea.

> A dataframe seems like overengineering, a dict (i.e. this is a hashtable problem) is fine?

I agree, dataframes are certainly near the end of the list of tools I'd reach for in this case. Is there a performant way to get a dict out of mendeleev without going through a dataframe? I would greatly prefer that.

I guess my bigger concern here is having all this be up at the `molecule` module level, because 1) it seems messy to run code at the module level to load the elements and process them into a different form (like a dict), and 2) that could make it user-discoverable, and then we've expanded the API (though this will be hard to avoid). 

So I'd be interested in learning how to solve this problem in a "standard" way. Some options I see are to:
* Turn it into its own module in the Toolkit
* Incorporate this into openff-units
* Use something other than mendeleev whose API offers faster direct lookups

j-wags:
Ok -- I think this isn't perfect (like, we may choose not to use mendeleev at all in the long run), but this gets the branch to a usable state where runtime stops being a hindrance for other development. The only other failing test is already fixed in #1174 and so will turn green when that's merged.

PR #1179, opened 2022-01-20
[WIP] Private TypedMolecule class for Interchange development 

j-wags:
This PR introduces a prototype of the "TypedMolecule" class to keep Interchange from needing to continue supporting/refactoring the "mdtop" class, and allow it to begin working directly with the OFF toolkit codebase. This is important to maintain the Foyer interface and not do fully duplicate work. 

Our current thinking is to keep these classes private for the 0.11.0 release, just using them as a way to unblock Interchange development. If we don't support these here, we'll have to continue using one-off hacks for developing gromacs/amber imports, which we will eventually have to fully REdo when we support them natively in the toolkit. 

Using this as temporary scratch space for some simple tests:
```python
import mbuild as mb
from openff.units import unit
from mendeleev import element


def _compound_to_typed_molecule(compound: mb.Compound) -> TypedMolecule:
    typed_molecule = TypedMolecule()
    id2index = dict()
    for index, particle in enumerate(compound.particles()):
        typed_molecule.add_atom(
            #TypedAtom(
                atomic_number=element(particle.name).atomic_number,
                symbol=particle.name,
            #)
        )
        id2index[particle] = index

    for bond in compound.bonds():
        atom1_index = id2index[bond[0]]
        atom2_index = id2index[bond[1]]
        typed_molecule.add_bond(
            atom1_index, atom2_index
        )
    return typed_molecule
        

def _typed_molecule_to_compound(typed_molecule: TypedMolecule) -> mb.Compound:
    compound = mb.Compound()
    for atom in typed_molecule.atoms:
        compound.add(
            mb.Compound(name=atom.symbol)
        )

    for bond in typed_molecule.bonds:
        compound.add_bond((compound[bond.atom1_index], compound[bond.atom2_index]))

    if typed_molecule.n_conformers > 0:
        compound.xyz = typed_molecule.conformers[0].m_as(unit.nanometer)
    
    return compound

def test_compound_to_typed_molecule():
    compound = mb.load("CCO", smiles=True)
    typed_molecule = _compound_to_typed_molecule(compound)

    assert typed_molecule.n_atoms == 9, typed_molecule.n_atoms
    assert typed_molecule.n_bonds == 8

    assert [atom.atomic_number for atom in typed_molecule.atoms] == [6, 6, 8, 1, 1, 1, 1, 1, 1]
    
def test_typed_molecule_compound_roundtrip():
    roundtrip = _typed_molecule_to_compound(
        _compound_to_typed_molecule(mb.load("CCO", smiles=True))
    )
    
    assert roundtrip.n_particles == 9
    assert roundtrip.n_bonds == 8, roundtrip.n_bonds

test_compound_to_typed_molecule()
test_typed_molecule_compound_roundtrip()
```

```python
from openff.toolkit.topology.molecule import Molecule

tm = TypedMolecule()
tm.add_atom(atomic_number=8)
tm.add_atom(atomic_number=1)
tm.add_atom(atomic_number=1, name='bob')
tm.add_bond(0,1,name='sam')
tm.add_bond(0,2,order=1)

tm_dict = tm.to_dict()
print(tm_dict)
tm2 = TypedMolecule.from_dict(tm_dict)

assert tm.n_atoms == tm2.n_atoms
assert tm.n_bonds == tm2.n_bonds

off_mol = Molecule.from_smiles('CCO')
off_top = off_mol.to_topology()
off_top.add_molecule(tm2)

assert off_top.n_molecules == 2
assert off_top.n_atoms == 12, off_top.n_atoms
assert off_top.n_particles == 12, off_top.n_particles
print(off_top.atom(1))
print(off_top.atom(11))
```


j-wags:
I feel like (7/10) strongly that we should keep this as `TypedX` over `SimpleX` - "Typed" could be interpreted as literally correct "Oh, it's an atom-typed atom, and a bond-typed bond", whereas "simple" may rub people the wrong way and hinder interoperability efforts.

mattwthompson:
The major point against `_TypedX` from my perspective is that it's not tied to being typed at all - storing i.e. Amber-style atom types is something that _could_ be enabled what's really happening here is that we're developing classes to store molecules with less chemical information than the main API. I'm also worried it implies that these classes **must** be used with molecules that have atom types assigned or that they are meant to be objects that store the result of atom-typing with SMIRNOFF - which is not something that the toolkit does.

I'm open to replacing "simple" with a synonym, but it's an adjective that I believe accurately captures what this functionality is supposed to do.

j-wags:
Ah, good point. That makes sense. I’ve thought of a few others here:

Custom molecule
Container molecule
External molecule
Basic molecule
Property molecule
Information molecule
Placeholder molecule

From this list, I think I would be most in favor of “container molecule “or
“external molecule”. What do you think?

mattwthompson:
Of that list I like Basic the most, it's the nearest synonym and most strongly communicates that the scope is smaller; of your top picks I'd prefer External

lilyminium:
Just dropping in to point out that Basic does have another meaning in chemistry, if that could be a concern.

openff/toolkit/topology/molecule.py
```diff
@@ -275,7 +275,7 @@ def add_bond(self, bond):
 
         Parameters
         ----------
-        bond: an openff.toolkit.topology.molecule.Bond
+        bond: an penff.toolkit.topology.molecule.Bond
```
jchodera, line 278: Typo?

mattwthompson: Yep, typing too quickly while moving around in files

mattwthompson:
I think the basic stuff is in place here, and I've been able to hack away using it. Since it's not going to land in the public API soon and I'm the only user, II think in general this is in a state that is ready for review.

The major shortcoming I've come across is the at ton of the `Topology` API trusts its molecules to have certain methods and properties at several low-level points, many of which are not implemented in `_SimpleMolecule`. I've chosen to refactor functionality into module-level private functions instead of try to base `_SimpleMolecule` off of `FrozenMolecule` for the simple fact that I don't want to need to track state everywhere in the current API; this approach allows for development with minimal or zero impact to existing functionality and duplicated code. I think if this was merged as-is or with fewer changes, I'd still be able to work on slowly adding these in while avoiding this branch drifting away from its upstream.

openff/toolkit/topology/mm_molecule.py
```diff
@@ -0,0 +1,475 @@
...
+
+    def to_networkx(self) -> "nx.Graph":
+        import networkx as nx
+
+        graph = nx.Graph()
+
+        for atom_index, atom in enumerate(self.atoms):
+            graph.add_node(
+                atom_index,
+                atomic_number=atom.atomic_number,
```
j-wags, line 176: We should attach any custom attributes that users have added to atoms or bonds to the nodes and edges. 

If possible, we should do the same in `from_networkx`

openff/toolkit/topology/mm_molecule.py
```diff
@@ -0,0 +1,475 @@
...
+        If any atoms are non-elemental, defined by having an atomic number of 0,
+        return INVALID instead.
+        """
+        return self.to_hill_formula()
+
+    def to_hill_formula(self) -> str:
+        atom_nums: List[int] = [atom.atomic_number for atom in self.atoms]
+
+        if min(atom_nums) <= 0:
+            return "INVALID"
```
j-wags, line 164: It may be good to have atomic_number 0 become an `X` in the resulting hill formula. If people are getting an error message about a molecule that can't even be represented in a hill formula, they need all the help they can get.

openff/toolkit/topology/mm_molecule.py
```diff
@@ -0,0 +1,475 @@
+"""
```
j-wags, line 1: Let's prepend this filename with an underscore to ensure that it is very private :-)

j-wags (commented):
I can't approve this because I'm technically the author, but I left a few non-blocking comments, and I approve merging this PR. I'm really excited about the functionality that this will unlock for Interchange. 

For future changes to this functionality, feel free to merge PRs without review. Just use your judgement that they don't touch any code paths that could arise from use of the Toolkit's public API.

openff/toolkit/topology/mm_molecule.py
mattwthompson: :+1: Added in 6b96338c263e08e14e4de3989d44d8582757d634 - this change does impact `Molecule.hill_formula`, but only in undefined behavior. If a user somehow gets a `Molecule` containing atoms with zero atomic number, there will almost surely be other issues that crop up earlier.

openff/toolkit/topology/mm_molecule.py
mattwthompson: 👍  5833f2058d92f1dd72b7ae02b7586b67a8cb8aa2 / 67474c741f75adbce08d556a925edb5cde3493d2

openff/toolkit/topology/mm_molecule.py
mattwthompson: https://github.com/openforcefield/openff-toolkit/pull/1179/commits/a8329aa2515cc00a1efb869f03480e30fe078fcc

mattwthompson:
Thanks for the review - test failure is the persistent `openff/toolkit/tests/test_molecule.py::TestMolecule::test_chemical_environment_matches_OE` from upstream so I'm hitting the go button on this one.

PR #1180, opened 2022-01-25
Switch from Numpydoc to Napoleon for rendering numpy-style docstrings

Yoshanuikabundi:
Numpydoc seems to be broken at the moment. In Interchange and BespokeFit, we use Napoleon, which is built in to Sphinx, to do the same things. This this has one or two direct benefits, like adding links to some types specified in docstrings, but mostly it brings the Toolkit in line with our other projects and simplifies the build environment. It's also distributed with Sphinx, so is less likely to break when it updates.

This PR also includes some corrections and clean ups to docstrings. All of these are existing errors, not the result of any difference in rendering between numpydoc and napoleon.

### DIFF
--- a/openff/toolkit/typing/engines/smirnoff/forcefield.py
+++ b/openff/toolkit/typing/engines/smirnoff/forcefield.py
@@ -113,16 +113,16 @@ def get_available_force_fields(full_paths=False):
     ``openff-forcefields`` package is installed, this should include several
     .offxml files such as ``openff-1.0.0.offxml``\ .
 
-     Parameters
-     ----------
-     full_paths : bool, default=False
-         If False, return the name of each available \*.offxml file.
-         If True, return the full path to each available \*.offxml file.
+    Parameters
+    ----------
+    full_paths : bool, default=False
+        If False, return the name of each available \*.offxml file.
+        If True, return the full path to each available \*.offxml file.
 
-     Returns
-     -------
-     available_force_fields : List[str]
-         List of available force field files
+    Returns
+    -------
+    available_force_fields : List[str]
+        List of available force field files
 
     """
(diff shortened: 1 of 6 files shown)


openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -167,25 +167,9 @@ class ForceField:
     The force field definition is processed by these handlers to populate the ``ForceField`` object model data
     structures that can easily be manipulated via the API:
 
-    Processing a :class:`Topology` object defining a chemical system will then call all :class`ParameterHandler`
+    Processing a :class:`Topology` object defining a chemical system will then call all :class:`ParameterHandler`
     objects in an order guaranteed to satisfy the declared processing order constraints of each
-    :class`ParameterHandler`.
-
-    Attributes
```
j-wags, line 174: Wow. Great catch, this is actively misinforming users. Thanks for deleting it.

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1046,11 +1038,10 @@ def parse_smirnoff_from_source(self, source):
 
         Parameters
         ----------
-        source : str or bytes
-            sources : string or file-like object or open file handle or URL (or iterable of these)
-            A list of files defining the SMIRNOFF force field to be loaded
+        source : str or bytes or file-like object
+            File define the SMIRNOFF force field to be loaded
```
j-wags, line 1042:
```suggestion
            File defining the SMIRNOFF force field to be loaded
```

j-wags (approved):
Wonderful! Thanks @Yoshanuikabundi

PR #1182, opened 2022-01-27
Refactor to openff.units.elements

### LINKED ISSUE #1178: Element handling
The current refactor switches out element handling from OpenMM to [`mendeleev`](https://mendeleev.readthedocs.io/en/stable/index.html), checks off a lot of our boxes - it has all of the features we need (and more), it's actively developed and well-documented, and it's available on `conda-forge`. But it's also a little heavy at import time and with respect to simple conversions, i.e. atomic number -> element symbol. See https://github.com/openforcefield/openff-toolkit/pull/1177 for some timings after optimization. AFAICT
* The toolkit only needs to know the atomic number of an atomic, sometimes also needs the element symbol and/or mass, but nothing more
* The toolkit does not currently support isotopes and it's unclear if it will (see a discussion https://github.com/openforcefield/openff-toolkit/issues/974)
* Anything that a downstream user/library (would want to look up about an element can be done using the atomic number as a starting point.

There's an easy tendency to want to pull in big libraries to do relatively simple things, which can be a poor match between maintainability/packaging/import times and features added. My counter-proposals, either of which I'd be happy to contribute

1. Use a simpler, smaller element package, like [`ele`](https://github.com/rsdefever/ele). The author has recently left for industry, but the MoSDeF ecosystem uses it and I expect we'd be able to maintain it via contributions - or simply fork it and go from there.
2. Don't use an element package at all - just have a couple of dicts hanging around in some file, probably `{atomic_number: symbol}` and `{atomic_number: mass}` mappings. These would map between simple types (int, str, and int/quantity) and the toolkit would not explicitly have any objects representing elements. **It would also allow us to keep the same values reported by OpenMM without importing it**, whereas another unit package likely introduces small changes.

My only strong preference is to get away from using `openmm.app.element` - it's a module that has worked great for our purposes but I would ***love*** our typical conda-based users to not need to install the CUDA toolkit to parse a SMILES string. Switching to `mendeleev` satisfies this request, as does the other ideas above.

 Keep in mind we're already breaking the API of the `element` and `mass` properties, so if we wanted to make a big change, now is the time to consider it.

mattwthompson:
Resolves #1178 if accepted.

This PR proposes to not use an external elements package and further remove the concept of atoms containing elements. It seems that every use case can be covered by knowing only the atomic number. 

* `Atom.element` is removed 
* `Atom.element.symbol` becomes `Atom.symbol`
* `Atom.element.mass` becomes `Atom.mass`
* `Atom.element.atomic_number` probably should never have been used, but was also probably always in sync with `Atom.atomic_number`

Upsides:
* No new dependencies
* Quick import (~6 ms on my aging hardware), which could probably be optimized
* Fast (< 100 ns) atomic mass and symbol lookups

Downsides:
* Breaks the API (though we're allowed to for 0.11.0!),  which was already going to happen for `Atom.mass`
* Potentially feature-light for feature development (ambiguous)

Quick profiling script
```
import time
from openff.units import unit
start_time = time.time()
from openff.units.elements import MASSES, SYMBOLS
print(time.time() - start_time)
```

- [ ] Check examples - some are not passing on the refactor branch, not clear if they should be fixed here or later while prepping an RC build
- [x] Update [tests](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/tests)


devtools/conda-envs/test_env.yaml
```diff
@@ -49,3 +47,4 @@ dependencies:
     - types-toml
     - types-PyYAML
     - mongo-types
+    - git+https://github.com/openforcefield/openff-units.git@element-module
```
mattwthompson, line 50: This would be in a release if this PR is approved; I just don't want to make a release for an idea that might not see the light of day.

mattwthompson:
@j-wags it might take multiple iterations of reviews but it's ready for a first review - I hope it's material enough to make a decision on if this is a direction we want to proceed with.

One new test (`openff/toolkit/tests/test_molecule.py::TestMolecule::test_chemical_environment_matches_OE`) is failing for reasons that aren't obvious to me.

openff/toolkit/utils/_cif_to_substructure_dict.py
```diff
@@ -165,15 +166,22 @@ def _generate_atom_symbol_number_dictionary(cif_entry):
...
-            atomic_nums_list.append(element_.atomic_number)
-        # Create dictionary
-        symbol_to_num_dict = {
-            symbol: number for symbol, number in zip(atom_symbol_list, atomic_nums_list)
+        # Is this de-duplicated? i.e. can it be ["C", "C", "H"] or only ["C", "H"] ?
+        atom_symbol_list: List[str] = cif_entry["_chem_comp_atom.type_symbol"]
+
+        # Is this dict a de-duplicated mapping between element symbols and atomic numbers
+        # for the elements in this cif file? If so, can maybe do it in one step
+        """
```
j-wags, line 174: (not blocking) In the interest of time I'm not going to give feedback on this. I forgot exactly what this function expects and we don't have tests for this file yet. This is not expected to be run by users and so runtime isn't super critical as long as it's under an hour. If you're interested in optimizing/refactoring this, feel free to do so at your discretion. The best test I can suggest would be ensuring that the `openff/toolkit/utils/make_*_substructures.py` scripts yield the same outputs.

j-wags (approved):
Thanks, @mattwthompson. This looks really solid. I left a few comments, mostly non-blocking. The one blocking thing is probably a misunderstanding on my part, let's discuss at our check in tomorrow.

The only additional thing I'd request is a few sentences of description/migration guide in the releasenotes. This is the first mention of `openff-units` in this repo, so it'd be good to give some info (differences in getting and setting unit bearing quantities, how to adapt scripts if users are already using openmm units, what is openff units (with a link) and pint (with a link) )

devtools/conda-envs/test_env.yaml
mattwthompson: Thanks - did you want to review https://github.com/openforcefield/openff-units/pull/14, or should I merge + release? Given this PR is in good shape, I don't think it's required.

openff/toolkit/topology/molecule.py
```diff
@@ -2463,7 +2451,7 @@ def generate_unique_atom_names(self):
 
         element_counts = defaultdict(int)
         for atom in self.atoms:
-            symbol = atom.element.symbol
+            symbol = SYMBOLS[atom.atomic_number]
```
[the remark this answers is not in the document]
mattwthompson: Agree; this is probably a holdover from thinking about this step before realizing the `Atom.symbol` property exists.

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1306,7 +1306,8 @@ def create_openmm_system(self, topology, **kwargs):
         # This means that even though virtual sites may have been created via
         # the molecule API, an empty VirtualSites tag must exist in the FF
         for atom in topology.topology_atoms:
-            system.addParticle(atom.mass)
+            # addParticle(mass.m_as(unit.dalton)) would be safer but slower
```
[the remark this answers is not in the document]
mattwthompson: Surprisingly, there is no conversion to an OpenMM quantity; internally the quickest way through this method is to pass it an int. Here is a snippet from the SWIG-generated file I found in `.../site-packages/openmm/openmm.py`:


```
    def addParticle(self, mass):
        r"""
        addParticle(self, mass) -> int
        Add a particle to the System. If the mass is 0, Integrators will ignore the particle and not modify its position or velocity. This is most often used for virtual sites, but can also be used as a way to prevent a particle from moving.

        Parameters
        ----------
        mass : double
            the mass of the particle (in atomic mass units)

        Returns
        -------
        int
            the index of the particle that was added
        """

        if unit.is_quantity(mass):
            mass = mass.value_in_unit(unit.amu)


        return _openmm.System_addParticle(self, mass)
```

So, giving it something that looks like a number is fastest. If it [thinks it sees](https://github.com/openmm/openmm/blob/7.7.0/wrappers/python/openmm/unit/quantity.py#L787-L791) something wrapped with OpenMM units, it will do the necessary conversion.

```

In [1]: from openff.units import unit

In [2]: from openmm import System, unit as openmm_unit

In [3]: dummy_mass_openff = 1.0 * unit.dalton

In [4]: dummy_mass_openmm = 1.0 * openmm_unit.amu

In [5]: system = System()

In [6]: %timeit system.addParticle(mass=1.0)
592 ns ± 81 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [7]: %timeit system.addParticle(mass=dummy_mass_openff.m)
756 ns ± 136 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [8]: %timeit system.addParticle(mass=dummy_mass_openmm)
4.76 µs ± 1.46 µs per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [9]: %timeit system.addParticle(mass=dummy_mass_openff.m_as(unit.dalton))
17.2 µs ± 2.16 µs per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

`Quantity.m` is fast because it does no unit checking, it just reports the magnitude in whatever it's currently set as. Ensuring that the values are in Dalton is slower, but only so much slower than allowing OpenMM do the analogous check internally.

Of course, the problem here is dealing with other masses - not something I've ever run into, but it could be considered. The current implementation will **always** be in Daltons as long as we don't change [this](https://github.com/openforcefield/openff-units/blob/033f624d21095aada3f62f589e6ba8c90488e034/openff/units/elements.py#L29-L32) upstream. There is no setter on the `Atom` class, so I don't even know it could get to that state without other changes that themselves might significant behavior changes.

If we can't assume Daltons, I'm having a hard time figuring out any way around the time it takes to do this conversion. The only middle ground I can think of now is using the setter to ensure its units; this would safeguard against similar cases in which we wrote code without considering the non-Dalton case. It has the downside of pushing the cost elsewhere, of course.

In conclusion, I think the current approach is the best one (if accidentally - I did not think this all through while working on this). It assumes the units, but those are hardcoded in `openff-units`. It's also virtually as performant as possible, clocking in around ~750 ms for adding 1,000,000 atoms compared to ~600 ms if everything was already as floats and ~4700 ms with the current approach.

devtools/conda-envs/test_env.yaml
mattwthompson: merged & released!

openff/toolkit/utils/_cif_to_substructure_dict.py
mattwthompson: I'll not tinker with this more, then. Except maybe to suggest it shouldn't be in the API, and instead tucked away in another older?

mattwthompson:
The failing test is also failing upstream on #951

```
openff/toolkit/tests/test_molecule.py::TestMolecule::test_chemical_environment_matches_OE

**Comment by mattwthompson** [comment]
I also can't reproduce the collection errors in the "CI / Test on ubuntu-latest, Python 3.7, RDKit=false, OpenEye=true".

File: openff/toolkit/tests/test_molecule.py [hunk 5]
```diff
@@ -302,28 +302,29 @@ def test_atom_constructor(self):
...
-        # should be fixed when a new release is made
+        assert atom.atomic_number == atomic_number
+        # this API point might be removed?
         # assert atom.element == this_element
         assert atom.formal_charge == formal_charge
         assert atom.is_aromatic == is_aromatic
-        assert atom.name == this_element.name
-        assert atom.mass == this_element.mass
+        assert atom.symbol == expected_symbol
+        assert atom.mass == expected_mass
```
j-wags, line 326: Since our use of `atom.mass` below requires it to always be in Daltons/amus, let's add a test for that here.

openff/toolkit/typing/engines/smirnoff/forcefield.py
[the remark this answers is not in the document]
j-wags: Thanks for the detailed response. 10 us * 100,000 atoms = 1 second, so using explicit units will become a performance bottleneck at some point. 

As a compromise, how about we leave this as-is, but make it so that we guarantee that `Atom.mass` always returns values in daltons/amus? I've added a comment for where we could test that here: https://github.com/openforcefield/openff-toolkit/pull/1182/files#r803188577

Adding the test described above would resolve the blocker here, and you could feel free to merge. Optionally, it'd be great if you update the `atom.mass` docstring to echo the unit guarantee.

openff/toolkit/utils/_cif_to_substructure_dict.py
j-wags: I was under the impression that the underscore prepended to the file name communicated the privateness/here-be-dragons-ness of the contents. Though I'm not sure if that's a python convention in reality.

ijpulidos: As far as I remember when we first planned this, we didn't want to support a public infrastructure to do this, only just what's "enough" for the toolkit. I think I'm still in favor of this stance, since we know how this problem can be pretty complex. Also, I guess it's worth noticing that this `.py` file is expected to be used only once in a while. My 2 cents :). BTW, I'm loving how this is evolving, thanks for the good work!

lilyminium: > Though I'm not sure if that's a python convention in reality.

It is :) https://www.python.org/dev/peps/pep-0008/?#public-and-internal-interfaces

mattwthompson: I'd like to clarify a bit the distinction between two different things. I'm not sure the right jargon (happy to be told what the terminology is, I can't think of anything but "private API" and "scripts") so I'll resort to being descriptive:

1. Code that's for core toolkit functionality and might get called at runtime while interacting with the public API, but is either too dangerous to be explicitly exposed to users or meant to hide away implementation details. Stuff like `Molecule._add_trivalent_lone_pair_virtual_site` or `OpenEyeToolkitWrapper._find_smarts_matches`. These code paths have intrinsic value (in the sense that they **are** the product), **should** be covered by unit tests, will be included in conda tarballs, and fall under the umbrella of code we're responsible for maintaining/documenting/etc.
2. Code that's  for one-off data cleaning or developer-facing operations, but not a part of a core feature and will not be called by a user (directly or indirectly) at runtime. These files are valuable  in large part for what they produce (i.e.  big CIF file I don't understand the details of), but aren't covered by unit tests, may or may not be included in conda tarballs, and don't have the same maintenance burden.

What I'm getting at here is that code that cannot be accessed at runtime should be moved `/utilities/` or somewhere that would not be packaged as part of the Python module. That folder is from before my time, but it looks like those files didn't contribute as much to maintenance burden.

There's no danger to the user as is since it's definitely not part of the public API. But I think functions that should be used once in a while should not be in the core module (though no less important, of course).

openff/toolkit/tests/test_molecule.py
mattwthompson: This might be implicitly covered, but I've added an explicit check

mattwthompson:
There are two failures in CI, each of which I believe are unrelated to these changes here

1. `openff/toolkit/tests/test_molecule.py::TestMolecule::test_chemical_environment_matches_OE` - reported upstream, unclear
2. Tests with RDKit not installed and OpenEye installed are failing to collect the molecule fixtures. I can't reproduce this locally

I'm going to merge this and we can fix those later

PR #1185, opened 2022-01-30
Update ValenceDict

### LINKED ISSUE #1183: Remove length check in ValenceDict

ValenceDict has a number of restrictions that require the length of the key to be no more than 4 atoms. Is there a real reason for this?

Remove length check here:

https://github.com/openforcefield/openff-toolkit/blob/c32527bbdd98e5141c0f7a98c968bf0e514217b0/openff/toolkit/topology/topology.py#L94-L103

Refactor this to just take the forward or backward permutation. Also, it's much more idiomatic python to make this a classmethod instead of a staticmethod that uses `__class__`.

```permutations = OrderedDict({refkey: 0, refkey[::-1]:1})```

https://github.com/openforcefield/openff-toolkit/blob/c32527bbdd98e5141c0f7a98c968bf0e514217b0/openff/toolkit/topology/topology.py#L105-L165

**Additional context**

### PR DESCRIPTION
**Pull request by lilyminium**
Fixes #1183 

### CODE REVIEW

File: openff/toolkit/tests/test_topology.py [hunk 1]
```diff
@@ -1271,3 +1271,62 @@ def test_tagsorted_dict_clear(tsd):
...
+            [[0, 1], None, 0],
+            [[1, 0], None, 1],
+            [(1, 3, 2, 4), None, 0],
+            [(4, 2, 3, 1), None, 1],
+            [
+                (4, 2, 3, 1),
+                [
+                    (4, 2, 3, 1),
+                ],
+                0,
```
lilyminium, line 1303: This is how black wanted it -- I personally would not keep the code formatted like this, but I'll leave it to y'all

openff/toolkit/topology/topology.py
```diff
@@ -94,16 +94,13 @@ class ValenceDict(_TransformedDict):
     @staticmethod
     def key_transform(key):
         """Reverse tuple if first element is larger than last element."""
-        # Ensure key is a tuple.
         key = tuple(key)
-        assert len(key) > 0 and len(key) < 5, "Valence keys must be at most 4 atoms"
-        # Reverse the key if the first element is bigger than the last.
```
j-wags, line 100: (not blocking) This PR opts to delete a lot of comments, a few of which are indeed incorrect, but many of which are correct-but-only-marginally-useful. In the future I'd kinda appreciate having fewer marginally-useful comments removed.

openff/toolkit/tests/test_topology.py
j-wags: :shakes fist at uncompromising formatter: "You were supposed to bring peace!"

Yeah, that's ugly. Ah well, I'll take the good with the bad with `black`

j-wags (approved):
Cool - Thanks @lilyminium! I left some comments, but none are blocking, so this is good to merge as-is. Though if you have time it'd be great to also make similar classmethod/staticmethod changes in [`ImproperDict.index_of`](https://github.com/openforcefield/openff-toolkit/pull/1185/files#diff-b4190c8f5b6310de138566c74edfbd81dc7f7a6afbb172f0b688ea6800fb855aR295) (no need for me to re-review after that, you can merge those changes directly). But since static and class methods are so similar, that's not blocking to merging this PR.

The releasenotes look good. I kinda waffled on whether these are API breaking changes, but I don't think static <--> classmethod changes need to be reported, so they should be in good shape. 

Also, super nice tests - I'm still confused how the valencedict ordering problem never got caught, but I'm glad that this PR puts in tests to ensure they remain well behaved!

openff/toolkit/tests/test_topology.py
mattwthompson: Also not but one can split out the arguments into multiple decorators and the function will gobble everything up


```
import pytest

@pytest.mark.parametrize("x", [(1, 2)])
@pytest.mark.parametrize("y", [(1, 2)])
def test_foo(x, y):
    assert x == y

#     test_foo.py .                                                            [100%]
# 
# ============================== 1 passed in 0.06s ===============================

File: openff/toolkit/topology/topology.py (hunk 2 above)
Comment (lilyminium) [reply openff/toolkit/topology/topology.py:100]: I was always taught not to have low-level comments that are immediately obvious from the nearby code, like `Ensure key is a tuple` -- both because they introduce unneeded clutter and because they easily go out of date when the code is changed. I thought about leaving some of the comments in the "im/possible" section but then it seemed that most of them were covered in the docstring or associated error messages.

File: openff/toolkit/topology/topology.py [hunk 3]
```diff
@@ -125,57 +122,31 @@ def index_of(key, possible=None):
...
-                    (refkey[0], refkey[1], refkey[2]): 0,
-                    (refkey[2], refkey[1], refkey[0]): 1,
-                }
-            )
-        else:
-            # For a proper, only forward/backward makes sense
-            permutations = OrderedDict(
-                {
-                    (refkey[0], refkey[1], refkey[2], refkey[3]): 0,
-                    (refkey[3], refkey[1], refkey[2], refkey[0]): 1,
```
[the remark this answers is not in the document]
lilyminium: This method may not be used anywhere, to be honest. If you grep `index_of` the only instances seem to be associated with VirtualSites, which either uses and ImproperDict or the virtual site one.

mattwthompson:
Is there anything blocking this from being merged?

j-wags (approved):

PR #1187, opened 2022-01-31
Update pre-commit config and black

mattwthompson:
I recently scaled back pre-commit's autoupdater to quarterly since _most_ patch changes of linters/autoformatters are inconsequential for our codebase. This is not one such case; black changed its opinion on how the `**` power operator should look, resulting in a relatively large number of code changes. They [pinky promise](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy) changes will only happen once a year now that it's formally out of beta, so this is presumably our once annual `black openff` call.


openff/toolkit/topology/molecule.py
```diff
@@ -4586,7 +4586,8 @@ def from_iupac(
             molecule = toolkit.from_iupac(
                 iupac_name,
                 allow_undefined_stereo=allow_undefined_stereo,
-                _cls=cls ** kwargs,
+                _cls=cls,
+                **kwargs,
```
mattwthompson, line 4590: This ... probably wasn't working before.

mattwthompson: This is the only manual change I made (adding a comma).

j-wags: (not blocking) Oy vey. I hate this "registry or wrapper" pattern, so many untested cases! Let's never use this design pattern ever again.

openff/toolkit/utils/serialization.py
```diff
@@ -318,7 +318,7 @@ def from_yaml(cls, serialized):
         yaml.SafeDumper.add_representer(
             OrderedDict,
             lambda dumper, value: self._represent_odict(
-                dumper, u"tag:yaml.org,2002:map", value
+                dumper, "tag:yaml.org,2002:map", value
```
j-wags, line 321: (not blocking) this diff looks scary but it just seems to be cosmetic https://stackoverflow.com/questions/2464959/whats-the-u-prefix-in-a-python-string

j-wags (approved):
LGTM! Thanks @mattwthompson

openff/toolkit/utils/serialization.py
mattwthompson: I was a little spooked by this as well

PR #1194, opened 2022-02-09
Add `Topology.__add__`

mattwthompson:
I need this functionality and will use it whether or not it lives here. But pushing it upstream forces me to think about it a bit more carefully and actually write tests.


openff/toolkit/topology/topology.py
```diff
@@ -610,13 +644,46 @@ def is_periodic(self, is_periodic):
...
+        """
+        Get the partial charge model applied to all molecules in the topology.
+
+        Returns
+        -------
+        charge_model : str
+            Charge model used for all molecules in the Topology.
+
+        """
+        return self._charge_model
```
j-wags, line 657: Sorry, I may have screwed something up in the merge conflict resolution, because I can't reason around this part of the diff. Could you remake this branch/PR? This is totally my fault for taking so long that this fell into conflict.

mattwthompson: This is a red herring, it was removed in #1213 and founds its way back via some merge

mattwthompson: Did some fiddling with my local branch, things are a bit cleaner now.

openff/toolkit/topology/topology.py
```diff
@@ -441,6 +444,37 @@ def _initialize(self):
...
+
+        """
+        if self.aromaticity_model != other.aromaticity_model:
+            raise InvalidAromaticityModelError(
+                "Mismatch in aromaticity models. Trying to add a Topology with aromaticity model "
+                f"{other.aromaticity_model} to a Topology with aromaticity model "
+                f"{self.aromaticity_model}"
+            )
+
+        self._constrained_atom_pairs = dict()
```
j-wags, line 463: (blocking) Hmmm, after some thought I think that resetting constraints is a big footgun waiting to go off. While I'd LIKE to think that constraints live entirely in post-parameterization-land (and therefore wouldn't be important to keep track of) we DO currently offer some guarantee that a constraint defined directly on the Topology before parameterization will [be put into the resulting system](https://github.com/openforcefield/openff-toolkit/blob/921fbe2d48f3069adc3af6e919349dee6deefa26/openff/toolkit/typing/engines/smirnoff/parameters.py#L2841-L2856). So there could be workflows out there that define constraints ahead of time, and this would be a silent total break for them.

So, for this PR to be accepted, it would need to keep track of the constraints during topology addition. While the `+` operator should be magic, it shouldn't be quite _this_ magic.

openff/toolkit/topology/topology.py
j-wags: Thanks! This is much more readable now.

openff/toolkit/topology/topology.py
mattwthompson: Added in https://github.com/openforcefield/openff-toolkit/pull/1194/commits/3452b2d137f065521b4d5ca228bc5be3936944cd

j-wags (approved):
LGTM! Thanks @mattwthompson

PR #1198, opened 2022-02-24
Port #896 to biopolymer-topology-refactor branch

mattwthompson:
This PR ports https://github.com/openforcefield/openff-toolkit/pull/896 (by @SimonBoothroyd) to the "refactor" branch according to my proposal [here](https://github.com/openforcefield/openff-toolkit/pull/896), thereby closing it.

Two modifications include
1. Changes after clarifying that `switch_width` is the difference between the "switching distance" (i.e. `setSwitchingDistance`) and cutoff distance. For example, typical values are 9 Angstrom cutoff, 1 Angstrom switch with, and therefore an 8 Angstrom switching distance.
2. Raising an exception (`InvalidSwitchingDistanceError`) if the switch width is greater than the switch distance, which is not a sensible pairing of parameters.

### DIFF
--- a/openff/toolkit/tests/test_forcefield.py
+++ b/openff/toolkit/tests/test_forcefield.py
@@ -3647,6 +3647,12 @@ class TestForceFieldParameterAssignment:
         ff = ForceField(
             "test_forcefields/test_forcefield.offxml", off_gbsas[gbsa_model]
         )
+
+        # OpenMM 7.7 and older don't properly handle parsing prmtop files with a GBSA model and
+        # a switching function; this bug may be fixed but for testing purposes, simply turn off
+        # the switching function
+        ff["vdW"].switch_width = 0.0 * unit.angstrom
+
         off_top = molecule.to_topology()
         if is_periodic:
             off_top.box_vectors = (
@@ -3695,6 +3701,13 @@ class TestForceFieldParameterAssignment:
             amber_nb_method = openmm.app.forcefield.NoCutoff
             amber_cutoff = None
 
+        # OpenMM will process False to mean "don't use one" and a quantity as "use with this distance"
+        switching_distance: Union[
+            bool, openmm_unit.Quantity
+        ] = off_nonbonded_force.getUseSwitchingFunction()
+        if switching_distance:
+            switching_distance = off_nonbonded_force.getSwitchingDistance()
+
         (
             amber_omm_system,
             amber_omm_topology,
@@ -3707,6 +3720,7 @@ class TestForceFieldParameterAssignment:
             nonbondedCutoff=amber_cutoff,
             gbsaModel="ACE",
             implicitSolventKappa=0.0,
+            switchDistance=switching_distance,
         )
 
         # Retrieve the GBSAForce from both the AMBER and OpenForceField systems
@@ -3781,8 +3795,6 @@ class TestForceFieldParameterAssignment:
             off_nonbonded_force.setNonbondedMethod(openmm.NonbondedForce.NoCutoff)
 
         off_nonbonded_force.setReactionFieldDielectric(1.0)
-
-        # TODO: look into if switching distance is relevant #882
 
         # Create Contexts
         integrator = openmm.VerletIntegrator(1.0 * openmm_unit.femtoseconds)
--- a/openff/toolkit/typing/engines/smirnoff/parameters.py
+++ b/openff/toolkit/typing/engines/smirnoff/parameters.py
@@ -92,6 +92,7 @@ from openff.toolkit.utils.exceptions import (
     FractionalBondOrderInterpolationMethodUnsupportedError,
     IncompatibleParameterError,
     IncompatibleUnitError,
+    InvalidSwitchingDistanceError,
     MissingIndexedAttributeError,
     NonintegralMoleculeChargeException,
     NotEnoughPointsForInterpolationError,
@@ -3647,7 +3648,7 @@ class vdWHandler(_NonbondedHandler):
         """
         float_attrs_to_compare = ["scale12", "scale13", "scale14", "scale15"]
         string_attrs_to_compare = ["potential", "combining_rules", "method"]
-        unit_attrs_to_compare = ["cutoff"]
+        unit_attrs_to_compare = ["cutoff", "switch_width"]
 
         self._check_attributes_are_equal(
             other_handler,
@@ -3676,6 +3677,17 @@ class vdWHandler(_NonbondedHandler):
                 force.setCutoffDistance(to_openmm(self.cutoff))
                 force.setEwaldErrorTolerance(1.0e-4)
 
+                if self.switch_width >= self.cutoff:
+                    raise InvalidSwitchingDistanceError(
+                        "Attempted to generate a NonbondedForce with a switching width greater than the or equal to the cutoff "
+                        f"distance. Found switch_width={self.switch_width} and cutoff={self.cutoff}."
+                    )
+                else:
+                    force.setSwitchingDistance(
+                        to_openmm(self.cutoff - self.switch_width)
+                    )
+                    force.setUseSwitchingFunction(self.switch_width < self.cutoff)
+
         # If method is cutoff, then we currently support openMM's PME for periodic system and NoCutoff for nonperiodic
         elif self.method == "cutoff":
             # If we're given a nonperiodic box, we always set NoCutoff. Later we'll add support for CutoffNonPeriodic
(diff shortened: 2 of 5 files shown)


j-wags:
I approve making new [reference energies](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/data/reference_energies) using the code in this branch to get the tests passing.

mattwthompson:
Oh wow, I didn't see that GBSA tests are also failing ... those might take some more thinking. Presumably the switching function isn't applied in the Amber files, but I haven't looked into it nor am I as familiar with that  code as others.

j-wags:
Ohhh darn, I was too hasty. The reference energies thing won't fix these tests. We'll need to update [these lines](https://github.com/openforcefield/openff-toolkit/blob/485414eeb2f45784abfb5fbb44218d3dc702923d/openff/toolkit/tests/test_forcefield.py#L3710-L3718) to feed in new cutoff/switching kwargs OpenMM's [AmberPrmtopFile.createSystem](http://docs.openmm.org/development/api-python/generated/openmm.app.amberprmtopfile.AmberPrmtopFile.html?highlight=prmtop) call

mattwthompson:
Well, I tried to thread that through, but ran up against it not being supported and/or valid (I don't yet have the familiarity with GBSA to say which):

```
>>> from openmm import GBSAOBCForce
>>> x = GBSAOBCForce()
>>> x.getUseSwitchingFunction()
AttributeError: 'GBSAOBCForce' object has no attribute 'getUseSwitchingFunction'
>>> dir(x)
['CutoffNonPeriodic', 'CutoffPeriodic', 'NoCutoff', '__class__', '__copy__', '__deepcopy__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setstate__', '__sizeof__', '__str__', '__subclasshook__', '__swig_destroy__', '__weakref__', 'addParticle', 'getCutoffDistance', 'getForceGroup', 'getName', 'getNonbondedMethod', 'getNumParticles', 'getParticleParameters', 'getSoluteDielectric', 'getSolventDielectric', 'getSurfaceAreaEnergy', 'setCutoffDistance', 'setForceGroup', 'setName', 'setNonbondedMethod', 'setParticleParameters', 'setSoluteDielectric', 'setSolventDielectric', 'setSurfaceAreaEnergy', 'this', 'thisown', 'updateParametersInContext', 'usesPeriodicBoundaryConditions']
```

More verbose demo here: https://gist.github.com/mattwthompson/77df8268933462d6dcfa100d3bf9f2aa

mattwthompson:
Might be an actual bug in OpenMM (https://github.com/openmm/openmm/issues/3488). Whether or not it is, pinning to a yet-to-be-released version of OpenMM is not tenable, so I might have to add some logic to work around this.

openff/toolkit/tests/test_forcefield.py
```diff
@@ -3782,8 +3796,6 @@ def test_freesolv_gbsa_energies(
 
         off_nonbonded_force.setReactionFieldDielectric(1.0)
 
-        # TODO: look into if switching distance is relevant #882
-
```
mattwthompson, line 3786: It does - unsurprisingly. I'm not interested in documenting it but several tests have failed energy comparisons  locally. When the switching was applied in the `System` created by the toolkit but not in the one that went through Amber files, the energies differed.

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -3676,6 +3677,17 @@ def create_force(self, system, topology, **kwargs):
                 force.setCutoffDistance(to_openmm(self.cutoff))
                 force.setEwaldErrorTolerance(1.0e-4)
 
+                if self.switch_width >= self.cutoff:
```
mattwthompson, line 3680: This case .... doesn't make much as much sense once the interpretation of `switch_width` matches up with the intended meaning. I'll have to update this logic and tests.

mattwthompson:
I decided against including every possible safeguards in either the `switch_width` validator or the lines that actually set it on the force. If somebody wants to muck around with a negative switch width, a switch width greater than the cutoff, etc., there should be some alarm bells going off at other steps such that it's easy to diagnose. With updated wording in the SMIRNOFF spec coming soon, sane defaults written in the source code of the parameter handler (9 A cutoff, 1 A switch width), and the high likelihood that users will simply copy+paste whatever is in our mainline force fields (also 9 A, 1 A) I think this ranks low on potential points of failure to be concerned about.


Barring surprise results in CI, these changes to not require updating the (minimal) reference energies in this repo. It's not too surprising, since they only cover systems of singe small molecules. So the behavior is only tested via the unit tests.

To get the GBSA tests online, I decided to go with an imperfect but low-friction solution. I simply turned off the switching functions in those tests and _only_ those tests. The failures were due to OpenMM's `AmberPrmtopFile` not being able to create a system with a switching function on its non-bonded force and an implicit solvent model [1]. Note that this **only** happens with `AmberPrmtopFile.createSystem` and does not seem to be a functional limitation of the systems created with `ForceField.create_openmm_system`. Other solutions are more involved (switching to ParmEd inside of this non-API function) or induce headache timebombs (hardcoding different behavior based on the version of OpenMM installed [2].


Now, the only failing tests is our good friend `test_chemical_environment_matches_OE`.

----
1. I've submitted a bugfix to OpenMM to fix this, but this PR **does not** depend on that.
2. Eventually I want to move many of these tests into Interchange / deleted from here - they constitute a large portion of some of the larger files in this codebase, many are skipped, and they take longer than most of the other unit tests.

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -3706,6 +3712,15 @@ def create_force(self, system, topology, **kwargs):
             atom_matches, topology, [(atom,) for atom in topology.atoms]
         )
 
+    def _apply_switching_function(self, force: "openmm.NonbondedForce"):
```
j-wags: I'm not smart enough to really follow the type annotation here, or how it related to the `if TYPE_CHECKING: import openmm` line above, so maybe we could remove this. It's not a techncal issue, I'm just not smart enough to follow, and that means that average contributors may not be smart enough either.

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -3647,7 +3650,7 @@ def check_handler_compatibility(self, other_handler):
         """
         float_attrs_to_compare = ["scale12", "scale13", "scale14", "scale15"]
         string_attrs_to_compare = ["potential", "combining_rules", "method"]
-        unit_attrs_to_compare = ["cutoff"]
+        unit_attrs_to_compare = ["cutoff", "switch_width"]
```
j-wags, line 3650: Woah, nice catch.

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -3706,6 +3712,15 @@ def create_force(self, system, topology, **kwargs):
             atom_matches, topology, [(atom,) for atom in topology.atoms]
         )
 
+    def _apply_switching_function(self, force: "openmm.NonbondedForce"):
+        """Apply a switching function to a NonbondedForce if self.switch_width is nonzero."""
+        from openff.units.openmm import to_openmm
+
```
j-wags, line 3719: Per our extensive discussion, it's nonsensical to have a switch width if we're using LJPME, since it's not used anywhere. But it's also not clear that it should be an error since the SMIRNOFF spec doesn't explicitly forbid it. Let's make a note here:
```suggestion
        # TODO: Should this raise an error if switch_width is nonzero but we're using LJPME?
```

j-wags (approved):
Thanks for the great PR, @mattwthompson. We talked a lot about whether we should raise an error for trying to set a `switch_width` for LJPME, but ultimately I don't think it's worth changing the behavior to raise an error or do anything else. With regard to what Interchange should do in that case, let's discuss that in the context of the "interchange as export machinery" project plan and approval. 

CI-wise, feel free to ignore the failing OpenEye test - That one's on me. However the failing OpenMM tests on Mac ARE blocking to merging this.

openff/toolkit/typing/engines/smirnoff/parameters.py
mattwthompson: Added comments at L3690 in https://github.com/openforcefield/openff-toolkit/pull/1198/commits/fdf37159c1eaeea703ce885e35810c47e5f25c51

Issue #1199, opened 2022-02-25
Water LibraryCharges aren't used in output OpenMM systems in Sage


lilyminium:
**Describe the bug**

Hi there, I'm not certain if this is a bug, but I noticed it when I was trying to do stuff with water. It looks like water LibraryCharges are shipped with OpenFF 2.0.0, but it doesn't look like they get used when creating an OpenMM system? I have no clue why because the smirks pattern matches with `chemical_environment_matches`, unless there's a special case exception -- and in that case I don't understand why it's in the force field.

**To Reproduce**

```python
In [2]: # %load mwe.py
   ...: import openff.toolkit
   ...: from openff.toolkit.typing.engines.smirnoff import ForceField
   ...: from openff.toolkit.topology import Molecule
   ...:
   ...: print(openff.toolkit.__version__)
   ...:
   ...: ff = ForceField("openff-2.0.0.offxml")
   ...: handler = ff.get_parameter_handler("LibraryCharges")
   ...: water_parameters = handler.parameters[-2:]
   ...: print("water parameters:")
   ...: print(water_parameters)
   ...:
   ...: water = Molecule.from_smiles("O")
   ...:
   ...: print("Matches")
   ...: for parameter in water_parameters:
   ...:     print(water.chemical_environment_matches(parameter.smirks))
   ...:
   ...: system = ff.create_openmm_system(water.to_topology())
   ...:
   ...: print("OpenMM charges")
   ...: nbforce = system.getForces()[0]
   ...: for i in range(3):
   ...:     print(water.atoms[i].element, nbforce.getParticleParameters(i)[0])
   ...:
   ...: print("AM1BCC charges")
   ...: water.assign_partial_charges("am1bcc")
   ...: print(water.partial_charges)
0.10.2+9.gc29c9359
water parameters:
[<LibraryChargeType with smirks: [#1]-[#8X2H2+0:1]-[#1]  charge1: -0.834 e  id: q-tip3p-O  >, <LibraryChargeType with smirks: [#1:1]-[#8X2H2+0]-[#1]  charge1: 0.417 e  id: q-tip3p-H  >]
Matches
[(0,), (0,)]
[(1,), (2,)]
Warning: No principle axes found during inertial alignment
OpenMM charges
<Element oxygen> -0.7850599884986877 e
<Element hydrogen> 0.39252999424934387 e
<Element hydrogen> 0.39252999424934387 e
AM1BCC charges
Warning: No principle axes found during inertial alignment
[...]

PR #1200, opened 2022-02-25
Fix incorrect assignment of TIP3P parameters in Sage

mattwthompson:
#1199 

### DIFF
--- a/openff/toolkit/tests/test_forcefield.py
+++ b/openff/toolkit/tests/test_forcefield.py
@@ -3274,6 +3274,21 @@ class TestForceFieldChargeAssignment:
 
         compare_partial_charges(using_kwarg, using_library_charges)
 
+    def test_sage_tip3p_charges(self):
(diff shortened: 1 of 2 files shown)


j-wags:
Thanks for the quick action, @mattwthompson! Seeing how quickly unintended side effects can spread, I'm going to take a few minutes to review this carefully.

j-wags (approved):
Thanks for opening this, @mattwthompson 

I made some direct changes and am pretty sure that the current state is cleaner - looking around, we already had two tests for TIP3P LibraryCharge application, it would have been a bit much to add a third :-) 

I think this is ready to merge - Could you update the release notes? Then we can cut 0.10.3 on Monday.

mattwthompson:
I definitely wrote them in the first set of changes ... but they're also not there, so my git foo must have botched that. Oh well.

I'm not going to dwell on the text of the change here and delay hitting the green button here - we can revise while making the release.

PR #1204, opened 2022-03-02
Remove some internal calls to deprecated API points

mattwthompson:
I noticed while working on downstream tests that several deprecated API points are called internally to the toolkit, mostly stuff like `topology_molecules`, `n_topology_molecules`, etc. This PR removes some of them, but probably not all.

### DIFF
--- a/openff/toolkit/typing/engines/smirnoff/parameters.py
+++ b/openff/toolkit/typing/engines/smirnoff/parameters.py
@@ -3461,7 +3461,7 @@ class _NonbondedHandler(ParameterHandler):
             system.addForce(force)
             # Create all atom particles. Virtual site particles are handled in
             # in its own handler
-            for _ in topology.topology_atoms:
+            for _ in topology.atoms:
                 force.addParticle(0.0, 1.0, 0.0)
         else:
             force = existing[0]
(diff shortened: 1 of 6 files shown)


j-wags (approved):
LGTM. Thanks for the cleanup. Before you merge, could you verify that we keep at least one test for all the `Topology.topology_X` and `.n_topology_X` methods? They should just throw a deprecation warning and redirect to the new api points, so I'd like to ensure that they remain tested.

mattwthompson:
Good idea - I was too focused on not getting deprecation warnings downstream I didn't realize we want to ensure we have them here.

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -4693,10 +4689,10 @@ def create_force(self, system, topology, **kwargs):
         # To keep it simple, we DO NOT pre-populate the particles in the GBSA force here.
         # We call addParticle further below instead.
         # These lines are commented out intentionally as an example of what NOT to do.
-        # for topology_particle in topology.topology_particles:
+        # for particle in topology.particles:
         # gbsa_force.addParticle([0.0, 1.0, 0.0])
 
-        params_to_add = [[] for _ in topology.topology_particles]
+        params_to_add = [[] * topology.n_particles]
```
mattwthompson, line 4695: Interesting - I broke this because of a gotcha I'm un familiar with - multiply an empty list doesn't work in the same as as multiplying one with contents:

```
ipdb> [[] for x in range(2)]
[[], []]
ipdb> [[] * 2]
[[]]
ipdb> [['ok'] for x in range(2)]
[['ok'], ['ok']]
ipdb> [['ok'] * 2]
[['ok', 'ok']]

PR #1205, opened 2022-03-03
Debug macOS failures in CI

mattwthompson:


lilyminium:
Could be this: https://github.com/openmm/openmm/issues/3495

mattwthompson:
MDAnalysis has a more elegant solution in pinning precisely where it's needed: https://github.com/MDAnalysis/mdanalysis/pull/3551 AFAIK there's not a clean way to do this with our environment files without rewriting a bunch of the CI infrastructure, and I don't want to do that. At the same time, I don't want to wait for fixes to be coordinated and deployed upstream before having useful CI. So I'm pinning to version 7.6 while it's being worked on - a pin that should not noticeably impact development (see https://github.com/openmm/openmm/releases/tag/7.7.0) and **never make it into a release build**.

PR #1207, opened 2022-03-04
Update unit checks in Topology.to_file

mattwthompson:
Today I needed to write a PDB file from a `Topology` and I found that the docstring, behavior, and tests don't really line up. This is probably my fault, being the last person to touch this code.

```python3
>>> from openff.toolkit import Topology
>>> print(Topology.to_file.__doc__)

        Save coordinates and topology to a PDB file.

        Reference: https://github.com/openforcefield/openff-toolkit/issues/502

        Notes:

        1. This doesn't handle virtual sites (they're ignored)
        2. Atom numbering may not remain same, for example if the atoms
           in water are numbered as 1001, 1002, 1003, they would change
           to 1, 2, 3. This doesn't affect the topology or coordinates or
           atom-ordering in any way.
        3. Same issue with the amino acid names in the pdb file, they are
           not returned.

        Parameters
        ----------
        filename : str
            name of the pdb file to write to
        positions : n_atoms x 3 numpy array or openmm.unit.Quantity-wrapped n_atoms x 3 iterable
            Can be an openmm 'quantity' object which has atomic positions as a list of Vec3s along with associated units, otherwise a 3D array of UNITLESS numbers are considered as "Angstroms" by default
        file_format : str
            Output file format. Case insensitive. Currently only supported value is "pdb".
```
The docstring doesn't  indicate that `openff.units.unit.Quantity` is accepted, and the logic of how it's handled is a little dangerous (again from yours truly). In particular, there is no logic that handles positions tagged with our Pint-style units, so it's automatically stripped of its units when it gets to `PDBFile.writeFile`. This silently works if they're in Angstroms but will silently break if they're in nanometers.

This PR updates the behavior to explicitly and more safely check the type of the `positions` input and tests some of the allowed use cases.

### DIFF
--- a/openff/toolkit/tests/test_topology.py
+++ b/openff/toolkit/tests/test_topology.py
@@ -763,10 +763,15 @@ class TestTopology:
     @requires_rdkit
     def test_to_file_units_check(self):
         """
-        Checks whether writing pdb with unitless positions, Angstrom positions,
-        nanometer positions, result in the same output
+        Checks that writing a PDB file with different coordinate representations results in the same output.
+        - Angstrom "openff units" (default behavior if using Molecule.conformers[0])
+        - nanometer "openff units"
+        - unitless NumPy array
+        - converted OpenMM quantity
         """
         from tempfile import NamedTemporaryFile
+
+        from openff.units.openmm import to_openmm
 
         from openff.toolkit.topology import Molecule, Topology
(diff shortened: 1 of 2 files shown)


j-wags (approved):
Looks great. Thanks, @mattwthompson!

PR #1211, opened 2022-03-07
Fixes #1206.

### LINKED ISSUE #1206: BondCharge virtual sites are not overriding correctly
**Describe the bug**
SMIRNOFF hierarchy rules are somewhat complicated for virtualsites. A virtualsite should only override another virtualsite if they match the same atoms[1], and have the same name and type. 

**To Reproduce**

This code makes two vsites with the same name and type, which should apply to the same atoms, but in a different order:

```
from openff.units import unit
from openmm import unit as omm_unit

mol = Molecule.from_mapped_smiles('[Cl:1][Na:2]')
ff = ForceField()
vsh = ff.get_parameter_handler('VirtualSites', {"version":0.3})

# Define two vsites that apply in opposite directions to the same atoms of the test molecule. 
# The vsites may collide (since they have the same name), but they should be 
# distinguishable in the output based on their `distance` values
vsh.add_parameter({'type':'BondCharge', 
                      'smirks':'[Cl:1][Na:2]',
                      'match':'once',
                      'charge_increment':[0.100, -0.100] * unit.elementary_charge,
                      'distance': [0.1010] * unit.angstrom,
                      'name':'aaa'})
vsh.add_parameter({'type':'BondCharge', 
                      'smirks':'[Na:1][Cl:2]',
                      'match':'once',
                      'charge_increment':[0.100, -0.100] * unit.elementary_charge,
                      'distance': [0.220] * unit.angstrom,
                      'name':'aaa'})

sys = ff.create_openmm_system(mol.to_topology())

# For each vsite, print out which particles are involved, and at which distance.
for vsite_idx in range(sys.getNumParticles()):
    if not(sys.isVirtualSite(vsite_idx)):
        continue
    vsite = sys.getVirtualSite(vsite_idx)
    vsite_parents = []
    for vsite_parent_idx in range(vsite.getNumParticles()):
        vsite_parent = vsite.getParticle(vsite_parent_idx)
        vsite_parents.append(vsite_parent)
    distance = vsite.getLocalPosition()[0].value_in_unit(omm_unit.angstrom)
    print(vsite_parents, distance)
```

> [0, 1] -0.101

The specific problems shown in this code are:
1) The first vsite applies to atoms (0,1), and the second applies to atoms (1,0). Since the atoms are in a different order, I don't think that one should overwrite the other, and so I'd expect both vsites should be in the output[1]. Instead only a single vsite is in the output.
2) If the second virtualsite (with distance `0.220`) MIGHT overwrite the first (with distance `0.101`), then I expect that the output should EITHER have one vsite with each distance, OR one vsite with `0.220`. Instead we get just one vsite with `0.101`.

[1] It's not specifically mentioned in the SMIRNOFF spec, but I assume this means "one vsite will overwrite another if they match the same atoms _in the same order_"

 - Mac OS
 - OFFTK topology biopolymer refactor branch

trevorgokey:
- [x] Fixes #1206

### DIFF
--- a/openff/toolkit/typing/engines/smirnoff/parameters.py
+++ b/openff/toolkit/typing/engines/smirnoff/parameters.py
@@ -5228,7 +5228,7 @@ class VirtualSiteHandler(_NonbondedHandler):
         def __eq__(self, obj):
             if type(self) != type(obj):
                 return False
-            A = ["name"]
+            A = ["name", "smirks"]
             are_equal = [getattr(self, a) == getattr(obj, a) for a in A]
             return all(are_equal)
 
@@ -5626,23 +5626,43 @@ class VirtualSiteHandler(_NonbondedHandler):
                         for new_match in v:
                             unique = True
                             new_item = new_match._parameter_type
-                            for idx, (name, existing_match) in enumerate(
-                                matches[k].items()
-                            ):
-                                existing_item = existing_match._parameter_type
-                                same_parameter = False
+                            for existing_indices in matches:
+                                # only go over existing indices which are a
+                                # permutation of our current match
+                                if set(k).difference(existing_indices):
+                                    continue
+                                # the goal here is to go over all current match
+                                # indices, and potentially clear them out if
+                                # our new parameter is a match
+                                for idx, (name, existing_match) in enumerate(
+                                    list(matches[existing_indices].items())
+                                ):
+                                    existing_item = existing_match._parameter_type
+                                    same_parameter = False
 
-                                same_type = type(existing_item) == type(new_item)
-                                if same_type:
-                                    same_parameter = existing_item == new_item
+                                    same_type = type(existing_item) == type(new_item)
+                                    if same_type:
+                                        same_parameter = existing_item == new_item
 
-                                # same, so replace it to have a FIFO priority
-                                # and the last parameter matching wins
-                                if same_parameter:
-                                    matches[k][new_item.name] = new_match
-                                    unique = False
+                                    # same, so replace it to have a FIFO priority
+                                    # and the last parameter matching wins
+                                    if same_parameter:
+                                        matches[k][new_item.name] = new_match
+                                        unique = False
+                                    else:
+                                        # If we are here, we have the same
+                                        # set of indices as the new match,
+                                        # but the parameter is not the same.
+                                        # In this case, we need to clear out
+                                        # any existing vsites, since this
+                                        # new incoming vsite needs to replace
+                                        # them
+                                        matches[existing_indices].pop(
+                                            existing_item.name
+                                        )
                             if unique:
                                 marginal_matches.append(new_match)
+
                         matches[k].update(
                             {p._parameter_type.name: p for p in marginal_matches}
                         )


j-wags:
Thanks for taking a swing at this @trevorgokey!

The failing test here seems to be significant - It applies [this ff](https://github.com/openforcefield/openff-toolkit/blob/fa03d80a6e568fd95edda05aacf23f8b195bac26/openff/toolkit/tests/test_forcefield.py#L566-L596) to an N2 molecule, but the [charge on each N ends up being 0.2 (indicating that just one vsite was assigned) instead of the expected 0.3 (if both vsites were applied)](https://github.com/openforcefield/openff-toolkit/runs/5456639996?check_suite_focus=true#step:14:429)

I'd be happy to review+merge this once it gets to a passing state, but there are a few (looking like maybe 2) other vsite issues I'm finding that may be advantageous to fix in a single PR. If you can wait a day or two I can write those up and publish the more thorough tests.

j-wags:
(misclicked, didn't mean to close this!)

trevorgokey:
That was my mistake in how I was clearing out the old sites, I was working on the fix while you were looking over. I think my modification should get everything going, but will tinker some more if something else catches on fire.

j-wags:
Unfortunately, I think this still fails the case in #1206. While the correct distance is now assigned, it seems to put the parent atoms in the wrong order.

```python
from openff.units import unit
from openmm import unit as omm_unit

mol = Molecule.from_mapped_smiles('[Cl:1][Na:2]')
ff = ForceField()
vsh = ff.get_parameter_handler('VirtualSites', {"version":0.3})

# Define two vsites that apply in opposite directions to the same atoms of the test molecule. 
# The vsites may collide (since they have the same name), but they should be 
# distinguishable in the output based on their `distance` values
vsh.add_parameter({'type':'BondCharge', 
                      'smirks':'[Cl:1][Na:2]',
                      'match':'once',
                      'charge_increment':[0.100, -0.100] * unit.elementary_charge,
                      'distance': [0.1010] * unit.angstrom,
                      'name':'aaa'})
vsh.add_parameter({'type':'BondCharge', 
                      'smirks':'[Na:1][Cl:2]',
                      'match':'once',
                      'charge_increment':[0.100, -0.100] * unit.elementary_charge,
                      'distance': [0.220] * unit.angstrom,
                      'name':'aaa'})

sys = ff.create_openmm_system(mol.to_topology())

# For each vsite, print out which particles are involved, and at which distance.
for vsite_idx in range(sys.getNumParticles()):
    if not(sys.isVirtualSite(vsite_idx)):
        continue
    vsite = sys.getVirtualSite(vsite_idx)
    vsite_parents = []
    for vsite_parent_idx in range(vsite.getNumParticles()):
        vsite_parent = vsite.getParticle(vsite_parent_idx)
        vsite_parents.append(vsite_parent)
    distance = vsite.getLocalPosition()[0].value_in_unit(omm_unit.angstrom)
    print(vsite_parents, distance, vsite.getOriginWeights(), vsite.getXWeights(), vsite.getYWeights())
```

> [0, 1] -0.22000000000000003 (1.0, 0.0) (-1.0, 1.0) (-1.0, 1.0)

This should have the Na (index 1) first, then the Cl (index 0),  so it should be `[1, 0]` instead of `[0, 1]`. 

(I... could be mistaken about this, so I've started printing some other important-looking vsite info, but it's indistinguishable from a vsite that's pointed in the wrong direction)

trevorgokey:
Confirmed that checking the vsite particle orientation in the TK vsite shows 0,1, and since Cl is index 0 this is incorrect. Currently, the orientations need to be correct for all of the weights/positions to work, so that is what needs to be fixed. Managing indices has been challenging, but I have an idea where to look/how to fix. (I get the feeling nothing works :| many apologies. It appeared to work back when it was first merged!)

The divalent example in the issue also calculated the position incorrectly, but I punted it until we felt the original parameter assignment issue was cleared. Hopefully the parameter selection problem is out of the way now.

mattwthompson:
Just to put my mind at ease a little, I've added the minrex ini the linked issue

mattwthompson:
The linked issue was resolved by #1252 and this is out of date, so we have to close this one.

PR #1213, opened 2022-03-08
Remove un-used Topology attributes

### LINKED ISSUE #784: Unset public attributes in Topology
**Describe the bug**
Two (potentially more?) getters of `Topology` are wired up to point to attributes that don't exist. `Topology.fractional_bond_order_model` and `Topology.charge_model` have setters and getters, but neither are set in the `__init__` and do not seem to directly have a use in other classes. It looks like molecules know their charge model and force fields know their fractional bond order methods, but topologies are never asked if the have their own (at least in our tests - maybe some users care about this?). I couldn't find any open issues about these.

These could be set to `None`, set some default, or wired up to attempt to report a set of all models by iterating over the constituent molecules and bonds. But maybe it's better to remove them altogether? @j-wags do you envision these being _topology-level_ attributes in your preferred set of models? At very least I'd rather have a more descriptive exception that indicates its temporary state.

**To Reproduce**
```python3
>>> from openforcefield.topology import Molecule, Topology
>>> top = Topology.from_molecules(2 * [Molecule.from_smiles("CCO")])
>>> top.fractional_bond_order_model
  File "/Users/mwt/software/openforcefield/openforcefield/topology/topology.py", line 1496, in fractional_bond_order_model
    return self._fractional_bond_order_model
AttributeError: 'Topology' object has no attribute '_fractional_bond_order_model'
>>> top.charge_model
  File "/Users/mwt/software/openforcefield/openforcefield/topology/topology.py", line 1452, in charge_model
    return self._charge_model
AttributeError: 'Topology' object has no attribute '_charge_model'
```

**Output**

 - Operating system:

**Additional context**
Specs for other classes than `ForceField`, like `Molecule` and `Topology`, might be useful to clear up some of these ambiguities, even if they're imperfect and not complete.

mattwthompson:
Resolves #784 

### DIFF
--- a/openff/toolkit/tests/test_topology.py
+++ b/openff/toolkit/tests/test_topology.py
@@ -579,7 +579,6 @@ class TestTopology:
             )
             assert mod_imp in top.amber_impropers
 
-    # test_get_fractional_bond_order
     # test_two_of_same_molecule
     # test_two_different_molecules
     # test_to_from_dict
--- a/openff/toolkit/topology/topology.py
+++ b/openff/toolkit/topology/topology.py
@@ -41,8 +41,6 @@ from openff.toolkit.utils.exceptions import (
 from openff.toolkit.utils.serialization import Serializable
 from openff.toolkit.utils.toolkits import (
     ALLOWED_AROMATICITY_MODELS,
-    ALLOWED_CHARGE_MODELS,
-    ALLOWED_FRACTIONAL_BOND_ORDER_MODELS,
     DEFAULT_AROMATICITY_MODEL,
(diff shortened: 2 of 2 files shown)


j-wags (approved):
Looks good. I spent a little while trying to figure out if we should also pull the aromaticity model and constraints info from topology objects, but I didn't end up coming to a conclusion there, so this PR should be fine to move ahead with its current scope.

Feel free to ignore tests failing due to the RDKit/cairo and OE stereochemistry issues.

mattwthompson:
Thanks!

PR #1214, opened 2022-03-10
Fix #837

### LINKED ISSUE #837: Use ELF10 for system creation when OpenEyeToolkitWrapper is available
**Describe the bug**
Original message by @SimonBoothroyd:

> Am I right in thinking that since the 0.7.0 release of the toolkit we don't actually use ELF10 to compute partial charges, even when OE is installed and licensed?
> (i.e. in #471 the call to a toolkit compute_partial_charges_am1bcc function by a molecules compute_partial_charges_am1bcc function was replaced by a call to assign_partial_charges with am1bcc explicitly stated)

The above is a regression introduced in the 0.7.0 release. We should fix it by having `ToolkitAM1BCCHandler.create_force` do a `try/except` where it attempts to use `partial_charge_method='am1bccelf10'` first, and then, if there's a `ChargeMethodUnavailableError`, drops down to `partial_charge_method='am1bcc'`. This will cause ELF10 to be used (if available), and vanilla AM1-BCC otherwise. 

**Additional context**
cc #447

j-wags:
- [x] Fix #837, making ToolkitAM1BCCHandler use ELF10 if OpenEye is available

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -100,6 +100,10 @@ print(value_roundtrip)
   `fractional_bondorder_method` kwarg to the 
   [`BondHandler`](openff.toolkit.typing.engines.smirnoff.parameters.BondHandler) initializer 
   was being ignored.
+- [PR #1214](https://github.com/openforcefield/openforcefield/pull/1214): A long overdue fix
+  for [Issue #837](https://github.com/openforcefield/openff-toolkit/issues/837)! If OpenEye is
+  available, the `ToolkitAM1BCCHandler` will use the ELF10 method to select conformers for AM1BCC
+  charge assignment. 
 
 ### Examples added
 
--- a/openff/toolkit/tests/test_forcefield.py
+++ b/openff/toolkit/tests/test_forcefield.py
@@ -3274,6 +3274,43 @@ class TestForceFieldChargeAssignment:
 
         compare_partial_charges(using_kwarg, using_library_charges)
 
+    @requires_openeye
+    def test_toolkit_am1bcc_uses_elf10_if_oe_is_available(self):
+        """Ensure that the ToolkitAM1BCCHandler assigns ELF10 charges if OpenEye is available."""
+        # Can't just use CCO - Molecule needs to be big enough to realistically
+        # result in more than one conformer when ELF10 is requested
+        mol = Molecule.from_smiles("OCCCCCCO")
+
+        # The test forcefield "out of the box" just uses ToolkitAM1BCC, so this
+        # should produce ELF10 charges
+        test_forcefield = ForceField("test_forcefields/test_forcefield.offxml")
+        sys = test_forcefield.create_openmm_system(topology=mol.to_topology())
+
+        # Make another system with a force field that has the ToolkitAM1BCCHandler
+        # removed, and is explicitly told to use a ChargeIncrementModelHandler
+        # with AM1BCCELF10 charges. Ensure that the resulting charges are identical.
+        test_forcefield.deregister_parameter_handler("ToolkitAM1BCC")
+        cimh = test_forcefield.get_parameter_handler(
+            "ChargeIncrementModel",
+            {"version": 0.3, "partial_charge_method": "am1bccelf10"},
+        )
+
+        sys_a1b_elf10 = test_forcefield.create_openmm_system(topology=mol.to_topology())
+
+        compare_partial_charges(sys, sys_a1b_elf10)
+
+        # Now modify the ChargeIncrementModelHandler to specifically use single-conformer
+        # AM1BCC and ensure that the resulting charges are different from the first system.
+        cimh.partial_charge_method = "am1bcc"
+        assert test_forcefield["ChargeIncrementModel"].partial_charge_method == "am1bcc"
+
+        sys_a1b_single_conf = test_forcefield.create_openmm_system(
+            topology=mol.to_topology()
+        )
+        # Ensure that we get different results with AM1BCC ELF10 and single conf AM1BCC
+        with pytest.raises(AssertionError):
+            compare_partial_charges(sys_a1b_elf10, sys_a1b_single_conf)
+
 
 # ======================================================================
 # TEST CONSTRAINTS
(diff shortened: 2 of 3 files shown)


mattwthompson (approved):
LGTM. This is a behavior changed so I wanted to be thorough, but I couldn't find anything broken or surprising.

If I'm understanding the specification of this change correctly, it's simple: use `amb1bccelf10` if OpenEye Toolkits are available, `am1bcc` otherwise. I created some systems locally with and without `OpenEyeToolkitWrapper` available in my development environment and this behavior is encoded as expected: the `openmm.System` has charges consistent `molecule.assign_partial_charges(partial_charge_method="am1bccelf10")` before uninstalling OpenEye and `partial_charge_method="am1bcc"` after.

I noticed that the ELF-10 variant of AM1-BCC is slower, at least when I call it directly on a molecule (presumably because it generates more conformers along the way?). I assume this was considered years ago when the SMIRNOFF spec was written with this behavior in mind. It just came as a small surprise to me.

The only issues I can imagine are out of scope here and arise from cases in which "OpenEye Toolkits are available" is not an identical condition to "`OpenEyeToolkitWrapper` is in the registry of toolkits". This assumption is already so heavily baked in everywhere that bigger problems would probably arise before we'd need to worry about ELF10 silently not being used - and adding any more testing around the license being available is, again, out of scope here.

j-wags:
Thanks for the quick review!

> The only issues I can imagine are out of scope here and arise from cases in which "OpenEye Toolkits are available" is not an identical condition to "OpenEyeToolkitWrapper is in the registry of toolkits". This assumption is already so heavily baked in everywhere that bigger problems would probably arise before we'd need to worry about ELF10 silently not being used - and adding any more testing around the license being available is, again, out of scope here.

Agreed. There are lots of complex failure cases possible with weird OE conditions. I think there was an idea a while back of having different decorators/checks for different OE products. If partially-available-OE-products end up causing a lot of trouble then we could consider committing a few days to fix that. But I'll wait for user reports since that would take a significant amount of time to implement.

mattwthompson:
This has caused a minor regression in cases in which the ELF10 variant of AM1-BCC cannot assign partial charges but the vanilla variant can. Few and far between, I ran into this in a virtual site test using molecular oxygen (with OpenEye Toolkits installed and licensed):


```python3
>>> from openff.toolkit.topology import Molecule
>>> molecule = Molecule.from_smiles("O=O")
>>> molecule.assign_partial_charges(partial_charge_method="am1bcc")
>>> molecule.partial_charges
<Quantity([0. 0.], 'elementary_charge')>
>>> molecule.assign_partial_charges(partial_charge_method="am1bccelf10")
  File "/Users/mwt/software/openforcefield/openff/toolkit/topology/molecule.py", line 3666, in assign_partial_charges
    toolkit_registry.call(
  File "/Users/mwt/software/openforcefield/openff/toolkit/utils/toolkit_registry.py", line 380, in call
    raise ValueError(msg)
ValueError: No registered toolkits can provide the capability "assign_partial_charges" for args "()" and kwargs "{'molecule': Molecule with name '' and SMILES 'O=O', 'partial_charge_method': 'am1bccelf10', 'use_conformers': None, 'strict_n_conformers': False, 'normalize_partial_charges': True, '_cls': <class 'openff.toolkit.topology.molecule.Molecule'>}"
Available toolkits are: [ToolkitWrapper around OpenEye Toolkit version 2021.2.0, ToolkitWrapper around The RDKit version 2021.09.5, ToolkitWrapper around AmberTools version 21.0, ToolkitWrapper around Built-in Toolkit version None]
 ToolkitWrapper around OpenEye Toolkit version 2021.2.0 <class 'openff.toolkit.utils.exceptions.ChargeCalculationError'> : Unable to assign charges: Warning: BCIChargeCorrector: BCI charge corrections for mol
Warning:                     missing bcc parameter for the O1-O2 bond (BCIType 310231)
Warning: Unable to assign all BCCs in

 ToolkitWrapper around The RDKit version 2021.09.5 <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : partial_charge_method 'am1bccelf10' is not available from RDKitToolkitWrapper. Available charge methods are ['mmff94']
 ToolkitWrapper around AmberTools version 21.0 <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : partial_charge_method 'am1bccelf10' is not available from AmberToolsToolkitWrapper. Available charge methods are ['am1bcc', 'am1-mulliken', 'gasteiger']
 ToolkitWrapper around Built-in Toolkit version None <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : Partial charge method "am1bccelf10"" is not supported by the Built-in toolkit. Available charge methods are ['zeros', 'formal_charge']
```

It happens that we don't even have parameters to completely parametrize it:

```python3
>>> from openff.toolkit.typing.engines.smirnoff import ForceField
>>> sage = ForceField("openff-2.0.0.offxml")
>>> sage.create_openmm_system(molecule.to_topology())
  File "/Users/mwt/software/openforcefield/openff/toolkit/utils/utils.py", line 85, in wrapper
    return function(*args, **kwargs)
  File "/Users/mwt/software/openforcefield/openff/toolkit/typing/engines/smirnoff/forcefield.py", line 1268, in create_openmm_system
    return self._old_create_openmm_system(topology, **kwargs)
  File "/Users/mwt/software/openforcefield/openff/toolkit/utils/utils.py", line 85, in wrapper
    return function(*args, **kwargs)
  File "/Users/mwt/software/openforcefield/openff/toolkit/typing/engines/smirnoff/forcefield.py", line 1358, in _old_create_openmm_system
    parameter_handler.create_force(system, topology, **kwargs)
  File "/Users/mwt/miniconda3/envs/openff-dev/lib/python3.9/site-packages/openff/utilities/utilities.py", line 75, in wrapper
    return function(*args, **kwargs)
  File "/Users/mwt/software/openforcefield/openff/toolkit/typing/engines/smirnoff/parameters.py", line 2894, in create_force
    self._check_all_valence_terms_assigned(
  File "/Users/mwt/software/openforcefield/openff/toolkit/typing/engines/smirnoff/parameters.py", line 2434, in _check_all_valence_terms_assigned
    raise exception
openff.toolkit.utils.exceptions.UnassignedBondParameterException: BondHandler was not able to find parameters for the following valence terms:

- Topology indices (0, 1): names and elements ( O), ( O),
```

I don't know how many molecules exhibit this behavior. But I was able to assign partial charges before, and presumably still would in an environment that only had open source cheminformatics toolkits installed.

j-wags:
Thanks for reporting - I was a little concerned about this but I've been waiting to see if we have such a case for a druglike molecule. Right now I don't think we support parameterizing O2 (the second output in your post above is missing a bond parameter, unrelated to charge assignment), but if a druglike molecule comes along that works with OE AM1BCC "vanilla"but not AM1BCC ELF19 we can build in some fallback behavior.

There were some tricky cases where ELF10 would sometimes fail for carboxylates but those should have been fixed in #1171.

PR #1217, opened 2022-03-15
Performance improvements in from_openeye

mattwthompson:
OpenMM quantities are generally as quick and lightweight as
corresponding NumPy arrays, but Pint units tend to be slower. Most of
these changes simply defer unit tagging until a final step, carrying
values with implicit units in some internal functions where the units
are known and can be expected not to vary.

I used a quick benchmark that I think is representative of a typical use case and dropped the execution time of `Molecule.from_openeye` from roughly 22 ms to 5 ms. This is still a regression from roughly 3 ms in v0.10.3, but it's much more palatable.

https://gist.github.com/mattwthompson/3a085559a9e5d040b8439870e01f9410

There's surely more improvements to be found here and elsewhere.

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -250,7 +250,8 @@ class Atom(Particle):
         self._atomic_number = atomic_number
         # Use the setter here, since it will handle either ints or Quantities
         if hasattr(formal_charge, "units"):
-            if formal_charge.units == unit.dimensionless:
+            # Faster check than ` == unit.dimensionless`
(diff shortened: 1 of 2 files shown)


j-wags (approved):
Thanks - This looks good to me. I don't have any changes to recommend, but I'm only 7/10 confident that our test suite would catch this breaking anything, so I'm glad that it waited for review. 

**How could I have determined that this does/doesn't need review?**

On this one, I'd say "the number of lines changed in this PR would indicate that it needs review".

mattwthompson:
Thanks!

Issue #1215, opened 2022-03-11
macOS-only failures in CI with RDKit visualization


mattwthompson:
Log: https://github.com/openforcefield/openff-toolkit/runs/5504391595?check_suite_focus=true

Traceback:
```
________________ TestMoleculeVisualization.test_visualize_rdkit ________________
openff/toolkit/tests/test_molecule.py:3797: in test_visualize_rdkit
    assert isinstance(mol.visualize(backend="rdkit"), IPython.core.display.SVG)
openff/toolkit/topology/molecule.py:6479: in visualize
    from rdkit.Chem.Draw import (  # type: ignore[import]
../../../miniconda3/envs/test/lib/python3.9/site-packages/rdkit/Chem/Draw/__init__.py:14: in <module>
    from rdkit.Chem.Draw import rdMolDraw2D
E   ImportError: dlopen(/Users/runner/miniconda3/envs/test/lib/python3.9/site-packages/rdkit/Chem/Draw/rdMolDraw2D.so, 2): Library not loaded: @rpath/libfontconfig.1.dylib
E     Referenced from: /Users/runner/miniconda3/envs/test/lib/libcairo.2.dylib
E     Reason: Incompatible library version: libcairo.2.dylib requires version 14.0.0 or later, but libfontconfig.1.dylib provides version 13.0.0
```

I can't reproduce this locally (macOS, Python 3.9) with the same versions of `cairo`, `pycairo`, and `rdkit`, the latter of which was recently bumped to 2021.09.5. I checked the RDKit and its feedstock's issue tracker and nobody else has reported it yet.

mattwthompson:
A similar issue was reported and fixed by the maintainer of `rdkit-pypi`: https://github.com/kuelumbus/rdkit-pypi/pull/33

mattwthompson:
Looks like everything's been fixed upstream. No changes to our code needed. I kicked the tires on it in #1225 and the import no longer causes a crash. If this causes deployment issues in the future we can deal with them in packaging.

mattwthompson:
[...]

PR #1227, opened 2022-03-24
Fix a redundant unit tag

mattwthompson:
A small error in #1217 was producing a redundant trip between array and unit-bearing quantity. Conveniently, a helpful warning is thrown:

```python3
>>> Molecule.from_smiles("O")
/Users/mwt/software/openforcefield/openff/toolkit/utils/openeye_wrapper.py:1032: UnitStrippedWarning: The unit of the quantity is stripped when downcasting to ndarray.
  positions[off_atom_index, :] = off_atom_coords
Molecule with name '' and SMILES '[H]O[H]'
```


j-wags (approved):
LGTM, and no need for releasenotes! In the future a change like this can go without review.

mattwthompson:
Thanks - I was going to hit the button but I got sidetracked noticing that #1215 is back 😞

PR #1229, opened 2022-03-25
Performance improvements in/near Molecule API and RDKit wrapper

mattwthompson:
So far I've uncovered two easy ways to reduce runtime of molecule conversions:
* Defer unit wrapping until later in loops, more or less following what we did in #1217 
* `Molecule._invalidate_cached_properties` is called every time an atom or a bond is added to a molecule, with no clear effect happening inside of those loops. I snuck in an argument to allow the molecule conversion methods to skip these checks and instead called it outside of the loop (i.e. just twice, instead of once per atom and bond). Behavior with the public API should be unchanged.

There were some docstrings ranging from mildly to dangerously out of date; I only modified those that I came across.

This should result in something like a 50% and 20% reduction in the runtime of `Molecule.from_rdkit` and `Molecule.from_openeye`, respectively. Relevant portion of the notebook I used to spot-check:

```python
molecule = Molecule.from_smiles(
    "C=C(CN1Cc2ccc(Cl)cc2C2(CCN(c3cncc4ccc(OCCNCC(F)(F)F)cc34)C2=O)C1)C(=O)NC",
    allow_undefined_stereo=True,
)
rdmol = molecule.to_rdkit()
oemol = molecule.to_openeye()
```

```python
%%timeit
molecule.from_rdkit(rdmol, allow_undefined_stereo=True)
# rdkit-perf branch: 8.33 ms ± 382 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# upstream/topology-biopolymer-refactor: 17.5 ms ± 691 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

```python
%%timeit
molecule.from_openeye(oemol, allow_undefined_stereo=True)
# rdkit-perf branch: 3.99 ms ± 121 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# upstream/topology-biopolymer-refactor: 5.05 ms ± 339 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```


j-wags:
Thanks for opening this. I'd like to take a quick look at this if my schedule allows, but please merge this on Monday if I haven't responded!

j-wags (approved):
LGTM!

PR #1232, opened 2022-03-26
Run examples workflow in biopolymer feature branch

mattwthompson:
I don't think this will run until merged, not sure.


j-wags:
Thanks Matt. Feel free to merge this when you think it's ready. You mentioned in our call that the diff will be terrible, and I agree. If you'd like a brief (not total) review, feel free to ping @Yoshanuikabundi.

mattwthompson:
I gave everything a once-over and I don't think it's worth us spending more time reviewing it; the diffs inside the cells are few and far between as far as I can tell/recall.

PR #1245, opened 2022-04-06
Let Dependabot update actions in the 0.11.0 branch

mattwthompson:
On a temporary basis. It's probably not a big deal but some of our actions are multiple major versions out of date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,6 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


mattwthompson:
I assume it's not magic enough to run updates from inside of a PR, but maybe 🤞

mattwthompson:
Not elegant but I'm going to merge this so it can do its thing. We should update this once the dust has settled from the 0.11.0 release and the default branch has been updated to `main`.

PR #1246, opened 2022-04-06
Add OpenMM 7.6+ requirement in release notes

mattwthompson:
A common user question is, and will continue to be, what versions of OpenMM are supported by different versions of the toolkit. It is possible that something like this should go in an FAQ or developers' guide, but here I'm just adding it to the release notes of the (only?) change that introduces a major compatibility shift. I authored this change (#1230) according to ours plans months ago, but somehow neglected to include it in the release notes then.

I'll merge this in 48 hours if not reviewed by then, as it's a minor change.


j-wags (approved):
This looks great. Thanks @mattwthompson!

Issue #1249, opened 2022-04-07
Molecule.add_conformer doesn't accept openff-unit units


richardjgowers:
**Describe the bug**

Openff seems to only accept openmm units.

**To Reproduce**

```
In [1]: from openff.toolkit.topology import Molecule

In [2]: from openff.units import unit

In [3]: import numpy as np

In [4]: m = Molecule.from_smiles('CCO')

In [5]: coords = np.random.random((len(m.atoms), 3))

In [6]: m.add_conformer(coords * unit.angstrom)

```
**Output**

`Exception: Coordinates passed to Molecule._add_conformer without units. Ensure that coordinates are of type openmm.units.Quantity`

**Additional context**

I tried looking through issues and I think there's a move away from openmm units in progress(?), but I'm not sure if that's actually happening on the topology refactor branch.  Sorry if this has been fixed there.  I think it wouldn't be too nasty to have some sort of decorator around functions with typed inputs which converted the openff-unit units to openmm units.  I think an alternative solution might be to have openff-units quack like an openmm unit when required, but maybe that's more of an openff-unit repo issue than a toolkit issue.

mattwthompson:
Are you using version 0.10.x?

richardjgowers:
@mattwthompson 0.10.3 of tk and 0.1.5 of units

mattwthompson:
Okay yeah, sorry, that behavior isn't landing until version 0.11.0 and unfortunately we're only backporting critical bugfixes to 0.10.x.  [This](https://github.com/openforcefield/openff-toolkit/blob/d5c465e21c1fc900c156dc697afc3be3562f4718/openff/toolkit/tests/test_molecule.py#L2493) is the test that's supposed to check that behavior and somehow I neglected to cover that - my oversight.  The relevant source code looks fine and runs your reproduction without error, so I guess that's why I forgot to add it in the test:

>  I think it wouldn't be too nasty to have some sort of decorator around functions with typed inputs which converted the openff-unit units to openmm units.

Hmm, I hadn't considered this. It sure seems like it would be an improvement over the branching logic I've used this far ...

> I think an alternative solution might be to have openff-units quack like an openmm unit when required, but maybe that's more of an openff-unit repo issue than a toolkit issue.

I agree, although I've had a hard time pulling that off for reasons that escape memory (https://github.com/openforcefield/openff-units/pull/9)

richardjgowers:
[...]

PR #1251, opened 2022-04-07
Fix adding conformers with OpenMM units

mattwthompson:
In #1249 it was made obvious that I didn't test the case of adding a conformer with OpenMM units. The functionality was there but that particular test case was missing. I also found the code to be hard to follow and updated it in hopes of it being more maintainable.

### DIFF
--- a/openff/toolkit/tests/test_molecule.py
+++ b/openff/toolkit/tests/test_molecule.py
@@ -27,7 +27,6 @@ import numpy as np
 import pytest
 from openff.units import unit
 from openff.units.elements import MASSES, SYMBOLS
-from openmm import unit as openmm_unit
 
 from openff.toolkit.tests.create_molecules import (
     create_acetaldehyde,
@@ -2492,6 +2491,8 @@ class TestMolecule:
 
     def test_add_conformers(self):
         """Test addition of conformers to a molecule"""
+        from openmm import unit as openmm_unit
+
         # Define a methane molecule
         molecule = Molecule()
         molecule.name = "methane"
@@ -2606,7 +2607,6 @@ class TestMolecule:
         assert molecule.n_conformers == 3
         assert molecule.conformers[2][0][0] == 10.0 * unit.angstrom
 
-        # Add a conformer with units of nanometers
         conf_nonsense_units = unit.Quantity(
             np.array(
                 [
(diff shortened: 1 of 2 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -4246,50 +4246,56 @@ def _add_conformer(self, coordinates):
         index: int
             The index of this conformer
         """
-        new_conf = unit.Quantity(
-            np.zeros(shape=(self.n_atoms, 3), dtype=float), unit.angstrom
-        )
```
mattwthompson, line 4251: I found this confusing - this `new_conf` read to me like the conformer passed through as an argument, but it's just a temporary variable used for  input validation. `coordinates` is the argument.

openff/toolkit/tests/test_molecule.py
```diff
@@ -2492,6 +2491,8 @@ def test_equality(self):
 
     def test_add_conformers(self):
         """Test addition of conformers to a molecule"""
+        from openmm import unit as openmm_unit
+
```
j-wags, line 2495: 👍 

At some point in the future we'll need to figure out how to handle test skips for openmm-free builds. But I don't think we need to handle that in this PR.

j-wags (approved):
Excellent. Thanks, @mattwthompson! Just one blocking comment, this looks great otherwise. Thanks for jumping on this so quickly and requesting my review.

openff/toolkit/tests/test_molecule.py
mattwthompson: We should be able to get away with liberal use of [`skip_if_missing`](https://github.com/openforcefield/openff-utilities/blob/6db2421dec43f9e6e34ca38fa39f9bac86e00c85/openff/utilities/testing.py#L6) but I won't worry about that now

openff/toolkit/tests/test_molecule.py
```diff
@@ -2622,6 +2622,26 @@ def test_add_conformers(self):
...
+            ),
+            openmm_unit.angstrom,
+        )
+        molecule.add_conformer(conf_openmm_units)
+
+        assert molecule.conformers[-1][0][0].m_as(unit.angstrom) == 1.0
+        assert molecule.conformers[-1][-1][-1].m_as(unit.angstrom) == 15.0
+
+        with pytest.raises(Exception) as excinfo:
+            molecule.add_conformer(conf_unitless)
```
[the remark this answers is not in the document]
mattwthompson: Wow, that's bad. You're right, and this is a side effect of the janky `with pytest.rasises(Exception):` which doesn't really do what we want.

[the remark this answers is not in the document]
mattwthompson: I'll resist the temptation to fix this in _this_ PR, but:

```shell
$ grep -r "pytest.raises(Exception)" openff/toolkit/tests | wc -l
      15

PR #1156, opened 2021-12-16
Remove deprecated exceptions

mattwthompson:
In #1021 we added some deprecated exceptions with the plan to remove them in January 2022.

- [x] Tag issue being addressed
- [ ] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/tests)
- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable
- [x] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [ ] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/master/docs/releasehistory.rst)

PR #1266, opened 2022-04-13
Port #1156

mattwthompson:


mattwthompson:
Codecov is not functional, but this introduces no new failures.

PR #1272, opened 2022-04-14
Add section to FAQ on AM1-BCC issues with large ligands

mattwthompson:
This has come up twice in the past week or so, which I think warrants an entry here. I chose to leave out a few topics in order to be brief, but can add any back if they're seen as necessary. That includes

* Explaining that biopolymer force fields will use library charges, bypassing this issue for proteins
* That the AM1-BCC calculation only takes place once per unique molecule, so this isn't an issue for boxes of many not-incredibly-large molecules
* `charge_from_molecules` (I'm personally not a huge fan of this patten, but I see why it must exist)

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -84,6 +84,11 @@ and then try rerunning and/or reinstalling the Toolkit.
 
 No! This is the intended behavior. The force field parameters of a molecule should be independent of both their chemical environment and conformation so that they can be used and compared across different contexts. When applying AM1BCC partial charges, the toolkit achieves a deterministic output by ignoring the input conformation and producing several new conformations for the same molecule. Partial charges are then computed based on these conformations. This behavior can be controlled with the `use_conformers` argument to the [assign_partial_charges()](openff.toolkit.topology.Molecule.assign_partial_charges) and [compute_partial_charges_am1bcc()](openff.toolkit.topology.Molecule.compute_partial_charges_am1bcc) methods of the [Molecule](openff.toolkit.topology.Molecule) class.
 
+## Parameterizing my system, which contains a large molecule, is taking forever. What's wrong?
+
+The mainline OpenFF force fields use AM1-BCC to assign partial charges (via the `<ToolkitAM1BCCHandler>` tag in the OFFXML file). This method unfortunately scales poorly with the size of a molecule and ligands roughly 100 atoms or larger may take so long (i.e. 10 minutes or more) that it seems like your code is simply hanging indefinitely. If you have OpenEye license and OpenEye Toolkits installed, the OpenFF Toolkit will instead use `quacpac`, which can offer better performance on large molecules. (Otherwise, it uses AmberTool's `sqm`, which is free to use.)
+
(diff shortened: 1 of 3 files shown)


FAQ.md
```diff
@@ -84,6 +84,11 @@ and then try rerunning and/or reinstalling the Toolkit.
...
 
+## Parameterizing my system, which contains a large molecule, is taking forever. What's wrong?
+
+The mainline OpenFF force fields use AM1-BCC to assign partial charges (via the `<ToolkitAM1BCCHandler>` tag in the OFFXML file). This method unfortunately scales poorly with the size of a molecule and ligands roughly 100 atoms or larger may take so long (i.e. 10 minutes or more) that it seems like your code is simply hanging indefinitely. If you have OpenEye license and OpenEye Toolkits installed, the OpenFF Toolkit will instead use `quacpac`, which can offer better performance on large molecules. (Otherwise, it uses AmberTool's `sqm`, which is free to use.)
```
mattwthompson, line 89: I wanted to link to https://open-forcefield-toolkit.readthedocs.io/en/stable/installation.html#openeye, but I'm not sure if that should be a hard link or something that sphinx would smoothly pick up.

mattwthompson: This is working locally after d7af4ebe. I used the below link, which is pretty slick and not overly verbose in syntax.

https://myst-parser.readthedocs.io/en/latest/syntax/syntax.html#targets-and-cross-referencing

FAQ.md
```diff
@@ -84,6 +84,11 @@ and then try rerunning and/or reinstalling the Toolkit.
...
 
+## Parameterizing my system, which contains a large molecule, is taking forever. What's wrong?
+
+The mainline OpenFF force fields use AM1-BCC to assign partial charges (via the `<ToolkitAM1BCCHandler>` tag in the OFFXML file). This method unfortunately scales poorly with the size of a molecule and ligands roughly 100 atoms or larger may take so long (i.e. 10 minutes or more) that it seems like your code is simply hanging indefinitely. If you have OpenEye license and OpenEye Toolkits [installed](installation/openeye), the OpenFF Toolkit will instead use `quacpac`, which can offer better performance on large molecules. (Otherwise, it uses AmberTool's `sqm`, which is free to use.)
```
j-wags, line 89:
```suggestion
The mainline OpenFF force fields use AM1-BCC to assign partial charges (via the `<ToolkitAM1BCCHandler>` tag in the OFFXML file). This method unfortunately scales poorly with the size of a molecule and ligands roughly 100 atoms (about 40 heavy atoms) or larger may take so long (i.e. 10 minutes or more) that it seems like your code is simply hanging indefinitely. If you have an OpenEye license and OpenEye Toolkits [installed](installation/openeye), the OpenFF Toolkit will instead use `quacpac`, which can offer better performance on large molecules. Otherwise, it uses AmberTool's `sqm`, which is free to use.
```

j-wags (approved):
LGTM @mattwthompson  - Thanks for proactively documenting this!

PR #1273, opened 2022-04-14
Clarify `compute_partial_charges_am1bcc` deprecation

### LINKED ISSUE #447: Document partial charge assignment scheme used by Molecule.compute_partial_charges_am1bcc()
Currently, the documentation for [`Molecule.compute_partial_charges_am1bcc()`](https://open-forcefield-toolkit.readthedocs.io/en/0.5.1/api/generated/openforcefield.topology.Molecule.html#openforcefield.topology.Molecule.compute_partial_charges_am1bcc) says almost nothing informative about what is going on under the hood, which is actually quite complex:

* The [canonical AM1-BCC charging method](https://docs.eyesopen.com/toolkits/cookbook/python/modeling/am1-bcc.html) is used to assign ELF10 charges if the OpenEye Toolkit is available.
* Otherwise, RDKit is used to assign a conformation, and [`antechamber -c bcc`](http://ambermd.org/antechamber/) from the [AmberTools]](http://ambermd.org/AmberTools.php) distribution (which uses the `sqm` semiempirical quantum chemical package) is used to assign AM1-BCC charges.

We should also document anything we do to work around issues with OpenEye charging, such as the fallback to standard AM1-BCC and workarounds for the [carboxylic acid issue](https://github.com/oess/openmm_orion/blob/STMD_Analysis/MDOrion/LigPrep/ff_utils.py#L31-L106).

mattwthompson:
Sorry to step on your toes if you finished this one @j-wags yesterday - I had a bit of momentum after being burned by ELF10 differences and wanted to make sure these improvements got in.

I know we considered logging, but I think falls under the description of "use in library code if the issue is avoidable and the client application should be modified to eliminate the warning" in the [docs](https://docs.python.org/3/howto/logging.html#when-to-use-logging).

Closes #447


openff/toolkit/topology/molecule.py
```diff
@@ -3792,6 +3800,19 @@ def assign_partial_charges(
...
+            - ``"am1-mulliken"``
+            - ``"mmff94"``
+            - ``"gasteiger"``
+
+        For more supported charge methods and details, see the corresponding methods in each toolkit wrapper:
+            - OpenEyeToolkitWrapper.assign_partial_charges
+            - RDKitToolkitWrapper.assign_partial_charges
+            - AmberToolsToolkitWrapper.assign_partial_charges
+            - BuiltInToolkitWrapper.assign_partial_charges
+
```
j-wags, line 3815: 👍

j-wags (approved):
LGTM! There's some new test failure on the py37 true/true builds that should be addressed before merging. Thanks for updating this!

mattwthompson:
Now back to "no new test failures," with the exception of one failed `git clone` that I don't want to wait for a retry of.

Issue #1084, opened 2021-09-20
Parametrizing non-periodic topologies with mainline force fields changes non-bonded settings


mattwthompson:
**Describe the bug**
This is a long-standing mismatch between the SMIRNOFF spec and the implementation of `ForceField.create_openmm_system`. I'm not sure how to fix it here, nor am I sure it should be fixed here. I took a stab at something that might fix it in #920, but it blew up in complexity and won't be implemented here. It's also been discussed several times in scattered locations, but I don't think there is a stub issue for it. This is similar to https://github.com/openforcefield/openff-toolkit/issues/882 in its subtlety, impact to users, and general tangled-up-ness.

**To Reproduce**

Using the README snippet as a starting point:
```python3
>>> import openmm
>>> from openff.toolkit.topology import Molecule
>>> from openff.toolkit.utils import get_data_file_path
>>> sdf_file_path = get_data_file_path('molecules/ethanol.sdf')
>>> molecule = Molecule.from_file(sdf_file_path)
>>>
>>> # Create an OpenFF Topology object from the molecule
>>> from openff.toolkit.topology import Topology
>>> topology = Topology.from_molecules(molecule)
>>>
>>> # Load the latest OpenFF force field definition
>>> from openff.toolkit.typing.engines.smirnoff import ForceField
>>> forcefield = ForceField('openff-1.2.0.offxml')
>>> forcefield['vdW'].method
'cutoff'
>>> forcefield['Electrostatics'].method
'PME'
>>> topology.box_vectors  # Returns nothing, since it's None
>>> openmm_system = forcefield.create_openmm_system(topology)
```
We have now parametrized a non-periodic topology with an electrostatics section that's supposed to use PME, which is not valid for non-periodic topologies. This is handled by converting the entire non-bonded force (recall that out of the box, OpenMM combines vdW and electrostatics interactions into one object) into something that does not have a cutoff:
```python3
>>> for force in openmm_system.getForces():
...     if type(force) == openmm.openmm.NonbondedForce:
...         print(force.getNonbondedMethod())
...         print(force.getCutoffDistance())
...
0
0.9 nm
>>> import openmm
>>> openmm.NonbondedForce.NoCutoff
0
```

[...]

PR #1277, opened 2022-04-20
Support version 0.4 of SMIRNOFF Electrostatics section

### LINKED ISSUE #1288: `ElectrostaticsHandler.switch_width` converter does nothing
**Describe the bug**
A [typo](https://github.com/openforcefield/openff-toolkit/blob/52542c604be17f43e696c840dd794c44507c11a8/openff/toolkit/typing/engines/smirnoff/parameters.py#L3776) prevents  the switch width of the electrostatics handler from doing anything. It's not used with PME electrostatics so it's unlikely anybody was using this.

**To Reproduce**

```shell
(openff-dev) [openff-toolkit] git status -uno                                             9:28:25  ☁  52542c60 ☂ ✭
HEAD detached at upstream/topology-biopolymer-refactor
nothing to commit (use -u to show untracked files)
(openff-dev) [openff-toolkit] python                                                      9:28:27  ☁  52542c60 ☂ ✭
Python 3.9.12 | packaged by conda-forge | (main, Mar 24 2022, 23:23:20)
[Clang 12.0.1 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from openff.toolkit.typing.engines.smirnoff.parameters import ElectrostaticsHandler
Warning: Unable to load toolkit 'OpenEye Toolkit'. The Open Force Field Toolkit does not require the OpenEye Toolkits, and can use RDKit/AmberTools instead. However, if you have a valid license for the OpenEye Toolkits, consider installing them for faster performance and additional file format support: https://docs.eyesopen.com/toolkits/python/quickstart-python/linuxosx.html OpenEye offers free Toolkit licenses for academics: https://www.eyesopen.com/academic-licensing
>>> handler = ElectrostaticsHandler(version=0.3)
>>> from openff.units import unit
>>> handler.switch_width = 1.234 * unit.nanometer
>>> handler.switch_width
>>> handler.switch_width is None
True
>>> handler.version
0.3
```

I'll try to fix this in #1277.

mattwthompson:
Resolves https://github.com/openforcefield/openff-toolkit/issues/1084 in part

Fixes #1288

- [x] ~~Blocked by~~ https://github.com/openforcefield/standards/pull/34
- [x] Add up-converters
- [x] Test up-converters
- [x] ~~Handle `switch_width` in electrostatics handler~~ Do not handle, but explicitly do not handle
- [x] Band-aid old code paths?
- [x] ~~Update Interchange~~ - I'd like to do this _after_ this PR to avoid too many feature branches being split off
- [x] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable - I did a scan for `.method` and similar patterns and didn't find anything obviously outdated.


openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -3783,15 +3805,80 @@ def scale15(self, attrs, new_scale15):
...
+                    '`nonperiodic_potential="Coulomb"`, and `exception_potential="Coulomb"`.'
+                )
+            else:
+                # This clause shouldn't be accessible; the only previously-allowed values
+                # (`converter=_allow_only(["Coulomb", "PME", "reaction-field"])`)
+                # should all be handled by the above up-conversion
+                logger.warning(
+                    "Failed to up-convert Electrostatics section from 0.3 to 0.4. Did not know "
+                    f"how to up-convert `method={kwargs['method']}`. Continuing to use version 0.3."
+                )
```
j-wags, line 3880: This seems dangerous and untested/able - I'm thinking it should be a NotImplementedError.

j-wags (changes requested):
Ok, I hammered on this a bit and it looks like everything works well. The only thing this PR still needs to add the `solvent_dielectric` ParameterAttribute required in the 0.4 Electrostatics spec. Per the spec, it should default to "none", and it's fine if we have the setter just raise a `NotImplementedError` if someone tries to set it to anything else.

Also (per inline comment) I don't think we should even try to support unexpected 0.3 `methods` - It would be fine if anything fishy just raises an error.

openff/toolkit/typing/engines/smirnoff/parameters.py
mattwthompson: Sure, this was only a fallback because I thought raising a new exception was not going to be favored.

j-wags (approved):
Nice - Thanks @mattwthompson! Feel free to merge once tests pass.

PR #1279, opened 2022-04-21
Use `packaging.version` to manage SMIRNOFF section versions

mattwthompson:
The toolkit currently tracks the version of each SMIRNOFF section in a corresponding `ParameterHandler.version`. These are stored as floats, however:

```python3
>>> from openff.toolkit.typing.engines.smirnoff import ElectrostaticsHandler
>>> type(ElectrostaticsHandler(version=0.3).version)
<class 'float'>
```

This is fine for now, but will cause issues in the future. This is commonly handled with PyPA's `packaging`, which is already a dependency used elsewhere. In a meeting today, @j-wags agreed to this change.

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -58,6 +58,12 @@ print(value_roundtrip)
 
 ## Current Development
 
+- [PR #1279](https://github.com/openforcefield/openforcefield/pull/1279):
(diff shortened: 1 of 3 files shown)


j-wags (approved):
Looks great, thanks @mattwthompson. We should document the API change in the `ParameterHandler.version` getter in the release notes, then this should be good to go!

Issue #966, opened 2021-06-01
[Request] Easier selection of tookit registries


lilyminium:
**Is your feature request related to a problem? Please describe.**

The toolkit allows you to select which toolkit_registry you would like to use for individual methods, but as soon as we get into more advanced functions that call these methods it is quite difficult to choose your toolkit. To give an example case, I wanted to `create_openmm_system`. Using the OpenEye toolkit to match smarts was hanging, whereas using RDKit worked. However, the function call goes `create_force -> find_matches -> _find_matches -> chemical_environment_matches` so my only options for modifying the toolkit used by `chemical_environment_matches` were:

1. Modify the GLOBAL_TOOLKIT_REGISTRY (except this is used by other functions that I did *not* want to change, e.g. partial charge calculation)
2. Go into the code and change that part of it -- not very transferrable and not accessible to novice programmers
3. (what I ended up doing) replacing `chemical_environment_matches` as below, also not very accessible to novice programmers.

```python
chemical_environment_matches = Topology.chemical_environment_matches

RDKIT_TOP_REGISTRY = ToolkitRegistry(
    toolkit_precedence=[
        RDKitToolkitWrapper,
        OpenEyeToolkitWrapper,
        AmberToolsToolkitWrapper,
        BuiltInToolkitWrapper,
    ],
    exception_if_unavailable=False,
)



def new_chemical_environment_matches(*args, **kwargs):
    toolkit_registry = kwargs.get("toolkit_registry", RDKIT_TOP_REGISTRY)
    kwargs["toolkit_registry"] = toolkit_registry
    return chemical_environment_matches(*args, **kwargs)

Topology.chemical_environment_matches = new_chemical_environment_matches
```

**Describe the solution you'd like**

Pass keyword arguments such as `toolkit_registry` through the function call chain.

**Describe alternatives you've considered**

**Additional context**

mattwthompson:
>  the function call goes create_force -> find_matches -> _find_matches -> chemical_environment_matches

It looks like this chain was broken pretty early, since `**kwargs` stops being passed through once `find_matches` is called? I don't know what the best pattern is, but the current design of passing a toolkit registry around does kinda fall apart if somebody wanted to use different toolkits for different steps (i.e. here you want to use OpenEye for charge assignment but RDKit for SMARTS matching?).

> Pass keyword arguments such as toolkit_registry through the function call chain.

In this case, you wouldn't be able to pick which toolkit is used for what - is that okay?

[...]

PR #1281, opened 2022-04-22
Add private `ToolkitRegistry` context manager

mattwthompson:
Relevant to https://github.com/openforcefield/openff-toolkit/issues/966 and some other discussions.

This uses some of @SimonBoothroyd's code from [`openff-fragmenter`](https://github.com/openforcefield/openff-fragmenter/blob/4a290b866a8ed43eabcbd3231c62b01f0c6d7df6/openff/fragmenter/utils.py#L97-L123) to enable registry selection without needing to implement it in Interchange.

- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable - AFAIR this change only adds a private function


j-wags (commented):
This could use some tests, for example:

* `create_openmm_system` with OpenEye vs with AT/RDK - Expect different charges
* `create_openmm_system` with AT/RDK, asking for AM1BCCELF10, ensuring that we get an error
* Pass an invalid registry (like an empty one, or only RDKit) and expect an error from charge assignment

mattwthompson:
This isn't quite working the way I hoped in that `GLOBAL_TOOLKIT_REGISTRY` is not restored to its original state when an exception is raise inside the context of `_toolkit_registry_manager`.


```python3
In [1]: from openff.toolkit import *

In [2]: from openff.toolkit.utils.toolkit_registry import _toolkit_registry_manager

In [3]: from openff.toolkit.utils import GLOBAL_TOOLKIT_REGISTRY, BuiltInToolkitWrapper

In [4]: topology = Molecule.from_smiles("CCOCCCOCCOC").to_topology()

In [5]: sage = ForceField("openff-2.0.0.offxml")

In [6]: sage.create_openmm_system(topology, toolkit_registry=BuiltInToolkitWrapper(), use_interchange=True)
Input In [6], in <cell line: 1>()
----> 1 sage.create_openmm_system(topology, toolkit_registry=BuiltInToolkitWrapper(), use_interchange=True)

<SNIPPING SOME VERBOSE TRACEBACK>

File ~/software/openff-toolkit/openff/toolkit/utils/toolkit_registry.py:382, in ToolkitRegistry.call(self, method_name, raise_exception_types, *args, **kwargs)
    380 for toolkit, error in errors:
    381     msg += " {} {} : {}\n".format(toolkit, type(error), error)
--> 382 raise ValueError(msg)

ValueError: No registered toolkits can provide the capability "find_smarts_matches" for args "(Molecule with name '' with bad SMILES and Hill formula 'C8H18O3', '[#6X4:1]-[#6X4:2]')" and kwargs "{'unique': False}"
Available toolkits are: [ToolkitWrapper around Built-in Toolkit version None]


In [7]: GLOBAL_TOOLKIT_REGISTRY
Out[7]: ToolkitRegistry containing Built-in Toolkit
```

Everything is as expected until the last line - I thought the point of these lines

https://github.com/openforcefield/openff-toolkit/blob/9b4f0e05671fca4da6fc86b416ef9535bf5dfbc6/openff/toolkit/utils/toolkit_registry.py#L412-L418

was to undo the modifications that happened a few lines above.

j-wags:
Oof, you're right. I looked into it and context managers don't handle exceptions gracefully by default. 

So, one option to move forward would be to not use the contextlib approach and instead make our own contextmanager class with a custom `__enter__` and `__exit__` method.

Or, the other option, which I tried implementing in that last commit, is to check if there's an exception in the `yield` block, and then reset the GLOBAL_TOOLKIT_REGISTRY and raise the exception once everything's back to normal. 

Apologies that I was wrong about this initially. I think the manually-raise-the-exception approach should be fine if that works for you.

openff/toolkit/utils/toolkit_registry.py
```diff
@@ -383,3 +385,41 @@ def __repr__(self):
...
+        exc = e
+
+    for toolkit in context_toolkits:
+        GLOBAL_TOOLKIT_REGISTRY.deregister_toolkit(toolkit)
+
+    for toolkit in original_toolkits:
+        GLOBAL_TOOLKIT_REGISTRY.register_toolkit(toolkit)
+
+    if exc is not None:
+        raise exc
```
SimonBoothroyd, line 425:
```suggestion
    try:
        yield
    finally:
        for toolkit in context_toolkits:
            GLOBAL_TOOLKIT_REGISTRY.deregister_toolkit(toolkit)
        for toolkit in original_toolkits:
            GLOBAL_TOOLKIT_REGISTRY.register_toolkit(toolkit)
```

SimonBoothroyd:
Ah you're right I missed the finally statement! See e.g. [here](https://github.com/openforcefield/openff-utilities/blob/f9f623823b4c49162bdd68ad88c8bf0469063036/openff/utilities/utilities.py#L150) were we correctly handle potential errors in a context manager like this

mattwthompson:
Thanks @SimonBoothroyd!

mattwthompson:
Okay, I think this is ready to go now.

j-wags (approved):
Looks great!

PR #1282, opened 2022-04-24
fix fluorine atomic num in OE test

j-wags:


mattwthompson:
Wow, a bad copy+paste on my part that also slipped my eyes when re-reading it. Oops.

PR #1286, opened 2022-04-26
Use flake8

mattwthompson:
This is my attempt at getting the toolkit passing a `flake8 openff/` check in addition to removing some legacy constructs. The key improvement here is that we can run `flake8` in development (or watch pre-commit.ci run it) and get useful feedback in a few seconds. This is an improvement over LGTM, which uses rules we cannot modify and comments on PRs only after changes have been pushed.

This includes the major big changes happening as of late; I think everything is up to date with the biopolymer branch. A lot of the diff is unreadable due to a combination of squash-merges and other changes not yet merged. This could probably be cleaned up with some care, I think squashing everything here into one commit would clarify the attributions.


j-wags (approved):
Cool -- This is as good a time as any to tear off the bandaid. Let's merge this once CI passes (though it's fine if the vsite notebook fails)

j-wags:
Hm, something is weird here. One of the major points of #1195 was to make `Molecule.from_pdb` not require rdkit.

mattwthompson:
Okay, I'll revert that and work from there.

PR #1289, opened 2022-04-28
Speed up CI with `pytest-xdist`

mattwthompson:
The unit tests are taking 12 minutes (not counting setup time) and I think we get 2 cores on each runner for free. This should be a free speedup.


mattwthompson:
Looking at the logs, I'm seeing more than 2x speedup (10-12 mins down to ~4 mins).

PR #1297, opened 2022-05-06
Drop Python 3.7 from testing

mattwthompson:
Python 3.7 support is being phased out in upstream packages:
* Pandas in January: https://github.com/pandas-dev/pandas/releases/tag/v1.4.0
* Pint in April: https://github.com/hgrecco/pint/blob/0.19/CHANGES#L41
* NumPy was scheduled for last December but some builds are still being made: https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table

Generally people follow NEP 29 and are phasing it out. The core, not-necessarily-for-scientific-uses support completely ends [next summer](https://devguide.python.org/#status-of-python-branches). Our core compiled upstreams (AmberTools, RDKit, and OpenEye Toolkits) all have recent releases with 3.7 support. However, I am growing concerned that a 0.11.0 release with official support for 3.7 already has an expiration date and the other API breaks provide a good opportunity to support.

This PR increments all Python versions in tests by 1 minor version, meaning versions 3.8-3.10 would be supported.


mattwthompson:
The 3.10 runs downgrade to 3.9 (on both macOS and Linux) for reasons that are completely unclear from the logs alone. I tried this locally (using the [same file](https://gist.github.com/mattwthompson/b70791b0881fc8b0d608efa300ec2fd9) with an added `python =3.10` pint) and got

```
$ CONDA_SUBDIR=osx-64 mamba env create --name tmp --file devtools/conda-envs/test_env.yaml
conda-forge/osx-64                                          Using cache
conda-forge/noarch                                          Using cache
pkgs/main/noarch                                              No change
pkgs/r/osx-64                                                 No change
pkgs/main/osx-64                                              No change
pkgs/r/noarch                                                 No change
openeye/osx-64                                                No change
openeye/noarch                                                No change


Looking for: ['python=3.10', 'pip', 'packaging', "openmm[version='>=7.7']", 'openff-forcefields', "openff-units[version='>=0.1.6']", 'openff-utilities', 'smirnoff99frosst', 'numpy', 'networkx', 'ambertools', 'rdkit', 'xmltodict', 'python-constraint', 'pytest', 'pytest-cov', 'pytest-xdist', 'pytest-rerunfailures', 'nbval', 'codecov', 'coverage', 'openeye-toolkits', 'mdtraj', 'cachetools', 'cached-property', 'pyyaml', 'toml', 'bson', 'msgpack-python', 'qcelemental', "qcportal[version='>=0.15']", 'qcengine', 'mdtraj', 'mypy', 'typing-extensions']


Encountered problems while solving:
  - package openmm-7.7.0-py37hc3a70d1_0_apple requires python >=3.7,<3.8.0a0, but none of the providers can be installed

$ CONDA_SUBDIR=osx-arm64 mamba env create --name tmp --file devtools/conda-envs/test_env.yaml
conda-forge/osx-arm64                                       Using cache
conda-forge/noarch                                          Using cache
pkgs/main/noarch                                              No change
pkgs/r/osx-arm64                                              No change
openeye/osx-arm64                                             No change
pkgs/main/osx-arm64                                           No change
pkgs/r/noarch                                                 No change
openeye/noarch                                                No change


Looking for: ['python=3.10', 'pip', 'packaging', "openmm[version='>=7.7']", 'openff-forcefields', "openff-units[version='>=0.1.6']", 'openff-utilities', 'smirnoff99frosst', 'numpy', 'networkx', 'ambertools', 'rdkit', 'xmltodict', 'python-constraint', 'pytest', 'pytest-cov', 'pytest-xdist', 'pytest-rerunfailures', 'nbval', 'codecov', 'coverage', 'openeye-toolkits', 'mdtraj', 'cachetools', 'cached-property', 'pyyaml', 'toml', 'bson', 'msgpack-python', 'qcelemental', "qcportal[version='>=0.15']", 'qcengine', 'mdtraj', 'mypy', 'typing-extensions']


Encountered problems while solving:
  - nothing provides requested ambertools
  - nothing provides requested openeye-toolkits
  - package rdkit-2021.03.1-py38hbcbf861_0 requires python >=3.8,<3.9.0a0, but none of the providers can be installed
  ```

so on macOS, you can pick between `osx-64` which fails to provide OpenMM or `osx-arm64` which fails to provide any useful cheminformatics toolkit (sorry `BuiltInToolkitWrapper`).

I will drop Python 3.10 from the matrix and look to add it back when at least OpenMM's builds are re-created. I think dropping Python 3.7 is worth the change; even if it (temporarily?) drops our support to 2 versions, we can actually be confident we support them on 0.11.0 release and will be able to for a decent period of time into the future.

j-wags (approved):
Looks good - Per our discussion today let's keep our eyes open and add 3.10 when possible!

PR #1300, opened 2022-05-10
Update test environments

mattwthompson:
* I have tagged a new alpha version of Interchange (`v0.2.0-alpha.4`)
* https://github.com/openmm/openmm/issues/3495 is not resolved, but the broken packages are unavailable so we should be able to back the pins down to lower constraints
* It's possible that a recent version of `openmmforcefields` is compatible with our stack, so I'm giving that a spin

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -105,7 +105,7 @@ jobs:
           # force uninstall openff-toolkit-base since it's only pulled in when
           # Interchange is installed with conda. Un-comment this when testing
           # against a conda build of Interchange.
-          # conda remove --force openff-toolkit openff-toolkit-base
+          conda remove --force openff-toolkit openff-toolkit-base
           python setup.py develop --no-deps
 
       - name: Install test plugins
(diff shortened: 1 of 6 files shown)


mattwthompson:
These failures are confusing; numerous, nebulous, nothing obviously wrong in the logs, and not reproducible on my machine. I'll have a closer look later. My guess is that some change in Interchange is not accounted for here, but it's hard to say.

j-wags:
I think that re-enabling openmmforcefields may be pulling in the `openff-toolkit` conda packages. So I've re-enabled the command to remove those before the pip install.

j-wags:
Ok, I think that was it, and my last commit may get the unit tests passing again. The examples are failing because of a genuine incompatibility with openmmforcefields. So we probably need to pip install that from the [new branch](https://github.com/openmm/openmmforcefields/pull/191) (and rejigger the envs again since a pip install won't pull in the openff toolkit, oy vey!)

mattwthompson:
I was too ambitious here and blew up the initial objective. I should have thought of it given how many times I've run into it, my apologies for leaving it dangling when I could have fixed it.

PR #1303, opened 2022-05-19
Deprecate particles

mattwthompson:
Implements https://github.com/openforcefield/openff-toolkit/pull/1302#issuecomment-1132243705

There are a lot of method, variable, etc. names in the hierarchy implementation that name things `particle`, and I think maybe these should be `atom` instead?

- [x] Tag issue being addressed
- [x] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/tests)
- [x] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable
- [x] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [x] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/master/docs/releasehistory.rst)

PR #1310, opened 2022-06-02
Fix more calls to deprecated methods

mattwthompson:
This is a bit of a follow-on to #1303, some common toolkit calls result in invalid deprecation warnings.

### DIFF
--- a/openff/toolkit/topology/_mm_molecule.py
+++ b/openff/toolkit/topology/_mm_molecule.py
@@ -65,10 +65,6 @@ class _SimpleMolecule:
         return len(self.atoms)
 
     @property
-    def n_particles(self) -> int:
-        return len(self.atoms)
-
-    @property
     def n_bonds(self) -> int:
         return len(self.bonds)
 
@@ -97,10 +93,6 @@ class _SimpleMolecule:
 
     def particle(self, index) -> int:
         return self.atom(index)
-
-    @property
-    def particles(self):
-        return self.atoms
 
     def particle_index(self, particle) -> int:
         return self.atom_index(particle)
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -110,7 +110,7 @@ class Particle(Serializable):
 
     @property
     def molecule(self):
-        """
+        r"""
         The ``Molecule`` this particle is part of.
 
         .. todo::
@@ -178,7 +178,7 @@ class AtomMetadataDict(UserDict):
 
 class Atom(Particle):
     """
-    A particle representing a chemical atom.
+    A chemcical atom.
 
     .. todo::
 
@@ -513,7 +513,6 @@ class Atom(Particle):
     def molecule_atom_index(self):
         """
         The index of this Atom within the the list of atoms in the parent ``Molecule``.
-        Note that this can be different from ``molecule_particle_index``.
         """
         if self._molecule is None:
             raise ValueError("This Atom does not belong to a Molecule object")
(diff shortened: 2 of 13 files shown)


openff/toolkit/topology/_mm_molecule.py
```diff
@@ -98,10 +94,6 @@ def get_bond_between(self, atom1_index, atom2_index):
     def particle(self, index) -> int:
         return self.atom(index)
```
j-wags, line 96: We can delete all of these since `_mm_molecule` isn't in the public API.

openff/toolkit/topology/molecule.py
```diff
@@ -178,7 +178,7 @@ def __setitem__(self, key, value):
 
 class Atom(Particle):
     """
-    A particle representing a chemical atom.
+    A chemcical atom.
```
j-wags, line 181:
```suggestion
    A chemical atom.
```

openff/toolkit/topology/molecule.py
```diff
@@ -522,37 +521,15 @@ def molecule_atom_index(self):
...
-    #     Note that this can be different from ``particle_index``.
-    #
-    #     """
-    #     if self._topology is None:
-    #         raise ValueError('This Atom does not belong to a Topology object')
-    #     # TODO: This will be slow; can we cache this and update it only when needed?
-    #     #       Deleting atoms/molecules in the Topology would have to invalidate the cached index.
-    #     return self._topology.atoms.index(self)
+    # I can deprecate this instead of removing it ... but it seems incredibly unlikely to be used directly
+    # as part of the public API
```
j-wags, line 525: Agreed. A GitHub search for this comes up blank, so I don't think we need a deprecation warning. 

```suggestion
```

openff/toolkit/topology/molecule.py
```diff
@@ -5414,9 +5390,9 @@ def to_dict(self):
 
     def perceive_hierarchy(self):
         """
-        Groups the particles of the parent of this HierarchyScheme according to their
+        Groups the atoms of the parent of this HierarchyScheme according to their
         metadata, and creates HierarchyElements suitable for iteration over the parent.
-        Particles missing the metadata fields in this HierarchyScheme's
+        atoms missing the metadata fields in this HierarchyScheme's
```
j-wags, line 5395:
```suggestion
        Atoms missing the metadata fields in this HierarchyScheme's
```

openff/toolkit/topology/molecule.py
```diff
@@ -5483,16 +5459,16 @@ def __repr__(self):
...
 
 class HierarchyElement:
-    def __init__(self, scheme, identifier, particle_indices):
+    def __init__(self, scheme, identifier, atom_indices):
         """
         scheme : HierarchyScheme
         id : tuple of str and int
             uniqueness tuple
-        particle_indicess : iterable int
+        atom_indicess : iterable int
```
j-wags, line 5467:
```suggestion
        atom_indices : iterable of int
```

openff/toolkit/utils/builtin_wrapper.py
```diff
@@ -123,14 +122,12 @@ def assign_partial_charges(
...
             )
 
-        partial_charges = unit.Quantity(
-            np.zeros((molecule.n_particles)), unit.elementary_charge
-        )
         if partial_charge_method == "zeroes":
-            pass
+            partial_charges = unit.Quantity(
+                molecule.n_atoms * [0], unit.elementary_charge
+            )
```
j-wags, line 128: Nice - This lets us drop numpy as a dependency from the whole file!

```suggestion
            partial_charges = unit.Quantity(
                molecule.n_atoms * [0.], unit.elementary_charge
            )
```

openff/toolkit/utils/builtin_wrapper.py
```diff
@@ -123,14 +122,12 @@ def assign_partial_charges(
...
         elif partial_charge_method == "formal_charge":
-            for part_idx, particle in enumerate(molecule.particles):
-                partial_charges[part_idx] = particle.formal_charge
+            partial_charges = [atom.formal_charge for atom in molecule.atoms]
```
j-wags, line 130: I'm a little cautious here. Because the previous code instantiated this with `np.zeros`, these would have set to float64s, so I'd like to make sure that we don't change these to ints in the new code if we don't have to. 

```suggestion
            partial_charges = [float(atom.formal_charge) for atom in molecule.atoms]
```

j-wags (approved):
Looks great - Awesome cleanups. Left a few polishing suggestions, none are blocking. The ones about ensuring partial charges remain floats after removing `np.zeroes` are just short of blocking, though - I'd strongly prefer those changes go in.

openff/toolkit/topology/molecule.py
mattwthompson: Oof

openff/toolkit/topology/_mm_molecule.py
mattwthompson: Nice catch, I totally forgot about them and it didn't occur to me that they're not in any API.

PR #1311, opened 2022-06-03
Add FAQ section about `allow_undefined_stereo=True`

mattwthompson:
`allow_undefined_stereo` is not quite a secret argument but it something that commonly comes up in user support. I added a short blurb about it to the FAQ with some warnings about how it should generally be avoided.

While nervously checking for typos of my own (and boy did I find some) I also did a pass with `aspell check` and found some typos in other sections.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -64,7 +64,7 @@ The toolkit does not accept molecules with undefined stereochemistry. Applying p
 
 ## I understand the risks and want to load a molecule from a SMILES string that might not fully define stereochemistry.
 
-Pass `allow_undefined_stereo=True` to molecule loading methods like [Molecule.from_smiles](openff.toolkit.topology.Molecule.from_smiles) to downgrade the exception to a warning. As an example, see the "SMILES without stereochemistry" section in the [Molecule cookbook](users/molecule_cookbook).
+Pass `allow_undefined_stereo=True` to molecule loading methods like [Molecule.from_smiles](openff.toolkit.topology.Molecule.from_smiles) to downgrade the exception to a warning. As an example, see the "SMILES without stereochemistry" section in the [Molecule cookbook](smiles_no_stereochemistry).
 
 ## My conda installation of the toolkit doesn't appear to work. What should I try next?
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -58,6 +58,14 @@ If you are unable to provide a molecule in the formats recommended above and wan
 - [MDAnalysis' RDKit converter](https://docs.mdanalysis.org/stable/documentation_pages/converters/RDKit.html?highlight=rdkit#module-MDAnalysis.converters.RDKit), with an [example here](https://github.com/openforcefield/openff-toolkit/issues/1126#issuecomment-969712195)
 - the Jensen group's [xyz2mol program](https://github.com/jensengroup/xyz2mol/)
 
+## I'm getting stereochemistry errors when loading a molecule from a SMILES string.
+
+The toolkit does not accept molecules with undefined stereochemistry. Applying parameters based on [direct chemical perception](https://pubs.acs.org/doi/pdf/10.1021/acs.jctc.8b00640) requires knowing the stereochemistry of chiral centers. Molecule representations without defined stereochemistry might not be given all appropriate parameters specified from the force field, particularly valence terms, around the stereocenters.
+
+## I understsand the risks and want to load a molecule from a SMILES string that might not fully define stereocehmistry.
+
+Pass `allow_undefined_stereo=True` to molecule loading methods like [Molecule.from_smiles](openff.toolkit.topology.Molecule.from_smiles) to downgrade the exception to a warning. As an example, see the "SMILES without stereochemistry" section in the [Molecule cookbook](users/molecule_cookbook).
+
 ## My conda installation of the toolkit doesn't appear to work. What should I try next?
 
 We recommend that you install the toolkit in a fresh conda environment, explicitly passing the channels to be used, in-order:
--- a/FAQ.md
+++ b/FAQ.md
@@ -54,17 +54,15 @@ Essentially, anything which provides the full identity of what you want to simul
 
 If you are unable to provide a molecule in the formats recommended above and want to attempt to infer the bond orders and atomic formal charges, there are tools available elsewhere that can provide guesses for this problem. These tools are not perfect, and the inference problem itself is poorly defined, so you should review each output closely (see our [Core Concepts](users/concepts) for an explanation of what information is needed to construct an OpenFF Molecule). Some tools we know of include:
 
-- the OpenEye Toolkits' [`OEPerceiveBondOrders`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEPerceiveBondOrders.html) functionality
+- the OpenEye Toolkit's [`OEPerceiveBondOrders`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEPerceiveBondOrders.html) functionality
 - [MDAnalysis' RDKit converter](https://docs.mdanalysis.org/stable/documentation_pages/converters/RDKit.html?highlight=rdkit#module-MDAnalysis.converters.RDKit), with an [example here](https://github.com/openforcefield/openff-toolkit/issues/1126#issuecomment-969712195)
 - the Jensen group's [xyz2mol program](https://github.com/jensengroup/xyz2mol/)
 
 ## I'm getting stereochemistry errors when loading a molecule from a SMILES string.
 
-The toolkit does not accept molecules with undefined stereochemistry by default. This is because the stereochemistry of a molecule may affect its partial charges, and assigning parameters using [direct chemical perception](https://pubs.acs.org/doi/pdf/10.1021/acs.jctc.8b00640) may require knowing the stereochemistry of chiral centers. Because the OpenFF Toolkit doesn't know which force field will be used when molecule is loaded (and the main-line OpenFF force fields use AM1BCC for charge assignment, which is stereochemistry-dependent) , it defaults to throwing an error if a molecule with undefined stereochemistry is loaded. 
+The toolkit does not accept molecules with undefined stereochemistry by default. This is because the stereochemistry of a molecule may affect its partial charges, and assigning parameters using [direct chemical perception](https://pubs.acs.org/doi/pdf/10.1021/acs.jctc.8b00640) may require knowing the stereochemistry of chiral centers. Because the main-line OpenFF force fields use a stereochemistry-dependent charge generation method, the Toolkit defaults to throwing an error if a molecule with undefined stereochemistry is loaded. This behavior is in line with OpenFF's general attitude of requiring users to explicitly acknowledge actions that may cause silent errors later on.
 
-## I understand the risks and want to load a molecule from a SMILES string that might not fully define stereochemistry.
-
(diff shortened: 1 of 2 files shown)


FAQ.md
```diff
@@ -54,10 +54,18 @@ Essentially, anything which provides the full identity of what you want to simul
...
 - the Jensen group's [xyz2mol program](https://github.com/jensengroup/xyz2mol/)
 
+## I'm getting stereochemistry errors when loading a molecule from a SMILES string.
+
+The toolkit does not accept molecules with undefined stereochemistry. Applying parameters based on [direct chemical perception](https://pubs.acs.org/doi/pdf/10.1021/acs.jctc.8b00640) requires knowing the stereochemistry of chiral centers. Molecule representations without defined stereochemistry might not be given all appropriate parameters specified from the force field, particularly valence terms, around the stereocenters.
+
+## I understand the risks and want to load a molecule from a SMILES string that might not fully define stereochemistry.
+
+Pass `allow_undefined_stereo=True` to molecule loading methods like [Molecule.from_smiles](openff.toolkit.topology.Molecule.from_smiles) to downgrade the exception to a warning. As an example, see the "SMILES without stereochemistry" section in the [Molecule cookbook](users/molecule_cookbook).
```
mattwthompson, line 67: Not quite sure this will link well (and I don't think there's a way to link to a section of a Jupyter notebook); I saw `(see our [Core Concepts](users/concepts)` elsewhere and copied it.

Yoshanuikabundi: We user MyST-notebook in the toolkit, so you can use ordinary MyST code in the notebook. The most robust way to do this is to add a reference target to the notebook (inside the markdown cell):

```markdown
(smiles_wo_stereochem)=
### SMILES without stereochemistry
```

And then link to it:

```markdown
[Molecule cookbook](smiles_wo_stereochem)
```

This way, even if the notebook or section is renamed, moved, broken up etc. the link still works as long as the target is left in the right spot. Ordinarily I'd suggest letting MyST fill in the section name (`[](smiles_wo_stereochem)` renders as "SMILES without stereochemistry"), but in this case we want the FAQ to be degrade gracefully on GitHub, so link names should be explicit.

FAQ.md
```diff
@@ -54,10 +54,18 @@ Essentially, anything which provides the full identity of what you want to simul
...
 
-- the OpenEye toolkits' [`OEPerceiveBondOrders`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEPerceiveBondOrders.html) functionality
+- the OpenEye Toolkits' [`OEPerceiveBondOrders`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEPerceiveBondOrders.html) functionality
 - [MDAnalysis' RDKit converter](https://docs.mdanalysis.org/stable/documentation_pages/converters/RDKit.html?highlight=rdkit#module-MDAnalysis.converters.RDKit), with an [example here](https://github.com/openforcefield/openff-toolkit/issues/1126#issuecomment-969712195)
 - the Jensen group's [xyz2mol program](https://github.com/jensengroup/xyz2mol/)
 
+## I'm getting stereochemistry errors when loading a molecule from a SMILES string.
+
+The toolkit does not accept molecules with undefined stereochemistry. Applying parameters based on [direct chemical perception](https://pubs.acs.org/doi/pdf/10.1021/acs.jctc.8b00640) requires knowing the stereochemistry of chiral centers. Molecule representations without defined stereochemistry might not be given all appropriate parameters specified from the force field, particularly valence terms, around the stereocenters.
```
j-wags, line 63:
```suggestion
The toolkit does not accept molecules with undefined stereochemistry by default. This is because the stereochemistry of a molecule may affect its partial charges, and assigning parameters using [direct chemical perception](https://pubs.acs.org/doi/pdf/10.1021/acs.jctc.8b00640) may require knowing the stereochemistry of chiral centers. Because the OpenFF Toolkit doesn't know which force field will be used when molecule is loaded (and the main-line OpenFF force fields use AM1BCC for charge assignment, which is stereochemistry-dependent) , it defaults to throwing an error if a molecule with undefined stereochemistry is loaded. 
```

FAQ.md
```diff
@@ -54,10 +54,18 @@ Essentially, anything which provides the full identity of what you want to simul
...
+
+## I understand the risks and want to load a molecule from a SMILES string that might not fully define stereochemistry.
+
+Pass `allow_undefined_stereo=True` to molecule loading methods like [Molecule.from_smiles](openff.toolkit.topology.Molecule.from_smiles) to downgrade the exception to a warning. As an example, see the "SMILES without stereochemistry" section in the [Molecule cookbook](smiles_no_stereochemistry).
```
j-wags, line 67:
```suggestion
Pass `allow_undefined_stereo=True` to molecule loading methods like [Molecule.from_smiles](openff.toolkit.topology.Molecule.from_smiles) to downgrade the exception to a warning. As an example, see the "SMILES without stereochemistry" section in the [Molecule cookbook](smiles_no_stereochemistry). Where possible, our parameter assignment infrastructure will gracefully handle molecules with undefined stereochemistry. 
```

j-wags (commented):
I'd like @Yoshanuikabundi be the final reviewer on this, but I suggested adding some sentences with more clarity about our stereochemistry predicament.

FAQ.md
mattwthompson: Thanks! This is working in the PR builds now.

FAQ.md
```diff
@@ -54,10 +54,18 @@ Essentially, anything which provides the full identity of what you want to simul
...
 - [MDAnalysis' RDKit converter](https://docs.mdanalysis.org/stable/documentation_pages/converters/RDKit.html?highlight=rdkit#module-MDAnalysis.converters.RDKit), with an [example here](https://github.com/openforcefield/openff-toolkit/issues/1126#issuecomment-969712195)
 - the Jensen group's [xyz2mol program](https://github.com/jensengroup/xyz2mol/)
 
+## I'm getting stereochemistry errors when loading a molecule from a SMILES string.
+
+The toolkit does not accept molecules with undefined stereochemistry by default. This is because the stereochemistry of a molecule may affect its partial charges, and assigning parameters using [direct chemical perception](https://pubs.acs.org/doi/pdf/10.1021/acs.jctc.8b00640) may require knowing the stereochemistry of chiral centers. Because the OpenFF Toolkit doesn't know which force field will be used when molecule is loaded (and the main-line OpenFF force fields use AM1BCC for charge assignment, which is stereochemistry-dependent) , it defaults to throwing an error if a molecule with undefined stereochemistry is loaded.
```
Yoshanuikabundi, line 63:
```suggestion
## I'm getting stereochemistry errors when loading a molecule from a SMILES string.

The toolkit does not accept molecules with undefined stereochemistry by default. This is because the stereochemistry of a molecule may affect its partial charges, and assigning parameters using [direct chemical perception](https://pubs.acs.org/doi/pdf/10.1021/acs.jctc.8b00640) may require knowing the stereochemistry of chiral centers. Because the main-line OpenFF force fields use a stereochemistry-dependent charge generation method, the Toolkit defaults to throwing an error if a molecule with undefined stereochemistry is loaded. This behavior is in line with OpenFF's general attitude of requiring users to explicitly acknowledge actions that may cause silent errors later on.
```

I think this gets the central issue across with a little less convolution, and I think its useful to acknowledge this is a general principle of our tools.

FAQ.md
```diff
@@ -54,10 +54,18 @@ Essentially, anything which provides the full identity of what you want to simul
...
+
+## I understand the risks and want to load a molecule from a SMILES string that might not fully define stereochemistry.
+
+Pass `allow_undefined_stereo=True` to molecule loading methods like [Molecule.from_smiles](openff.toolkit.topology.Molecule.from_smiles) to downgrade the exception to a warning. As an example, see the "SMILES without stereochemistry" section in the [Molecule cookbook](smiles_no_stereochemistry). Where possible, our parameter assignment infrastructure will gracefully handle molecules with undefined stereochemistry that are loaded this way.
```
Yoshanuikabundi, line 67:
```suggestion
If you're confident a `Molecule` with unassigned stereochemistry is acceptable, pass `allow_undefined_stereo=True` to molecule loading methods like [Molecule.from_smiles](openff.toolkit.topology.Molecule.from_smiles) to downgrade the exception to a warning. As an example, see the "SMILES without stereochemistry" section in the [Molecule cookbook](smiles_no_stereochemistry). Where possible, our parameter assignment infrastructure will gracefully handle molecules with undefined stereochemistry that are loaded this way. 
```

I think this is better as a second paragraph in the same answer for a few reasons:

- Our FAQ already has a lot of questions, and we want to make it easy to find the one readers want
- The first answer implies that the default behavior is just a default, but gives no clues about how to override it
- When I read FAQs, I tend to look for the first entry that suits my question, and treat it as independent from the other questions. I can easily imagine readers reading the first answer, thinking "well that wasn't very useful", and missing the very next question.

FAQ.md
```diff
@@ -106,7 +114,7 @@ Where `get_my_new_force_field_paths` is a function in the `my_package` module pr
 
 ## What does "unconstrained" mean in a force field name?
 
-Each release of an [OpenFF force field](https://github.com/openforcefield/openff-forcefields/tree/master/openforcefields/offxml) has two associated `.offxml` files: one unadorned (for example, `openff-2.0.0.offxml`) and one labelled "unconstrained" (`openff_unconstrained-2.0.0.offxml`). This reflects the presence or absence of holonomic constraints on hydrogen-involving bonds in the force field specification.
+Each release of an [OpenFF force field](https://github.com/openforcefield/openff-forcefields/tree/master/openforcefields/offxml) has two associated `.offxml` files: one unadorned (for example, `openff-2.0.0.offxml`) and one labeled "unconstrained" (`openff_unconstrained-2.0.0.offxml`). This reflects the presence or absence of holonomic constraints on hydrogen-involving bonds in the force field specification.
```
Yoshanuikabundi, line 117: TIL a single 'l' is preferable in American English. Thanks!

Yoshanuikabundi (approved):
Looks good! This is an important addition, thanks @mattwthompson. One typo, and two optional rephrasings/reorganizations.

FAQ.md
mattwthompson: 👍

FAQ.md
```diff
@@ -54,10 +54,18 @@ Essentially, anything which provides the full identity of what you want to simul
 
 If you are unable to provide a molecule in the formats recommended above and want to attempt to infer the bond orders and atomic formal charges, there are tools available elsewhere that can provide guesses for this problem. These tools are not perfect, and the inference problem itself is poorly defined, so you should review each output closely (see our [Core Concepts](users/concepts) for an explanation of what information is needed to construct an OpenFF Molecule). Some tools we know of include:
 
-- the OpenEye toolkits' [`OEPerceiveBondOrders`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEPerceiveBondOrders.html) functionality
+- the OpenEye Toolkits' [`OEPerceiveBondOrders`](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEPerceiveBondOrders.html) functionality
```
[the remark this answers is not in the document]
mattwthompson: I take it to be plural when phrased this way because that's how they communicate it (https://docs.eyesopen.com/toolkits/python/index.html, https://anaconda.org/openeye/openeye-toolkits) as I'm not sure a single toolkit from them exists. I'm also not sure where one toolkit begins and the other ends ...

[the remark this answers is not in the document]
Yoshanuikabundi: Ok yeah good call 🙂

PR #1314, opened 2022-06-08
Hierarchy scheme ergonomics

Yoshanuikabundi:
This PR adds detailed docs to the hierarchy scheme objects, adds a `perceive_hierarchy()` call to the `Molecule.add_hierarchy_scheme()` method, and adds the residue hierarchy scheme at the end of the `perceive_residues()` method. This means that when a hierarchy is added, it is perceived automatically, and when residues are perceived, they can be iterated over immediately.

Tests are currently not passing locally - there seems to be an incorrect behaviour encoded into the tests (the residues of an ACE-ALA peptide are repeated in the iterator). I'm still working on this.


docs/topology.md
```diff
@@ -33,13 +33,9 @@ including offering serialization to a variety of standard formats (including [XM
...
-    VirtualSite
-    VirtualParticle
-    BondChargeVirtualSite
-    MonovalentLonePairVirtualSite
-    DivalentLonePairVirtualSite
-    TrivalentLonePairVirtualSite
     ValenceDict
     ImproperDict
+    ~molecule.HierarchyScheme
+    ~molecule.HierarchyElement
```
j-wags, line 39: Let's re-export this at the `topology` level so that we don't have to do the `~molecule` thing here

openff/toolkit/topology/molecule.py
```diff
@@ -1420,14 +1476,19 @@ def delete_hierarchy_scheme(self, iter_name):
 
     def perceive_hierarchy(self, iter_names=None):
```
j-wags, line 1477: Let's refactor this to be called `update_hierarchy_schemes`

openff/toolkit/topology/molecule.py
```diff
@@ -5508,26 +5638,23 @@ def to_dict(self):
         return return_dict
 
     @property
-    def particles(self):
+    def particles(self) -> Particle:
```
j-wags, line 5641:
```suggestion
    def particles(self) -> Generator[Particle]:
```

j-wags (commented):
Thanks for the discussion today. I think we agreed that there were some test/behavior issues that still need to be shaken out. I left notes about the renamings we talked about. Happy to review this later when it's ready!

openff/toolkit/topology/molecule.py
mattwthompson: We're hoping to have this deprecated or removed (I'm not actually sure which). Ref #1303 / #1310

j-wags: Oof, yeah this found a conflict really quickly. This conflicts directly with [this diff](https://github.com/openforcefield/openff-toolkit/pull/1310/files#diff-68cb3f9873e00994ccf957cc089b18cbaf34302d5eeb90c34ca289803e10a69eL5511), where we will no longer allow the user to manipulate virtual sites before parameterization, so there's no sense in talking about `Molecule.particles` because all of the particles must be atoms. Since this never reached the public API, @mattwthompson's changes should take precedence - This conflict doesn't block merging #1310, and this PR should be updated to use `HierarchyElement.atoms` instead of `.particles`.

Yoshanuikabundi:
OK this testing bug is WEIRD. I've tracked it down to the behaviour of the standard library's `copy.deepcopy()` differing based on a single line in `openff/toolkit/topology/molecule.py`

On 29787e7, `deepcopy` does not produce an exact copy, but adds a repeat of all residues in the residues hierarchy scheme:

```python
>>> from copy import deepcopy
>>> mol = Molecule.from_smiles('[H][C@@](C(=O)[O-])(C([H])([H])[H])N([H])C(=O)C([H])([H])[H]')
>>> mol.perceive_residues()
>>> mol.update_hierarchy_schemes() 
>>> mol.residues
[HierarchyElement ('None', 1, 'ALA') of iterator 'residues' containing 11 particle(s),
 HierarchyElement ('None', 2, 'ACE') of iterator 'residues' containing 6 particle(s)]
>>> deepcopy(mol).residues
[HierarchyElement ('None', 1, 'ALA') of iterator 'residues' containing 11 particle(s),
 HierarchyElement ('None', 2, 'ACE') of iterator 'residues' containing 6 particle(s),
 HierarchyElement ('None', 1, 'ALA') of iterator 'residues' containing 11 particle(s),
 HierarchyElement ('None', 2, 'ACE') of iterator 'residues' containing 6 particle(s)]
```

But if I comment out line 1436 in molecule.py:

```diff
          self._hierarchy_schemes[iterator_name] = new_hier_scheme
-         self.update_hierarchy_schemes([iterator_name])
+         # self.update_hierarchy_schemes([iterator_name])
          return new_hier_scheme
```

then the deepcopy is correct:

```python
>>> from copy import deepcopy
>>> mol = Molecule.from_smiles('[H][C@@](C(=O)[O-])(C([H])([H])[H])N([H])C(=O)C([H])([H])[H]')
>>> mol.perceive_residues()
>>> mol.update_hierarchy_schemes() 
>>> mol.residues
[HierarchyElement ('None', 1, 'ALA') of iterator 'residues' containing 11 particle(s),
 HierarchyElement ('None', 2, 'ACE') of iterator 'residues' containing 6 particle(s)]
>>> deepcopy(mol).residues
[HierarchyElement ('None', 1, 'ALA') of iterator 'residues' containing 11 particle(s),
 HierarchyElement ('None', 2, 'ACE') of iterator 'residues' containing 6 particle(s)]
```

`__deepcopy__` is not defined on `Molecule` nor anywhere else in the module, and `.residues` is just a list. Calling 
`update_hierarchy_schemes()` on the copy fixes the problem and the copy doesn't seem to be otherwise different - for example, `deepcopy(mol)==mol` returns `True`. Copying the molecule with `Molecule(mol)` has the same behaviour. This strange doubling has been present in every commit since I added `update_hierarchy_schemes()` to `add_hierarchy_scheme()`, though the above scripts might need to be changed a bit as the names of some of these methods has changed in that time. Changing the hierarchy schemes from adding an attribute for iterators to implementing __getattr__ and __dir__ doesn't make a difference either (see eg. commit 24a26361 for the old strategy).

The elements in the `HierarchyScheme` have been duplicated as well, but deep copying the hierarchy scheme itself does not produce the duplication.

@mattwthompson @j-wags any ideas? I'm stumped.

mattwthompson:
I think the problem is somewhere in or near `FrozenMolecule._initialize_from_dict`. Making the deep copy of the dict doesn't seem to obviously be duplicating the residues in dictionary form, unless I'm missing something obvious:

```
In [47]: deepcopy(mol.to_dict())['hierarchy_schemes']
Out[47]:
{'residues': {'uniqueness_criteria': ('chain',
   'residue_number',
   'residue_name'),
  'iterator_name': 'residues',
  'hierarchy_elements': [{'identifier': ('None', 1, 'ALA'),
    'particle_indices': [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]},
   {'identifier': ('None', 2, 'ACE'),
    'particle_indices': [11, 12, 13, 14, 15, 16]}]}}
```

I spent a bit of time with the loop around L1252 and found that removing the calls to `HierarchyScheme.add_hierarchy_element` fixes your minimal reproduction. 

```diff
diff --git a/openff/toolkit/topology/molecule.py b/openff/toolkit/topology/molecule.py
index 651abdce..547a912f 100644
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -1257,10 +1257,6 @@ class FrozenMolecule(Serializable):
                 iter_name,
             )
             # hierarchy_scheme = self._hierarchy_schemes[iter_name]
-            for element_dict in hierarchy_scheme_dict["hierarchy_elements"]:
-                new_hier_scheme.add_hierarchy_element(
-                    element_dict["identifier"], element_dict["particle_indices"]
-                )

     def __repr__(self):
         """Return a summary of this molecule; SMILES if valid, Hill formula if not."""
```
It does't make any sense to me, however, as I don't see where else the hierarchy elements get added, nor do I see where else they'd get added. (If I'm understanding `add_hierarchy_scheme` correctly, it adds the machinery for the object to understand what a residue is, but not the actual data of what the residues actually are.)

The test still fails, although in a different way:

```
(openff-dev) [openff-toolkit] python -m pytest openff/toolkit/tests/test_topology.py -vv --lf          14:12:43  ☁  29787e74 ☂ ⚡ ✭
	======================================================= test session starts ========================================================
platform darwin -- Python 3.9.13, pytest-7.1.2, pluggy-1.0.0 -- /Users/mattthompson/miniconda3/envs/openff-dev/bin/python
cachedir: .pytest_cache
Using --randomly-seed=1091171244
rootdir: /Users/mattthompson/software/openff-toolkit
plugins: anyio-3.6.1, xdist-2.5.0, forked-1.4.0, rerunfailures-10.2, nbval-0.9.6, qcfractal-0.15.8.1, randomly-3.12.0, cov-3.0.0
collected 65 items / 64 deselected / 1 selected
run-last-failure: rerun previous 1 failure

openff/toolkit/tests/test_topology.py::TestTopology::test_topology_hierarchy_iterators FAILED                                [100%]

============================================================= FAILURES =============================================================
__________________________________________ TestTopology.test_topology_hierarchy_iterators __________________________________________

self = <toolkit.tests.test_topology.TestTopology object at 0x178719a90>

    def test_topology_hierarchy_iterators(
        self,
    ):
        top = Topology()
        # Ensure that an empty topology has no residues defined
        residues = list(top.hierarchy_iterator("residues"))
        assert residues == []
        # Ensure that a topology with no metadata has no residues defined
        top.add_molecule(dipeptide())
        residues = list(top.hierarchy_iterator("residues"))
        assert residues == []
        # Ensure that a topology with residues perceived has residues but not chains
        top.add_molecule(dipeptide_residues_perceived())
        residues = list(top.hierarchy_iterator("residues"))
        chains = list(top.hierarchy_iterator("chains"))
        assert chains == []
        assert [res.identifier for res in residues] == [
            ("None", 1, "ACE"),
            ("None", 2, "ALA"),
        ]
        # Ensure that adding molecules WITH hierarchy perceived DOES give the
        # topology residues and chains to iterate over
        top.add_molecule(dipeptide_hierarchy_added())
        top.add_molecule(cyx_hierarchy_added())
        residues = list(top.hierarchy_iterator("residues"))
        chains = list(top.hierarchy_iterator("chains"))
>       assert [res.identifier for res in residues] == [
            ("None", 1, "ACE"),
            ("None", 2, "ALA"),
            ("None", 1, "ACE"),
            ("None", 2, "ALA"),
            ("None", 4, "ACE"),
            ("None", 5, "CYS"),
            ("None", 6, "NME"),
        ]
E       AssertionError: assert [('None', 1, 'ACE'),\n ('None', 2, 'ALA'),\n ('None', 1, 'ACE'),\n ('None', 2, 'ALA'),\n ('None', 1, 'ACE'),\n ('None', 2, 'CYS'),\n ('None', 3, 'NME'),\n ('None', 4, 'ACE'),\n ('None', 5, 'CYS'),\n ('None', 6, 'NME')] == [('None', 1, 'ACE'),\n ('None', 2, 'ALA'),\n ('None', 1, 'ACE'),\n ('None', 2, 'ALA'),\n ('None', 4, 'ACE'),\n ('None', 5, 'CYS'),\n ('None', 6, 'NME')]
E         At index 4 diff: ('None', 1, 'ACE') != ('None', 4, 'ACE')
E         Left contains 3 more items, first extra item: ('None', 4, 'ACE')
E         Full diff:
E           [
E            ('None',
E             1,
E             'ACE'),
E            ('None',
E             2,
E             'ALA'),
E            ('None',
E             1,
E             'ACE'),
E            ('None',
E             2,
E             'ALA'),
E            ('None',
E         +   1,
E         +   'ACE'),
E         +  ('None',
E         +   2,
E         +   'CYS'),
E         +  ('None',
E         +   3,
E         +   'NME'),
E         +  ('None',
E             4,
E             'ACE'),
E            ('None',
E             5,
E             'CYS'),
E            ('None',
E             6,
E             'NME'),
E           ]

openff/toolkit/tests/test_topology.py:1094: AssertionError
========================================================= warnings summary =========================================================
openff/toolkit/tests/test_topology.py::TestTopology::test_topology_hierarchy_iterators
  /Users/mattthompson/software/openff-toolkit/openff/toolkit/topology/molecule.py:92: MoleculeDeprecationWarning: Molecule.particles is deprecated. Use Molecule.atoms instead.
    warnings.warn(

openff/toolkit/tests/test_topology.py: 40 warnings
  /Users/mattthompson/miniconda3/envs/openff-dev/lib/python3.9/site-packages/packaging/version.py:111: DeprecationWarning: Creating a LegacyVersion has been deprecated and will be removed in the next major release
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
===================================================== short test summary info ======================================================
FAILED openff/toolkit/tests/test_topology.py::TestTopology::test_topology_hierarchy_iterators - AssertionError: assert [('None', ...
========================================== 1 failed, 64 deselected, 41 warnings in 1.85s ===========================================
```

I apologize for not understanding this code well enough to say if that's an improvement or not.

I don't see any obvious paths the elements could get added twice; `FrozenMolecule.__init__` does nothing once it's found it's loaded something and the loop around L1252 really looks like it can only add the elements once. Overriding `__setstate__` and `__getstate__` makes me nervous but I can't isolate that as the root cause. The best I can offer at the moment is showing that deleting those lines gets your simple case working, but I can't really make sense of it.

```python3
from copy import deepcopy

from openff.toolkit.topology.molecule import FrozenMolecule, Molecule

mol = Molecule.from_smiles(
    "[H][C@@](C(=O)[O-])(C([H])([H])[H])N([H])C(=O)C([H])([H])[H]"
)
mol.perceive_residues()
mol.update_hierarchy_schemes()

# Just make sure residue scheme didn't get added twice or anything / general input check
assert len(mol.to_dict()["hierarchy_schemes"]) == 1

# And make sure the deepcopy of the dict only has the two elements (residues)
assert len(mol.to_dict()["hierarchy_schemes"]["residues"]["hierarchy_elements"]) == 2

f = FrozenMolecule()
f._copy_initializer(mol)

assert len(f.residues) == 2, len(f.residues)
```

```shell
(openff-dev) [openff-toolkit] python 1314.py                                                              14:27:17  ☁  29787e74 ☂ ✭
(openff-dev) [openff-toolkit] git reset --hard                                                         14:27:43  ☁  29787e74 ☂ ⚡ ✭
HEAD is now at 29787e74 Avoid overwriting existing methods with hierarchy iterators
(openff-dev) [openff-toolkit] python 1314.py                                                              14:27:45  ☁  29787e74 ☂ ✭
  File "/Users/mattthompson/software/openff-toolkit/1314.py", line 17, in <module>
    assert len(f.residues) == 2, len(f.residues)
AssertionError: 4

**Comment by Yoshanuikabundi** [comment]
Ahhhh thank you!! I think I know the problem: Extending the bit you changed looks like this:

```python

        for iter_name, hierarchy_scheme_dict in molecule_dict[
            "hierarchy_schemes"
        ].items():
            new_hier_scheme = self.add_hierarchy_scheme(
                hierarchy_scheme_dict["uniqueness_criteria"],
                iter_name,
            )
            # hierarchy_scheme = self._hierarchy_schemes[iter_name]
            for element_dict in hierarchy_scheme_dict["hierarchy_elements"]:
                new_hier_scheme.add_hierarchy_element(
                    element_dict["identifier"], element_dict["atom_indices"]
                )
```

So right before the hierarchy elements are added, the new hierarchy scheme is perceived by my changes inside the call to `add_hierarchy_scheme`. So the first copy of the residues is from perceiving them and the second is from the dictionary! They just happen to be the same when the source hierarchy schemes are up to date.

Thanks for your help @mattwthompson!!! I can fix this by clearing the hierarchy schemes or adding some way to add a hierarchy scheme without perceiving it.

Yoshanuikabundi:
@j-wags I think this is ready for review :)

mattwthompson:
I should mention I have a change coming down the pipeline in #1320 that might possibly interact with this; to be cautious I made a new branch based off of that and these changes (#1321) and it does not seem to vocally break anything.

openff/toolkit/topology/molecule.py
```diff
@@ -1380,31 +1437,42 @@ def delete_hierarchy_scheme(self, iter_name):
...
             hierarchy_scheme.perceive_hierarchy()
-            self._expose_hierarchy_scheme(iter_name)
 
-    def _expose_hierarchy_scheme(self, iter_name):
-        assert iter_name in self._hierarchy_schemes
-        setattr(self, iter_name, self._hierarchy_schemes[iter_name].hierarchy_elements)
+    def __getattr__(self, name: str):
+        """If a requested attribute is not found, check the hierarchy schemes"""
+        try:
+            return self.__dict__["_hierarchy_schemes"][name].hierarchy_elements
```
j-wags, line 1467: (not blocking) I see that calling to `__dict__` in `__getattr__` is the recommended use, however this spooked me a little bit because I hadn't considered the difference between `__dict__` and `to_dict`. 

The way that a lot of our serialization works is that we call `to_dict`, which recursively converts the object and everything it "owns" to a serializable representation (consisting only of dicts, lists, tuples, ints, strings, and floats). Some things that are attributes of an object aren't "owned" by the object. For example, I can call `hierarchy_scheme.parent` and get a Molecule, but when I call `hierarchy_scheme.to_dict()`, there is no `parent` in the keys of the resulting dict. The deliberate separation of "attributes" from "things that get serialized in `to_dict`" is important to prevent infinite recursion (which I see is causing problems recently, though this particular situation may not be the root cause). 

It would be bad if `to_dict` were called here, since then this wouldn't return a list of `HierarchyElements`, instead it would return a list of _serialized_ `HierarchyElements` (that is, dicts). 

So, in the end, I think I've convinced myself that this is OK, and probably even correct. Thanks for iterating on this!

openff/toolkit/topology/molecule.py
```diff
@@ -1223,16 +1223,17 @@ def _initialize_from_dict(self, molecule_dict):
         for iter_name, hierarchy_scheme_dict in molecule_dict[
             "hierarchy_schemes"
         ].items():
-            new_hier_scheme = self.add_hierarchy_scheme(
+            new_hier_scheme = HierarchyScheme(
```
j-wags, line 1226: (not blocking) Good job on this whole section - It's possible that the molecule that we're deserializing here had its HierarchyElements defined, and then later had some `atom.metadata` info changed to that the original HierarchyElement definition was no longer current. In that case it would be important that we DON'T re-perceive the hierarchy during deserialization, and instead initialize the molecule to have the "outdated" hierarchy information. Nice work!
```suggestion
            # It's important that we do NOT call `add_hierarchy_scheme` here, since we 
            # need to deserialize these HierarchyElements exactly as they were serialized,
            # even if that conflicts with the current values in atom metadata. 
            new_hier_scheme = HierarchyScheme(
```

j-wags (approved):
Thanks for the wonderful work here, @Yoshanuikabundi and @mattwthompson. I went down the non-orthogonal rabbit holes of "good serialization practices" and "however it is that our serialization works" this morning, and while they're not the same thing, the recent PRs do bring them much closer. Also, these are MASSIVE improvements to the docs, and I really appreciate the attention to detail @Yoshanuikabundi!

openff/toolkit/topology/molecule.py
```diff
@@ -1328,17 +1354,39 @@ def add_hierarchy_scheme(
         """
         Use the molecule's metadata to facilitate iteration over its atoms.
 
+        This method will add an attribute with the name given by the
+        ``iterator_name`` argument that provides an iterator over groups of
+        atoms. Atoms are grouped by the values in their ``atom.properties``
```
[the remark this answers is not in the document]
Yoshanuikabundi: How did I get that mixed up... is this a recent change?

openff/toolkit/topology/molecule.py
Yoshanuikabundi: Yeah I think `__dict__` and `to_dict` are two completely different things - the former is just the dict that every Python object uses to store its attributes, and the latter is for serialization.

openff/toolkit/topology/molecule.py
Yoshanuikabundi: Yeah, that was exactly the bug that I needed help from Matt with - it was leading to a duplication of the hierarchy schemes in tests!

openff/toolkit/topology/molecule.py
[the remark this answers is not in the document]
Yoshanuikabundi: Ahhh it's `Molecule.properties` but `Atom.metadata`

Issue #1136, opened 2021-11-29
Molecule attributes in multiprocessing lost


schallerdavid:
Dear openff team,

we are really happy with the openff-toolkit and successfully use it in our KinoML framework. Lately, we used the openff Molecule class in some experiments with multiprocessing ([#97](https://github.com/openkinome/kinoml/issues/97)). We were surprised to see that custom attributes added to Molecule instances get lost when doing multiprocessing. Some online research pointed us to possible serialization issues of the Molecule class. Here is a simple script to reproduce the behavior:

```
from multiprocessing import Pool

from openff.toolkit.topology import Molecule


def dummy_function(molecule):
    return molecule


if __name__ == '__main__':

    molecules = []
    for i, smiles in enumerate(["CCC", "CCCC", "COC"]):
        molecule = Molecule.from_smiles(smiles)
        molecule.custom = i
        molecules.append(molecule)

    print("Getting attributes without multiprocessing ...")
    for molecule in molecules:
        print(molecule.custom)

    print("Doing some multiprocessing ...")
    with Pool(processes=2) as pool:
        molecules = pool.map(dummy_function, molecules)

    print("Getting attributes after multiprocessing ...")
    for molecule in molecules:
        print(molecule.custom)

    print("Finished!")
```
After doing some multiprocessing, the "custom" attribute is lost and we receive an `AttributeError`.

We actually also had problems in terms of serialization and came up with a strategy to serialize everything that is serializable. Maybe this could be a workaround to be considered ([see here](https://github.com/openkinome/kinoml/blob/master/kinoml/features/core.py#L235)).

Thanks for looking into this,
David

mattwthompson:
Thanks for the report; the problem seems to be that the custom arguments are lost during the pickle roundtrip.

```python3
>>> import pickle
>>> from openff.toolkit.topology import Molecule
>>> molecule = Molecule.from_smiles("O")
>>> molecule.custom = 4
>>> pickle.loads(pickle.dumps(molecule))
Molecule with name '' and SMILES '[H]O[H]'
>>> pickle.loads(pickle.dumps(molecule)).custom
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'Molecule' object has no attribute 'custom'
```

This seems to be a common feature of pickling; we don't have much custom code there. We do have other serialization avenues, but unfortunately they also don't keep track of these at the moment, either:

```python3
[...]

PR #1318, opened 2022-06-09
Do not override state dunder methods in `FrozenMolecule`

mattwthompson:
This PR reverts the overrides of `FrozenMolecule.__setstate__` and `FrozenMolecule.__getstate__`. The [history](https://github.com/openforcefield/openff-toolkit/blame/topology-biopolymer-refactor/openff/toolkit/topology/molecule.py) indicates these have not been touched since the early days of OpenFF and were done in an effort to [support serialization pathways](https://github.com/openforcefield/openff-toolkit/commit/eb82878f81b87085b9f9b48f4ce4e0b7b3b702c8). This causes issues with [pickling](https://github.com/openforcefield/openff-toolkit/issues/1136) and [might](https://github.com/openforcefield/openff-toolkit/pull/1314#issuecomment-1150892719) impact some copying behavior; it's called in [a few](https://github.com/python/cpython/blob/3.10/Lib/copyreg.py#L80) [places](https://github.com/python/cpython/blob/3.10/Lib/copy.py#L159) in the standard library that I don't fully understand but make me nervous about overriding dunder methods for features that do not require it.

This is a small change but one that would make me sleep easier at night; I think there should be a moderately high bar for overriding default methods like these and I don't think that's met in this case.

Credit to @lilyminium who discovered this about a year before I did (which is more than enough time for me to forget what I've read): https://github.com/openforcefield/openff-toolkit/issues/1136#issuecomment-981789758


j-wags (approved):
Ok, this is a scary piece of debt to take out (since I don't know if this is "load bearing debt", and removing it is going to break or corrupt something, somewhere that we don't know to test). But it's clearly a bad practice/debt, and this is our one API break this year (well, year and a half at this pace), so lets go ahead and remove it. If this breaks something downstream we can try to catch it in the RC period and revert this PR.

mattwthompson:
Thanks!

If a future reader finds that this broke something please ping me and I'll address it.

PR #1325, opened 2022-06-21
Update virtual site notebook

mattwthompson:


mattwthompson:
Something's wrong with either the SMIRNOFF definitions used here or the charges as assigned:

image

mattwthompson:
Above should be fixed by https://github.com/openforcefield/openff-interchange/pull/480

examples/virtual_sites/vsite_showcase.ipynb
```diff
@@ -10,17 +10,16 @@
    },
```
j-wags, line 10: <pre class="ql-syntax" spellcheck="false">Line #1.    molecule = Molecule.from_smiles("c1cc(-Cl)ccc1C(=O)CS[C]1=CO[C](F)(F)CC1")
(not blocking) It's strange that this SMILES contains <code>(-Cl)</code> and <code>(F)</code> - Why is the single bond explicit in the first but not in the second?
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1325/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

j-wags, line 10: <pre class="ql-syntax" spellcheck="false">Line #18.    positions = to_openmm(
</pre><p>(not blocking) This has become a chonker of a line, can it be broken into a few lines to improve readability?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1325/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

j-wags, line 10: <pre class="ql-syntax" spellcheck="false">Line #24.        # IMPORTANT! The new system has the virtual sites, but the oFF top does not as there is no API yet to do so
</pre><p>This could use another look - Maybe interchange can handle this more cleanly at this time. Would be nice to update this for future us, even if we don't change the code (also if we don't need to set <code>allow_nonintegral_charges</code> to a non-default value below, we can remove mention of it too) </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1325/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

j-wags (approved):
Left a few small suggestions, but I'm mostly concerned that the output numbers at the very bottom of the notebook changed - This may indicate that something was broken before this PR, after this PR, or both. Could you certify that the current behavior is sane before merging? In particular, I wonder why:

* the coordinate norm at the very end changed, and 
* whether the new vsite behavior completely zeroes all intramolecular vdW and ES, so we could be missing a significant energy difference. This could be inspected by increasing Nx to capture intermolecular interactions in the outputs.

mattwthompson:
Here are the different `State` object generated by each code path:

[openff.xml.txt](https://github.com/openforcefield/openff-toolkit/files/9426361/openff.xml.txt)
[openmm.xml.txt](https://github.com/openforcefield/openff-toolkit/files/9426362/openmm.xml.txt)


I'm working on figuring out if there are any substantive differences between them or if the energy difference (-0.2 kJ/mol) is due to something else.

mattwthompson:
By manual inspection:

* 🟢 Periodic boxes are precisely identical
* 🟢 Particle masses are precisely identical
* 🟡 Virtual sites use different types
  * OpenMM uses `OutOfPlaneSite`
  * OpenFF uses `LocalCoordinatesSite`
* 🟢 Non-bonded parameters are precisely identical
* 🟢 Non-bonded settings are precisely identical
  * `diff *xml | grep Nonb` returns nothing
* 🟢 Angle parameters are precisely identical
* 🟢 Bond parameters are precisely identical
* 🟢 Exceptions differ only in unimportant ways
  * Ordering of each exception differs
  * Some exceptions have non-zero sigma, but all have zero epsilon
* 🟢 The `CMMotionRemover` differs in that OpenMM generates one but OpenFF does not.
  * However, removing this force does not affect the energy

mattwthompson:
As far as I'm aware, OpenMM does not have a clean way to serialize their `Topology` object for easy comparison, nor do I know how to quickly compare their states in memory. It's not intuitive to me that there could be a difference in the topologies causing a small difference in energies.

The positions passed to `sim.context.setPositions(particle_positions)` look the same to me:

OpenFF:
```
[[ 0.0061120212  0.0388656706  0.0058900423]
 [ 0.0723783553 -0.0311629355 -0.0038235087]
 [-0.0784903705 -0.0077027328 -0.0020665338]
 [ 0.0000000000  0.0000000000  0.0000000000]
 [ 0.0000000000  0.0000000000  0.0000000000]
 [ 0.3061120212  0.0388656706  0.0058900423]
 [ 0.3723783553 -0.0311629355 -0.0038235087]
 [ 0.2215096295 -0.0077027328 -0.0020665338]
 [ 0.0000000000  0.0000000000  0.0000000000]
 [ 0.0000000000  0.0000000000  0.0000000000]] nm
```
OpenMM:
```
[[ 0.0061120212  0.0388656706  0.0058900423]
 [ 0.0723783553 -0.0311629355 -0.0038235087]
 [-0.0784903705 -0.0077027328 -0.0020665338]
 [ 0.0000000000  0.0000000000  0.0000000000]
 [ 0.0000000000  0.0000000000  0.0000000000]
 [ 0.3061120212  0.0388656706  0.0058900423]
 [ 0.3723783553 -0.0311629355 -0.0038235087]
 [ 0.2215096295 -0.0077027328 -0.0020665338]
 [ 0.0000000000  0.0000000000  0.0000000000]
 [ 0.0000000000  0.0000000000  0.0000000000]] nm
```

And that's all the information passed to each `_evaluate_positions_and_energy` call.

mattwthompson:
I've pushed a dev-only version of my local notebook to a separate branch: 47088c712d8b282c56764b617bd39e6000d88e2b

j-wags:
Thanks for all the careful work here. I'm looking into this now.

mattwthompson:
FWIW I'm more than open to refactoring the Interchange code to use `openmm.OutOfPlaneSite` and friends

mattwthompson:
@j-wags and I worked on this for a while, looking side-by-side-by-side at the OpenMM implementation and the 0.10.6 version of this notebook. We have verified that every parameter has only the expected changes - only the differences in atom indexing that are expected. We aren't completely sure that the systems generated by OpenMM's TIP5P force field produce identical physics as ours, but we have generally high confidence that the OpenMM developers have accurately implemented TIP5P. As an extra measure, @j-wags did some back-of-the-envelope math, starting with the `openmm.System` XML file produced by a version of this notebook and calculating the geometries of the resulting water, including virtual sites. He found that the distance between the oxygen and each virtual site is 0.7 Angstroms, which is precisely what is specified in [the model](https://en.wikipedia.org/wiki/Water_model#5-site). So, running with the reasonable assumption that OpenMM's TIP5P is faithful to the source, all three code paths should be producing identical physics. This is basically my traffic light comment above, except with a green light on the virtual sites lines.

It's still somewhat mysterious why the coordinates and energies are different - by numbers that are too large to totally brush away - but at this point I'm willing to chalk this up to something _other_ than the produced `System` objects (even though we're using the same functions to run the simulations and passing them the same positions).

j-wags (dismissed):
LGTM! Still not sure why we're getting slightly different coords+energies than openMM, but our implementation seems to match 0.10.6, and doing the math by hand shows us defining these vsites with the right geometries. If this ends up being a bigger deal we can dive deeper after 0.11.0!

mattwthompson:
Let me just say again I despise GitHub's new rules for "dismissing" "stale" reviews. I understand that reviews from prior commits don't necessarily apply to future states, but the language they added to the UI is way too strong ...

mattwthompson:
Everything is passing except for one conda-related failure that I expect to be spurious. Merging!

PR #1330, opened 2022-06-24
Warn when using AM1-BCC or similar methods on large molecules

mattwthompson:
There is no linked issue for this, but it's a common pain point for which emitting a warning is better than beefing up the FAQ (#1272).

I decided to put this in `Molecule.assign_partial_charges` directly since that will delegate out to the underlying calls and it's how `Interchange` [currently calls them](https://github.com/openforcefield/openff-interchange/blob/d4aaec28f036a3361f90b7deb0610bfee7cfb306/openff/interchange/components/smirnoff.py#L1183-L1190). The other approach would be to add it to each `ToolkitWrapper.assign_partial_charges` but I think that's unnecessarily specific (a user calling a wrapper directly is more likely to know about this limitation, as distinct from the typical user that sees `ForceField.create_openmm_system` hang) and more likely to fall out of sync with increasing toolkit functionality.

I didn't write a test for this, since it would require a little bit of kludging  - this warning is only hit when entering a dubious state, and getting there without causing other issues like a hanging unit test is tricky. But I have an idea how to do it, if that is desired.

I tested locally like so:

```python3

In [1]: SMILES = 10 * "C(CCCC)CCC"

In [2]: from openff.toolkit import Molecule

In [3]: molecule = Molecule.from_smiles(SMILES, allow_undefined_stereo=True)
Warning (not error because allow_undefined_stereo=True): OEMol has unspecified stereochemistry. oemol.GetTitle():
Problematic atoms are:
<SNIP>

In [4]: molecule.assign_partial_charges(partial_charge_method='am1bccelf10')
/Users/mattthompson/software/openff-toolkit/openff/toolkit/topology/molecule.py:2459: UserWarning: Warning! Partial charge method 'am1bccelf10' is not designed for use on large (i.e. > 150 atoms) molecule and may crash or take hours to run on this molecule (found 242 atoms). For more, see https://docs.openforcefield.org/projects/toolkit/en/topology-biopolymer-refactor/faq.html#parameterizing-my-system-which-contains-a-large-molecule-is-taking-forever-whats-wrong
  warnings.warn(
Warning: : Failed due to unspecified stereochemistry
^C^C^C[1]    20109 terminated  ipython
```

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -2444,6 +2444,26 @@ class FrozenMolecule(Serializable):
         openff.toolkit.utils.toolkits.AmberToolsToolkitWrapper.assign_partial_charges
         openff.toolkit.utils.toolkits.BuiltInToolkitWrapper.assign_partial_charges
         """
+
+        # Raise a warning when users try to apply these charge methods to "large" molecules
+        WARN_LARGE_MOLECULES: Set[str] = {
+            "am1bcc",
+            "am1bccelf10",
+            "am1-mulliken",
+            "am1bccnosymspt",
+            "am1elf10",
+        }
+
+        if partial_charge_method in WARN_LARGE_MOLECULES:
+            if self.n_atoms > 150:
+                warnings.warn(
+                    f"Warning! Partial charge method '{partial_charge_method}' is not designed "
+                    "for use on large (i.e. > 150 atoms) molecule and may crash or take hours to "
+                    f"run on this molecule (found {self.n_atoms} atoms). For more, see "
+                    "https://docs.openforcefield.org/projects/toolkit/en/topology-biopolymer-refactor/faq.html"
+                    "#parameterizing-my-system-which-contains-a-large-molecule-is-taking-forever-whats-wrong",
+                )
+
         if isinstance(toolkit_registry, ToolkitRegistry):
             # We may need to try several toolkitwrappers to find one
             # that supports the desired partial charge method, so we
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -2458,7 +2458,7 @@ class FrozenMolecule(Serializable):
             if self.n_atoms > 150:
                 warnings.warn(
                     f"Warning! Partial charge method '{partial_charge_method}' is not designed "
-                    "for use on large (i.e. > 150 atoms) molecule and may crash or take hours to "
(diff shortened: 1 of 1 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -2444,6 +2444,26 @@ def assign_partial_charges(
...
         """
+
+        # Raise a warning when users try to apply these charge methods to "large" molecules
+        WARN_LARGE_MOLECULES: Set[str] = {
+            "am1bcc",
+            "am1bccelf10",
+            "am1-mulliken",
+            "am1bccnosymspt",
+            "am1elf10",
+        }
```
mattwthompson, line 2455: I'm pretty sure I hit everything here - Gasteiger, `mmff94`, etc ran in a few ms on my molecule - but I could have missed something.

openff/toolkit/topology/molecule.py
```diff
@@ -2444,6 +2444,26 @@ def assign_partial_charges(
...
+
+        if partial_charge_method in WARN_LARGE_MOLECULES:
+            if self.n_atoms > 150:
```
mattwthompson, line 2458: I don't care what the threshold is set at; 100 might hit some false positives, especially with OpenEye, but too large like 300 would miss some bad inputs.

openff/toolkit/topology/molecule.py
```diff
@@ -2444,6 +2444,26 @@ def assign_partial_charges(
...
+                warnings.warn(
+                    f"Warning! Partial charge method '{partial_charge_method}' is not designed "
+                    "for use on large (i.e. > 150 atoms) molecules and may crash or take hours to "
+                    f"run on this molecule (found {self.n_atoms} atoms). For more, see "
+                    "https://docs.openforcefield.org/projects/toolkit/en/topology-biopolymer-refactor/faq.html"
+                    "#parameterizing-my-system-which-contains-a-large-molecule-is-taking-forever-whats-wrong",
+                )
```
mattwthompson, line 2465: It would be nice to add a section covering alternatives, like
* If you already know your charges, use `charge_from molecules` ...
* If your charges fit a pattern, here are some instructions on how to write up a `LibraryChargeType` ...

but I didn't want to write too much here. Maybe it's worth having a paragraph on that in this FAQ entry.

j-wags (approved):
LGTM!

mattwthompson:
In hindsight, this warning should have been a custom one, making it easier to capture than the vanilla `UserWarning`

PR #1331, opened 2022-06-27
Enhanced errors for biopolymers

Yoshanuikabundi:
This PR improves and adds clear exceptions for `from_polymer_pdb()`.

Goals:
- [ ] Better error messages when loading proteins (either OFF or OMM should raise a descriptive error about the following):
    - [x] “This protein is missing explicit Hs”
    - [x] “This looks like a modified amino acid”
    - [ ] “I don’t like this cap”
    - [x] "There can only be a single molecule in a PDB file"
    - [x] missing atoms
    - [ ] unresolved loops
    - [x] Crystallographic water
    - [x] Incorrect atom names
- [ ] Good tests could be:
    - [x] a protein without Hs
    - [x] A protein with a modified AA
    - [ ] a protein with unresolved atoms
    - [x] a pdb with two proteins
    - [x] a pdb with a protein and xtal waters (suggest how to separate them using eg. mdanalysis or pdbfixer or something before loading)

    - [x] Update `from_polymer_pdb` docstring with exceptions it can raise
    - [x] Update `from_polymer_pdb` docstring with expectations about input
    - [x] Add exceptions page to API docs


Yoshanuikabundi:
@j-wags So far I've got a few things working. The way the toolkit assigns residues is very different to (at least my) chemical intuition - I think of residues as links in a chain, but the toolkit looks at the whole molecule at once for each substructure. So getting useful errors out is taking quite a lot of work, both for me and my computer!

I've added an exception to `RDKitToolkitWrapper._add_chemical_info()` that is raised when atoms or bonds couldn't be assigned, and attempts to diagnose the problem. The code is currently in `exceptions.py`, but I imagine we'll want to find a better home for it...

At the moment it looks at a few potential problems and gives hints about common stuff, and then spams the user with info about what's missing. So for an alanine helix with a dye in the middle, you get an error like:

```
MissingChemistryFromPolymerError: Some bonds or atoms in the input could not be 
identified.

The following residue names with unassigned atoms were not found in the 
substructure library. While the OpenFF Toolkit identifies residues by matching 
chemical substructures rather than by residue name, it currently only supports 
the 20 'canonical' amino acids.
    DYE

The following 40 atoms exist in the input but could not be assigned chemical 
information from the substructure library:
    Atom    63 (C) in residue DYE#0007
    Atom    64 (C) in residue DYE#0007
    Atom    67 (C) in residue DYE#0007
    Atom    68 (C) in residue DYE#0007
    Atom    69 (C) in residue DYE#0007
    Atom    70 (C) in residue DYE#0007
    Atom    71 (C) in residue DYE#0007
    Atom    72 (C) in residue DYE#0007
    Atom    73 (C) in residue DYE#0007
    Atom    74 (C) in residue DYE#0007
    Atom    75 (C) in residue DYE#0007
    Atom    76 (C) in residue DYE#0007
    Atom    77 (C) in residue DYE#0007
    Atom    78 (C) in residue DYE#0007
    Atom    79 (C) in residue DYE#0007
    Atom    80 (C) in residue DYE#0007
    Atom    81 (C) in residue DYE#0007
    Atom    82 (C) in residue DYE#0007
    Atom    83 (C) in residue DYE#0007
    Atom    84 (C) in residue DYE#0007
    Atom    85 (C) in residue DYE#0007
    Atom    88 (O) in residue DYE#0007
    Atom    89 (O) in residue DYE#0007
    Atom    90 (O) in residue DYE#0007
    Atom    91 (O) in residue DYE#0007
    Atom    92 (O) in residue DYE#0007
    Atom    94 (H) in residue DYE#0007
    Atom    95 (H) in residue DYE#0007
    Atom   100 (H) in residue DYE#0007
    Atom   101 (H) in residue DYE#0007
    Atom   102 (H) in residue DYE#0007
    Atom   103 (H) in residue DYE#0007
    Atom   104 (H) in residue DYE#0007
    Atom   105 (H) in residue DYE#0007
    Atom   106 (H) in residue DYE#0007
    Atom   107 (H) in residue DYE#0007
    Atom   108 (H) in residue DYE#0007
    Atom   109 (H) in residue DYE#0007
    Atom   110 (H) in residue DYE#0007
    Atom   111 (H) in residue DYE#0007

The following residues have atoms that were assigned a name that does not match 
the residue name in the input, or could not be assigned at all. This may 
indicate that atoms are missing from the input or some other error. The OpenFF 
Toolkit requires all atoms, including hydrogens, to be explicit in the input to 
avoid ambiguities in protonation state or bond order:
    Input residue DYE#0007 matched substructures {'CYS', None, 'PEPTIDE_BOND'}

```

The problem with missing atoms is that there's no way AFAIK to see that a SMIRKS *nearly* matches. All we can really say is "I couldn't match anything to this part of the molecule" and "the assigned residue doesn't match the input residue name", or else add specific hints by hand. So if a PDB is missing all its hydrogens, the error looks like this:

```
MissingChemistryFromPolymerError: Some bonds or atoms in the input could not be 
identified.

Hint: There are no hydrogens in the input. The OpenFF Toolkit requires explicit 
hydrogens to avoid ambiguities in protonation state or bond order. Try 
generating hydrogens with another package and trying again.

The following residues have atoms that were assigned a name that does not match 
the residue name in the input, or could not be assigned at all. This may 
indicate that atoms are missing from the input or some other error. The OpenFF 
Toolkit requires all atoms, including hydrogens, to be explicit in the input to 
avoid ambiguities in protonation state or bond order:
    Input residue ACE#0001 matched substructures {None, 'PEPTIDE_BOND'}
    Input residue ALA#0002 matched substructures {None, 'PEPTIDE_BOND'}
    Input residue ALA#0003 matched substructures {None, 'PEPTIDE_BOND'}
    Input residue NME#0004 matched substructures {'PEPTIDE_BOND'}

The following 3 atoms exist in the input but could not be assigned chemical 
information from the substructure library:
    Atom     2 (C) in residue ACE#0001
    Atom     7 (C) in residue ALA#0002
    Atom    12 (C) in residue ALA#0003

The following 2 bonds exist in the input but could not be assigned chemical 
information from the substructure library:
    Bond between atom     4 (C) in ALA#0002 and atom     5 (C) in ALA#0002
    Bond between atom     9 (C) in ALA#0003 and atom    10 (C) in ALA#0003
Note: Bonds involving atoms that could not be assigned chemical information 
have been elided from this list.
```

If you take out the "there are no hydrogens" hint - like, say only one residue is missing hydrogens, or the polar hydrogens are present - this error is correct, and it would be useful to an experienced user ("oh, all my residues aren't matching"), but its also kinda misleading ("why are only three of the atoms unable to be assigned?" - actually conceptually none of them are, but the PEPTIDE_BOND substructure is hiding this)

This approach makes errors kinda backwards. It's fine if someone accidentally puts an extra atom in their PDB file, but the much more common case is that an atom is missing. This causes the residue SMARTS not to match, so atoms are left unassigned, which is reported basically as "there are too many atoms in the input" - exactly the opposite of the true problem. It's not clear to me how to make this easier, short of implementing a parallel system that tries to find missing atoms by iterating over the chain and trusting the residue names. My suspicion is that to get good errors and be confident that we're doing what the user wants we're going to need the information that's encoded in residue names.

I'll keep going with this approach - its definitely better than the status quo - but ideas would be great :)

j-wags (commented):
This is a really great start. I'm super happy with how you're organizing this - Putting the reporting logic in exceptions is wonderful and will help these thoughtfully-designed error messages extend to OpenEyeToolkitWrapper PDB loading. The one thing I'd like would be if the exceptions didn't get an RDMol, but instead got an OMMTop - This will help keep them generic and reduce complexity later on. We can merge `_add_chemical_info` and `_omm_top_to_rdmol` to make sure the `omm_top` is in scope when we need it for an error message. 

If some of these cases get hard to check for/test, feel free to request review before everything is covered. I don't know how hard all of these will end up being so if you hit decreasing returns I'd be more than happy to just merge the cases that are straightforward.

Yoshanuikabundi:
Here's a notebook that demonstrates the new errors: [new_errors.zip](https://github.com/openforcefield/openff-toolkit/files/9044166/new_errors.zip)

I'm not sure how to detect missing loops, and unknown caps should be caught by the general unknown residue hints.

I've also added a page for the exceptions to the docs. It automatically documents every exception in the exceptions module, so no one needs to remember to update it when they add an exception.

Yoshanuikabundi:
@j-wags @mattwthompson The more I think about this, the more I think the right way to go would be to iterate over the residues in the PDB and attempt to assign them chemistry by matching the same residue name in the substructure library. I think this makes exceptions easier, but it also makes the source of the chemical information consistent, predictable, and extensible, and may prevent a lot of nasty bugs down the line.

The reason we don't usually allow PDBs without SMILES is because they don't include enough chemical information to assign bond orders, aromaticity, chirality, and formal charges. We need to take PDBs for proteins because there isn't an alternative, but we're in luck; the PDB format *does* specify bond order, aromaticity, chirality, and formal charge for atoms *with particular names in particular residues*, as described in the chemical component dictionary. So the chemical information *is in the PDB*, encoded in the RESNAME and ATOMNAME columns.

At the moment, we:
- Read the PDB with OpenMM, which assigns connectivity (but not bond order, aromaticity, chirality, or formal charge) according to [this file](https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/data/residues.xml) as well as the CONECT records 
    - At this point, if an atom is not named correctly, its bonds are silently unassigned, which RDKit will later see as unconnected lone atoms (which, thankfully, won't match anything in the substructure library)
- Loop over every SMARTS for every residue in the substructure library and for each...
    - match the SMARTS to the entire protein according to connectivity alone
    - assign chemical information from the smarts to whatever it matches, as long as it hasn't already been assigned any chemistry
- If any atoms are left unassigned, raise an error and try to guess what went wrong

I thought we would use atom names somewhere since they're in the substructure dictionary, but I can't see any place they're used in the code (aside from the exceptions added in this PR)

I feel like it's easy to imagine this going wrong. Already this strategy only fails for multiple molecule input files if we add an explicit check for that (I expect the assignment is usually correct though). OpenMM assigning connectivity probably protects us from things going badly wrong and assigning an isomer of a residue completely wrong connectivity, but it also (a) will make it difficult to add new residues or allow users to extend the dictionary, and (b) means that we actually *do* use the residue and atom names to assign our parameters, it just happens in someone else's code.

If the user adds CONECT records of their own, all bets are off, especially if they're in addition to the standard bonds!

This is the same basic strategy we use for assigning parameters. I think this strategy works for assigning force field parameters to chemistry, because there's no ambiguity - we know exactly what the chemical is, and the correct parametrization is the one with all the parameters that match it. But here, there may be many possible chemistries for a particular connectivity, and they may overlap - especially as the substructure library grows. If we want to grow past the OpenMM-assigned connectivity it gets even worse, as we can't rely on any existing connectivity, and any isomer of a group of atoms might be fair game!

The way the PDB format disambiguates this is with the CCD, residue names, and atom names.

I'd suggest we do the following:

- Read the PDB with OpenMM, and ignore the connectivity it assigns
- Loop over every residue in the PDB and for each...
    - Index the substructure dictionary with the residue name to get the possible SMARTS-atomnames mappings
    - See if any of the maps has the exact list of atomnames we have in this residue
        - Use the SMARTS to assign connectivity and everything else if we have an exact match between atom names
        - Raise an error if we don't have an exact match

This will give us super clear error messages and make it extremely clear where all our information is coming from, letting us be explicit about our assumptions and expectations. Extending the substructure dictionary will be trivial. And we can guarantee that we never look at CONECT records.

thank you for listening to my ted talk

mattwthompson:
> Already this strategy only fails for multiple molecule input files if we add an explicit check for that (I expect the assignment is usually correct though). 

Separate from biopolymers, I'm not really sure our infrastructure around multi-molecule inputs is good. Like, we'll load up `Molecule.from_smiles("CCO.CCO")` and some functionality works, but others choke for reasons that would be confusing to a new user. (Try to assign partial charges or fractional bond orders and you'll get an error message that only we have an easy time understanding.) Outside of this PR, I'd actually be in support of a change that errors out if it sees a multi-molecule-ish input _anywhere_, like `if '.' in input_smiles: raise SomeException: ...` until something like [this](https://github.com/openforcefield/openff-toolkit/pull/1086#issuecomment-1173175518) is carefully implemented.

> [using OpenMM to assign connectivity] will make it difficult to add new residues or allow users to extend the dictionary [but using atom names makes this trivial]

I'm not sure I follow - are there residues that OpenMM won't know how to process? Or do we simply expect users to be happy extending the CCD? Specifically I'd be worried about the case of non-biological polymers; as a user I'd rather write up a chemical description in a behemoth SMARTS pattern than use atom names, but I'm not really sure.

I see extending the CCD as a requirement for the 0.11.0 release, but if that's a huge constraint on the design of the internals I'd like to see how hard or easy that is to actually pull off.

Yoshanuikabundi:
> Outside of this PR, I'd actually be in support of a change that errors out if it sees a multi-molecule-ish input anywhere

That sounds like a good idea to me

> I'm not sure I follow - are there residues that OpenMM won't know how to process?

Non-canonical amino acids might be the most obvious example of this.

> I'd rather write up a chemical description in a behemoth SMARTS pattern than use atom names

I think, for me, in either case I'd be writing code I don't totally trust, and a PDB that I can load into any visualization software I like would be easier to verify. Also, `from_polymer_pdb` definitely uses atom names, it just happens in OpenMM's code.

If we leave connectivity up to OpenMM, we both rely on atom names and make it harder for users to extend the substructure library, which feels like a lose-lose.

openff/toolkit/topology/molecule.py
```diff
@@ -3824,8 +3849,32 @@ def from_polymer_pdb(
...
         offmol.add_default_hierarchy_schemes()
 
+        if len(offmol.chains) > 1:
+            raise MultipleMoleculesInPDBError(
+                "Multiple polymer chains were detected in the input. The OpenFF "
+                + "Toolkit requires that only one polymer chain is present in a "
+                + "PDB, and that it is the only molecule present. Try splitting "
+                + "each polymer chain into its own PDB with another tool, and "
+                + "import any small molecules with Topology.from_pdb_and_smiles."
+            )
```
j-wags, line 3859: To the OpenFF toolkit, Chain IDs are just cosmetic. So it's fine if one Molecule has atoms with multiple chain IDs

```suggestion
```

openff/toolkit/topology/molecule.py
```diff
@@ -3824,8 +3849,32 @@ def from_polymer_pdb(
...
+        if offmol._has_multiple_molecules():
+            raise MultipleMoleculesInPDBError(
+                "This PDB has multiple molecules. The OpenFF Toolkit "
+                + "requires that only one polymer chain is present in a PDB, "
+                + "and that it is the only molecule present. Try splitting "
+                + "each polymer chain into its own PDB with another tool, and "
+                + "import any small molecules with Topology.from_pdb_and_smiles."
+            )
```
j-wags, line 3867: "chain" is kinda a loaded term for us, since we consider chain IDs to just be cosmetic. So maybe to make sure users understand this error, we should say "molecule", since that's a discrete concept in OpenFF.

```suggestion
            raise MultipleMoleculesInPDBError(
                "This PDB has multiple molecules. The OpenFF Toolkit "
                + "requires that only one polymer molecule is present in a PDB, "
                + "and that it is the only molecule present. Try splitting "
                + "each polymer molecule into its own PDB with another tool, and "
                + "import any small molecules with Topology.from_pdb_and_smiles."
            )
```

j-wags (commented):
Finally, in lines like 

```
    Atom   396 (C) in residue ILE#0050
```

It may be good to list the atom name instead of the element. Kinda your call, I'd approve it in either case :-)

openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -238,14 +240,15 @@ def from_pdb_and_smiles(
             raise InvalidConformerError("The PDB and SMILES structures do not match.")
 
     def _polymer_openmm_topology_to_offmol(self, omm_top, substructure_dictionary):
-        rdkit_mol = self._polymer_openmm_topology_to_rdmol(omm_top)
-        rdkit_mol = self._add_chemical_info(rdkit_mol, substructure_dictionary)
+        rdkit_mol = self._polymer_openmm_topology_to_rdmol(
+            omm_top, substructure_dictionary
+        )
```
j-wags, line 245: 👍

j-wags (approved):
Wonderful! This is a masterpiece of a PR :-)

j-wags:
Sorry, I saw a few things in the `from_polymer_pdb` docstring that I wanted to fix, so I pushed a commit to your PR so we can have it in the next release candidate. Thanks for understanding! 😄 🙏

PR #1339, opened 2022-07-04
Run CI on main branch

Yoshanuikabundi:
Updates CI to run on `main` (and not `master` or `topology-biopolymer-refactor`)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -3,15 +3,14 @@ name: CI
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


mattwthompson (approved):
Thanks for catching this - I anticipate in the next week that we'll shotgun a number of PRs that look like this

mattwthompson:
Not too surprised that CodeCov doesn't know what to do with this - but at least now PRs into 0.11.0 won't require manual upload.

PR #1350, opened 2022-07-12
Remove lint workflow

mattwthompson:
I think it's time to remove this workflow; its functionality is wholly replaced by `pre-commit` / https://pre-commit.ci/. I intentionally did not remove this workflow when adding `pre-commit` since I wanted to allow our developer base time to trust it. I now believe enough time has passed for us to make this change.


mattwthompson:
We decided against this

PR #1353, opened 2022-07-25
Fix conda CI

j-wags:
- [x] Separate nightly tests for 0.10.X and >=0.11.X lines
- [x] Rename test envs to be cleaner
- [x] Other small nightly test cleanups

### DIFF
--- a/devtools/conda-envs/conda.yaml
+++ b/devtools/conda-envs/conda.yaml
@@ -1,6 +1,8 @@
 name: test
 channels:
   - conda-forge
+  - conda-forge/label/openff-toolkit_rc
+  - conda-forge/label/openff-interchange_rc
 
 dependencies:
     # Base depends
--- a/devtools/conda-envs/conda.yaml
+++ b/devtools/conda-envs/conda.yaml
@@ -6,7 +6,7 @@ channels:
 
 dependencies:
     # Base depends
-  - openff-toolkit
+  - openff-toolkit-examples
 
   - openmmforcefields
   - pytest
--- a/devtools/conda-envs/conda.yaml
+++ b/devtools/conda-envs/conda.yaml
@@ -1,4 +1,4 @@
-name: test
+name: latest-deployment
 channels:
   - conda-forge/label/openff-toolkit_rc
   - conda-forge/label/openff-interchange_rc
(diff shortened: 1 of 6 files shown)


devtools/conda-envs/conda.yaml
```diff
@@ -1,10 +1,12 @@
 name: test
```
mattwthompson, line 1: Nit (similar for other files as well): when reproducing issues locally I've sometimes maybe had clashes with files like these and existing environments named `test`
```suggestion
name: deployment-env
```

j-wags: fixed in [8b4ebf3](https://github.com/openforcefield/openff-toolkit/pull/1353/commits/8b4ebf3b32cf6bd79e4ddc309fd59adf3ae01420)

j-wags:
Thanks for the earlier comments, @mattwthompson. This should be ready for review.

I've got runs of the conda tests with the final state of this PR started here:
0.10.x: https://github.com/openforcefield/openff-toolkit/actions/runs/2742593001
latest: https://github.com/openforcefield/openff-toolkit/actions/runs/2742593008

devtools/conda-envs/conda_010X.yaml
```diff
@@ -0,0 +1,14 @@
+name: 010x-latest-deployment
+channels:
+  - conda-forge
+
+
+dependencies:
+    # Base depends
+  - openff-toolkit-examples =0.10
```
mattwthompson, line 8: I assume this will typically solve to `0.10.6` or newer and I don't think the complexity of more constraints is necessary, but we'd want to check a couple logs to make sure it doesn't somehow grab one of the buggy versions.

mattwthompson (approved):
LGTM - we could probably add Python 3.10 to these matrices but I don't think that's necessary for this PR.

devtools/conda-envs/conda_010X.yaml
mattwthompson: Okay - looks like it doesn't now

j-wags: Yeah, I'd checked this manually during tests, and ensured that this action still finds the latest git tag via a separate command and ensures it matches with the installed version (just to be extra safe) :-)

j-wags:
Thanks! I'll merge once my manually-initiated conda runs go fully green.

PR #1360, opened 2022-08-03
Replace Miniconda with micromamba

mattwthompson:
Uses https://github.com/mamba-org/provision-with-micromamba

In my and others' testing, it solves environments in roughly half the time as https://github.com/conda-incubator/setup-miniconda but has some issues with channel priority (which should not affect us)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -76,17 +76,12 @@ jobs:
               echo "PACKAGES_TO_REMOVE=openeye-toolkits" >> $GITHUB_ENV
           fi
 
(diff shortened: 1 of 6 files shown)


j-wags:
Oh, nice! Let me know when this is ready for review.

mattwthompson:
The examples workflow took a bit of kludging, but I think everything is working now.

j-wags (approved):
Great - Thanks! The changes look good, and I went through the CI logs and ensured that OE was being installed in the appropriate builds. LGTM!

PR #1362, opened 2022-08-05
Test post please ignore

mattwthompson:


mattwthompson:
Code coverage reports should be useful again

j-wags:
Thanks @mattwthompson!

mattwthompson:
Unfortunately the history might be lost (https://app.codecov.io/gh/openforcefield/openff-toolkit) but I don't care much about that

PR #1368, opened 2022-08-11
Toolkit showcase update

Yoshanuikabundi:
This PR updates the showcase to use all the new magic in the toolkit and interchange. It will no longer use Parmed and OpenMM force fields.

It also includes performance optimizations for new features required to run the example in an acceptable time frame.


mattwthompson:
@Yoshanuikabundi please tag me or reach out to me directly if you have any questions or suggestions on integrating Interchange in this, including patches to other repos. I don't want to get in your way but I'm also interested and invested in getting this working nicely 🙂

Yoshanuikabundi:
I've added a few potentially controversial methods to Topology:

- `Topology.get_positions()`
- `Topology.set_positions()`
- `Topology.visualize()`

I'm happy to make them private or put them in the notebook itself, but I think all three make life a lot easier. The first two are just convenience methods for accessing and updating all the positions at once and work by iterating over molecules and assigning/reading the first conformers of each. At some point we should talk about adding an option to the PDB output methods to use standard atom names, because I haven't found any protein visualization software that can render a cartoon on our PDB outputs (and the atom names are not even unique, possibly because the field in a PDB isn't long enough), but for the moment the new `visualize()` method can fill in this gap.

@mattwthompson Thanks for the offer! Turns out I need you immediately. When I try to construct an OpenMM topology, Interchange complains about virtual sites (AFAIK there should be no virtual sites in this system, but Interchange seems to expect them):

```python
LookupError                               Traceback (most recent call last)
Input In [14], in <cell line: 2>()
      1 omm_system = interchange.to_openmm()
----> 2 omm_top = interchange.to_openmm_topology()

File ~/Documents/openff/openff-toolkit/.soap/examples/lib/python3.10/site-packages/openff/interchange/components/interchange.py:529, in Interchange.to_openmm_topology(self, ensure_unique_atom_names)
    526 """Export components of this Interchange to an OpenMM Topology."""
    527 from openff.interchange.interop.openmm import to_openmm_topology
--> 529 return to_openmm_topology(
    530     self, ensure_unique_atom_names=ensure_unique_atom_names
    531 )

File ~/Documents/openff/openff-toolkit/.soap/examples/lib/python3.10/site-packages/openff/interchange/interop/openmm.py:922, in to_openmm_topology(interchange, ensure_unique_atom_names)
    916 from openff.interchange.interop._virtual_sites import (
    917     _virtual_site_parent_molecule_mapping,
    918 )
    920 topology = interchange.topology
--> 922 virtual_site_molecule_map = _virtual_site_parent_molecule_mapping(interchange)
    924 molecule_virtual_site_map = defaultdict(list)
    926 for virtual_site, molecule in virtual_site_molecule_map.items():

File ~/Documents/openff/openff-toolkit/.soap/examples/lib/python3.10/site-packages/openff/interchange/interop/_virtual_sites.py:18, in _virtual_site_parent_molecule_mapping(interchange)
     13 def _virtual_site_parent_molecule_mapping(
     14     interchange: "Interchange",
     15 ) -> Dict[VirtualSiteKey, Molecule]:
     16     mapping = dict()
---> 18     for virtual_site_key in interchange["VirtualSites"].slot_map:
     19         virtual_site_key: VirtualSiteKey  # type: ignore[no-redef]
     20         parent_atom_index = virtual_site_key.orientation_atom_indices[0]

File ~/Documents/openff/openff-toolkit/.soap/examples/lib/python3.10/site-packages/openff/interchange/components/interchange.py:826, in Interchange.__getitem__(self, item)
    824     return self.handlers[item]
    825 else:
--> 826     raise LookupError(
    827         f"Could not find component {item}. This object has the following "
    828         f"potential handlers registered:\n\t{[*self.handlers.keys()]}"
    829     )

LookupError: Could not find component VirtualSites. This object has the following potential handlers registered:
	['Bonds', 'Constraints', 'Angles', 'ProperTorsions', 'ImproperTorsions', 'vdW', 'Electrostatics']
```

A minimal example is (with the attached [`topology.json`](https://github.com/openforcefield/openff-toolkit/files/9308190/topology.zip)):

```python
from openff.toolkit import Topology, ForceField
from openmm import unit as openmm_unit
from openff.units.openmm import from_openmm

with open("topology.json", "r") as f:
    top = Topology.from_json(f.read())

sage_ff14sb = ForceField("openff-2.0.0.offxml", "ff14sb_off_impropers_0.0.3.offxml")
interchange = sage_ff14sb.create_interchange(top)
omm_top = interchange.to_openmm_topology()
```

If you want to see how the topology is constructed, I've committed and pushed my current working version of the notebook. Note that it requires changes to the Toolkit itself that are included on this branch. Please ignore the prose, I'm working on the code first.

I'm excited to get this working!

mattwthompson:
My first-impression is that these new `Topology` methods should be standalone functions scoped to the notebook; I think we are getting close to the wire for significant changes to the API in this release and writing them as functions allows us to defer decisions into the future - enabling the notebook to work just about as well as if they were methods but requiring less testing and consensus-gathering. I'm moderately opposed to positions being on topologies here and my opinion on `Topology.visualize()` is closer to neutrality.

I'll fix that bug shortly, no reason that logic should be in place in the way it is. I thought I wrote a test for that, guess not.

Yoshanuikabundi:
Ok, I've moved those methods back into the notebook.

> I'm moderately opposed to positions being on topologies here
> We've been talking about exposing positions on topologies for a while but every option seems to be a steep tradeoff between horrible complexity and user confusion/footguns.

We need to make a decision about this, because I need to document (and already have) when users should introduce their positions. I thought we agreed in [openff-interchange/436](https://github.com/openforcefield/openff-interchange/issues/436#issuecomment-1104609747) that we would support velocities, visualization, and positions in topologies - hence me adding some of that support here! :sweat_smile: 

@j-wags I think just providing `Topology.get_positions()` and `Topology.set_positions()` is the compromise that avoids confusion without much added complexity. Both methods work by iterating over the molecules' first conformers. The get/set pattern makes it clear that you're working with a copy of the data, and we document that as well. Both methods together including docstrings are less than 60 lines of code and include all the safety checks you could hope for. I suspect if we don't provide these methods, our users will write them themselves. But we don't have to do that today!

<details><summary> Here's me ranting about how we should conceptually have positions on topologies (and a solvate method) for a few pages, you should probably skip it but I couldn't bear to delete it.</summary>

I'll just recap the current situation:

- Interchange stores the system positions and velocities in the `.positions` and `.velocities` arrays
- Interchange initializes its positions from the first conformation of each molecule in the topology iff every molecule has at least one conformer.
- Interchange works with `.positions` or `.velocities` set to `None`, but export methods requiring those values will fail
- Interchange supports a `.visualize()` method
- By design, methods for adding or removing molecules to an existing Interchange do not exist

In the showcase, this makes things a bit awkward. Our inputs are:

a. A PDB of the receptor protein with crystallographic waters (that we want to preserve)
b. An SDF of the small molecule ligand

To assemble the system, the current strategy is:

1. Solvate the PDB with crystallographic waters with PDBFixer, which takes a PDB as input
2. Split the protein out into its own PDB and import it with `Molecule.from_polymer_pdb()`
3. Import the fully solvated system directly from PDBFixer with `Topology.from_openmm()`, passing in the above protein molecule as one of the `unique_molecules`

At this point, we have:
       
a. a topology with everything except the ligand, without positions
b. an array of positions for everything except the ligand (from PDBFixer)
c. an SDF of the ligand, with positions

If topologies don't have positions, what do we do here? Something like identify the clashing waters, remove their coordinates from the array, remove the corresponding number of waters from the topology, concatenate the SDF coordinates to the rest of the coordinates, add the ligand to the topology, convert it to an interchange, and THEN add the coordinates to the interchange. Conceptually, having positions in the topology is much simpler:

4. Set the positions of the molecules in the topology with the array of positions from PDBFixer.
5. Import the ligand with `Molecule.from_file()` and add it to the topology
6. Remove water molecules that clash with the ligand

Note that I'm not talking about having a positions array in the topology; I'm just talking about what we suggest conceptually. Setting positions just means iterating over the molecules and setting the first conformer. This keeps everything object oriented, which is convenient for scientists. I think in most workflows involving proteins, the molecular identity and coordinates are stored together, and so they should enter the OpenFF ecosystem together. Plus, if topologies have coordinates, a user can easily parametrize the exact same system with different force fields.

This workflow is very complex. There is an important question as to whether the developers should pay that complexity cost or the users. I know a `Topology.solvate()` method is unpopular, but I would rather personally maintain a single solvation method in the toolkit than a separate one in every example, and I'm sure our users would appreciate it too. This would replace steps 1, 4, and 6 with a call to this method. It would mean either packaging equillibrated water boxes with the toolkit, or calling out to something like PDBFixer.

A `Topology.from_pdb()` method, which would take a list of unique molecules and just load the PDB with `openmm.app.PDBFile` and pass it through `Topology.from_openmm`, would be nice too, but this is simple enough that I'm happy for it to live in the examples.

Not asking for anything to change now, but if we're gonna say positions don't belong on topologies I need an absolutely rock solid reason because from a pedagogical and users' perspective I am 99.999% sure they do.

</details>

-------

I need to flag a few other things that are driving me nuts at the moment so we can sort them out after 0.11.0:

 - Visualization at the moment is a massive pain, because we attempt to uniquify atom names at the drop of a hat. PyMol, VMD, and NGLView all rely on atom names to identify proteins. Interchange automatically attempts to uniquify atom names, and this means that all outputs are unable to be visualized as ribbons or cartoons, and you can't use selection languages to show side chains in a different representation, and so on. Call `interchange.visualize()` on the toolkit showcase interchange and you'll see... two caps and a ligand in ball and stick rep and nothing else. This happens even in `Topology.to_file()`, which is essential for visualization outside the notebook - and there, it produces PDBs that don't even have unique atom names past 100 atoms of an element because the name has a fixed width! I think we should have options everywhere to preserve atom names.

- at the moment I have to write PDBs to disk a lot to pass information between libraries
    - It would be nice to be able to provide a file-like object to methods like `Topology.to_file`.
    - Exposing a method like `Molecule.from_polymer_openmm_topology()`, which is just `Molecule.from_polymer_pdb()` without the `PDBFile` step, would be wonderful too

------

@mattwthompson - What are the dangers of just doing `top.to_openmm()` rather than `interchange.to_openmm_topology()`? It seems to be working fine (and it lets me visualize things).

------

Also, `Topology.to_json()` and `Topology.from_json()` are wonderful methods whom I love.

Sorry if this is a grumpy comment.

mattwthompson:
> What are the dangers of just doing top.to_openmm() rather than interchange.to_openmm_topology()? 

Not a danger, but the toolkit's method does not include virtual sites (by design). Otherwise their behavior should be identical

j-wags:
> I thought we agreed in https://github.com/openforcefield/openff-interchange/issues/436#issuecomment-1104609747 that we would support velocities, visualization, and positions in topologies - hence me adding some of that support here! 😅

Woah, so I did. My mistake @Yoshanuikabundi, and thanks for correcting me. You're right - We agreed to add positions to topologies (and velocities, which doesn't need to be done in this PR). So I'll stick with that decision. Sorry for the confusion and thanks for linking to the written decision :-) 

> Visualization at the moment is a massive pain, because we attempt to uniquify atom names at the drop of a hat. PyMol, VMD, and NGLView all rely on atom names to identify proteins

Agreed. It looks like you've already switched `ensure_unique_atom_names=False` in the Toolkit calls, which is great. @mattwthompson, would you be open to a future PR to Interchange that prevents renaming in `Interchange.visualize`?

> It would be nice to be able to provide a file-like object to methods like Topology.to_file.

Oh, that'd be great. Happy to have that in this PR or in a future one. 

> Exposing a method like Molecule.from_polymer_openmm_topology(), which is just Molecule.from_polymer_pdb() without the PDBFile step, would be wonderful too

Agree, but that will take more quality assurance/documentation than we have time for in the near future. I think the current `from_polymer_pdb` behavior relies somewhat on the fact that the topology it's loading came from an OpenMM PDBFile object. So we'll need additional logic to handle/provide useful errors for loading general OpenMM topologies.

Yoshanuikabundi:
Sorry @mattwthompson  - I see `Interchange.to_openmm_topology(ensure_unique_atom_names=False)` does work already, and `Interchange.visualize()` does preserve atom names. I must have woken up on the wrong side of the bed the other day. The vsites error has also gone away! Thanks!

@j-wags - I've implemented the changes to `ensure_unique_atom_names` in commit f76a02a as discussed on slack. If we like it, I'll write tests and submit a PR for the equivalent functionality in Interchange, and if it needs more dev time I'll move it to its own PR.

What do we want to do if we can't get speedy GROMACS support working in time? Just remove the appendix?

examples/toolkit_showcase/toolkit_showcase.ipynb
```diff
@@ -28,47 +28,12 @@
     "</details>\n",
```
j-wags, line 28: <p>"<code>Molecule</code> objects are unusual in the computational chemistry world; they're not a thin wrapper around a SMILES, nor a collection of elements, coordinates, and formal charge" --&gt; I think I see what you're getting at but this is confusing as a first time reader. It might not be worth having this part.</p><p> </p><p>"SDF files include all a molecule's bond orders and information about aromaticity" --&gt; "SDF files include all a molecule's bond orders and formal charges"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

j-wags, line 28: <p>"inferring chemical information from residue and atom names and a substructure dictionary distributed with the toolkit" --&gt; "...information from element and connectivity". For example, we could load an ALA from PDB, even if it didn't have a residue name, just as long as it has CONECT records instead. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

j-wags, line 28: <p>Now that there's OE PDB reading support, we don't need to restrict to RDKit here. </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

j-wags, line 28: <pre class="ql-syntax" spellcheck="false">Line #1.    receptor = Molecule.from_polymer_pdb("receptor.pdb")
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

j-wags, line 28: <pre class="ql-syntax" spellcheck="false">Line #1.    def insert_molecule_and_remove_clashes(
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

j-wags, line 28: <pre class="ql-syntax" spellcheck="false">Line #45.        new_top.add_molecule(ligand)
</pre><p>Could instead build up a list of "surviving" molecules and run <code>Topology.from_molecules</code> right at the end on that list.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

j-wags, line 28: <pre class="ql-syntax" spellcheck="false">Line #49.    top = insert_molecule_and_remove_clashes(top, ligand)
</pre><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

j-wags, line 28: <p>"until the end of 2022" --&gt; "2023"</p><p> </p><p>"We don't recommend using this combination in production; instead, wait a few months for Rosemary." --&gt; Maybe consider taking this out? Seems especially redundant with the warning box at the bottom.</p><p> </p><p>Also "FF15SB"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

j-wags, line 28: <p>"Parametrize"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

j-wags, line 28: <p>People will have time to read a bit here - Maybe expand or link to the interchange docs?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

j-wags, line 28: <p>Let's drop the GROMACS appendix for now - Once we hammer on the export timing a bit we can put it back in. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1368/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='48'/>

j-wags (changes requested):
This is looking really, really good. Thanks for all the thought you put into it. It's really big and there's a lot of new functionality, so I'm asking for a few changes. Feel free to resolve comments as you get to them, but please keep the big ones (basically "add these tests") open so I can use them as a guide for the final review :-)

examples/toolkit_showcase/toolkit_showcase.ipynb
Yoshanuikabundi: <p>I've changed the thin wrapper thing to be much clearer: </p><p> </p><p>"<code>Molecule</code> objects represent a collection of atoms with specified formal charges, connected by bonds with specified bond orders, optionally including any number of conformer coordinates. This is intended to closely align with a chemist's intuitive understanding of a molecule, rather than simply wrap the minimal information needed for a calculation."</p>

Yoshanuikabundi: <p>New phrasing is "... for polymers the Toolkit supports inferring chemical information from PDB files. Connectivity can be taken from residue and atom names or CONECT records, and formal charges and bond orders are inferred from a substructure dictionary distributed with the Toolkit."</p>

Yoshanuikabundi: <p>Fixed here and simplified the imports accordingly.</p>

Yoshanuikabundi: <p>Done</p>

Yoshanuikabundi: <p>Brilliant. I've also added comments everywhere, since this function is intended to be an example for users. And I decided the <code>topology_from_pdb</code> function in the previous cell wasn't pulling its weight, so I've replaced it with pulling the OpenMM topology directly from the PDBFixer object. </p>

Yoshanuikabundi: <p>Fixed throughout.</p>

Yoshanuikabundi: <p>Done, and removed from the installation instructions at the top of the notebook.</p>

Yoshanuikabundi: <p>Done both! I've added an impassioned argument for the OpenFF workflow and bragged about Interchange a bunch.</p>

openff/toolkit/tests/test_topology.py
```diff
@@ -1448,3 +1451,116 @@ def test_impossible_error(self):
...
+        # Water molecules with empty list of conformers
+        water_emptyconfs = Molecule.from_mapped_smiles("[H:2][O:1][H:3]")
+        water_emptyconfs._conformers = []
+
+        # Compile topology
+        topology = Topology.from_molecules(
+            [
+                methane,
+                *[water] * 500,
+                *[water_emptyconfs] * 500,
```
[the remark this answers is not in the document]
Yoshanuikabundi: Done here and in the topology fixture.

openff/toolkit/tests/test_topology.py
```diff
@@ -1448,3 +1451,116 @@ def test_impossible_error(self):
...
+        # (because adding them to the topology copies them)
+        assert all(np.all(a == b) for a, b in zip(methane.conformers, init_conformers))
+        assert not water.conformers
+
+        # Now we're going to check all the molecules
+        mol_iter = topology.molecules
+
+        # Make sure the methane's first conformation is changed, but the others are left
+        first_mol = next(mol_iter)
+        print(first_mol.conformers[0])
```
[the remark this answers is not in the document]
Yoshanuikabundi: Oops

openff/toolkit/tests/test_topology.py
```diff
@@ -1448,3 +1451,116 @@ def test_impossible_error(self):
...
+    "generate_positions",
+    [
+        # Test an array
+        lambda n, rng: rng.random((n, 3)),
+        # Test a list of lists
+        lambda n, rng: [[rng.random(), rng.random(), rng.random()] for _ in range(n)],
+        # Test a list of tuples
+        lambda n, rng: [(rng.random(), rng.random(), rng.random()) for _ in range(n)],
+    ],
+)
```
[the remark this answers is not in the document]
Yoshanuikabundi: Now uses range/arange instead of rng. I also moved the generate_positions parameters to the test in which its actually necessary, rather than the whole class.

openff/toolkit/tests/test_topology.py
```diff
@@ -1448,3 +1451,116 @@ def test_impossible_error(self):
...
+        # Make sure the methane's first conformation is changed, but the others are left
+        first_mol = next(mol_iter)
+        print(first_mol.conformers[0])
+        assert np.all(first_mol.conformers[0] == positions[0:5, :])
+        assert np.all(first_mol.conformers[1:] == init_conformers[1:])
+
+        # Check the other molecules in the topology.
+        # Reusing the iterator to skip the methane
+        for i, mol in enumerate(mol_iter):
+            assert np.all(mol.conformers[0] == positions[i * 3 + 5 : i * 3 + 8, :])
```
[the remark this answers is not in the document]
Yoshanuikabundi: Done!

openff/toolkit/tests/test_topology.py
```diff
@@ -1448,3 +1451,116 @@ def test_impossible_error(self):
...
+        # Generate positions without units deterministically
+        rng = np.random.default_rng(seed=999)
+        positions_no_units = generate_positions(topology.n_atoms, rng)
+
+        with pytest.raises(
+            IncompatibleUnitError,
+            match="array should be an OpenFF Quantity with dimensions of length",
+        ):
+            topology.set_positions(positions_no_units)
+
```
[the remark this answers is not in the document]
Yoshanuikabundi: Done

openff/toolkit/tests/test_topology.py
```diff
@@ -1448,3 +1451,116 @@ def test_impossible_error(self):
...
+    def test_get_positions(self, topology, generate_positions):
+        # Generate positions deterministically
+        rng = np.random.default_rng(seed=999)
+        positions = Quantity(generate_positions(topology.n_atoms, rng), unit.nanometer)
```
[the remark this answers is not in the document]
Yoshanuikabundi: Done with above

openff/toolkit/topology/molecule.py
```diff
@@ -1874,7 +1863,25 @@ def are_isomorphic(
...
+        def _object_to_n_atoms(obj):
+            import networkx as nx
+
+            if isinstance(obj, FrozenMolecule):
+                return obj.n_atoms
+            elif isinstance(obj, nx.Graph):
+                return obj.number_of_nodes()
+            else:
+                raise RuntimeError(
+                    f"Unsupport object of type {type(obj)} passed to "
```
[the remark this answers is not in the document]
Yoshanuikabundi: Done, and also fixed the error I cribbed this code from.

openff/toolkit/topology/molecule.py
```diff
@@ -5653,34 +5663,39 @@ def __init__(self, scheme, identifier, atom_indices):
...
+    def n_atoms(self):
+        """
+        The number of atoms in this hierarchy element.
+        """
+        return len(self.atom_indices)
+
     @property
     def atoms(self):
+        """
+        Iterator over the atoms in this hierarchy element.
```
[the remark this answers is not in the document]
Yoshanuikabundi: XD

openff/toolkit/topology/topology.py
```diff
@@ -1575,42 +1613,159 @@ def to_file(
...
-
-        """
+        filename
+            Name of the file to write to. Mutually exclusive and collectively
+            exhaustive with ``file``; provide exactly one of ``filename`` or
+            ``file``.
+        file
+            A file-like object to write to. Mutually exclusive and collectively
+            exhaustive with ``filename``; provide exactly one of ``filename`` or
+            ``file``.
```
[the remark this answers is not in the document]
Yoshanuikabundi: Yay cool things! Do we want to rename `keepIds` to `keep_ids` while we're at it?

examples/toolkit_showcase/toolkit_showcase.ipynb
j-wags: Nice! Thanks.

j-wags: Perfection!

j-wags: :thumbsupparrot:

j-wags: Nice.

openff/toolkit/tests/test_topology.py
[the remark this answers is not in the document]
j-wags: Nailed it!

openff/toolkit/topology/topology.py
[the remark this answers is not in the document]
j-wags: Yeah, sure.

openff/toolkit/topology/topology.py
```diff
@@ -1079,15 +1099,19 @@ def identical_molecule_groups(self):
...
+        identity_maps = self._identify_chemically_identical_molecules()
+        groupings = defaultdict(list)
+        for molecule_idx, (unique_mol, atom_map) in identity_maps.items():
+            groupings[unique_mol] += [[molecule_idx, atom_map]]
+
+        self._cached_chemically_identical_molecules = groupings
+
+        return self._cached_chemically_identical_molecules
 
     def _identify_chemically_identical_molecules(self):
```
mattwthompson, line 1116: It's in the docstring but it might be useful to annotate it here (and of course please double-check that this is correct)

```suggestion
    def _identify_chemically_identical_molecules(self) -> Dict[int, Tuple[int, Dict[int, int]]]:
```

openff/toolkit/topology/topology.py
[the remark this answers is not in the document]
Yoshanuikabundi: Done

openff/toolkit/topology/topology.py
```diff
@@ -1557,7 +1589,13 @@ def to_openmm(self, ensure_unique_atom_names=True):
 
     @requires_package("openmm")
     def to_file(
```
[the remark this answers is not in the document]
Yoshanuikabundi: New functionality to test:
- [ ] `ensure_unique_atom_names` argument
    - [ ] `True` is previous behaviour
    - [ ] `False` skips uniquification
    - [ ] `str` is per hierarchy element
- [ ] Passing file-like object instead of path
    - [ ] Path still works
    - [ ] File like object works
- [ ] Optional positions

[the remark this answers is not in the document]
Yoshanuikabundi: There is a lot of repeated code in the existing tests for `to_file`. I don't want to change the existing functionality tests in a PR where I'm trying to avoid changing the functionality they test, but I think these tests could be improved (and reorganised) at a later date.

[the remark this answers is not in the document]
Yoshanuikabundi: Just recording a bug I found in the existing behavior while writing these tests: `Topology.to_openmm(ensure_unique_atom_names=False)` will raise an exception from the guts of OpenMM if any atom name is `None`, such as when a molecule is initialized from a SMILES. Might be frustrating for a user who is combining molecules from lots of different places and doesn't want to risk losing atom names. A simple fix might just be to call `str()` on the atom name before passing it to OpenMM, so that missing atom names are rendered in the PDB as the string "None" 😂

openff/toolkit/topology/topology.py
```diff
@@ -1414,7 +1434,7 @@ def from_openmm(cls, openmm_topology, unique_molecules=None):
         return topology
 
     @requires_package("openmm")
-    def to_openmm(self, ensure_unique_atom_names=True):
+    def to_openmm(self, ensure_unique_atom_names: Union[str, bool] = "residues"):
```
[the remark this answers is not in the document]
Yoshanuikabundi: Done

openff/toolkit/topology/topology.py
Yoshanuikabundi: Done! And I've given the public method that calls this one an annotation too.

Yoshanuikabundi:
OK well this PR is a monstrosity. Sorry you have to review it @j-wags. I think it is done though, and I expect the currently running tests to pass!

docs/releasehistory.md
```diff
@@ -55,6 +55,49 @@ print(value_roundtrip)
...
++ topology.to_file("out.pdb", xyz, keep_ids=True)
+```
+
+#### Non-breaking changes
+
+In addition to these breaking changes, the `positions` argument is now optional. If it is not provided, positions will be taken from the first conformer of each molecule in the topology. If any molecule has no conformers, an error will be raised.
+
+### Breaking change: List to Tuple in `Topology.identical_molecule_groups`
+
+The [`Topology.identical_molecule_groups`] property has been refactored for performance and improved typing support. It now returns a value of type `{int:[(int, {int: int})]}`, where it previously returned `{int:[[int, {int: int}]]}`; the interior 2-element list has been changed to a tuple to allow the type of each element to be specified separately. This is technically a breaking change but is unlikely to cause any issues.
```
j-wags, line 97: (blocking) It's nice that you thought to include this! But it's not actually an API break because it was never in a release. You can either remove it or announce identical_molecule_groups as a new API point.

openff/toolkit/tests/test_molecule.py
```diff
@@ -897,9 +897,11 @@ def test_create_from_file(self):
 
     def test_from_pathlib_path(self):
         ethanol = create_ethanol()
-        ethanol.to_file("ethanol.sdf", file_format="sdf")
+        with NamedTemporaryFile(suffix=".sdf") as outfile:
+            filename = str(outfile.name)
+            ethanol.to_file(filename, file_format="sdf")
 
-        Molecule.from_file(pathlib.Path("ethanol.sdf"))
+            Molecule.from_file(pathlib.Path(filename))
```
j-wags, line 904: 👍

openff/toolkit/tests/test_topology.py
```diff
@@ -820,6 +825,146 @@ def test_to_file_multi_molecule_different_order(self):
...
+    def test_to_file_ensure_uniqueness_residues(self):
+        """
+        Checks that ensure_unique_atom_names="residues" provides per-residue unique atom names
+        """
+        from io import StringIO
+
+        from openff.toolkit.topology import Molecule, Topology
+
+        mol = Molecule.from_polymer_pdb(
+            get_data_file_path("proteins/MainChain_ALA_ALA.pdb")
```
j-wags, line 946: (not blocking) The atom names in this file are already unique, so we wouldn't expect any atom renaming to happen in this test. It's still a useful test as-is though.

openff/toolkit/tests/test_topology.py
```diff
@@ -930,6 +1107,188 @@ def test_to_openmm_do_not_assign_unique_atom_names(self):
...
+        """
+        Test that to_openmm preserves atom names that are unique per-residue by default
+        """
+        # Create a topology from a capped dialanine
+        peptide = Molecule.from_polymer_pdb(
+            get_data_file_path("proteins/MainChain_ALA_ALA.pdb")
+        )
+        off_topology = Topology.from_molecules([peptide])
+
+        # Remove atom names from some residues, make others have duplicate atom names
```
j-wags, line 1157: Brilliant!

openff/toolkit/tests/test_topology.py
```diff
@@ -930,6 +1107,188 @@ def test_to_openmm_do_not_assign_unique_atom_names(self):
...
+        assert (
+            len(atom_names) == 32
+        ), "There should not be duplicate atom names in this output topology"
+
+    @pytest.mark.parametrize(
+        "ensure_unique_atom_names", [True, "residues", "chains", False]
+    )
+    def test_to_openmm_copies_molecules(self, ensure_unique_atom_names):
+        """
+        Check that generating new atom names doesn't affect the input topology
```
j-wags, line 1250: 👍

j-wags (approved):
What a labor of love. Left a few small comments but few are blocking. Overall this looks great. Merge when ready :-)

openff/toolkit/topology/topology.py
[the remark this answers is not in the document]
j-wags: Ah, that's a little obnoxious. I've opened an issue so we can handle it in a bugfix release.

openff/toolkit/tests/test_topology.py
Yoshanuikabundi: Yeah I guess the point of this test is just to make sure the argument is getting passed along to `to_openmm`. The exact behavior is tested in the tests for that method.

docs/releasehistory.md
Yoshanuikabundi: Oh nice! Easy

Yoshanuikabundi:
This must be a new GH feature - I think by making changes after you gave an approving review I've undone your approval. I'll merge with my admin privileges once tests pass and I've double checked the rendered docs but this might need addressing for the future.

PR #1376, opened 2022-08-24
Migration guide

Yoshanuikabundi:
This adds (nearly) every breaking change to the migration guide, as well as how to change user code in as many cases as possible.

I've prepared this by going over the diff between `main` and `0.10.6`. I've tried to catch every removed public API point, but I might have missed a couple and I have comments in the guide outlining the parts I still have to check.  I've also tried to explain why each change was done (though a lot of this was guesswork), and how to migrate code to the new version. I haven't tried to tie these to the PRs that are missing from the changelog - my goal was to give users a path to update, not to document every change

I have not documented behavior changes, coz its basically impossible to deduce from a diff. The behavior changes I implemented have been written up though, and if there are others people can point out I'm happy to add them too!

I think I've also caught the major new features in this release, and explained what they're for and how to find more info.

There are two changes to the code (and I have more suggestions for small deprecations, but I'll get to that another day):

- I removed `Molecule.from_pdb`. This is an API point that has been deprecated despite never being in a release. I'm happy to undo this, but I didn't want it to get released just because it slipped our minds.
- I've added greedy imports to `openff.toolkit`, but only in type check mode. This doesn't change the time it takes to import, but it does mean that type checkers can find the types of the variables that would be imported lazily. I've found this frustrating in some hobby projects that use the toolkit.

This PR also has some hopefully-uncontroversial changes to other parts of the docs.

Not quite finished, but done for today. I think when this is done, it might make sense to make the migration guide it's own page, rather than a section at the top of the changelog.

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -149,8 +149,7 @@ The following properties have been **deprecated** and will be removed in a futur
 - `Topology.topology_bonds` (use [`Topology.bonds`](Topology.bonds) instead)
 - `Topology.n_topology_particles` (use [`Topology.n_particles`](Topology.n_particles) instead)
 - `Topology.topology_particles` (use [`Topology.particles`](Topology.particles) instead)
-- `Topology.n_reference_molecules` (use [`Topology.n_unique_molecules`](Topology.n_unique_molecules) instead)
-- `Topology.reference_molecules` (use [`Topology.unique_molecules`](Topology.unique_molecules) instead).
+- `Topology.n_reference_molecules` (use [`Topology.n_molecules`](Topology.n_molecules) instead)
 - `Topology.n_topology_molecules` (use [`Topology.n_molecules`](Topology.n_molecules) instead)
 - `Topology.topology_molecules` (use [`Topology.molecules`](Topology.molecules) instead)
 - `Topology.n_particles` (use [`Topology.n_atoms`](Topology.n_atoms) instead)
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -118,13 +118,15 @@ Code for applying parameters to topologies has been removed from the Toolkit. Th
 
 The [`ForceField.create_interchange()`] method has been added, and the [`ForceField.create_openmm_system()`] method now uses Interchange under the hood.
 
-The following classes and methods have been removed from `openff.toolkit.typing.engines.smirnoff.parameters`:
-- NonintegralMoleculeChargeException
-- NonbondedMethod
-- ParameterHandler.assign_parameters()
-- ParameterHandler.postprocess_system()
-- ParameterHandler.check_partial_bond_orders_from_molecules_duplicates()
-- ParameterHandler.assign_partial_bond_orders_from_molecules()
+The following classes and methods have been **removed** from `openff.toolkit.typing.engines.smirnoff.parameters`:
+- `NonintegralMoleculeChargeException`
+- `NonbondedMethod`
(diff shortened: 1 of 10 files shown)


docs/releasehistory.md
```diff
@@ -55,6 +100,90 @@ print(value_roundtrip)
...
+`Topology` objects now store complete copies of their constituent `Molecule` objects, rather than using simplified classes specific to `Topology`. This dramatically simplifies the code base and allows the use of the full `Molecule` API on molecules inside topologies.
+
+The following classes have been removed:
+- `TopologyAtom`
+- `TopologyBond`
+- `TopologyMolecule`
+- `TopologyVirtualParticle`
+- `TopologyVirtualSite`
+
+In addition, the `Topology.identical_molecule_groups` property has been added, to facilitate iterating over copies of isomorphic molecules in a `Topology`.
```
mattwthompson, line 126: IMO it's not necessary to create links everywhere in this document, but this one stuck out to me as an "oh I'd like to click through to see what this method does"

docs/releasehistory.md
```diff
@@ -55,6 +100,90 @@ print(value_roundtrip)
...
+- `Molecule.add_trivalent_lone_pair_virtual_site()`
+- `Molecule.add_bond_charge_virtual_site()`
+- `Topology.n_topology_virtual_sites`
+- `Topology.topology_virtual_sites`
+- `Topology.virtual_site()`
+- `Topology.add_particle()`
+
+### Breaking change: Removed `Topology.charge_model` and `Topology.fractional_bond_order_model`
+
+To maintain a clear distinction between a model and the chemistry it represents, the `Topology.charge_model` and `Topology.fractional_bond_order_model` properties have been removed. Charge models and FBOs are now the responsibility of the ForceField.
```
mattwthompson, line 162: I wonder if it would be useful to add (a smithed version of) "these properties weren't even used anywhere lol"; it might be worth emphasizing that these never talked to `Molecule.assign_partial_charges`/`Molecule.assign_fractional_bond_orders` at all, in addition to the existing note about topological representations vs. force field settings

mattwthompson (commented):
I read through most of the [rendered diff](https://openff-docs--1376.org.readthedocs.build/projects/toolkit/en/1376/releasehistory.html) and am offering a few drive-by comments. Overall I think it's great! I'll let @j-wags be the final reviewer unless he wants me to 🪁

I thought a bit about what other stuff (like how I deleted thousands of lines of code in `parameters.py`) but I'm not immediately sure how much of that is public-facing enough to warrant space in a migration guide. Users' workflows depending heavily on `ParameterHandler.create_force` will need more guidance than a short blurb here can accomplish 🥲

j-wags (approved):
This PR looks great in any form (with or without further work done) - I'll meet with @mattwthompson this afternoon to determine whether there are further technical blockers. Once all the technical stuff is lined up I'll merge this and cut the release. I'll communicate the technical progress to @Yoshanuikabundi so you don't start on something and then have me merge this out from under your feet.

docs/releasehistory.md
Yoshanuikabundi: Done! And good thing too, turns out the docstring needed some help to render correctly. Thanks!

docs/releasehistory.md
Yoshanuikabundi: Mmm good smithing. Done!

j-wags:
Awesome! Thanks @Yoshanuikabundi!!

PR #1377, opened 2022-08-25
Deprecate `Topology.reference_molecules` and add `Topology.unique_molecules`

j-wags:
- [x] @Yoshanuikabundi pointed out that we redirected `Topology.reference_molecules` to `Topology.molecules`. This doesn't provide equivalent functionality, however `Topology.identical_molecule_groups` DOES allow us to do so. This PR redirects `Topology.reference_molecules` to a new API point, `Topology.unique_molecules`.


mattwthompson:
How important is this functionality? If it were removed, users could still do it themselves with the 1-liner build from the function contents.

Yoshanuikabundi:
I think the biggest pro for this change is that `unique_molecule_groups` is a complicated function with a complicated signature, and this is a much simpler way of interacting with it. We also need to deprecate `reference_molecules` one way or the other, and this preserves its core functionality rather than changing it for a couple of releases before removing it.

Yoshanuikabundi (approved):
LGTM! I just made a couple changes to fix lints and bring this in line with changes from the migration guide PR. Remaining test failures seem to be spurious or intermittent.

j-wags:
> How important is this functionality?

I shot down most of @Yoshanuikabundi's proposals during our check in yesterday, but this is the right mix of "useful" (replaces an existing API point that we'd otherwise remove) and "simple" (two line function) where it's worth doing. While it's quick for a user to write these two lines once they know that `identical_molecule_groups` exists, that would take them a good amount of digging to learn about. So we decided to go ahead with this one.

PR #1378, opened 2022-08-25
Docs and simplified kwarg handling for `create_openmm_system`

Yoshanuikabundi:
The `create_openmm_system` method does not document its keyword arguments, including that one is deprecated. This PR adds this documentation and updates the `create_interchange` docstring to match. I've also added the deprecation notice to the migration guide.

I also can't think of any reason to keep the bespoke keyword handling logic, except to avoid breaking users who rely on the exception for some reason? So I've refactored that. I've maintained the prevention of passing keyword arguments by position and documented the change of exception in the migration guide. I've kept this change all together and separate from the others in commit 962de55; in the likely case that we don't want it for some reason I haven't thought of, that single commit can be reverted and this PR will just have the documentation changes.

I think I'm done for the night - if release happens tomorrow, feel free to merge this without me!

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -118,6 +118,9 @@ Code for applying parameters to topologies has been removed from the Toolkit. Th
 
 The [`ForceField.create_interchange()`] method has been added, and the [`ForceField.create_openmm_system()`] method now uses Interchange under the hood.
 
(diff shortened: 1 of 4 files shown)


j-wags (approved):
Looks great - Thanks @Yoshanuikabundi

PR #1381, opened 2022-08-26
Recommend `interchange.to_openmm_topology()` in migration guide

Yoshanuikabundi:
This is an important API, and the previous migration suggestion might have led to a lot of missing virtual sites in the future from people doing `interchange.topology.to_openmm()` rather than `interchange.to_openmm_topology()`


Yoshanuikabundi:
Hey @mattwthompson  @j-wags, this PR and  #1380 should both be merged before 0.11.0 is released. They include important notes in the migration guide.

PR #1385, opened 2022-08-26
Re-enable examples tests

j-wags:


mattwthompson:
No clue why it's failing to import `pdbfixer`. It's definitely in the environment and nothing funny happens when I try the import locally from the same environment file.

mattwthompson:
I looked briefly at the PDBFixer source code and couldn't find anything weird about how it's packaged.

mattwthompson:
The short of it is that nothing in particular is wrong with PDBFixer - is was calling everything from the base environment and for whatever reason PDBFixer was the first import. This has _only_ happened 

There's something quirky with how `mamba env update` works with the activated environments[^1] - whatever the cause, this was not an issue before I pushed for the switch from `setup-miniconda` to `provision-with-micromamba`. I wasn't able to quickly think of and test a suitable workaround given the number of ways things could go wrong, so I'm just reverting the examples workflow back to `setup-miniconda`. It's still very slow, but should at least work.

[^1]: I'm able to drop into a debug environment using [`action-tmate`](https://github.com/mxschmitt/action-tmate), but it doesn't include the activated environment an something's up with the shell - I get the dreaded `CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.` despite ... it definitely being configured. There are ways to dig in in there I'll explore after the release.

.github/workflows/examples.yml
```diff
@@ -119,20 +121,12 @@ jobs:
...
-          NB_ARGS+=" --ignore=examples/virtual_sites/vsite_showcase.ipynb"
-
           # openmmforcefields incompatible with new toolkit
           NB_ARGS+=" --ignore=examples/external/swap_amber_parameters/swap_existing_ligand_parameters_with_openmmforcefields.ipynb"
 
-          # JM will do some smithing on this before 0.11.0 release
-          NB_ARGS+=" --ignore=examples/toolkit_showcase/toolkit_showcase.ipynb"
-
-          pytest $NB_ARGS examples
+          python -m pytest $NB_ARGS examples
```
mattwthompson, line 129: I can be talked out of this change; it's not needed here but I have a preference for running pytest and other tools as modules where possible

mattwthompson:
Haha, just kidding. It's failing for the same reason and I have no clue why.

mattwthompson:
Codecov is failing, probably because of being [rate-limited](https://github.com/codecov/codecov-action/issues/557#issuecomment-1216749652). I changed the upload to work using a repo-specific token. My process was

1. Copy the upload token from `https:// app (dot) codecov (dot) io/gh/ openforcefield / openff-toolkit / settings`
2. Paste the token into the _repository_ secrets at `https:// github (dot) com / openforcefield / openff-toolkit / settings / secrets / actions`
3. [Pass the token](https://github.com/openforcefield/openff-toolkit/pull/1385/commits/c3bd6701fffcc794eed43b76a460c21edd18172d) to the action [as suggested](https://github.com/codecov/codecov-action/issues/557#issuecomment-1224970469)

mattwthompson:
Not sure why the simulation in the toolkit showcase runs fine on macOS but appears to produce a mangled trajectory on Linux: https://github.com/openforcefield/openff-toolkit/actions/runs/2952241015

j-wags:
My top suspect is the netcdf format - Checking whether there's chatter on any of their [issue trackers](https://github.com/Unidata/netcdf4-python/issues?q=is%3Aissue+is%3Aclosed) about recent releases.

j-wags:
So at this point, Binder isn't able to minimize the Toolkit Showcase starting geometry in under 20 minutes. This is a problem, but it's out of scope for this PR. I've switched the binder env to something that Just Works and we can deal with the specific runtime/linux netcdf problems in a subsequent release.

mattwthompson (approved):
LGTM - we discussed the limitations and decided to push them off into after the release

* Something is weird with the NetCDF files in the toolkit showcase (but only on Linux runners)
* The Binder environment is pretty slow

PR #1390, opened 2022-08-31
Move testing of documentation notebooks from RTD to CI

Yoshanuikabundi:
The Molecule Cookbook page of the users guide is a Jupyter notebook. Currently, it is tested when the documentation is built by RTD. This frequently causes issues as errors in the notebook do not reliably cause Sphinx to fail the build. This PR bakes the correct output into `molecule_cookbook.ipynb` and adds a check to CI that notebooks in the documentation run and produce the correct output.


mattwthompson:
I suppose these warnings couple be emitted from within MDTraj?

```
/home/joshmitchell/Documents/openff/toolkit/.soap/test/lib/python3.10/site-packages/numpy/core/getlimits.py:500: UserWarning: The value of the smallest subnormal for <class 'numpy.float64'> type is zero. 
  setattr(self, word, getattr(machar, word).flat[0]) 
/home/joshmitchell/Documents/openff/toolkit/.soap/test/lib/python3.10/site-packages/numpy/core/getlimits.py:89: UserWarning: The value of the smallest subnormal for <class 'numpy.float64'> type is zero. 
  return self._float_to_str(self.smallest_subnormal) 
/home/joshmitchell/Documents/openff/toolkit/.soap/test/lib/python3.10/site-packages/numpy/core/getlimits.py:500: UserWarning: The value of the smallest subnormal for <class 'numpy.float32'> type is zero. 
  setattr(self, word, getattr(machar, word).flat[0]) 
/home/joshmitchell/Documents/openff/toolkit/.soap/test/lib/python3.10/site-packages/numpy/core/getlimits.py:89: UserWarning: The value of the smallest subnormal for <class 'numpy.float32'> type is zero. 
  return self._float_to_str(self.smallest_subnormal)

**Comment by Yoshanuikabundi** [comment]
> I suppose these warnings couple be emitted from within MDTraj?

I have no idea, but it seems like it could be.

File: .github/workflows/CI.yml [hunk 1]
```diff
@@ -161,6 +161,11 @@ jobs:
         run: |
           pytest -v --doctest-glob="docs/*.rst" docs/
 
+      - name: Run notebooks in docs
+        if: ${{ matrix.rdkit == true && matrix.openeye == true }}
+        run: |
+          pytest -v --nbval --sanitize-with docs/nbval-sanitize.cfg --ignore docs/_build/ docs/
```
mattwthompson, line 167: I picked up the `--nbval-lax` argument purely out of habit years ago, so I'm not sure which is right. Based on the current documentation, I'm pretty sure we could go with either, given that I don't know of any `#NBVAL_CHECK_OUTPUT` comments. I'd be nice to be standardized, though, and I think we should standardize on `--nbval`.

mattwthompson (changes requested):
I'm pretty sure this is complete with respect to what you set out to do originally, so I'm reviewing now. I found (two instances of) one blocking issue that should be posted from ReviewNB, and also found a few other details that are non-blocking but deserve a second look:

* Should we run this notebook in CI or the examples workflow? I don't have a strong preference but it might be nice to keep all of the notebook stuff in one workflow.
* We should try to standardize `--nbval` vs. `--nbval-lax`. I prefer the former but will let you decide

I won't claim to understand the other tooling you added into here - but it's clearly running in CI, which is plenty for me.

Yoshanuikabundi:
> Should we run this notebook in CI or the examples workflow? I don't have a strong preference but it might be nice to keep all of the notebook stuff in one workflow.

I wanted to run this in an environment with both OpenEye and RDKit to keep warnings to a minimum. OpenEye is a current requirement, as some cells rely on it, and its possible that RDKit will become a requirement if we ever add any molecule creation methods that aren't supported by OpenEye. That meant CI. I'd happily add an OE+RDKit run to the examples CI or create a new documentation CI though.

> We should try to standardize --nbval vs. --nbval-lax. I prefer the former but will let you decide

The purpose of this CI is somewhat different to the examples. In the examples, each cell is dependent on the previous one, and I think it is sufficient to make sure that the examples still run to completion. In the documentation, each cell is basically only dependent on the very first cell, and both the code and the output are curated and displayed, and it's important that they match exactly. So `--nbval` is a requirement here, but not in the examples.

I don't think we should standardize, because keeping the output exactly matched to the input in the examples would be an enormous amount of work both now and forever into the future. I think it would be reasonable not to include outputs in the examples at all.

> I won't claim to understand the other tooling you added into here 

It's basically just to specify which parts of the outputs have to match exactly and which can be a bit vague.

Yoshanuikabundi:
<p>This cell (and others that raise an exception) are tagged as raising an exception, and so pass CI only if they raise an exception. If we added exception handling code, the cells would pass CI whether or not they raised an exception. Exception handling code would also get in the way of the primary purpose of the notebook, which is to be clear documentation.</p>

---

 <a href='https://app.reviewnb.com/openforcefield/openff-toolkit/pull/1390/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

mattwthompson:
I think my feedback was a step backwards and I'd like to defer reviewing to @j-wags - Jeff, do you have time to review this one shortly?

j-wags (approved):
Thanks for the careful work @Yoshanuikabundi, and I would have asked the same questions @mattwthompson so thanks for going ahead with them. I think this PR is a clear improvement, and I'll try to keep in mind that we need to store the output in the notebook and update the sneaky tags for erroring cells. Also I'm super impressed that the NGLView widget renders in the docs preview. That's some real wizardry @Yoshanuikabundi. 

Two non-blocking concerns that don't require action:
1) I am a little worried about the size of the stored nglview widgets. 25 MB of nglview widget came and went in the earlier Toolkit Showcase PR and this will motivate cleaning up the git history. But in this case I think it's appropriate.
2) The `docs/` directory is getting a bit complex and may become harder to maintain. Not a blocker but I certainly wouldn't mind if future PRs reduced the amount of code complexity in that folder :-)

PR #1400, opened 2022-09-15
Fix `from_pdb_and_smiles`

Yoshanuikabundi:
`from_pdb_and_smiles` has a bug where it can mangle the atom ordering, leading to physically incorrect conformers:
mangled_ala5dyeala5_capped

Hilariously, this will happily energy minimize and simulate:

enermin_a5dyea5_capped

This was caused (as @j-wags pointed out) by the mapping between the PDB and SMILES being reversed. This was not caught by tests because the molecule being used was symmetric under this reversal, and because the conformations and atom ordering were not tested. This PR fixes this test, and then swaps the ordering so that it passes.


openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -217,8 +217,8 @@ def from_pdb_and_smiles(
         # check isomorphic and get the mapping if true the mapping will be
         # Dict[pdb_index: offmol_index] sorted by pdb_index
```
mattwthompson, line 218:
```suggestion
        # Dict[offmol_index, pdbmol_index] sorted by offmol index
```
I think this docstring is out of date? I'm trying to figure out the behavior change of this diff and am somewhat confused - now when atom ordering in the PDB file and SMILES disagree, the SMILES wins (by nature of the `Molecule` derived from SMILES being first in this method call)?

j-wags (commented):
I'm like 80% sure that this really was just an embarrassingly simple bug that we've overlooked for 3 years. It'd be good to get to the bottom of this so I'm following up with some users to confirm whether they're seeing the same issue.

openff/toolkit/utils/rdkit_wrapper.py
Yoshanuikabundi: Yeah that comment is out of date, thanks for catching that! I think [Jeff's explanation](https://github.com/openforcefield/openff-toolkit/pull/1400#discussion_r971550949) is exactly right - the previous behavior was incorrect, and the new behavior is the previously documented behavior (PDB wins).

Yoshanuikabundi:
OK I've re-written the tests for this method to be parametric over different molecules and also to include explicit property tests that are independent of the SDF file comparison. Hopefully running these tests with the old and new implementation (just swapping the first two arguments to are_isomorphic) makes it clear why the old behavior is wrong and the new behavior correct - it looks like just using a different SMILES for toluene was enough to break some of the new tests for the old behavior!

openff/toolkit/utils/rdkit_wrapper.py
mattwthompson: So the resulting atom ordering is NEITHER the atom ordering in the PDB file, the SMILES (as processed by RDKit + OpenFF), but instead a third atom ordering? Am I totally misunderstanding what remapping?

Yoshanuikabundi: I'm not sure I'm understanding you - the behavior prior to this PR is that the atoms are given a third ordering, but the coordinates are assigned as though the PDB order prevailed. This meant that we violated a guarantee we made in the docstring, AND that the coordinates were incorrect. This PR fixes that bug so that the coordinates are correct and the PDB ordering prevails.

mattwthompson: Okay, I misread.

j-wags (approved):
Looks great. Thanks @Yoshanuikabundi!

PR #1408, opened 2022-09-22
Update `examples/README.md` to include more examples

mattwthompson:
I wanted to look up the virtual sites example and I happened to already have the docs open in a tab. I figured I'd be able to get to the notebook with a couple of clicks, and I was partially right. I was right in that the docs link to examples, but it only lists those in `examples/README.md`, which is out of date. See what's missing: https://docs.openforcefield.org/projects/toolkit/en/stable/examples.html

I doubt this was intentional - or at least I can say I've never remembered to update this file while examples are moved around.


j-wags (approved):

mattwthompson:
Nice, this worked on the first try:

image

PR #1421, opened 2022-10-06
Allow D- and F- block radicals

### LINKED ISSUE #1420: Metal ions with unpaired electrons are rejected as radicals with RDKit
**Describe the bug**

Metal ions like `[Cu+2]` are rejected with the RDKit toolkit (but not openeye) because they have unpaired electrons. I expect this will affect at least one oxidation state for any element that has at least one odd and even oxidation state, which is a lot of metals.

Speaking with @j-wags on our check-in today, we agreed that the best resolution would be to figure out what OpenEye is doing and do that for RDKit so that we at least have consistency. If it turns out OpenEye is doing something we consider wrong, we can address that later.

**To Reproduce**

```python

Molecule.from_smiles("[Cu++]") # Raises RadicalsNotSupportedError
```

**Output**

```
RadicalsNotSupportedError                 Traceback (most recent call last)
Cell In [6], line 3
      1 from openff.toolkit import Molecule
----> 3 Molecule.from_smiles("[Cu++]")

File ~/Documents/openff/bespokefit/.soap/rdkit/lib/python3.10/site-packages/openff/toolkit/topology/molecule.py:1769, in FrozenMolecule.from_smiles(cls, smiles, hydrogens_are_explicit, toolkit_registry, allow_undefined_stereo)
   1741 """
   1742 Construct a Molecule from a SMILES representation
   1743 
   (...)
   1766 
   1767 """
   1768 if isinstance(toolkit_registry, ToolkitRegistry):
-> 1769     molecule = toolkit_registry.call(
   1770         "from_smiles",
   1771         smiles,
   1772         hydrogens_are_explicit=hydrogens_are_explicit,
   1773         allow_undefined_stereo=allow_undefined_stereo,
   1774         _cls=cls,
   1775     )
   1776 elif isinstance(toolkit_registry, ToolkitWrapper):
   1777     toolkit = toolkit_registry

File ~/Documents/openff/bespokefit/.soap/rdkit/lib/python3.10/site-packages/openff/toolkit/utils/toolkit_registry.py:359, in ToolkitRegistry.call(self, method_name, raise_exception_types, *args, **kwargs)
    357             for exception_type in raise_exception_types:
    358                 if isinstance(e, exception_type):
--> 359                     raise e
    360             errors.append((toolkit, e))
    362 # No toolkit was found to provide the requested capability
    363 # TODO: Can we help developers by providing a check for typos in expected method names?

File ~/Documents/openff/bespokefit/.soap/rdkit/lib/python3.10/site-packages/openff/toolkit/utils/toolkit_registry.py:355, in ToolkitRegistry.call(self, method_name, raise_exception_types, *args, **kwargs)
    353 method = getattr(toolkit, method_name)
    354 try:
--> 355     return method(*args, **kwargs)
    356 except Exception as e:
    357     for exception_type in raise_exception_types:

File ~/Documents/openff/bespokefit/.soap/rdkit/lib/python3.10/site-packages/openff/toolkit/utils/rdkit_wrapper.py:1007, in RDKitToolkitWrapper.from_smiles(self, smiles, hydrogens_are_explicit, allow_undefined_stereo, _cls)
    998         if atom.GetNumImplicitHs() != 0:
    999             raise ValueError(
   1000                 f"'hydrogens_are_explicit' was specified as True, but RDKit toolkit interpreted "
   1001                 f"SMILES '{smiles}' as having implicit hydrogen. If this SMILES is intended to "
   (...)
   1004                 f"Molecule.from_rdkit() to create the desired OFFMol."
   1005             )
-> 1007 molecule = self.from_rdkit(
   1008     rdmol,
   1009     _cls=_cls,
   1010     allow_undefined_stereo=allow_undefined_stereo,
   1011     hydrogens_are_explicit=hydrogens_are_explicit,
   1012 )
   1014 return molecule

File ~/Documents/openff/bespokefit/.soap/rdkit/lib/python3.10/site-packages/openff/toolkit/utils/rdkit_wrapper.py:1718, in RDKitToolkitWrapper.from_rdkit(self, rdmol, allow_undefined_stereo, hydrogens_are_explicit, _cls)
   1714 for rda in rdmol.GetAtoms():
   1715 
   1716     # See issues #1075 for some discussion on radicals
   1717     if rda.GetNumRadicalElectrons() != 0:
-> 1718         raise RadicalsNotSupportedError(
   1719             "The OpenFF Toolkit does not currently support parsing molecules with radicals. "
   1720             f"Found {rda.GetNumRadicalElectrons()} radical electrons on molecule {Chem.MolToSmiles(rdmol)}."
   1721         )
   1723     rd_idx = rda.GetIdx()
   1724     # if the molecule was made from a mapped smiles this has been hidden
   1725     # so that it does not affect the sterochemistry tags

RadicalsNotSupportedError: The OpenFF Toolkit does not currently support parsing molecules with radicals. Found 1 radical electrons on molecule [Cu+2].
```

 - BespokeFit CI and locally

Yoshanuikabundi:
Closes #1420.

OK I've dug into this issue a bit more now. The RDKit wrapper uses the `Atom.GetNumRadicalElectrons()` method, which looks at the atom's electron configuration and reports the number of unpaired electrons. If there are any unpaired electrons, we raise an error. This matches the chemical definition of radical exactly, but means that d- and f-block elements that we don't usually think about as radicals can be excluded.

The OpenEye wrapper is not equivalent, as openeye doesn't provide the equivalent  - it instead gets what RDKit calls the "standard valence", or what OpenEye calls the "MDL model valence" for the atom, and if it's different to the actual valence (including formal charge) then we call it a radical. This doesn't match the chemical definition of radical, as it's more about whether there are dangling bonds rather than unpaired electrons. However, it does pretty closely match what chemists usually mean when they say radical - I don't think I've ever heard a transition metal called a radical, even though plenty of them technically are.

I tried to think of weird radicals that would demonstrate the difference, but apart from d- and f-block elements the approach we use in the OpenEye wrapper seems to basically work. Unfortunately, the MDL valence model isn't available in RDKit. It's documented in the OE docs [here](https://docs.eyesopen.com/toolkits/python/oechemtk/valence.html#mdl-valence-model). I think we would end up matching it pretty closely just by skipping the radical check for d- and f- block elements (transition metals, lanthanides, actinides) and using the existing code for everything else.

So TL;DR:
 - RDKit's `Atom.GetNumRadicalElectrons()` is too expansive and doesn't have an equivalent in openeye
 - OpenEye's `oechem.OEMDLGetValence()` is about right and doesn't have an equivalent in RDKit
 - I think "Accept all d- and f- block elements, and other elements when they don't have unpaired electrons" will be about equivalent to the current OE implementation, so that's what I've written.

This approach is also safe with respect to #1075, as implicit hydrogens are never added to D- and F- block elements.

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -8,6 +8,9 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ## Current development
 
+## Behavior changes
+- [PR #1421](https://github.com/openforcefield/openff-toolkit/pull/1421): Allow `Molecule.from_rdkit()` to load D- and F- block radicals, which cannot have implicit hydrogens.
+
 ## 0.11.1 Minor release forbidding loading radicals
 
 ## Behavior changes
--- a/openff/toolkit/tests/test_toolkits.py
+++ b/openff/toolkit/tests/test_toolkits.py
@@ -2052,7 +2052,7 @@ class TestRDKitToolkitWrapper:
 
     def test_rdkit_from_smiles_transition_metal_radical(self):
         """Test that parsing an SMILES with a transition metal radical works."""
-        RDKitToolkitWrapper().from_smiles("[Zn2+]")
+        RDKitToolkitWrapper().from_smiles("[Zn+2]")
 
     def test_rdkit_from_smiles_hydrogens_are_explicit(self):
         """
--- a/openff/toolkit/tests/test_toolkits.py
+++ b/openff/toolkit/tests/test_toolkits.py
@@ -342,6 +342,10 @@ class TestOpenEyeToolkitWrapper:
         with pytest.raises(RadicalsNotSupportedError):
             OpenEyeToolkitWrapper().from_smiles("[CH3]")
 
+    def test_openeye_from_smiles_transition_metal_radical(self):
+        """Test that parsing an SMILES with a transition metal radical works."""
+        OpenEyeToolkitWrapper().from_smiles("[Zn+2]")
+
     # TODO: test_smiles_round_trip
 
     def test_smiles_add_H(self):
@@ -665,7 +669,7 @@ class TestOpenEyeToolkitWrapper:
         assert oechem.OEHasImplicitHydrogens(oe_molecule)
 
     def test_from_openeye_radical(self):
-        """Test that parsing an rdmol with a radical raises RadicalsNotSupportedError."""
+        """Test that parsing an oemol with a radical raises RadicalsNotSupportedError."""
         from openeye import oechem
 
         smiles = "[H][C]([H])[H]"
@@ -674,6 +678,16 @@ class TestOpenEyeToolkitWrapper:
 
         with pytest.raises(RadicalsNotSupportedError):
             OpenEyeToolkitWrapper().from_openeye(oemol)
+
+    def test_from_openeye_transition_metal_radical(self):
+        """Test that parsing an oemol with a transition metal radical works."""
+        from openeye import oechem
+
+        smiles = "[Zn+2]"
+        oemol = oechem.OEGraphMol()
+        oechem.OESmilesToMol(oemol, smiles)
+
+        OpenEyeToolkitWrapper().from_openeye(oemol)
 
     def test_from_openeye_implicit_hydrogen(self):
         """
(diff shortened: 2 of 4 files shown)


mattwthompson:
Well this certainly validates how difficult I found it to be to implement the first time. I kept thinking to myself "surely there's one-call parity for such a simple thing" - nope, guess not!

j-wags (approved):
Fantastic. Thanks @Yoshanuikabundi!

PR #1427, opened 2022-10-13
Remove star imports

mattwthompson:


setup.cfg
```diff
@@ -24,10 +24,10 @@ per-file-ignores =
     openff/toolkit/topology/__init__.py:F401
     openff/toolkit/topology/molecule.py:
     openff/toolkit/topology/topology.py:F821
-    openff/toolkit/typing/chemistry/__init__.py:F401,F403
-    openff/toolkit/typing/engines/smirnoff/__init__.py:F401,F403
+    openff/toolkit/typing/chemistry/__init__.py:F401
```
j-wags, line 27: (not blocking) This seems pretty brittle - Are there easy options to have this file not exist?

j-wags (approved):
LGTM!

setup.cfg
mattwthompson: The easiest route would be modifying the code to pass all `flake8` checks - something I'd be happy to do in a separate PR but not something that would be a trivial amount of work. I'd argue most of the checks `flake8` attempts to enforce are good ones.

mattwthompson:
I had to resolve a merge conflict related to #1429; I believe it should be fine given that we're still maintaining the the import order of


```python3
from openff.toolkit.utils.utils import ...
from openff.toolkit.utils.toolkits import ...

PR #1236, opened 2022-03-29
Raise exception when processing molecule with radical electrons

mattwthompson:
Here's a block to add to the changelog, keeping it out of the tree to make merges easier:

```
- [PR #1236](https://github.com/openforcefield/openff-toolkit/pull/1236) raises
  [`RadicalsNotSupportedError`](openff.toolkit.utils.exceptions.RadicalsNotSupportedError) when
  [`Molecule.from_openeye`](openff.toolkit.topology.Molecule.from_openeye),
  [`Molecule.from_rdkit`](openff.toolkit.topology.Molecule.from_rdkit),
  [`Molecule.from_smiles`](openff.toolkit.topology.Molecule.from_smiles), or similar methods
  detect that a molecule contains radical electrons.
```

Resolves https://github.com/openforcefield/openff-toolkit/issues/1075, assuming I haven't forgotten how chemistry works.

This is a follow-on to #1098 targeting the v0.11.x line.

Two remaining considerations may be:
* Performance evaluation - this adds an O(number of atoms) check to two key conversion methods.
* False positives - I have no idea if this will be too noisy and claim to have found radicals where none exist. Throwing this at a large dataset should give an indication of this.

- [x] Tag issue being addressed
- [x] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/master/openff/toolkit/tests)
- [x] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/master/docs), if applicable
- [x] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [x] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/master/docs/releasehistory.rst)

PR #1429, opened 2022-10-14
Fix possible import loop

mattwthompson:
In #1236 a couple of imports were changed to happen in a different order. This breaks _some_ things ([_some_ `conda-build` builds ](https://github.com/conda-forge/openff-interchange-feedstock/pull/29#issuecomment-1278099222)and [_some_ `pip`-based installs](https://readthedocs.org/api/v2/build/18345132.txt)) but does not seem to break anything we have distributed - or at least we have no reports of this.


mattwthompson:
This has caused me some headache and I believe it was introduced on accident, so I think it's worth merging now. I'll mea culpa if I'm wrong.

PR #1431, opened 2022-10-18
Document how to check installed toolkits

### LINKED ISSUE #1428: Document how to check installed toolkits
The first thing I thought when I saw this PR is that we should document how to check which toolkits are available. I think "Optional dependencies (toolkits)" section of the installation guide is a natural place for this - maybe something like:

````markdown
All available toolkits are automatically registered in the 
`GLOBAL_TOOLKIT_REGISTRY`. The available toolkits and their 
versions can be inspected through the `registered_toolkit_versions`
dictionary:

```python
from openff.toolkit import GLOBAL_TOOLKIT_REGISTRY
print(GLOBAL_TOOLKIT_REGISTRY.registered_toolkit_versions)
```
````

_Originally posted by @Yoshanuikabundi in https://github.com/openforcefield/openff-toolkit/pull/1426#pullrequestreview-1141742528_

mattwthompson:
Closes #1428

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -198,3 +198,14 @@ $ conda install -c openeye -c conda-forge openeye-toolkits
 
 Though OpenEye can be installed for free, using it requires a license file. 
 No essential `openff-toolkit` release capabilities *require* the OpenEye toolkit, but the Open Force Field developers make use of it in parameterizing new open source force fields.
+
+
+### Check installed toolkits
+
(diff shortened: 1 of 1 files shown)


mattwthompson:
https://openff-docs--1431.org.readthedocs.build/projects/toolkit/en/1431/installation.html#check-installed-toolkits

image

Yoshanuikabundi (approved):
LGTM! My first instinct would be to put this before the other toolkits rather than at the end of the section but it's up to you. Thanks!

mattwthompson:
I can see that - my thinking was that users would more often want to know how to install things vs. check them. I don't feel strongly and we can restructure at some point.

PR #1433, opened 2022-10-19
Finalize 0.11.2 release notes

mattwthompson:


mattwthompson:
The rendered docs look okay and the links go where one might expect. I left out some entries that I don't think are important to be user-facing.

image

PR #1444, opened 2022-10-27
Update for Pint 0.20

mattwthompson:
Ref https://github.com/openforcefield/openff-units/pull/51

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -134,9 +134,9 @@ jobs:
             done
           fi
 
-      - name: Install Pint 0.20.0, openff-units#51
+      - name: Install openff-units#51
         run: |
-          python -m pip install "pint==0.20.0"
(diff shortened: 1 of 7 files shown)


mattwthompson:
This it totally fine in the original PR but is totally broken when used outside of the package itself.

mattwthompson:
We actually will need a PR into this toolkit with this update. It can be this PR with most of the debug code stripped out, but I don't want to remove it until I make a new `openff-units` release.

In loading a force field, `object_to_quantity` sends SMIRKS patterns to Pint and if some error was thrown, it just accepted the string as-is. Now a `pint.DefinitionSyntaxError` is thrown, which for some reason was not caught in the existing `except` block.

mattwthompson:
The examples are failing because QCEngine has not been updated, though that it in the works. I think it would be unwise to wait for that before merging and releasing these changes.

j-wags (approved):
LGTM! Thanks for digging in.

Issue #1446, opened 2022-10-30
Unit problem in the toolkit showcase notebook?


ziyuanzhao2000:
In the "Minimize the combined system" step of the [notebook](https://github.com/openforcefield/openff-toolkit/blob/main/examples/toolkit_showcase/toolkit_showcase.ipynb), there's the code as follows:
```
simulation.minimizeEnergy(
    tolerance=openmm_unit.Quantity(value=50.0, unit=openmm_unit.kilojoule_per_mole)
)
minimized_state = simulation.context.getState(
    getPositions=True, getEnergy=True, getForces=True
)
```
However, as I understand, the OpenMM code is supposed to accept the tolerance in the unit of kJ/mol/nm, which has the unit of force rather than energy as currently it stands. Curiously, the code cell does run without problem, and I had no problem running other systems using the wrong unit. Shall we at least fix the notebook to use the correct unit, even if it might be an OpenMM but not OpenFF's problem with unit checking or implicit unit conversion?

ziyuanzhao2000:
By the way, I wish to ask a question related to energy minimization using OpenMM on the system produced from OpenFF's Interchange object. As of now it seems relatively slow, which is mentioned in another issue about the showcase notebook in September. I'm currently writing a tool based upon OpenFF's API for modifying chemical structures (will work for proteins, for sure). A function I've implemented is to procedurally grow functional groups on the existing structure. I hope to do a local energy minimization after every such modification, but as of now using OpenMM seems pretty slow relative to, say, similar minimization functionality provided by softwares like pymol and MOE. For example, for a modified dipeptide, I need 90s to minimize the system... Please kindly advise how I can make it faster, and maybe if we can discuss this in a separate issue. Thanks!

Yoshanuikabundi:
Good catch, thanks! Looks like this was previously [incorrectly documented](https://github.com/openmm/openmm/pull/2972) by OpenMM. We should definitely fix this - I can do it, or if you'd like to we'd welcome a PR!

I'm not aware of anything that would cause an OpenMM `System` produced by our tooling to be slower than any other `System`, but that doesn't mean it doesn't exist! If you have an example of a system that is slower to minimize with a force field produced by Interchange than by one of the default OpenMM force fields, I'm sure @mattwthompson would love you to raise an issue on the [Interchange issue tracker](https://github.com/openforcefield/openff-interchange/issues). 

[...]

PR #1447, opened 2022-10-31
Corrected tolerance to have the unit of force as defined in the OpenMM doc

ziyuanzhao2000:
This is just to fix the issue mentioned here #1446 in the showcase notebook.

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -9,6 +9,9 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ## Current development
 
+### Examples updates
+- [PR #1447](https://github.com/openforcefield/openff-toolkit/pull/1447): Fixed units of tolerance used in OpenMM minimization in Toolkit Showcase example notebook (from @ziyuanzhao2000)
+
 ### Miscellaneous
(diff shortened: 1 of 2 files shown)


j-wags (approved):
Excellent - Thanks for the quick PR @ziyuanzhao2000! As far as I can tell, this is the only place where we provide the tolerance to OpenMM's minimizer, so this should be the only change that's needed. It is a bit concerning that I don't know how OpenMM was interpreting the previous values... and in fact these examples CI runs are taking 5-10 minutes longer than usual...  I'll tinker around with this a bit before merging. 

Half the CI is failing because your PR is coming from a fork, so it can't access some org secrets - That's our problem, not yours :-) The half of the CI that is passing shows that your change works!

I've updated the releasenotes and will merge this once I've resolved the timing thing.

j-wags:
I'm unable to reproduce the timing difference locally, so I'll just assume that this is running slower because the CI machines are busy on a Monday morning :-)

ziyuanzhao2000:
Thanks for your prompt feedback! Glad to hear that there's no timing difference for your local run. I do agree it's a bit concerning how OpenMM handled the previous unit and didn't throw up an error.

PR #1448, opened 2022-11-01
[DNM] Switch CI env to try OpenMM 8 RC

j-wags:


mattwthompson:
For the record, to anybody else following along, this should be resolved by modifying the last cell in the "Parametrize the complex (OpenFF Toolkit)" section to

```
omm_system = interchange.to_openmm(combine_nonbonded_forces=True)
omm_top = interchange.to_openmm_topology()
```

Currently, `combine_nonbonded_forces=False` is the default (Interchange versions through 0.2.2) - although _not_ when wrapped by the toolkit (Toolkit versions 0.11.0-0.11.3):

https://github.com/openforcefield/openff-toolkit/blob/e1c0c7ddfe8cf45f2da98d8b0c4dcb09b7a2501e/openff/toolkit/typing/engines/smirnoff/forcefield.py#L1128

I'd like to change this in a future release as I think combining the non-bonded forces (currently a `NonbondedForce` containing electrostatics, a `CustomNonbondedForce` for vdW, and two `CustomBondForce`s, one for 1-4 electrostatics and the other for 1-4 vdW) into a single `NonbondedForce` can accurately be described as what one would expect to be the default behavior.

However, everything above can _probably_ be ignored because it's _probably_ been fixed upstream! https://github.com/openmm/openmm/issues/3832#issuecomment-1309440153 / https://github.com/openmm/openmm/pull/3840

mattwthompson:
Tagging https://github.com/conda-forge/openmm-feedstock/commit/262e8ca2a11201c1affffd89159d31cffd46388c

mattwthompson:
As of the last `dev` build, CI is passing and 3/4 example runs are passing. The run

> [ubuntu-latest, Python 3.9, RDKit=false, OpenEye=true](https://github.com/openforcefield/openff-toolkit/actions/runs/3451324542/jobs/5785844367#logs)

fails consistently after many restarts. Other example runs do not seem to fail. The issue is some sort of simulation crash in the toolkit showcase example. The `openmm.Platform` used is unspecified, tinkering with that may or may not resolve the issue (to the extent one exists).

j-wags:
Hmm, [here's the diff](https://www.diffchecker.com/rBV83C8q) between the working (RDKit+AmberTools) and non-working (OE) environments. Nothing obvious jumps out at me - Same exact openmm package and cudatoolkit.

What weird is that the failing env has fewer packages than the passing one. So the failures are either from one of the few packages that are unique to the OE environment, or it's because there's a switch in OpenMM that's selecting one of the packages that the RDKit env adds.

I opened up a recent `openeye-toolkits` conda package to see if it was maybe patching something in OpenMM, and I don't see any evidence that it was. 

So my current thinking is that's this is due to an optional package that OpenMM is picking in the RDKit env. But I'll keep looking.

mattwthompson:
At this point I think just toggling `combine_nonbonded_forces=True` could be the most efficient solution. I doubt the issue is directly related to behavior of upstream packages but some corner case involving `CustomNonbonedForce` that only our stack hits, as minor as the differences between wrapped toolkits are.

mattwthompson:
FTR: Reproducing this should now require pulling down Interchange 0.2.2

PR #1458, opened 2022-11-09
Fix collecting old script as a test

mattwthompson:
Nightly CI has been failing after #1442.

Our nightly CI adds the `--runslow` option, which has a handful of effects including opening up `FreeSolv.tar.gz`. Inside this archive are some number of files like `openff/toolkit/data/molecules/FreeSolv/recreate_0_0_2_systems.py` that serve only to document how these artifacts were generated. `pytest` will not usually see these file(s) as they don't match the `test_.*.py` sort of pattern it uses to collect tests. The `--doctest-modules` option, however, is designed to look through source code, so it found these file(s). This was not detectable during the review process since we don't normally run the "slow" tests, so it snuck by us. The path `openff/toolkit/tests` was ignored to avoid this sort of unintended collection, but `openff/toolkit/data/` was missed. There wasn't much reason to suspect it would cause this issue - it's supposed to be a collection of data files, not Python modules, and even looking at the [files on the web UI](https://github.com/openforcefield/openff-toolkit/tree/main/openff/toolkit/data/molecules) doesn't immediately present the issue. This can only be reproduced by running something like `pytest --runslow openff` _before_ running the doctests. I had to this locally to reproduce the error, and this happens to be the [order that tests are run in CI.](https://github.com/openforcefield/openff-toolkit/blob/e1c0c7ddfe8cf45f2da98d8b0c4dcb09b7a2501e/.github/workflows/CI.yml#L150-L171)

This is easy enough to fix, but I think we should drop the `__init__.py` files from the `/data/` directory and its subdirectories. It indicates to Python packaging tools that there's some sort of useful code there that could be imported - which is not the case, as far as I can tell. For example, this is all valid code (that makes me nervous):

```python3
>>> from openff.toolkit.data import *
>>> from openff.toolkit.data.test_forcefields import *
>>> from openff.toolkit.data.systems.amber import *
```

and it gets weirder when these paths have Python files in them:

```python3
>>> from openff.toolkit.data.molecules import FreeSolv
>>> FreeSolv
<module 'openff.toolkit.data.molecules.FreeSolv' (namespace)>
```
I don't think we want this to be a module exposed to the public


.github/workflows/CI.yml
```diff
@@ -151,9 +151,7 @@ jobs:
         run: |
           PYTEST_ARGS+=" --ignore=openff/toolkit/tests/test_examples.py"
           PYTEST_ARGS+=" --ignore=openff/toolkit/tests/test_links.py"
-          if [[ "$GITHUB_EVENT_NAME" == "schedule" ]]; then
-            PYTEST_ARGS+=" --runslow"
-          fi
+          PYTEST_ARGS+=" --runslow"
```
j-wags, line 154: Let's just be sure to revert this before merging!
```suggestion
          if [[ "$GITHUB_EVENT_NAME" == "schedule" ]]; then
            PYTEST_ARGS+=" --runslow"
          fi
```

j-wags (approved):
Good catch. Either solution is good by me - explicitly excluding these folders from tests or removing their inits.

Issue #1453, opened 2022-11-04
Molecule.from_polymer_pdb does not register double bond stereochemistry (E/Z)


ziyuanzhao2000:
**Describe the bug**

If we apply `Molecule.from_polymer_pdb` on a peptide with amino acid which sidechain has a E/Z stereobond (curiously, almost all amino acids do not), the function will complain and fail, even if the `substructure_file` contains the required bond stereochemistry.

**To Reproduce**

`Molecule.from_polymer_pdb('./chemeditor/resources/arginine_4.pdb')`

Use this for the structure of the neutral arginine (I tried the different protonation variants)
[arginine_4.pdb.zip](https://github.com/openforcefield/openff-toolkit/files/9942473/arginine_4.pdb.zip)

**Output**

```
---------------------------------------------------------------------------
UndefinedStereochemistryError             Traceback (most recent call last)
Input In [88], in <cell line: 1>()
----> 1 Molecule.from_polymer_pdb('./chemeditor/resources/arginine_4.pdb', substructure_file_path= './chemeditor/resources/temp_substructures_2.json')

File ~/Documents/OpenFF-chem-editor/chemeditor/molecule.py:266, in Molecule.from_polymer_pdb(cls, file_path, substructure_file_path, toolkit_registry)
    263 with open(substructure_file_path, "r") as subfile:
    264     substructure_dictionary = json.load(subfile)
--> 266 offmol = toolkit_registry.call(
    267     "_polymer_openmm_topology_to_offmol", pdb.topology, substructure_dictionary
    268 )
    270 coords = unit.Quantity(
    271     np.array(
    272         [
   (...)
    277     unit.angstrom,
    278 )
    279 offmol.add_conformer(coords)

File ~/opt/anaconda3/envs/openmm2/lib/python3.9/site-packages/openff/toolkit/utils/toolkit_registry.py:355, in ToolkitRegistry.call(self, method_name, raise_exception_types, *args, **kwargs)
    353             for exception_type in raise_exception_types:
    354                 if isinstance(e, exception_type):
--> 355                     raise e
    356             errors.append((toolkit, e))
    358 # No toolkit was found to provide the requested capability
    359 # TODO: Can we help developers by providing a check for typos in expected method names?

File ~/opt/anaconda3/envs/openmm2/lib/python3.9/site-packages/openff/toolkit/utils/toolkit_registry.py:351, in ToolkitRegistry.call(self, method_name, raise_exception_types, *args, **kwargs)
[...]

PR #1460, opened 2022-11-10
Band-aid #1432 and #1453

j-wags:
- [x] Band-aids (doesn't close) #1432 and #1453 by making RDKit stereo checks non-strict when loading PDBs


j-wags:
If I have time, add the neutral arginine thing here, but it's not a blocker.

PR #1466, opened 2022-11-18
Do not use `collections.OrderedDict`

mattwthompson:
The build-in `dict` has been ordered by key insertion since Python 3.6 so there is not much reason to use `collections.OrderedDict`. There are some corner cases in which they differ but we do not touch them.


openff/toolkit/topology/topology.py
```diff
@@ -429,7 +427,7 @@ def __init__(self, other=None):
             self.copy_initializer(other)
         elif isinstance(other, FrozenMolecule):
             self.from_molecules([other])
-        elif isinstance(other, OrderedDict):
+        elif isinstance(other, dict):
```
mattwthompson, line 430: For the record:

```python3
>>> from collections import OrderedDict
>>> isinstance(OrderedDict(), dict)
True

j-wags (approved):
Nice! I was worried that the removal of `attach_units` and `detach_units` would break the SMIRNOFF spec upgrader script (since the 0.1 SMIRNOFF spec has the units defined at the top of each section, rather than alongside each value). However it looks like that's still working (and thankfully we had tests for it too, I was just paranoid and also tested manually). To ease future-me's worries - the behavior that we need is handled by `recursive_attach_unit_strings`. 

So this looks great. Thanks @mattwthompson. My knees feel so much better not carrying around 260 unnecessary lines. Could you update the releasenotes to reflect the removal of these methods from the API before you merge?

PR #1469, opened 2022-11-18
Update internal references to `main` branch

mattwthompson:
I found some links pointing to the `master` branch ... which we should think about deleting altogether.


j-wags (approved):
LGTM!

Issue #1422, opened 2022-10-10
Label Molecules breaks with plugin vsite handlers


jthorton:
**Describe the bug**
When using a plugin vsite handler such as the one in [qubekit](https://github.com/qubekit/QUBEKit/blob/ac67d937d18c6562cd55d77dddf7ab4e7fd00342/qubekit/cli/utils.py#L283) the label molecules function in the force field fails due to a type check on the normal virtual site handler [here](https://github.com/openforcefield/openff-toolkit/blob/5887e1f1cb8408ea7548650d19d12acdaed9cc7e/openff/toolkit/typing/engines/smirnoff/forcefield.py#L1492) which changes how the matches are dealt with. Ideally, we could use some check that does not explicitly depend on the class to check how to deal with the matches. Maybe a try-except would do the job like

```python
try:
    for match in matches:
        parameter_matches[match] = matches[match].parameter_type
except: TypeError:
    for match in matches:
        parameter_matches[match] = [m.parameter_type for m in matches[match]]
```

**To Reproduce**

Run the following with qubekit master installed

```python
from openff.toolkit.typing.engines.smirnoff import ForceField
from openff.toolkit.topology import Molecule

methanol = Molecule.from_smiles("CO")
ff = ForceField("methanol.offxml", load_plugins=True)
ff.label_molecules(methanol.to_topology())
```
> File ~/Documents/Projects/openff/plugins/water_test/software/openff-toolkit/openff/toolkit/typing/engines/smirnoff/forcefield.py:1518, in ForceField.label_molecules(self, topology)
   1516     else:
   1517         for match in matches:
-> 1518             parameter_matches[match] = matches[match].parameter_type
   1520     current_molecule_labels[tag] = parameter_matches
   1522 molecule_labels.append(current_molecule_labels)

AttributeError: 'list' object has no attribute 'parameter_type'

Methanol offxml attached as [methanol.txt](https://github.com/openforcefield/openff-toolkit/files/9747326/methanol.txt)

**Output**

**Computing environment (please complete the following information):**
 - Operating system mac
 - Output of running `conda list`

**Additional context**

mattwthompson:
Which versions of the toolkit does this affect - just 0.10.6 vs earlier versions of 0.10.x?

PR #1470, opened 2022-11-22
VS handler fix for 0.10.6

bieniekmateusz:
This removes a bug which is the dependence on the order of the list. Ie when param_is_list is set to true, it stays true for each next handler. 

Is one way to addresses #1422 with minimal changes

### DIFF
--- a/openff/toolkit/typing/engines/smirnoff/forcefield.py
+++ b/openff/toolkit/typing/engines/smirnoff/forcefield.py
@@ -1489,9 +1489,6 @@ class ForceField:
             for tag, parameter_handler in self._parameter_handlers.items():
                 param_is_list = False
 
-                if type(parameter_handler) == VirtualSiteHandler:
-                    param_is_list = True
-
                 matches = parameter_handler.find_matches(top_mol)
                 # Remove the chemical environment matches from the
                 # matched results.
@@ -1505,6 +1502,8 @@ class ForceField:
 
                 # Now make parameter_matches into a dict mapping
                 # match objects to ParameterTypes
+
+                param_is_list = isinstance(parameter_handler, VirtualSiteHandler)
 
                 if param_is_list:
                     for match in matches:
--- a/devtools/conda-envs/openeye.yaml
+++ b/devtools/conda-envs/openeye.yaml
@@ -31,8 +31,8 @@ dependencies:
   - msgpack-python
   - xmltodict
   - qcelemental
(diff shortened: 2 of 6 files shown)


openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1507,6 +1503,8 @@ def label_molecules(self, topology):
                 # Now make parameter_matches into a dict mapping
                 # match objects to ParameterTypes
 
+                param_is_list = issubclass(type(parameter_handler), VirtualSiteHandler)
```
mattwthompson, line 1506: Perhaps
```suggestion
                param_is_list = isinstance(parameter_handler, VirtualSiteHandler)
```

mattwthompson:
Thanks for the fix - could you add a unit test(s) to ensure this fixes the linked issue?

mattwthompson:
Could you also modify these files to have `openff-interchange <0.2`? This is unrelated to your changes; the branch was previously failing

* `devtools/conda-envs/test_env.yaml`
* `devtools/conda-envs/rdkit.yaml`
* `devtools/conda-envs/openeye.yaml`

mattwthompson:
One more to get the environments updated: please add `qcportal >= 0.15` to the same files - thanks and sorry for our upstream being out of date

The `OpenEye=True` builds will all fail because the license is not accessible from a fork, so just ignore those runs

j-wags (commented):
This is a small enough change that we can go forward without a test if @bieniekmateusz doesn't have time right now. Then we could cut a 0.10.7 release with this change! I'll let @mattwthompson make the final call here, just adding my two cents.

bieniekmateusz:
Thanks, I have tested it locally as part of our pipeline and in a small script as suggested by @JoshHorton , and with a small snippet:
```
from qubekit.cli.utils import LocalCoordinateVirtualSiteHandler

methanol = Molecule.from_smiles("CO")
ff = ForceField("methanol.offxml", load_plugins=True)
ff.label_molecules(methanol.to_topology())
```
However, when I extract the class and use it directly for a test case, it returns a different structure/model, so I haven't had a look at that yet.

bieniekmateusz:
Thank you both and @jthorton, I added a naive test case that checks if the `param_is_list` is triggered.

mattwthompson:
Thanks! I double-checked to see that the test fails with 0.10.6.

PR #1471, opened 2022-11-28
[DNM] Update environments of 0.10 line

mattwthompson:


mattwthompson:
The test failures from 4b7cfc47ad91269c7b022570e64f81077d226e17 are from LAMMPS and GROMCS simulations crashing, which I have no interest in looking into further on this branch

Issue #1473, opened 2022-11-28
`ParameterHandler._VALENCE_TYPE` looks important but has no impact


mattwthompson:
The [docstrings](https://github.com/openforcefield/openff-toolkit/blob/1dd42519ef064dbc541785bfb37fbbf05108390b/openff/toolkit/typing/engines/smirnoff/parameters.py#L1637-L1642) make `ParameterHandler._VALENCE_TYPE` look important, but it's not clear that the behavior matches the comments.

I couldn't find [any discussion](https://github.com/search?q=repo%3Aopenforcefield%2Fopenff-toolkit+_VALENCE_TYPE&type=issues) around it. The only changes to the [lines it's used](https://github.com/openforcefield/openff-toolkit/blame/1dd42519ef064dbc541785bfb37fbbf05108390b/openff/toolkit/typing/engines/smirnoff/parameters.py) (in the base class and subclasses) have not undergone any functional changes in almost four years. This seems to date back to https://github.com/openforcefield/openff-toolkit/pull/173/files which is old enough that I don't grok all details. I [tried simply removing all lines](https://github.com/openforcefield/openff-toolkit/pull/1472) using it and the test suite passed locally.

These lines of code are copy-pasted into other projects; if it's not serving any purpose we should aim to remove it.

mattwthompson:
Removed in the upcoming 0.12.0 release, along with the deprecation of the associated module #1500

PR #1474, opened 2022-11-29
[WIP] Reinstate parameter smirks valence checks

j-wags:
- [ ] An alternative fix to #1473, which brings `ParameterType` back to it's former (/intended) glory
- [ ] Ensure that this doesn't increase FF loading time too much

### DIFF
--- a/openff/toolkit/tests/test_forcefield.py
+++ b/openff/toolkit/tests/test_forcefield.py
@@ -891,7 +891,9 @@ class TestForceField:
     @pytest.mark.parametrize("force_field_file", [*get_available_force_fields()])
     def test_get_available_force_fields_loadable(self, full_path, force_field_file):
         """Ensure get_available_force_fields returns load-able files"""
-        if "ff14sb" in force_field_file and "off_imp" not in force_field_file:
+        if ("ff14sb" in force_field_file and "off_imp" not in force_field_file) or (
+            "ff14sb_off_impropers_0.0.1.offxml" == force_field_file
+        ):
             pytest.skip(
                 "Only the variants of the ff14SB port with SMIRNOFF-style impropers "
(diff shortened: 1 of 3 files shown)


j-wags:
This has basically no effect with the OpenEye backend, but it makes the Toolkit run slower when using the RDKit backend.

Loading Sage using the RDKit backend used to take 0.25 seconds, but when doing valence+parseability checks it takes about 0.70 seconds. 

Loading ff14sb_off_impropers_0.0.3 used to take about 7 seconds, but when doing valence+parseability checks it takes 15. 

Ideally, we'd want to somehow make it so people doing like experimental FF work can run in "safe mode" and get these checks, while people loading Sage for the millionth time don't. 

Some solutions are:
* We could make the OFFXML loader skip SMIRKS validation, but have the direct eg. `BondChemicalEnvironment` init methods perform it. This would be a "surprising" behavior and doesn't seem like a good idea. 
* We could thread a new kwarg through the `ForceField` initializer that specifies whether SMIRKS validation should be skipped, and have it be `True` by default (this kwarg would need to be threaded through `ForceField.__init__`, `ForceField.parse_sources`, `ForceField._load_smirnoff_data`, and `ParameterHander.add_parameters`). But that seems like a lot of complexity for something we may not need. 
* We could just disable SMIRKS validation altogether because nobody seems to use it, and an unparseable SMIRKS will fail anyway when the user invokes `find_matches` (though a parseable-but-topologically-wrong SMIRKS could slip through...). This might let us remove `ToolkitWrapper.get_tagged_smarts_connectivity`, since it [doesn't seem to be used anywhere outside the Toolkit](https://github.com/search?q=get_tagged_smarts_connectivity&type=code). And maybe some other code too.

I'll poll around to see if anyone would object to this change. It would be an API break/feature removal, but for something which really nobody seems to use.

Issue #1493, opened 2022-12-15
Loading a ForceField from file substitutes units inappropriately


lilyminium:
**Describe the bug**

Loading a ForceField from file substitutes units inappropriately. For example, "F" seems to get converted into `1 * farad`, and K to `1 * kelvin`.

**To Reproduce**

The easiest way to see this is to load base Sage:

```python
>>> import openff.toolkit
>>> openff.toolkit.__version__
'0.11.4'
>>> from openff.toolkit import ForceField
>>> ff = ForceField("openff-2.0.0.offxml")
>>> charge_handler = ff.get_parameter_handler("LibraryCharges")
>>> charge_handler.parameters[2]
<LibraryChargeType with smirks: [#19+1:1]  charge1: 1.0 elementary_charge  name: 1 kelvin  >
```

**Output**

Checking https://github.com/openforcefield/openff-forcefields/blob/e2cb3b0746a173eb18d499bc680d9758ccaae657/openforcefields/offxml/openff-2.0.0.offxml#L363 shows the original name was just `K+`.

**Computing environment (please complete the following information):**
 - Operating system
 - Output of running `conda list`
 - toolkit version: 0.11.4

**Additional context**

The example given here isn't too bad. However, it's a tad inconvenient because it means users can't use element symbol smarts (`.to_smiles`) but have to use atomic numbers instead (have to pass through RDKit's `Chem.MolToSmarts`). It's also surprising, as I think [SMIRKS allows you to use symbols as element specifications](https://www.daylight.com/dayhtml/doc/theory/theory.smirks.html). However, parameters like the LibraryCharge below get interpreted as an unholy mishmash of coulomb and farad.

```
<LibraryCharge smirks="[C:1](=[C:2]([F:3])[F:4])([C:5]([F:6])([F:7])[F:8])[F:9]" charge1="-0.053769998252391815 * elementary_charge ** 1" charge2="0.3786599934101105 * elementary_charge ** 1" charge3="-0.12859000265598297 * elementary_charge ** 1" charge4="-0.12859000265598297 * elementary_charge ** 1" charge5="0.69309002161026 * elementary_charge ** 1" charge6="-0.21591000258922577 * elementary_charge ** 1" charge7="-0.21591000258922577 * elementary_charge ** 1" charge8="-0.21591000258922577 * elementary_charge ** 1" charge9="-0.11307000368833542 * elementary_charge ** 1"></LibraryCharge>
```

[...]

PR #1494, opened 2022-12-15
Add `ignore_keys` argument to `convert_all_strings_to_quantity`

mattwthompson:
This PR implements @lilyminium's suggested fix to #1493 and also adds a test to catch some of the behavior she observed, which was previously not covered in tests.

### DIFF
--- a/openff/toolkit/tests/test_forcefield.py
+++ b/openff/toolkit/tests/test_forcefield.py
@@ -998,6 +998,26 @@ class TestForceField:
         assert len(forcefield._parameter_handlers["ImproperTorsions"]._parameters) == 2
         assert len(forcefield._parameter_handlers["vdW"]._parameters) == 2
 
+    def test_load_do_not_convert_non_quantity_strings(self):
+        """Reproduce part of #1493"""
+        sage = ForceField("openff-2.0.0.offxml")
+
+        for parameter_handler_name in sage.registered_parameter_handlers:
+
+            parameter_handler = sage.get_parameter_handler(parameter_handler_name)
+
+            for parameter in parameter_handler.parameters:
+                assert isinstance(parameter.smirks, str)
+                assert not isinstance(parameter.smirks, unit.Quantity)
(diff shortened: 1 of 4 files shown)


openff/toolkit/tests/test_forcefield.py
```diff
@@ -998,6 +998,25 @@ def test_create_forcefield_from_xml_string(self):
...
+            parameter_handler = sage.get_parameter_handler(parameter_handler_name)
+
+            for parameter in parameter_handler.parameters:
+                assert isinstance(parameter.smirks, str)
+                assert not isinstance(parameter.smirks, unit.Quantity)
+
+                if (
+                    parameter_handler_name == "LibraryCharges"
+                    and getattr(parameter, "name") is not None
+                ):
```
j-wags, line 1016:
```suggestion
                # Ensure that, for example, F isn't converted to Farad 
                if (
                    parameter_handler_name == "LibraryCharges"
                    and getattr(parameter, "name") is not None
                ):
```

j-wags (approved):
Thanks for working with my extremely janky serialization code. Hopefully openff-models will enable us to make the migration to a cleaner, safer pydantic-based FF class.

mattwthompson:
Thanks! And it's _our_ (de)serialization code at this point 😁

I'd like @lilyminium to have a look, particularly if `ignore_keys=["smirks", "name"],` is the right way to call this

lilyminium:
Thanks @mattwthompson, that was a super fast fix! A more long-term solution might be to work with `ParameterAttribute.unit` somehow (i.e. not convert if there is no unit attached), but that looks like it'll take more work and this is an effective patch. LGTM!

With regards to additional keywords to ignore -- `id` might eventually become problematic if numbers aren't attached, but really `smirks` was the one breaking everything and iirc we don't guarantee anything about the `name` and `id` attributes.

mattwthompson:
Thanks!

Issue #1067, opened 2021-09-08
Make the failure case of ToolkitRegistry.call more useful


mattwthompson:
Right now, `ToolkitRegistry.call` tries the requested method in each of its wrapped toolkits and if they _all_ fail, it raises a `ValueError` that squishes together the messages in each exception:

https://github.com/openforcefield/openff-toolkit/blob/7a3fad5239bcdb940fec8b28a4d4d27235f0738e/openff/toolkit/utils/toolkit_registry.py#L376-L380

A small improvement would involve raising a custom exception instead of `ValueError`, using mostly the same message.

cc: @j-wags who can provide more thoughts on other solutions and possibly if it's worth refactoring the design in general.

Issue #1066, opened 2021-09-08
Disallow `ToolkitWrapper` objects passed to toolkit_registry


mattwthompson:
In #1050 @j-wags and I re-encountered strange behavior relating to the `toolkit_registry` argument in some `Molecule` API points. The behavior is different if a `ToolkitWrapper` and `ToolkitRegistry` argument is passed and the wrapper or all wrappers in the registry fail. I can expand on the failure cases later.

A solution would involve forbidding `ToolkitWrapper` objects passed to toolkit_registry; initially with a deprecation warning but later with an exception. The migration would involve simply passing a single-wrapper registry in place of a wrapper object, i.e.

```
Molecule.some_api_point(
    ...
    toolkit_registry=OpenEyeToolkitWrapper(),
    ...
)
```

will need to become
```
Molecule.some_api_point(
    ...
    toolkit_registry=ToolkitRegistry(OpenEyeToolkitWrapper()),
    ...
)
```

lilyminium:
Could you just do that under the hood -- check if a wrapper is passed in, and if so, make a single-wrapper registry?

mattwthompson:
Agreed, another solution would be to convert the wrappers for the user. This would be backwards-compatible at a high-level but would introduce some behavior changes, i.e. a specific exception that an individual wrapper raises upon the failure of one of its methods (like `ChargeAssignmentError` or something like that) would become the `ValueError` resulting from a _registry_ failing.

It's a tricky decision point since these methods are all accessible from the wrapper and registry objects themselves and these parts of the `Molecule` API are "just" convenience for the users - but super convenient, and also some of our most commonly-used API points!

mattwthompson:
This won't happen - but #1495 does what I think Lily suggested as a clean workaround.

PR #1495, opened 2022-12-19
Consolidate registry/wrapper argument validation

mattwthompson:
There's a lot of repeated code in the `Molecule` API that coaxes an argument that could be a `ToolkitWrapper` (or sometimes its class) into a `ToolkitRegistry`. The functionality here is identical across a few dozen uses except for some error messages that have subtly diverged over time. I took these all and wrapped them into a single function whose purpose should be clear from the signature:

```
def _ensure_toolkit_registry(
    toolkit_registry: Union[ToolkitRegistry, Type[ToolkitRegistry], ToolkitWrapper],
) -> ToolkitRegistry:
```

I don't think this breaks API. There might be some behavior changes resulting from a `ToolkitWrapper` being called instead of a `ToolkitRegistry` containing it, which arguably should have never happened.

This touches on some older discussions (https://github.com/openforcefield/openff-toolkit/issues/1066, #1067).

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -33,6 +33,7 @@ from copy import deepcopy
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     DefaultDict,
     Dict,
     Generator,
@@ -50,6 +51,7 @@ from openff.units import unit
 from openff.units.elements import MASSES, SYMBOLS
 from openff.utilities.exceptions import MissingOptionalDependencyError
 from packaging import version
+from typing_extensions import TypeAlias
 
 from openff.toolkit.utils.exceptions import (
     HierarchySchemeNotFoundException,
@@ -58,20 +60,21 @@ from openff.toolkit.utils.exceptions import (
     InvalidAtomMetadataError,
     InvalidBondOrderError,
     InvalidConformerError,
+    InvalidToolkitRegistryError,
     MultipleMoleculesInPDBError,
     SmilesParsingError,
+    UndefinedStereochemistryError,
     UnsupportedFileTypeError,
 )
 from openff.toolkit.utils.serialization import Serializable
+from openff.toolkit.utils.toolkit_registry import _ensure_toolkit_registry
 from openff.toolkit.utils.toolkits import (
     DEFAULT_AROMATICITY_MODEL,
     GLOBAL_TOOLKIT_REGISTRY,
-    InvalidToolkitRegistryError,
     OpenEyeToolkitWrapper,
     RDKitToolkitWrapper,
     ToolkitRegistry,
     ToolkitWrapper,
-    UndefinedStereochemistryError,
 )
 from openff.toolkit.utils.utils import get_data_file_path, requires_package
 
@@ -86,6 +89,9 @@ if TYPE_CHECKING:
 
 # TODO: Allow all OpenEye aromaticity models to be used with OpenEye names?
 #       Only support OEAroModel_MDL in RDKit version?
+
+
+TKR: TypeAlias = Union[ToolkitRegistry, ToolkitWrapper]
 
 
 def _molecule_deprecation(old_method, new_method):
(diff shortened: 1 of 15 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -88,6 +91,9 @@
 #       Only support OEAroModel_MDL in RDKit version?
 
 
+TKR: TypeAlias = Union[ToolkitRegistry, ToolkitWrapper]
```
mattwthompson, line 94: As I understand it, this is as simple as "hey, I'm re-using this type annotation a bunch of times, can I define it as variable-like thing so it's not so repetitive and verbose?"

https://peps.python.org/pep-0613/

openff/toolkit/topology/molecule.py
```diff
@@ -1661,23 +1640,12 @@ def to_inchi(self, fixed_hydrogens=False, toolkit_registry=GLOBAL_TOOLKIT_REGIST
...
-            inchi = toolkit.to_inchi(self, fixed_hydrogens=fixed_hydrogens)
-        else:
-            raise InvalidToolkitRegistryError(
-                "Invalid toolkit_registry passed to to_inchi. Expected ToolkitRegistry or ToolkitWrapper. "
-                f"Got {type(toolkit_registry)}"
-            )
+        toolkit_registry = _ensure_toolkit_registry(toolkit_registry)
 
-        return inchi
+        return toolkit_registry.call("to_inchi", self, fixed_hydrogens=fixed_hydrogens)
```
mattwthompson, line 1645: This pattern could be further golfed down to a one-liner


```
return _ensure_toolkit_registry(toolkit_registry)..call("method_name", ...)
```
but I think that's too far

j-wags: Haha yeah, it's already golfy enough :-P

j-wags (approved):
Thanks @mattwthompson - These PRs are like an advent calendar - I love getting negative lines of code in my stocking :-P 

This _technicalllly_ changes the exception type of some `Molecule` API calls by having all of these route through `ToolkitRegistry` instead of having some route through `ToolkitWrapper` (depending on the input type). For example, calling `RDKitToolkitWrapper.generate_conformers` could raise a `ConformerGenerationError`. But `ToolkitRegistry([RDKitToolkitWrapper()]).call('generate_conformers'...)` would raise [a "summary" `ValueError`](https://github.com/openforcefield/openff-toolkit/blob/efe91cd54b8bae0e5046b1366bb6b11a5447464d/openff/toolkit/utils/toolkit_registry.py#L359-L370) that describes the `ConformerGenerationError` raised by the toolkitwrappers. 

Anyway, I don't care that much, let's just be a responsible upstream by mentioning this in the releasehistory and make the next release be a minor version bump.

mattwthompson:
Do you care much about `FrozenMolecule.from_file`? I should have mentioned this before; that's the only method including this logic that I didn't touch. The special cases between file formats and wrapped toolkits looked pretty fragile to me.

j-wags (changes requested):
Oh, I'd forgotten about all that. Let me think a bit more and test some stuff out.

j-wags (commented):
Ok, so I think that anywhere we used to specify `raise_exception_types=[]` (namely `Molecule.assign_partial_charges` and `generate_conformers`) should be reverted to their original form, and also `Molecule.from_file` should remain in its original form. Those methods have specific behaviors when using toolkits/throwing errors that I think are useful enough to keep around. But all the other changes should be fine to stay.

mattwthompson:
Which are the existing behaviors you'd like to preserve?

mattwthompson:
I've run slow tests locally and everything is passing there as well. (With my luck something will fail in CI, of course.)

j-wags (commented):
Two hours ago I came by to give this a quick approval but I saw that some good questions were added about the architecture of the Molecule API. I started typing up a big response but after a while I realized that architecture discussion would be lost and architecture change is out of scope of this PR, which is just intended to simplify some reused code. But this PR doesn't really accomplish that because it's changing more than 300 lines of in order to remove maybe 40 lines of source, so I don't think it really succeeds there. It also adds type annotations, which I like, but I don't like the changes to get_tagged_smarts_connectivity, which seem off topic. I'm gonna close this PR because I think the cost/benefit here has gotten a bit too high, but I'd be happy to consider these changes as separate PRs in the future.

mattwthompson:
Nothing changed with `get_tagged_smarts_connectivity` but I can revert those if you wish?

PR #1497, opened 2022-12-20
Migrate `python setup.py` to `pip install`

mattwthompson:
`setup.py` is dead, long live `setup.py`!

https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

This should not affect deployment since `python setup.py` calls only happen in our automation. There was a single mention in docs, which I've updated. I couldn't find anything else in the public sections.

* `--no-deps` should not be needed since we don't (and can't and should not) define our requirements via `install_requires` in `setup.py`
* `--develop` is unneeded since we're not patching the toolkit inside of workflows
* `pip` can take a path, which is handy for installing the test plugins

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -46,7 +46,8 @@ jobs:
     env:
       OE_LICENSE: ${{ github.workspace }}/oe_license.txt
       PACKAGE: openff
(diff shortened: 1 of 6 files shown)


j-wags (approved):
Tested locally. Seems fine to me. Thanks @mattwthompson!

Issue #699, opened 2020-08-31
Make ff.label_molecules and ParameterHandler.find_matches accept EITHER Topology or FrozenMolecule


j-wags:
**Is your feature request related to a problem? Please describe.**
Currently, both of these methods require users to provide a Topology, even if they only have a single Molecule (as is frequently the case). Further, in VirtualSiteHandler, the `find_matches` method would be most efficient if allowed to operate on `FrozenMolecule`s, but instead we have to cast each `FrozenMolecule` to a `Topology` to fit the `find_matches` method's input requirement. 

**Describe the solution you'd like**
A quick solution would be, if a FrozenMolecule is provided to either function, immediately cast it into a `Topology` (using, for example, `Molecule.to_topology`).

**Describe alternatives you've considered**
To solve the `VirtualSiteHandler` problem, we could also refactor `FrozenMolecule.chemical_environment_matches` and `Topology.chemical_environment_matches` to return the same type (currently the former returns a list of tuples of integer, whereas the latter returns a list of `ChemicalEnvironmentMatch`, which contains both reference molecule atom indices and topology molecule atom indices). However the return value of `Topology.chemical_environment_matches` is used heavily internally, and would require a significant refactor if changes.

mattwthompson:
> (currently the former returns a list of tuples of integer, whereas the latter returns a list of ChemicalEnvironmentMatch, which contains both reference molecule atom indices and topology molecule atom indices).

Do you have any insight into which of these should be preferred?

mattwthompson:
This would be useful for parametrizing a topology on a "per-molecule" basis, since casting each molecule individually into a `Topology` is a wasteful transformation.

mattwthompson:
I'm finding that 20% of the time spent calling `ForceField.label_molecules` on a large protein is spent doing conversions between `Topology` and `Molecule`.

PR #1507, opened 2023-01-09
Speed up `Molecule` roundtrips

mattwthompson:
This is the first in a sequence of patches to improve performance when dealing with proteins or other large molecules/(bio-)macromolecules.

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -1180,12 +1180,12 @@ class FrozenMolecule(Serializable):
         self._initialize()
         self.name = molecule_dict["name"]
         for atom_dict in molecule_dict["atoms"]:
-            self._add_atom(**atom_dict)
+            self._add_atom(**atom_dict, invalidate_cache=False)
 
         for bond_dict in molecule_dict["bonds"]:
             bond_dict["atom1"] = int(bond_dict["atom1"])
             bond_dict["atom2"] = int(bond_dict["atom2"])
-            self._add_bond(**bond_dict)
+            self._add_bond(**bond_dict, invalidate_cache=False)
 
         if molecule_dict["partial_charges"] is None:
             self._partial_charges = None
@@ -1229,6 +1229,8 @@ class FrozenMolecule(Serializable):
                 new_hier_scheme.add_hierarchy_element(
                     tuple(element_dict["identifier"]), element_dict["atom_indices"]
                 )
+
+        self._invalidate_cached_properties()
 
     def __repr__(self):
         """Return a summary of this molecule; SMILES if valid, Hill formula if not."""
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -1230,8 +1230,6 @@ class FrozenMolecule(Serializable):
                     tuple(element_dict["identifier"]), element_dict["atom_indices"]
                 )
 
-        self._invalidate_cached_properties()
-
     def __repr__(self):
         """Return a summary of this molecule; SMILES if valid, Hill formula if not."""
         description = f"Molecule with name '{self.name}'"
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -242,10 +242,7 @@ class Atom(Particle):
         self._atomic_number = atomic_number
 
         # Use the setter here, since it will handle either ints or Quantities
-        if hasattr(formal_charge, "units"):
-            # Faster check than ` == unit.dimensionless`
-            if str(formal_charge.units) == "":
-                raise Exception
+        # and it is designed to quickly process ints
         self.formal_charge = formal_charge
         self._is_aromatic = is_aromatic
         self._stereochemistry = stereochemistry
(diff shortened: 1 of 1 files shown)


mattwthompson:
I'm finding significant speedups in the re-creation of `Topology` and `Molecule` objects. This might not seem flashy or the bottleneck, but these operations crop up in numbers. Interchange immediately executes a deepcopy on the `Topology` it's passed, and much of the toolkit relies on the fast interconversion between the two objects (i.e. #699)

mattwthompson:
```python3
In [1]: from openff.toolkit import Topology, Molecule

In [2]: t4 = Molecule.from_polymer_pdb("openff/toolkit/data/proteins/T4-protein.pdb").to_topology()

In [3]: %%timeit
   ...: Topology(t4)
   ...:
   ...:

# 91d03ae
187 ms ± 31.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# 0.12.0
1.26 s ± 44.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

mattwthompson:
It'll be hard to squeeze out performance in force field loading since most of it it spent waiting for Pint to parse strings into `Quantity`s. Fortunately that issue should not scale with the complexity of the proteins being studied.

mattwthompson:
I think this one's ready to go; the rest of this comment is a brief analysis. I used RDKit to generate proteins of different sizes from a naive sequence.

```python3
import string
import time

from rdkit import Chem

from openff.toolkit import Molecule, Topology

base_sequence = ""

for char in string.ascii_uppercase:
    try:
        Molecule.from_rdkit(Chem.MolFromSequence(char))
        base_sequence += char
    except:
        pass


for n in range(1, 51, 5):
    original_topology = Molecule.from_rdkit(
        Chem.MolFromSequence(n * base_sequence)
    ).to_topology()
    start = time.time()
    Topology(original_topology)
    diff = round(time.time() - start, 3)
    print(
        f"{n=}, {n * len(base_sequence)} residues, {original_topology.n_atoms} atoms, took {diff}"
    )
```

Running them and visualizing with a quick-and-dirty plot

```console
$git checkout upstream/main && python bench.py && git checkout protein-perf && python bench.py
HEAD is now at 5e4928b0 Fix "slow" gasteiger test (#1505)
n=1, 19 residues, 304 atoms, took 0.03
n=6, 114 residues, 1809 atoms, took 0.6
n=11, 209 residues, 3314 atoms, took 1.673
n=16, 304 residues, 4819 atoms, took 3.61
n=21, 399 residues, 6324 atoms, took 6.856
n=26, 494 residues, 7829 atoms, took 10.066
n=31, 589 residues, 9334 atoms, took 13.302
n=36, 684 residues, 10839 atoms, took 21.266
n=41, 779 residues, 12344 atoms, took 23.366
n=46, 874 residues, 13849 atoms, took 33.658
Previous HEAD position was 5e4928b0 Fix "slow" gasteiger test (#1505)
Switched to branch 'protein-perf'
n=1, 19 residues, 304 atoms, took 0.015
n=6, 114 residues, 1809 atoms, took 0.15
n=11, 209 residues, 3314 atoms, took 0.318
n=16, 304 residues, 4819 atoms, took 0.575
n=21, 399 residues, 6324 atoms, took 0.89
n=26, 494 residues, 7829 atoms, took 1.233
n=31, 589 residues, 9334 atoms, took 1.633
n=36, 684 residues, 10839 atoms, took 2.056
n=41, 779 residues, 12344 atoms, took 2.616
n=46, 874 residues, 13849 atoms, took 3.17
```

Overall, I think this is pretty good and not worth optimizing further. Doing anything interesting with a protein of a 1000 residues will surely be slow because of other steps in the pipeline.

image

This looks slightly superlinear, which is in line with our expectations. Plotting it on semi-log axes indicates this as well:

image

And plotting instead the time _per residue_ shows this more clearly:

image


The speedup probably asymptotes to a constant value with larger systems, but I didn't take the time to evaluate it. Most of the execution time is spend in or near RDKit and I'm not interesting in optimizing either here.

A main limitation of this benchmark is that the molecules had no conformers or partial charges - neither of which are trivial to generate with this approach, but each do contribute significantly to serialization performance as they're unit-bearing attributes. The other detail is that the scope is just `Topology(topology)`, which is important but only one step in the pipeline.

openff/toolkit/topology/molecule.py
```diff
@@ -5037,15 +5019,14 @@ def add_atom(
         'C'
 
         """
-        atom_index = self._add_atom(
+        return self._add_atom(
```
j-wags, line 5022: I think the original code here was a lot more readable - Could you revert it?

openff/toolkit/topology/molecule.py
```diff
@@ -5084,15 +5065,14 @@ def add_bond(
         --------
         For an example of use, see :py:meth:`add_atom`.
         """
-        bond_index = self._add_bond(
+        return self._add_bond(
```
j-wags, line 5068: I think the original code here was a lot more readable - Could you revert it?

j-wags (approved):
LGTM! This is a great speedup. Thanks!

Issue #1500, opened 2023-01-04
Remove `openff.toolkit.typing.engines.chemistry` module


mattwthompson:
#1472 deprecates `ChemicalEnvironment`, and by extension the module that contains it as its only public API member.

We should aim to remove this module altogether roughly 6 months from now, be it version 0.13, 0.14, or a different number.

PR #1472, opened 2022-11-28
Remove SMIRKS validation checks in `ForceField` construction

mattwthompson:
#1473
- [x] Remove `ParameterHandler._VALENCE_TYPE` and equivalents in subclasses
- [x] Deprecates `ChemicalEnvironment`
- [x] Updates `Molecule.strip_atom_stereochemistry` to not use `AtomChemicalEnvironment`
- [x] Updates `Topology.chemical_environment_matches` to not accept a `ChemicalEnvironment`
- [x] Re-routes some exceptions that were imported from aliases
- [x] Tag issue being addressed
- [x] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/main/openff/toolkit/tests)
- [x] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/main/docs), if applicable
- [x] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [ ] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/main/docs/releasehistory.rst)

PR #1509, opened 2023-01-17
Remove `openff.toolkit.typing.chemistry`

mattwthompson:
#1500, I looked through #1472 to find the changes that need to actually be removed. Grepping around for `typing.chemistry` or ` ChemicalEnvironment` don't trigger anything non-historical as far as I can tell.


j-wags:
This would [break ForceBalance](https://github.com/leeping/forcebalance/blob/10b9280cab9735634d25a4f9c8b3e84cb13a6bb1/src/smirnoff_hack.py#L76). It should be possible to remove the validate cache from ForceBalance entirely since it is no longer used and then move forward with this. 

@jthorton @cringrose94 @bieniekmateusz  This will [affect and probably break QUBEKit](https://github.com/qubekit/QUBEKit/blob/01165ef1ef14f2f74847c54aa4219e538b9ce68a/qubekit/torsions/utils.py#L27). It looks like you missed the deprecation warning because QUBEKit never instantiates a `ChemicalEnvironment` object but rather somehow calls `validate` as a classmethod. I'm thinking you could either remove the line and skip validation or copy `environment.py` to QUBEKit.

jthorton:
Thanks for the warning @j-wags, Ill look into changing this!

j-wags (approved):
This one is going to require a bit of pinning in our ecosystem for a while, but it's really important to me that we keep all of the deprecation followthroughs in the 0.14.0 release, so merging this now is the lesser of two evils.

j-wags:
After thinking some more, in the interest of not teaching the new science team hires to downpin their envs on their first day, I'm gonna revert this so we can get the other API deprecations out. We include these changes in an off-schedule API breaking release once ForceBalance is ready for this.

PR #1510, opened 2023-01-17
Add more type annotations

mattwthompson:
Some of my recent PRs have been made difficult to review by my desire to improve the annotation coverage of the codebase. It might be more tractable to separate these changes out from anything that might alter behavior.

This is sizable but virtually all changes are _only_ additions of type annotations. Many are easy ones to add. As a reminder, annotations are not evaluated at runtime (unless you try really hard) so they are functionally closer to comments than source code.

I can offer no guidance on how to review this, or even if it should be manually reviewed in any way.


j-wags (approved):
LGTM!

PR #1511, opened 2023-01-17
Constrain `ipywidgets`

mattwthompson:
At some point over the weekend, [nightly CI](https://github.com/openforcefield/openff-toolkit/actions/workflows/CI.yml?query=event%3Aschedule) started failing at the step of running through docs in the vanilla test environment, which was not constrained in a way that resolves the `ipywidgets`/NGLview situation.

I'm not immediately sure what caused it to go from regularly passing to failing, but here's the closest [before](https://github.com/openforcefield/openff-toolkit/actions/runs/3915866280) and [after](https://github.com/openforcefield/openff-toolkit/actions/runs/3920879965) I could find. So I'm not sure if this needs to be fixed in deployment as well.

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -37,6 +37,7 @@ dependencies:
   - qcengine
   - mdtraj
   - parmed
+  - ipywidgets =7
   - nbval
   - mypy
   - pip:


mattwthompson:
I've looked through the logs and can't find an obvious culprit.

* The versions of `libmamba` and `micromamba` are identical (1.1) in each case
* The same version of `nglview` is pulled down each time
  * The same _build_ of (nglview-3.0.3-pyh8a188c0_0) is pulled down each time
* Different versions of `ipywidgets` (7.x and 8.x) are given, causing the well-documented issue
* Each package is installed when the environment is created and not modified in later steps

Deployment tests are passing (we have `ipywidgets =7` in our conda builds) and no users are reporting issues, so I am moving forward with this minimal fix.

j-wags:
Thanks for the quick action, @mattwthompson! 

Just manually tested for the same problem in production and didn't have any issues.

Issue #697, opened 2020-08-29
Safer and more consistent aromaticity handling


j-wags:
(Continuing from #206)

**Describe the problem**

Problem 1:
Our current data structures and functions handle aromaticity in an unsafe way. There are three places where we store aromaticity information, but don't rigorously enforce that they are consistent (largely my fault [1]) or have mechanisms to flag/resolve inconsistencies. These three places are:

* In a Molecule's `Atom` or `Bond`'s `is_aromatic` attribute. 
* In a Topology's `aromaticity_model` attribute
* In a ForceField's `aromaticity_model` attribute

Problem 2:

A `Molecule` currently does not know its own aromaticity model, just whether its atoms and bonds ARE aromatic. Currently we have different degrees of enforcement, usually assuming the molecule conforms to `DEFAULT_AROMATICITY_MODEL`, but not all Molecule-creation pathways enforce this, and the public API allows invalidating this assumption. This means that the information in the `Atom`/`Bond.is_aromatic` attributes is missing essential context and can't safely be used for most purposes.

Problem 3: 

Currently, molecule equality comparisons don't check that all molecules involved have equivalent aromaticity models. It's possible that two molecules _are_ identical with one particular aromaticity model, but _not_ with another, since the aromaticity model is the bridge between a single molecule's multiple possible kekule structures. These molecule equality comparisons are core to our functionality, since they are how `Topology.add_molecule` knows whether a newly-added molecule is unique or not. 

Problem 4:
It's taken me almost two years to understand the subtlety of the aromaticity issue, and where in the codebase we are or aren't making assumptions about the use of identical aromaticity models. This will spell organizational trouble as the volume of contributors and PRs increases, so I think we need at least approachable documentation of this, and at best a change to our object model that doesn't allow implementation of dangerous behavior. 

So far this hasn't been an issue because we currently only support `OEAroModel_MDL` and its RDKit equivalent. But users constructing molecules atom-by-atom using the API [2] may have been unknowingly circumventing our "safe pathways" and getting weird results. 

@davidlmobley points out that aspects of this problem that affect _parameterization_ will be resolved by WBO interpolation of parameters. This is good news, but resolving this still won't help the problems with our `Molecule` and `Topology` classes.

**Solution specification**

Important requirements:
* The `Molecule` class either 
    * MAY store `atom/bond.is_aromatic` flags, but if so, it MUST know its own aromaticity model, or 
[...]

Issue #453, opened 2019-11-08
Topology.chemical_environment_matches has unused aromaticity_model arg.


SimonBoothroyd:
**Describe the bug**
The `Topology.chemical_environment_matches` method currently takes an `aromaticity_model` argument which doesn't seem to be used. 

Should this be implemented, or if not it would be good to remove it to avoid confusion.

PR #1513, opened 2023-01-17
Clarify aromaticity model support

mattwthompson:
Our infrastructure only supports one aromaticity model, but our documentation is not always clear about this and the internal logic around checking it is not unified.
 - [x] Make docstrings consistent and clear that `OEAroModel_MDL` is the only supported mode
 - [x] Raise `InvalidAromaticityModelError` where currently some `ValueError`s are raised
 - [x] Leave in place the distinction bewteen `SMIRNOFFAromaticityError` and `InvalidAromaticityModelError`
 - [x] Make internal checks somewhat amenable to expanding the list of supported models in the future (however unlikely)
 - [x] Update some import paths to be more direct

This does not break the API, it extends it very slightly. `InvalidAromaticityModelError` already inherits from `ValueError` so no code should be broken by this. This doesn't close any issues but it touches on #697 #771 #664 #453 (the last of which could be fixed here)

### DIFF
--- a/openff/toolkit/utils/rdkit_wrapper.py
+++ b/openff/toolkit/utils/rdkit_wrapper.py
@@ -1905,8 +1905,8 @@ class RDKitToolkitWrapper(base_wrapper.ToolkitWrapper):
 
         if aromaticity_model not in ALLOWED_AROMATICITY_MODELS:
             raise InvalidAromaticityModelError(
-                f"Given aromaticity model {aromaticity_model} which is not in the set of allowed aromaticity models:  "
-                f"{ALLOWED_AROMATICITY_MODELS}."
+                f"Given aromaticity model {aromaticity_model} which is not in the set of allowed aromaticity models: "
+                f"{ALLOWED_AROMATICITY_MODELS}"
             )
 
         # Create an editable RDKit molecule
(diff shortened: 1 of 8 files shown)


openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -1954,11 +1964,12 @@ def _connection_table_to_rdkit(
             Chem.SANITIZE_ALL ^ Chem.SANITIZE_ADJUSTHS ^ Chem.SANITIZE_SETAROMATICITY,
         )
 
-        # Fix for aromaticity being lost
         if aromaticity_model == "OEAroModel_MDL":
             Chem.SetAromaticity(rdmol, Chem.AromaticityModel.AROMATICITY_MDL)
         else:
-            raise ValueError(f"Aromaticity model {aromaticity_model} not recognized")
+            raise NotImplementedError(
+                "The aromaticity model is assumed to be OEAroModel_MDL"
```
j-wags, line 1971:
```suggestion
                "The only supported aromaticity model is OEAroModel_MDL"
```

j-wags (approved):
LGTM! One suggested change and please update the releasehistory before merging.

PR #1524, opened 2023-01-27
Remove unused test files

mattwthompson:
- [ ] Remove un-used Amber files
- [ ] Reduce the number of calls to test-only APIs


j-wags (approved):
LGTM!

PR #1526, opened 2023-01-30
Update pre-commit hooks

mattwthompson:
I'm manually updating hooks ahead of schedule because over the weekend, Poetry introduced a breaking change that broke `isort` hooks in a way that cannot simply be ignored by sticking to a previously-functional version.

https://github.com/PyCQA/isort/issues/2077


mattwthompson:
Merging ahead of completed CI checks given that the `pre-commit.ci` check passed and the new versions of the hooks do not modify any source code.

PR #1531, opened 2023-01-31
Test `openmmforcefields` example

mattwthompson:
Currently, we recommend use of `openmmforcefields` for some use cases but do not test that it works with the toolkit. We should change one of those things; here I am turning on the test.


mattwthompson:
From the logs, it looks like AmberTools is clobbering a standalone ParmEd installation and when AmberTools is removed, so is ParmEd ...

j-wags (approved):
LGTM - Thanks for catching this! If we can't figure out how to fix things to work with ParmEd 4 in a month let's consider adding a `run_constrained` to the package recipe until we do.

PR #1534, opened 2023-02-01
Turn on macOS + Python 3.10 tests

mattwthompson:
OpenMM 8 now includes a build for this combination of inputs.

https://anaconda.org/conda-forge/openmm/files


j-wags:
The test failures here may be resolved by merging #1531. I'll review this once that's merged and this branch is updated so tests go green.

j-wags (approved):
Ignore the previous statement about waiting for #1531 - This is approved pending tests passing (which we're pretty sure 1531 will cause)

PR #1535, opened 2023-02-01
Reformat with new Black

mattwthompson:
Black released a new version - which introduces significant changes - [last night](https://pypi.org/project/black/#history). I like the empty line after function definitions, but I guess that's a thing of the past.

```console
$ pre-commit autoupdate && pre-commit run --all-files  && git add .
```

To be crystal clear, the issue stems from the linting workflow duplicating a configuration and ; since it blindly pulls down the most recent version of its tools (`pip install black isort`) development is immediately hindered by changes, whereas with `pre-commit` we are able to adopt a [slower pace](https://github.com/openforcefield/openff-toolkit/blob/5a7bafb85d37cc23c0cbc553737b25896f6722ef/.pre-commit-config.yaml#L1-L2) of these changes. At least with `pre-commit`, the configuration and source code can be updated with a one-liner.


j-wags (approved):
LGTM!

PR #1536, opened 2023-02-01
Test on Python 3.11

mattwthompson:
The new OpenMM builds include Python 3.11 (released October 2022) so let's see what else is missing.

https://anaconda.org/conda-forge/openmm/files

Related #1534


mattwthompson:
Nah

```
        - package openmmforcefields-0.11.2-pyhd8ed1ab_0 requires ambertools >=20.0, but none of the providers can be installed
```
```
        - package openeye-toolkits-2022.2.2-py310_0 requires python >=3.10,<3.11.0a0, but none of the providers can be installed
```

PR #1537, opened 2023-02-01
Allow setting charges per-atom

### LINKED ISSUE #466: Molecule.atoms[index].partial_charge does not support setting
Currently, setting partial charges on an atom via the obvious route does not work:
```python
>>> molecule.atoms[0].partial_charge *= 2
AttributeError: can't set attribute
```
Instead, the only way to modify the partial charges in a `Molecule` object is via the less obvious
```python
>>> molecule.partial_charges[0] = 1.0 * unit.elementary_charge
```

It would be useful for the obvious route to work as well.

mattwthompson:
Closes #466

Allows setting `Atom.partial_charge` is it is part of a molecule, if that molecule already has partial charges, and if the passed value is int, float, or unit-wrapped int or float.


openff/toolkit/tests/test_molecule.py
```diff
@@ -370,6 +371,53 @@ def test_set_molecule_error(self):
         with pytest.raises(AssertionError, match="already has an associated molecule"):
             atom.molecule = mol
 
+    @pytest.fixture()
+    def water_without_charges(self):
+        return Molecule.from_smiles("O")
```
j-wags, line 376: (not blocking) Since molecules in test code tend to get reused a lot, it may be cleaner to use from_mapped_smiles here so we know what to expect at each atom index.

openff/toolkit/topology/molecule.py
```diff
@@ -362,6 +363,35 @@ def partial_charge(self):
...
+    @partial_charge.setter
+    def partial_charge(self, charge):
+        if self.molecule.partial_charges is None:
+            raise MissingPartialChargesError(
+                "Cannot set individual atom's partial charge if it is in a molecule with no partial charges. "
+                "Instead, use the `Molecule.partial_charges` setter. If this behavior is important to you, "
+                "please raise an issue describing your use case."
+            )
+
+        if not isinstance(charge, (unit.Quantity, int, float)):
```
j-wags, line 375: (blocking) We should disallow `int` here to avoid users confusing this for `Atom.formal_charge`

j-wags (approved):
Thanks @mattwthompson! I would strongly prefer that we not accept ints for the setter, otherwise this looks great. Thanks for carefully testing for the edge cases if `Molecule.partial_charges is None`.

PR #1539, opened 2023-02-02
Use `flake8-bugbear` to safeguard against default mutable args

### LINKED ISSUE #780: Globally instantiated mutable args
There are several places in the toolkit such as [here](https://github.com/openforcefield/openforcefield/blob/1eb4e3feeff07d8a3281e3d6889262326b15259f/openforcefield/utils/toolkits.py#L4900) (found by @swails) with mutable args instantiated in the function definition. In these cases the argument is global and appending/extend to this value will expand the mutable value in all instantiations. It looks like there are `if arg is None` statements below which would be correct if it was instantiated as None.

mattwthompson:
Closes #780 and prevents this code smell from being introduced in the future (as I did in #1494).

### DIFF
--- a/openff/toolkit/utils/utils.py
+++ b/openff/toolkit/utils/utils.py
@@ -26,7 +26,7 @@ __all__ = [
 import contextlib
 import functools
 import logging
-from typing import Dict, List, Tuple, Union
+from typing import Dict, Iterable, List, Tuple, Union
 
 import numpy as np
 import pint
@@ -209,7 +209,7 @@ def string_to_quantity(quantity_string) -> Union[str, int, float, unit.Quantity]
 
 def convert_all_strings_to_quantity(
     smirnoff_data: Dict,
-    ignore_keys: List[str] = list(),
+    ignore_keys: Iterable[str] = tuple(),
 ):
     """
     Traverses a SMIRNOFF data structure, attempting to convert all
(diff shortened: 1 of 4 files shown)


openff/toolkit/utils/utils.py
```diff
@@ -209,7 +209,7 @@ def string_to_quantity(quantity_string) -> Union[str, int, float, unit.Quantity]
 
 def convert_all_strings_to_quantity(
     smirnoff_data: Dict,
-    ignore_keys: List[str] = list(),
+    ignore_keys: Iterable[str] = tuple(),
```
mattwthompson, line 212: This is a simpler fix than letting it be `Optional[List[str]]` and adding a layer of logic around it possibly being `None` vs. `list()`.

setup.cfg
```diff
@@ -17,7 +17,7 @@ exclude_lines =
 [flake8]
 # Flake8, PyFlakes, etc
 max-line-length = 119
-ignore = E203,W605,W503
+ignore = E203,W605,W503,B007,B009,B011,B014,B017
```
mattwthompson, line 20: These are other errors brought in by `flake8-bugbear` - most of which are quick improvements that I will introduce in subsequent changes, but did not want to bring into this PR.

mattwthompson:
Docs build failure is unrelated to these changes, I think

https://github.com/openforcefield/openff-sphinx-theme/issues/76

j-wags (approved):

PR #1540, opened 2023-02-03
More `flake8-bugbear`: fix B007, B009, B011

mattwthompson:
A small follow-on to #1539; nothing struck me as significant here

### DIFF
--- a/openff/toolkit/tests/test_links.py
+++ b/openff/toolkit/tests/test_links.py
@@ -44,7 +44,7 @@ def test_readme_links(readme_link):
     # Try to connect 5 times, keeping track of exceptions so useful feedback can be provided.
     success = False
     exception = None
-    for retry in range(5):
+    for retry in range(5):  # noqa: B007
         try:
             urlopen(request)
             success = True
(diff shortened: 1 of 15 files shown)


openff/toolkit/tests/test_links.py
```diff
@@ -44,7 +44,7 @@ def test_readme_links(readme_link):
     # Try to connect 5 times, keeping track of exceptions so useful feedback can be provided.
     success = False
     exception = None
-    for retry in range(5):
+    for retry in range(5):  # noqa: B007
```
mattwthompson, line 47: In this case I think it's more readable to keep the loop variable in; if I was reading `for _ in range(5):` I'd have to do a small amount of work to figure out that it's just doing 5 retries.

I don't feel strongly about this and am fine properly "fixing" it

j-wags: Totally agree with this reasoning and I'm in favor of keeping it readable. Good call.

openff/toolkit/topology/molecule.py
```diff
@@ -5450,7 +5450,7 @@ def perceive_residues(self, substructure_file_path=None, strict_chirality=True):
             substructure_dictionary_no_chirality = deepcopy(substructure_dictionary)
             # Update inner key (SMARTS) maintaining its value
             for res_name, inner_dict in substructure_dictionary.items():
-                for smarts, atom_types in inner_dict.items():
+                for smarts in inner_dict:
```
j-wags, line 5453: (not blocking) Slight preference here, this improves readability to my eyes.

```suggestion
                for smarts in inner_dict.keys():
```

j-wags (approved):
Looks great. One non-blocking suggestion, otherwise this is good to go. Thanks @mattwthompson!

PR #1541, opened 2023-02-03
More `flake8-bugbear`: fix B014, B017

mattwthompson:
Another  follow-on to https://github.com/openforcefield/openff-toolkit/pull/1539; in this case `pytest.raises(Exception)` was masking some (in my opinion) subpar error handling, so I implemented some more discrete cases.

A couple of the other smaller changes would have been caught by https://github.com/asottile/pyupgrade, a tool I've used for a long time and think does a good job of keeping old (i.e. Python 2 and <3.8) constructs out.  It's easy to add - don't need to do that now.

### DIFF
--- a/openff/toolkit/utils/openeye_wrapper.py
+++ b/openff/toolkit/utils/openeye_wrapper.py
@@ -173,7 +173,7 @@ class OpenEyeToolkitWrapper(base_wrapper.ToolkitWrapper):
         for tool, license_func in cls._license_functions.items():
             try:
                 module = importlib.import_module("openeye." + tool)
-            except (ImportError, ModuleNotFoundError):
+            except ImportError:
                 continue
             else:
                 if getattr(module, license_func)():
@@ -203,7 +203,7 @@ class OpenEyeToolkitWrapper(base_wrapper.ToolkitWrapper):
                     cls._is_installed = True
                     try:
                         importlib.import_module("openeye." + tool)
-                    except (ImportError, ModuleNotFoundError):
+                    except ImportError:
                         cls._is_installed = False
             if cls._is_installed:
                 if cls._is_licensed:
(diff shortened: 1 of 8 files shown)


openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -203,7 +203,7 @@ def is_available(cls) -> bool:
                     cls._is_installed = True
                     try:
                         importlib.import_module("openeye." + tool)
-                    except (ImportError, ModuleNotFoundError):
+                    except ImportError:
```
j-wags, line 206: I can't manage to get `importlib.import_module` to raise an `ImportError` - All of the following give me `ModuleNotFoundError`:

* `module = importlib.import_module("openeye.oechem")` raises no error, even when missing an OE license
* `module = importlib.import_module("openeye.zzz")` gives a `ModuleNotFoundError`
* `module = importlib.import_module("zzzz")` gives a `ModuleNotFoundError`

But I'm also not confident that this won't raise an `ImportError` - There's a lot of voodoo around python imports and OE Toolkit performance that I don't have a lot of confidence about. 

I'd propose that we revert these ones to `except (ImportError, ModuleNotFoundError):` and live with the bugbear we know.

mattwthompson: `ModuleNotFoundError` is a subclass of `ImportError`

```python3
>>> issubclass(ModuleNotFoundError, ImportError)
True
```

Installing OpenEye Toolkits doesn't require an active license, nor does importing `openeye.oechem`; if you uninstall it that will raise a `ModuleNotFoundError`

j-wags: Ahhh, thanks for the explanation!

j-wags (approved):
LGTM! Thanks for the cleanups!

PR #1543, opened 2023-02-06
Fix #1542

### LINKED ISSUE #1542: Plugins not reliably tracked if loaded externally
**Describe the bug**

The logic I introduced in #1467 only works if its call to `load_handler_plugins` is the first one.

**To Reproduce**

The simple behavior, which works fine, is

```python3
>>> from openff.toolkit import ForceField
>>> ForceField(load_plugins=True)._plugin_parameter_handler_classes
[<class 'custom_plugins.handler_plugins.CustomHandler'>]
```

If anything is done out of order, however, custom handlers are not tracked. This includes, for example, if the plugins are loaded before creating the `ForceField`:

```python3
>>> from openff.toolkit import ForceField
>>> from openff.toolkit.typing.engines.smirnoff.plugins import load_handler_plugins
>>> from openff.toolkit.typing.engines.smirnoff import ParameterHandler
>>> load_handler_plugins()
[<class 'custom_plugins.handler_plugins.CustomHandler'>]
>>> ForceField(load_plugins=True)._plugin_parameter_handler_classes
[]
```
This might sound like an esoteric case but downstream developers might want to know which plugins are available prior to creating a force field and this is unexpected behavior.

I think this is due to an interaction with `all_subclasses`:
```python3
>>> from openff.toolkit.typing.engines.smirnoff import ParameterHandler
>>> from openff.toolkit.utils.utils import all_subclasses
>>> from openff.toolkit.typing.engines.smirnoff.plugins import load_handler_plugins
>>> assert load_handler_plugins()[0] in ParameterHandler.__subclasses__()  # does not raise
```

This is a private attribute but one I'd like to behave more in line with expectations.

mattwthompson:

### DIFF
--- a/openff/toolkit/typing/engines/smirnoff/forcefield.py
+++ b/openff/toolkit/typing/engines/smirnoff/forcefield.py
@@ -309,11 +309,11 @@ class ForceField:
             parameter_handler_classes = default_handlers
 
         if load_plugins:
-            plugins = load_handler_plugins()
+            plugin_classes = load_handler_plugins()
 
-            for plugin in plugins:
-                parameter_handler_classes.append(plugin)
-                self._plugin_parameter_handler_classes.append(plugin)
+            for plugin_class in plugin_classes:
+                parameter_handler_classes.append(plugin_class)
+                self._plugin_parameter_handler_classes.append(plugin_class)
 
         self._register_parameter_handler_classes(parameter_handler_classes)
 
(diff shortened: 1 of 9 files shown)


openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -287,14 +288,32 @@ def __init__(
         # otherwise, we can't define two different ParameterHandler subclasses to compare for a new type of energy term
         # since both will try to register themselves for the same XML tag and an Exception will be raised.
         if parameter_handler_classes is None:
-            parameter_handler_classes = all_subclasses(ParameterHandler)
+            # TODO: What is the point of this argument? The default behavior includes a lot of logic.
+            #       Unaware of cases in which the argument is used.
```
j-wags, line 292: Let's keep this TODO in. I also don't know of anyone using a non-default argument here, but the behavior of `all_subclasses` is just so colossally automagic that I think there's decent value in letting users have access to a manual-entry mode.

j-wags (approved):
LGTM!

j-wags:
Please update the releasenotes before you merge!

PR #1544, opened 2023-02-06
Add more unit tests for handler plugins

mattwthompson:
While working on #1543 I came across some behavior in the plugin system that's not tested. We should either remove those lines or add these tests.


j-wags (approved):
Looks great. Thanks for the testing improvements @mattwthompson!

PR #1545, opened 2023-02-07
Do not rely on removed `packaging.version.LegacyVersion` for sorting hierarchy elements

### LINKED ISSUE #1487: `packaging =0.21` breaks things
https://packaging.pypa.io/en/latest/changelog.html#id1

mattwthompson:
Fixes #1487 

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -5785,9 +5785,20 @@ class HierarchyScheme:
         their identifiers.
         """
 
-        # hard-code the sort_func value here, since it's hard to serialize safely
         def sort_func(x):
-            return version.parse(".".join([str(i) for i in x.identifier]))
+            """Sort by int-like tags, using version.Version as a quick hack."""
+            tag = ""
+            for i in x.identifier:
+                try:
+                    int(i)
+                except ValueError:
+                    continue
+                tag += i + "."
+
+            if tag.endswith("."):
+                tag = tag[:-1]
+
+            return version.parse(tag)
 
         self.hierarchy_elements.sort(key=sort_func)
 
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -5812,7 +5812,7 @@ class HierarchyScheme:
 
             return tuple()
 
-        if False:
+        if True:
             # Using chain- and residue-specific sorting functions might be safer against some corner
             # cases, but the logic built into the fallback `_sort_other` works for existing tests
             _sort_functions = {
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -50,7 +50,6 @@ import numpy as np
 from openff.units import Quantity, unit
 from openff.units.elements import MASSES, SYMBOLS
 from openff.utilities.exceptions import MissingOptionalDependencyError
-from packaging import version
 from typing_extensions import TypeAlias
 
 from openff.toolkit.utils.constants import DEFAULT_AROMATICITY_MODEL
@@ -5789,31 +5788,7 @@ class HierarchyScheme:
         Semantically sort the HierarchyElements belonging to this object, according to
         their identifiers.
         """
-
-        def sort_func(x: HierarchyElement) -> version.Version:
-            """Sort by int-like tags, using version.Version as a quick hack."""
-            tag = ""
-            for i in x.identifier:
-                if isinstance(i, int):
-                    _str_i = str(i)
-                elif isinstance(i, str):
-                    try:
-                        int(i)
-                    except ValueError:
-                        continue
-                    _str_i = i
-                else:
-                    raise Exception(f"Unexpected type {type(i)} as identifier")
-                tag += _str_i + "."
-
-            if tag.endswith("."):
-                # If something was smooshed together, return that without the trailing dot,
-                return version.Version(tag[:-1])
-            else:
-                # otherwise just return a zero version placeholder
-                return version.Version("0")
-
-        self.hierarchy_elements.sort(key=sort_func)
+        self.hierarchy_elements.sort(key=lambda x: str([*x.identifier]))
 
     def __str__(self):
         return (
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -30,6 +30,7 @@ import pathlib
 import warnings
 from collections import UserDict
 from copy import deepcopy
+from functools import cmp_to_key
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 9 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -5785,9 +5790,28 @@ def sort_hierarchy_elements(self):
...
+                if isinstance(i, int):
+                    _str_i = str(i)
+                elif isinstance(i, str):
+                    try:
+                        int(i)
+                    except ValueError:
+                        continue
+                    _str_i = i
+                else:
+                    raise Exception(f"Unexpected type {type(i)} as identifier")
```
mattwthompson, line 5806: This could maybe do without existing, but non-`Union[int, str]` sneaking into `HierarchyElement.identifier` makes me nervous and I haven't explored how possible it might be with sloppy inputs

mattwthompson:
This is not the most elegant solution but it passes all existing tests and works in the context of what I think is supposed to happen. We might want to think long-term about a cleaner solution and having this behavior covered by unit tests (i.e. tests that directly cover the behavior of `sort_func`) but I don't think that should prevent a short-term fix from being included.

j-wags (changes requested):
Thanks for starting this! I checked it out and hammered on it a little. I've added a test for some cases where the behavior differs from what it was before.

mattwthompson:
What is the correct behavior?

j-wags:
Sorting should ensure that residues in chain "Z" should come after residues in chain "A".

openff/toolkit/topology/molecule.py
```diff
@@ -5785,11 +5789,42 @@ def sort_hierarchy_elements(self):
...
+            for identifier in element.identifier:
+                # If an identifier is int-ish, cast it to int,
+                # otherwise let it remain as a string
+                try:
+                    x.append(int(identifier))
+                except ValueError:
+                    x.append(identifier)
+
+            return tuple(x)
+
```
mattwthompson, line 5814: This is inelegant but the best I can come up with at the moment given the number of constraints

* One has no idea what the length of `HierarchyElement.identifier` is a prior
* It is implicit but not required that elements should be sorted by identifiers in the order they're provided at construction
  * For chains, this is unimportant given there is only one identifier
  * For residues, this shouldn't be important unless a single residue number in a chain maps to multiple insertion codes and/or residue names (?)
* Residue numbers can be stored as string representations of numbers
* Empty values can sneak in as `" "` and `"None"` but never `None`
* 100% of the time, our users care about chains and residues but some effort must be made to sort other arbitrary hierarchy schemes

mattwthompson:
One of the tests is failing because a duplicate `HierarchyScheme` is added and stuffing residues with `"None"` everywhere, which breaks my assumption that `residue_number` is either an int or a string that can be cast to one.

This is using the `main` branch`:

```python3
In [24]: smiles = (
    ...:             "[H]c1c(c(c(c(c1[H])[H])C([H])([H])[C@]2(C(=O)N([C@](C(=O)N([C@@](C(=O)N3[C@@](C(=O)N2[H])"
    ...:
    ...:             "(C(C(C3([H])[H])([H])[H])([H])[H])[H])([H])C([H])([H])[H])[H])([H])C([H])([H])C4=C(N(c5c4c
    ...: (c(c(c5[H])"
    ...:             "[H])[H])[H])[H])[H])[H])[H])[H])[H]"
    ...:         )
    ...: offmol = Molecule.from_smiles(smiles)
    ...: offmol.perceive_residues(strict_chirality=True)

In [25]: offmol.residues
Out[25]:
[HierarchyElement ('None', 'None', 'None', 'None') of iterator 'residues' containing 38 atom(s),
 HierarchyElement ('None', '1', ' ', 'PHE') of iterator 'residues' containing 20 atom(s),
 HierarchyElement ('None', '2', ' ', 'ALA') of iterator 'residues' containing 10 atom(s)]
```

Should I safeguard around this?

j-wags (approved):
Sorry this is taking so long. Setting residue numbers to `None` is weird but we don't require that residue number be an integer (the PDB spec requires it, but we intentionally don't tie our `metadata` behavior to the PDB spec). So we should handle the `"None"` case for residue number.

I tested locally using just `_sort_other` and it worked fine on the existing tests. So this is approved pending just using the `_sort_other` logic throughout.

mattwthompson:
How do you want "None" sorted in this case? Other residues are sorted by numerical comparison after casting to int, which fails here since `"None"` is allowed but cannot be casted.


```python3
In [1]: def f(values):
   ...:     values.sort()
   ...:     return values
   ...:

In [2]: print(f([2, 1, 3]))
[1, 2, 3]

In [3]: print(f([2, 1, 3, "None"]))
TypeError                                 Traceback (most recent call last)
Cell In[3], line 1
----> 1 print(f([2, 1, 3, "None"]))

Cell In[1], line 2, in f(values)
      1 def f(values):
----> 2     values.sort()
      3     return values

TypeError: '<' not supported between instances of 'str' and 'int'
```

If we don't require residue numbers to be or look like ints, but want them sorted like they are, we'll have to add special behavior around this case. Right now `openff/toolkit/tests/test_molecule.py::TestMoleculeResiduePerception::test_perceive_residues_cyclic_peptide_chirality[True]` is failing for this reason.

mattwthompson:
We could just force `"None"` into zeros (as ints)?

mattwthompson:
That fails to handle another case (chain IDs can be `"None"`):

```python3
ipdb> self
HierarchyScheme with uniqueness_criteria '('chain_id',)', iterator_name 'chains', and 2 elements
ipdb> self.hierarchy_elements
[HierarchyElement ('None',) of iterator 'chains' containing 16 atom(s), HierarchyElement ('Z',) of iterator 'chains' containing 1 atom(s)]
ipdb> for element in self.hierarchy_elements:
    print(_sort(element))

(0,)
('Z',)
```

Going back to dealing with chains and residues with specific functions and falling back to `_sort_other` would make fixing this more tractable, I think

j-wags:
I'm working on this locally - Intended to refactor but now I'm mostly categorizing edge cases and adding tests. Thanks for your patience!

j-wags:
Allllright, I took a swing at this. Ultimately it's questionable whether my attempt is an improvement, though it uses the `cmp_to_key` kwarg to `list.sort` in a clean-seeming way, and I kinda like that it doesn't have separate logic for specific iterator names. Having dealt with this, I almost wish I'd restricted `atom.metadata` to only allow strings (I think I chose what I did to make some interoperability roundtrips happy, but ugggh, what a pain)

Anyway, I've lost perspective on this today. Let's take a look at this tomorrow during our check in.

openff/toolkit/topology/molecule.py
```diff
@@ -5785,11 +5790,45 @@ def sort_hierarchy_elements(self):
...
+                        return 1
+                    else:
+                        continue
+                else:
+                    # Otherwise, assume that ints are "greater than" strings.
+                    if type(val1) is int:
+                        return 1
+                    else:
+                        return -1
+            return 0
```
j-wags, line 5829: Comment this/make more explicit

PR #1556, opened 2023-02-27
0.12.1 release prep

j-wags:


j-wags:
There was a new release of bugbear two weeks ago and we've run afoul of B028. Silencing for now.

PR #1565, opened 2023-03-23
Topology.from_pdb

j-wags:
- [x] Implement `Topology.from_multicomponent_pdb`
- [x] Ensure that loading with multiple unique mols works
- [ ] ~Implement water shortcut in SMARTS matching?~
- [ ] ~Add kwarg for capped modified residues, where cap atoms rre identified using a special property~
- [x] Ensure that different stereoisomers in the same file are supported
- [x] Test timing on "real sized" water box (15 seconds for 75k atoms... not great but good enough for a first release)
- [x] Add tests for all ions (including two-letter symbols, both a monoatomic and as part of unique mols)
- [x] Ensure this works when both OE+RDK backends are available
- [x] Test for more failure cases
    - [x] what if user provides both phosphate ions and a phosphorylated molecule as unique mols?
    - [x] If a user provides ethene in the unique_molecules kwarg, how could we keep this from matching something like a leucine sidechain? (could make sure that unique_mol matches only match entire mol?)
- [x] Unmangle copy-pasted docstrings


openff/toolkit/topology/topology.py
```diff
@@ -1524,6 +1524,149 @@ def _ensure_unique_atom_names(self, ensure_unique_atom_names: Union[str, bool]):
...
+        file_path: Union[str, TextIO],
+        unique_molecules: Optional[Iterable[Molecule]] = None,
+        toolkit_registry=GLOBAL_TOOLKIT_REGISTRY,
+    ):
+        """
+        Loads polymers, waters, monoatomic ions, and user-defined molecules from a PDB file.
+
+        Molecules in the PDB file have the following requirements:
+          * Polymer molecules must use the standard atom names described in the
+            `PDB Chemical Component Dictionary <https://www.wwpdb.org/data/ccd>`_.
```
j-wags, line 1540: @Yoshanuikabundi - If you have docstring-reviewing energy I'd love a glance at this. Feel free to directly modify+push any changes/polish you like. I'll merge this early on my Monday (if tests pass) so no worries if you don't get to it before then.

j-wags: (But I am NOT asking for a full PR review, this one is big and complex so I'm self-approving it as product owner. But a glance at the docstring would be great if you are in the mood)

Yoshanuikabundi (approved):
This looks awesome!!! Love the API - elegant, simple, covers common cases with just a file name, eliminates most (see below) sources of ambiguity, clear and reasonable requirements. Have made docstring changes to clear up some ambiguities and hopefully make the logical flow a bit clearer, but these are changes I think are worth considering but beyond the scope of my review:

- The rendered signature is messed up - this is due to a known Sphinx bug with class methods. Fixing it in our docs relies on https://github.com/openforcefield/openff-sphinx-theme/pull/82, which is blocked by https://github.com/openforcefield/openff-recharge/pull/127
- It would be good to set the Topology box vectors from the PDB crystal records (if this is already done we should document it!)
- "Stereochemistry is assigned based on the geometry in the PDB (even if this differs from the stereochemistry in the unique molecule)" - This seems like a major footgun to me. We've tried to cultivate a stereochemistry-safe culture elsewhere and I think users will assume that extends here (and won't read a long docstring to find out it doesn't). I think it would be more consistent (and safer for our users) if a conflict between unique molecule stereochemistry and PDB geometry was an error (possibly with an `ignore_stereochemistry` argument or similar). I've added a noisy warning to the docstring in the meantime.

Yoshanuikabundi:
Oh! And we should update the toolkit showcase with this - since we don't need the unique molecule for the protein, we can just load the receptor with crystal waters into pdbfixer, solvate it, save it to pdb, and then load that from the toolkit. That'll remove two cells I think. I can do that when this is merged if you like.

Is there any appetite for extending the polymer recognition in this method to work on `Topology.from_openmm`?

j-wags:
Sweet - Thanks!

The stereo thing was really intentional. I think this will save us from far more footguns than we introduce. 

Loading box vectors is a great idea. I'll do that before merging.

We should update the example, but I can feel myself getting pulled toward the event horizon of annual meeting coordination, and once I fall in I won't have much technical productivity until late May. So I need to get this in a release before I go over the line, and I'd love to delegate example updates to you or handle it later.

Issue #1028, opened 2021-07-26
Allow Molecule.from_smiles() to accept an optional argument to assign the name or title of the molecule


jchodera:
**Is your feature request related to a problem? Please describe.**
A frequent use case is to create a `Molecule` object from SMILES and then assign the `Molecule.name` field.

**Describe the solution you'd like**
It would be useful if an optimal `name` argument to `Molecule.from_smiles()` enabled this to be done in one line.

**Describe alternatives you've considered**
The standard idiom 
```python
molecule = Molecule.from_smiles(smiles)
molecule.name = name
```
takes two lines.

mattwthompson:
It's also be handy to be able to set the residue name of a molecule at instantiation time, so that users don't have to do things like

```python
for atom in water.atoms:
    atom.metadata["residue_name"] = "HOH"
```

Arguably this would push the `Molecule` class in the direction of having too many arguments and options  - I'll be doing this downstream one way or another.

https://github.com/openforcefield/openff-interchange/issues/819

PR #1567, opened 2023-04-03
Allow setting `Molecule.name` in constructor and class methods

### LINKED ISSUE #1566: Allow setting `Molecule.name` in constructor and some methods

`Molecule.name` is not relevant in all use cases but critical in some, i.e. exporting to GROMACS and LAMMPS files. When constructing a `Topology` or collection of `Molecule`s that need names, it's a bit clunky.

```python3

ligand = Molecule.from_smiles(SMILES)
ligand.name = "LIG"
```

or when dealing with multiple molecules
```python3
topology = Topology.from_molecules(
    molecules=[Molecule.from_smiles(smiles) for smiles in ["C", "CC", "CO"]]
)

for index, name in enumerate(["MET", "ETH", "ETO"]):
    topology.molecule(index).name = name
```

I also found that `Molecule.__init__` allows anything to be passed into its constructor, but only keyword arguments go nowhere and only certain named arguments can be processed:

```python
class FrozenMolecule(Serializable):
    def __init__(
        self,
        other=None,
        file_format: Optional[str] = None,
        toolkit_registry: TKR = GLOBAL_TOOLKIT_REGISTRY,
        allow_undefined_stereo: bool = False,
    ):
        ...

class Molecule(FrozenMolecule):
    def __init__(self, *args, **kwargs):
```

As a result, this trick doesn't work:

```python
>>> Molecule(name='test').name
  File "/Users/mattthompson/mambaforge/envs/openff-interchange-env/lib/python3.10/site-packages/openff/toolkit/topology/molecule.py", line 5150, in __init__
TypeError: FrozenMolecule.__init__() got an unexpected keyword argument 'name'
```

It's be nice to do this in one pass, i.e.

```python3
ligand = Molecule.from_smiles(SMILES, name="LIG")

topology = Topology.from_molecules(
    molecules=[
        Molecule.from_smiles(smiles, name=name)
        for smiles, name in zip(["C", "CC", "CO"], ["MET", "ETH", "ETO"])
    ]
)
```

Would a patch that adds `name` as a named argument to `Molecule.__init__` be accepted? It would default to the current value of an empty string. This could also be wired to some `Molecule.from_X` methods without breaking existing behavior, I think.

**Additional context**

Related #1028 for a more specific use case

mattwthompson:
Resolves #1566 #1028 


openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -2002,12 +2009,19 @@ def from_smiles(
...
         )
         return molecule
 
     def from_inchi(
-        self, inchi: str, allow_undefined_stereo: bool = False, _cls=None
+        self,
+        inchi: str,
+        allow_undefined_stereo: bool = False,
+        _cls=None,
+        name: str = "",
```
j-wags, line 2024: Is this necessary? It seems to be unused in the method.

openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -956,6 +965,7 @@ def from_smiles(
         hydrogens_are_explicit: bool = False,
         allow_undefined_stereo: bool = False,
         _cls=None,
+        name: str = "",
```
j-wags, line 968: ditto here and many other places in this file - a `name` kwarg is added but then never used.

j-wags (changes requested):
I think either I'm misunderstanding something or there are a few loose threads here (see comments) - Let's chat today.

Bigger picture - It may be good to restrict the scope of these changes to `Molecule.__init__`, `from_inchi`, and `from_smiles`. This is because `from_rdkit/openeye`, `from_file`, and `from_file_obj` all potentially introduce conflicting names and even if we documented which one wins, allowing the conflict to happen in the first place invites user confusion/surprise.

openff/toolkit/topology/molecule.py
```diff
@@ -1034,6 +1037,8 @@ def __init__(
                     msg += str(value_error)
                 raise ValueError(msg)
 
+        self._name = name
```
j-wags, line 1040: Oh, wow, I now see how this is a disaster. Most of the code paths above that could have loaded a molecule have their own logic for setting that molecule's name, and this would unconditionally overwrite all of that. But to do anything else would require deciding, implementing, documenting, and testing how the priority works out in the event of each conflict.

I hate this initializer so much.

I've been going in circles for like an hour trying to figure out how to make this work and my train of thought is either fully derailed or this is actually hard, but I figure this will at least get partial credit for putting `name` in `Molecule.from_smiles` and I'm gonna redefine this PR to only do that and `from_inchi`. We can figure out how to handle this toxic waste pile of an `__init__` in a different PR.

mattwthompson: Agree, modulo any details I've forgotten since the last time I looked at this code.

In this PR I tried to set the object's name in as many code paths as possible with a little changes to the existing signatures as possible, i.e. some use kwargs and some don't. The major exceptions are `Molecule.from_file` since that umbrellas across so much behavior and some stuff like `.from_file_obj` because I wanted to avoid looking into it. We could just as easily split this into smaller, more targeted PRs for the various input paths. (I forget how many I changed and right now don't want to take the time to survey it.)

I'd be in support of a more substantial rewrite of the initializer and some of the related code paths, even if that meant some behavior changes. The pattern of passing a string to the initializer is possibly the source of most of the issues, given the expansive set of possible inputs it needs to reason through. It'd be as easy to claw back as a toy from a toothing toddler, though ...

mattwthompson:
At some point (not now, not here) we might want to think about having a somewhat fleshed-out other molecule class for testing purposes, one that still subclasses from `FrozenMolecule` and has a few API points, but still much smaller than the existing `Molecule` class. There are a few scattered unit tests that build out something like this for one-off uses, but whenever we think about what a hypothetical "other subclass of `FrozenMolecule`" might be structured like, it feels like I'm back to the drawing board. Something to think about.

j-wags (approved):

PR #1568, opened 2023-04-03
Add more `Bond` tests

mattwthompson:
I noticed a couple of strange/untested bits in `Bond`.


j-wags (approved):
Toyed around with this and the added TODOs are correct - We can get by without `Bond.from_dict`, but to remove it in a non-confusing way we'd either want to make `Bond` not inherit from `Serializable` or replace use of `Molecule._initialize_bonds_from_dict` with `Bond.from_dict`. I went a little ways down each of these refactors and it looks like those would expand the scope of this PR a lot, so it'll be easiest to leave these TODOs in but not block merging of the new tests.

PR #1575, opened 2023-04-12
Simplify Toolkit Showcase example with `Topology.from_pdb`

Yoshanuikabundi:
This PR changes the Toolkit Showcase to use `Topology.from_pdb()` to load a receptor and its crystallographic waters instead of a complicated trip through MDTraj, OpenMM and back again.

It also fixes two bugs I found in `Topology.from_pdb()`:

1. Atom names were not stored in the OFFTK atoms. This meant that NGLView didn't know how to represent the protein. The documentation stated that atom names were stored.
2. When multiple identical molecules were in the PDB, every instance had the same coordinates loaded into the Topology. I think this is because of the logic to only run `from_rdkit` once on each molecule. I fixed this in the dumbest way possible: Adding a call to `top.set_positions()` to the end of the `_polymer_openmm_pdbfile_to_offtop` method.

It's possible the `top.set_positions()` should go in `from_pdb`, as the logic will be shared between RDKit and other toolkits, but the box vectors were set in `_polymer_openmm_pdbfile_to_offtop` and that logic would also be shared so I put it with that.

I have not added tests for these two bugs but I can if wanted!

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -139,6 +139,7 @@ jobs:
             NB_ARGS+=" --ignore=examples/check_dataset_parameter_coverage"
             NB_ARGS+=" --ignore=examples/conformer_energies"
             NB_ARGS+=" --ignore=examples/using_smirnoff_in_amber_or_gromacs"
+            NB_ARGS+=" --ignore=examples/toolkit_showcase"
           fi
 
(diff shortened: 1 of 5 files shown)


Yoshanuikabundi:
Ahhhh because `Topology.from_pdb` only exists in RDKit, this breaks the showcase when that's not installed. That's probably not what we want.

PR #1581, opened 2023-04-14
Add isomorphism checks on `_SimpleMolecule`

### LINKED ISSUE #1580: Add isomorphism checks for `_SimpleMolecule`

When I do something like `Interchange.from_openmm`, I end up with a topology of `_SimpleMolecule`. Exporting these to GROMACS loosely requires an isomorphism check that is currently unsupported by the toolkit. It's called from `Topology.identical_molecule_groups` or something similar, but ultimately fails on `Molecule.are_isomorphic` not knowing how to deal with this input.

I'd like `Molecule.is_isomorphic_with` to work on `_SimpleMolecule`.

I can monkey-patch this, but not easily or without copying sizable chunks of code.

Another backdoor is allowing GROMACS to not de-duplicate isomorphic molecules, but this solution is unworkable for large systems with many similar molecules, i.e. water or solvent.

**Additional context**

mattwthompson:
Resolves #1580 

- [x] Ensure `Topology.identical_molecule_groups` functions on a topology containing ONLY `_SimpleMolecule`
  - [x] Add `_SimpleMolecule.are_isomorphic`
  - [x] Add `_SimpleMolecule.is_isomorphic_with`
  - [x] Short-circuit `Molecule.are_isomorphic` on input of different subclass
  - [x] Short-circuit `_SimpleMolecule.are_isomorphic` on input of different subclass
- [x] Add tests
  - [x] Unit tests for new `_SimpleMolecule` methods
  - [x] Unit tests for short-circuits on each method
  - [x] create a Topology of _SimpleMolecule ensure that `Topology.identical_molecule_groups` behaves as expected

### DIFF
--- a/openff/toolkit/topology/_mm_molecule.py
+++ b/openff/toolkit/topology/_mm_molecule.py
@@ -344,13 +344,16 @@ class _SimpleMolecule:
         _cls = _SimpleMolecule
 
         if isinstance(mol1, nx.Graph) and isinstance(mol2, nx.Graph):
-            pass
+            mol1 = _SimpleMolecule._from_subgraph(mol1)
+            mol2 = _SimpleMolecule._from_subgraph(mol2)
 
         elif isinstance(mol1, nx.Graph):
             assert isinstance(mol2, _cls)
+            mol1 = _SimpleMolecule._from_subgraph(mol1)
 
         elif isinstance(mol2, nx.Graph):
             assert isinstance(mol1, _cls)
+            mol2 = _SimpleMolecule._from_subgraph(mol2)
 
         else:
             # static methods (by definition) know nothing about their class,
--- a/openff/toolkit/topology/_mm_molecule.py
+++ b/openff/toolkit/topology/_mm_molecule.py
@@ -341,15 +341,21 @@ class _SimpleMolecule:
     ) -> Tuple[bool, Optional[Dict[int, int]]]:
         import networkx as nx
 
+        _cls = _SimpleMolecule
+
         if isinstance(mol1, nx.Graph) and isinstance(mol2, nx.Graph):
             pass
+
+        elif isinstance(mol1, nx.Graph):
+            assert isinstance(mol2, _cls)
+
+        elif isinstance(mol2, nx.Graph):
+            assert isinstance(mol1, _cls)
 
         else:
             # static methods (by definition) know nothing about their class,
             # so the class to compare to must be hard-coded here
-            if not (
-                isinstance(mol1, _SimpleMolecule) and isinstance(mol2, _SimpleMolecule)
-            ):
+            if not (isinstance(mol1, _cls) and isinstance(mol2, _cls)):
                 return False, None
 
         if mol1.n_atoms != mol2.n_atoms:
@@ -358,7 +364,7 @@ class _SimpleMolecule:
         if mol1.n_bonds != mol2.n_bonds:
             return False, None
 
-        for this_atom, other_atom in zip(mol1.atoms, mol2.atoms, strict=True):
+        for this_atom, other_atom in zip(mol1.atoms, mol2.atoms):
             if this_atom.atomic_number != other_atom.atomic_number:
                 return False, None
 
--- a/openff/toolkit/topology/_mm_molecule.py
+++ b/openff/toolkit/topology/_mm_molecule.py
@@ -9,7 +9,7 @@ TypedMolecule TODOs
   deserialize a Molecule or a TypedMolecule.
 
 """
-from typing import TYPE_CHECKING, Dict, List, NoReturn, Optional, Union
+from typing import TYPE_CHECKING, Dict, List, NoReturn, Optional, Tuple, Union
 
 from openff.units import unit
 from openff.units.elements import MASSES, SYMBOLS
@@ -25,6 +25,8 @@ from openff.toolkit.utils.utils import deserialize_numpy, serialize_numpy
 if TYPE_CHECKING:
     import networkx as nx
     from openff.units.unit import Quantity
+
+    from openff.toolkit.topology.molecule import FrozenMolecule
 
 
 class _SimpleMolecule:
@@ -311,9 +313,11 @@ class _SimpleMolecule:
             "an OpenFF Molecule with sufficiently specified chemistry."
         )
 
-    def _is_isomorphic_with(self, other) -> bool:
+    def is_isomorphic_with(
+        self, other: Union["FrozenMolecule", "_SimpleMolecule", "nx.Graph"], **kwargs
+    ):
         """
-        Untrustworthy check for pseudo-isomorphism.
+        Check for pseudo-isomorphism.
 
         This currently checks that the two molecules have
         * The same number of atoms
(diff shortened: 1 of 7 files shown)


openff/toolkit/topology/topology.py
```diff
@@ -1417,12 +1421,12 @@ def from_openmm(
...
                 if isomorphic:
                     # Take the first valid atom indexing map
-                    first_topology_atom_index = min(mapping.keys())
+                    first_topology_atom_index = min(mapping.keys())  # type: ignore[union-attr]
                     topology_molecules_to_add.append(
                         (
                             first_topology_atom_index,
                             unq_mol_G,
-                            mapping.items(),
+                            mapping.items(),  # type: ignore[union-attr]
```
mattwthompson, line 1429: These could be fixed by overloading `Molecule.are_isomorphic` with an absolute mess of gore that I'm not inclined to suggest (or an API break to return an empty dictionary instead of `None`):


```python
    @overload
    @staticmethod
    def are_isomorphic(
        mol1: Union["FrozenMolecule", "_SimpleMolecule", nx.Graph],
        mol2: Union["FrozenMolecule", "_SimpleMolecule", nx.Graph],
        return_atom_map: Literal[False] = False,
        aromatic_matching: bool = True,
        formal_charge_matching: bool = True,
        bond_order_matching: bool = True,
        atom_stereochemistry_matching: bool = True,
        bond_stereochemistry_matching: bool = True,
        strip_pyrimidal_n_atom_stereo: bool = True,
        toolkit_registry: TKR = GLOBAL_TOOLKIT_REGISTRY,
    ) -> Tuple[bool, None]:
        ...

    @overload
    @staticmethod
    def are_isomorphic(
        mol1: Union["FrozenMolecule", "_SimpleMolecule", nx.Graph],
        mol2: Union["FrozenMolecule", "_SimpleMolecule", nx.Graph],
        return_atom_map: Literal[True] = False,
        aromatic_matching: bool = True,
        formal_charge_matching: bool = True,
        bond_order_matching: bool = True,
        atom_stereochemistry_matching: bool = True,
        bond_stereochemistry_matching: bool = True,
        strip_pyrimidal_n_atom_stereo: bool = True,
        toolkit_registry: TKR = GLOBAL_TOOLKIT_REGISTRY,
    ) -> Tuple[bool, Dict[int, int]]:
        ...

    @staticmethod
    def are_isomorphic(
        mol1: Union["FrozenMolecule", "_SimpleMolecule", nx.Graph],
        mol2: Union["FrozenMolecule", "_SimpleMolecule", nx.Graph],
        return_atom_map: bool = False,
        aromatic_matching: bool = True,
        formal_charge_matching: bool = True,
        bond_order_matching: bool = True,
        atom_stereochemistry_matching: bool = True,
        bond_stereochemistry_matching: bool = True,
        strip_pyrimidal_n_atom_stereo: bool = True,
        toolkit_registry: TKR = GLOBAL_TOOLKIT_REGISTRY,
    ) -> Tuple[bool, Optional[Dict[int, int]]]:
        # Okay, now you're allowed to do the implementation :)

File: openff/toolkit/topology/molecule.py [hunk 2]
```diff
@@ -1956,11 +1956,23 @@ def are_isomorphic(
             [Dict[int,int]] ordered by mol1 indexing {mol1_index: mol2_index}
             If molecules are not isomorphic given input arguments, will return None instead of dict.
         """
+        import networkx as nx
 
-        def _object_to_n_atoms(obj):
-            import networkx as nx
+        if isinstance(mol1, nx.Graph) and isinstance(mol2, nx.Graph):
+            pass
```
mattwthompson, line 1962: Since we allow this method to take in graphs and want to let `FrozenMolecule` subclasses through, I thought it'd be best to explicitly handle this case separately

openff/toolkit/topology/topology.py
```diff
@@ -1175,9 +1175,13 @@ def _identify_chemically_identical_molecules(
                 if mol2_idx in already_matched_mols:
                     continue
                 mol2 = self.molecule(mol2_idx)
-                are_isomorphic, atom_map = Molecule.are_isomorphic(
-                    mol1, mol2, return_atom_map=True
-                )
+                if isinstance(mol1, type(mol2)) and isinstance(mol2, type(mol1)):
```
mattwthompson, line 1178: This turned out uglier than I thought it would be ... I think this basically just checks if they're the same class since I'm not sure there's any other way two classes can "be" subclasses of each other. I guess just having it try this logic if one is a subclass of the other (but not necessarily both) might handle what you're after:


```suggestion
                if isinstance(mol1, type(mol2)) or isinstance(mol2, type(mol1)):
```

This should catch the case of comparing `Molecule` to somebody else's custom class that subclasses out of `FrozenMolecule`

openff/toolkit/topology/_mm_molecule.py
```diff
@@ -323,18 +327,55 @@ def _is_isomorphic_with(self, other) -> bool:
...
+        if mol1.n_atoms != mol2.n_atoms:
+            return False, None
+
+        if mol1.n_bonds != mol2.n_bonds:
+            return False, None
+
+        for this_atom, other_atom in zip(mol1.atoms, mol2.atoms):
             if this_atom.atomic_number != other_atom.atomic_number:
-                return False
+                return False, None
```
j-wags, line 373: (blocking) For the use cases I can foresee, false negatives in this check are fairly benign (just lead to duplicated atom types), but false positives would be really bad (could silently lead to incorrect energies). I'm pretty certain this could return a false positive when comparing, for example, n-propanol and iso-propano, if both happen to be indexed starting with CCCO. 

So this should probably also evaluate the bond graph by comparing something like `sorted([neighbor.molecule_atom_index for neighbor in this_atom.bonded_atoms()])` for each atom.

I hate to call this blocking because the previous implementation also had this issue, but leaving this in its current state will very likely lead to a Class I Silent Wrong Energy Bug sooner or later.

openff/toolkit/topology/topology.py
j-wags: Good call, I like the "or". Thanks for the nice explanation.

openff/toolkit/topology/topology.py
```diff
@@ -1175,9 +1175,13 @@ def _identify_chemically_identical_molecules(
...
+                    are_isomorphic, atom_map = mol1.__class__.are_isomorphic(
```
j-wags, line 1179: (not blocking) I think we can call staticmethods from any instance of the class, so this may work and remove the unsightly `.__class__` bit

```suggestion
                    are_isomorphic, atom_map = mol1.are_isomorphic(
```

openff/toolkit/topology/topology.py
j-wags: Gore indeed! I'll be keeping the `@overloads` on the outside of the airlock for now.

j-wags (approved):
Approved pending resolution of one blocking comment (or maybe I'm misunderstanding the context and we can chat).

Thanks for getting this implemented so quickly! This is a really graceful implementation.

openff/toolkit/topology/_mm_molecule.py
mattwthompson: I've implemented a different check - atomic number of each bonded neighbor of each atom, accounting for order - that passes the propanol case. I'm not sure in which other cases two of these types of molecules might no be isomorphic but pass these requirements

j-wags: I think there's still a plausible edge case where this could return a false positive. I tried to type it up but realized it would be quicker to just commit a minimal reproducing test. The edge case is where two mols could seem isomorphic via the previous logic depending on the path used to set the indexing. Since SMILES are impossible to read, here's a picture of what the new test compares (with the C atoms labeled with their indices)

Screen Shot 2023-04-21 at 10 29 21 AM

j-wags:
Glorious!

PR #1583, opened 2023-04-17
[WIP] Add a warning to Topology.from_pdb if the stereo of the molecule in the PDB is different from the substructure template or unique mol

j-wags:

### DIFF
--- a/openff/toolkit/utils/rdkit_wrapper.py
+++ b/openff/toolkit/utils/rdkit_wrapper.py
@@ -266,7 +266,12 @@ class RDKitToolkitWrapper(base_wrapper.ToolkitWrapper):
         return offmol
 
     def _polymer_openmm_pdbfile_to_offtop(
-        self, topology_class, pdbfile, substructure_dictionary, coords_angstrom, _ignore_stereo: bool = False
+        self,
+        topology_class,
+        pdbfile,
+        substructure_dictionary,
+        coords_angstrom,
+        _ignore_stereo: bool = False,
     ):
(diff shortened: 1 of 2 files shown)


openff/toolkit/utils/rdkit_wrapper.py
```diff
@@ -401,8 +451,16 @@ def _polymer_openmm_topology_to_rdmol(
...
                         atom_j = mol.GetAtomWithIdx(j)
                         # copy over chirality
+                        # TODO: This is most likely wrong. This will copy over CW/CCW tags but there's no
+                        #  checking that the substituent order in the molecule and substructure are the same,
+                        #  so this is basically random assignment. Thankfully Molecule.from_polymer_pdb and
+                        #  Topology.from_pdb both overwrite this assignment based on 3D coordinates. The only
+                        #  way I can currently think of doing this right would be to arbitrarily assign CW,
+                        #  then try to run a strict substructure match on it, and if that fails then assign
+                        #  CCW. This would be monstrously inefficient and all it would let us do is compare
+                        #  the substructure stereo to 3D stereo, which is of dubious value anyway.
```
j-wags, line 462: Big blocker here, this basically summarizes it. This will need a lot more work and I don't think it's worth holding up the release for.

j-wags:
This has switched into a "won't do" - This behavior has been out in the wild long enough that I think we can just say the behavior is "stereo is always taken from 3D, even if it contradicts the reference mol/substructure, and no warnings will be emitted"

PR #1584, opened 2023-04-18
Remove `TopologyAtom` from docstrings, add some annotations

mattwthompson:

### DIFF
--- a/openff/toolkit/topology/topology.py
+++ b/openff/toolkit/topology/topology.py
@@ -725,7 +725,7 @@ class Topology(Serializable):
 
         Returns
         -------
-        atoms : Generator of TopologyAtom
+        atoms : Generator of Atom
         """
         for molecule in self._molecules:
             for atom in molecule.atoms:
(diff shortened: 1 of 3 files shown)


openff/toolkit/topology/topology.py
```diff
@@ -2494,6 +2494,11 @@ def atom(self, atom_topology_index):
                 return molecule.atom(atom_molecule_index)
             this_molecule_start_index += molecule.n_atoms
 
+        raise AtomNotInTopologyError(
+            f"No atom with index {atom_topology_index} exists in this topology, "
+            f"which contains {self.n_atoms} atoms."
+        )
```
mattwthompson, line 2500: This function's logic has an expression after the `return` statement, which confused the type checker. I threw this in not because it should ever be hit, but maybe it's possible.

j-wags (approved):
LGTM! Thanks for the update :-)

j-wags:
I see this isn't marked "Ready for review" but IMO this is mergeable in its current state (only other thing I could think of is releasenotes, but that's not a blocker). Please merge when you're ready!

mattwthompson:
My oversight - I was unsure if the added exception block was too ugly to get through, but that's the opposite of not asking for feedback 🙃

mattwthompson:
Thanks for the review - and allowing me to do something ugly in the advancement of non-ugly things

PR #1594, opened 2023-04-26
Fix 1593

mattwthompson:
This PR prevents AmberTools 20 and ParmEd 4 from being installed, which breaks some tests.


mattwthompson:
Since AmberTools vendors ParmEd, the constraints in the environment YAML files have no effect except for the OpenEye-only builds.

Based on Jason's comment [here](https://github.com/conda-forge/ambertools-feedstock/pull/105#issuecomment-1523429272) I think we should consider constraining everything to 22 & 3.4 and bumping to 23 & 4 once things are ready (particularly the examples) and not until then.

The differences between 4.0 and 4.1 seem small: https://github.com/ParmEd/ParmEd/releases

j-wags (approved):
Dope. I'd thought about making the failing test require RDKit so that it'd be skipped in OE-only builds but this is a way cleaner and more visible solution.

PR #1606, opened 2023-05-09
Fix CI, update examples

mattwthompson:
- [ ] Tag issue being addressed
- [ ] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/main/openff/toolkit/tests)
- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/main/docs), if applicable
- [ ] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [ ] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/main/docs/releasehistory.rst)

PR #1613, opened 2023-05-15
Finish #1606

mattwthompson:
Some stuff in nightly tests were not fixed in #1606 because they're not run in normal CI.


openff/toolkit/tests/test_molecule.py
```diff
@@ -3405,21 +3405,18 @@ def test_assign_partial_charges(self, toolkit, method):
...
-        # print(
-        #     "molecule {}    charge_sum {}     molecule.total_charge {}".format(
-        #         molecule.name, charges_sum_unitless, total_charge_unitless
-        #     )
-        # )
-        np.allclose(charges_sum_unitless, total_charge_unitless, atol=0.002)
+
+        np.testing.assert_allclose(
+            charges_sum_unitless, total_charge_unitless, atol=0.002
+        )
```
mattwthompson, line 3411: I actually kinda wonder if this line was doing anything (as it was missing an `assert`)

PR #1614, opened 2023-05-15
Remove binder

mattwthompson:


mattwthompson:
In #1613 the binder links started to fail (at a rate less than 100%) so we may want to remove them.

This can all be replaced with Google Colab, if we want to use it.

j-wags (approved):
Thanks @mattwthompson - This is good to go since we'll have colab on the centralized examples page.

PR #1619, opened 2023-05-24
Update partial_charges.setter to convert type and give errors

### LINKED ISSUE #1618: Setting Molecule.partial_charges can fail silently with non-Unit quantities
**Describe the bug**

Noticed while reproducing #1616 that setting partial_charges with non-Unit values just fails silently. I would expect either that the value gets set, or an error gets raised. 

**To Reproduce**

```python
In [1]: %load untyped-charges.py

In [2]: # %load untyped-charges.py
   ...: from openff.toolkit import Molecule
   ...: import numpy as np
   ...:
   ...: molecule = Molecule.from_smiles("C")
   ...: molecule.partial_charges = np.ones(molecule.n_atoms)
   ...: assert molecule.partial_charges is not None
AssertionError                            Traceback (most recent call last)
Cell In[2], line 7
      5 molecule = Molecule.from_smiles("C")
      6 molecule.partial_charges = np.ones(molecule.n_atoms)
----> 7 assert molecule.partial_charges is not None

AssertionError:

```
**Output**

 - Operating system

**Additional context**

Might need a catch-all `else` in the setter for this.

Edit: looking at the code, the cases where `_partial_charges` is actually set is quite specific -- charges with the wrong shape will also run into this same bug.

```
    @partial_charges.setter
    def partial_charges(self, charges):
        """
        Set the atomic partial charges for this molecule.

        Parameters
        ----------
        charges : None or a openff.unit.Quantity - wrapped numpy array [1 x n_atoms]
            The partial charges to assign to the molecule. If not None, must be in units compatible with
            openff.unit.elementary_charge

        """
        if charges is None:
            self._partial_charges = None
        elif charges.shape == (self.n_atoms,):
            if isinstance(charges, unit.Quantity):
                if charges.units in unit.elementary_charge.compatible_units():
                    self._partial_charges = charges
            if hasattr(charges, "unit"):
                from openmm import unit as openmm_unit

                if not isinstance(charges, openmm_unit.Quantity):
                    raise IncompatibleUnitError(
                        "Unsupported type passed to partial_charges setter. "
                        "Found object of type {type(charges)}."
                    )

                elif isinstance(charges, openmm_unit.Quantity):
                    from openff.units.openmm import from_openmm

                    converted = from_openmm(charges)
                    if converted.units in unit.elementary_charge.compatible_units():
                        self._partial_charges = converted
```

lilyminium:
Fixes #1618 


mattwthompson:
🎵🎵🎵 Hello There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 502 - my old friend 🎵🎵🎵

openff/toolkit/topology/molecule.py
```diff
@@ -3073,25 +3075,47 @@ def partial_charges(self, charges):
...
+                )
 
-                elif isinstance(charges, openmm_unit.Quantity):
-                    from openff.units.openmm import from_openmm
+            else:
+                from openff.units.openmm import from_openmm
+
+                converted = from_openmm(charges)
+                if converted.units in unit.elementary_charge.compatible_units():
+                    self._partial_charges = converted.astype(float)
```
mattwthompson, line 3111: I noticed this case doesn't have an `else:` clause, so I think it could silently fail if somebody gets lost and passes an OpenMM quantity with non-charge-like units. This is a weird corner case but you started this PR by finding a similar failure case so here would be a great time to safeguard against it.

openff/toolkit/tests/test_molecule.py
```diff
@@ -3418,6 +3420,25 @@ def test_assign_partial_charges(self, toolkit, method):
...
+        int_charges = np.zeros(molecule.n_atoms, dtype=int)
+        molecule.partial_charges = int_charges * unit.elementary_charge
+        assert molecule.partial_charges.dtype == float
+
+    @pytest.mark.parametrize(
+        "value, error",
+        [
+            (3, IncompatibleTypeError),
+            (np.zeros(5), IncompatibleTypeError),
+            (np.zeros(2), IncompatibleShapeError),
```
mattwthompson, line 3434:
```suggestion
            (np.zeros(2), IncompatibleShapeError),
            (np.zeros(2) * unit.elementary_charge, IncompatibleShapeError),

```
Unless it's already covered (I didn't look too hard) it'd be nice to ensure the shape failure is caught whether or not it's the correct data type

mattwthompson (changes requested):
It's up to @j-wags but if I can stand in for him, this is great! I think I was the last person to work on this setter and this change fixes things I left totally uncovered.

It'd be nice to expand the tests to cover the user passing in OpenMM quantities, including a couple of failure cases covered in the code, since those are currently [missing](https://app.codecov.io/gh/openforcefield/openff-toolkit/pull/1619/blob/openff/toolkit/topology/molecule.py#L3097) in test coverage.

I spot-checked the performance of the OpenEye conversion and this _sped it up_ by a little bit:

```python
In [1]: from openff.toolkit.topology import Molecule

In [2]: oemol = Molecule("aceto.sdf").to_openeye()

In [3]: %%timeit
   ...: Molecule.from_openeye(oemol)
   ...:
   ...:
# fix-1618: 842 µs ± 4.5 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
# upstream/main: 961 µs ± 12.8 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

File: openff/toolkit/topology/molecule.py [hunk 3]
```diff
@@ -3073,25 +3075,47 @@ def partial_charges(self, charges):
...
-                        "Found object of type {type(charges)}."
-                    )
+            return
+
+        if not hasattr(charges, "shape"):
+            raise IncompatibleTypeError(
+                "Unsupported type passed to partial_charges setter. "
+                f"Found object of type {type(charges)}."
+            )
+
```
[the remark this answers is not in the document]
lilyminium: I just updated the error message to say what it *is* expecting. (There's probably more elegant logic to avoid repeating the errors over and over...)

j-wags:
Thanks for jumping in, @mattwthompson - I started working on this yesterday and got distracted by a sea of DMs and emails 🤦 

I'll give final approval on this one. I'm just wrapping my head around whether this is an API/behavior change - Checking whether the old setter could take lists/arrays of int/float as raw values/OFF quantities/OMM quantities and do the right thing, and whether the new setter restricts any of these cases.

docs/releasehistory.md
```diff
@@ -17,6 +17,7 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ### Bugfixes
 - [PR #1617](https://github.com/openforcefield/openff-toolkit/pull/1617): Fixes [#1616](https://github.com/openforcefield/openff-toolkit/issues/1616), by converting NAGL charges to float and converting partial_charges to float before converting to_openeye()
+- [PR #1619](https://github.com/openforcefield/openff-toolkit/pull/1619): Fixes [#1618](https://github.com/openforcefield/openff-toolkit/issues/1618), by making partial_charges.setter more comprehensive in type and shape checking
```
j-wags, line 20:
```suggestion
- [PR #1619](https://github.com/openforcefield/openff-toolkit/pull/1619): Fixes silent partial-charge-setting error [#1618](https://github.com/openforcefield/openff-toolkit/issues/1618), by making partial_charges.setter more comprehensive in type and shape checking.
```

j-wags (approved):
This doesn't appear to restrict the API in any ways that were previously _actually working_, and it handles a bug I was about to report. Great work. This will be a good template to use for other quantity setters like `Molecule.conformers` :-) 

I spotted that this was going to make openmm globally required for test_molecule so I did a little quick protection against that. I also increased the messaging about the severity of this bug in the releasenotes. 

Approved pending tests passing!

PR #1631, opened 2023-06-07
Test on Python 3.11

mattwthompson:
- [x] Add NAGL back if its recipe is split (somehow this works now?)
- [x] Think about if there's an easier way to workaround Interchange adhering to the toolkit's packaged Python pin
- [ ] ~~Update examples environments/workflow~~ already only running on only one Python version, can do this if desired


devtools/conda-envs/test_env.yaml
```diff
@@ -18,9 +18,9 @@ dependencies:
   - smirnoff99Frosst
   - openff-units >=0.2
   - openff-utilities >=0.1.5
-  - openff-interchange-base >=0.3.3
-  - openff-nagl >=0.2.2
-  - openff-nagl-models >=0.0.2
```
mattwthompson, line 23: Ref https://github.com/conda-forge/openff-nagl-feedstock/issues/6

devtools/conda-envs/openeye.yaml
```diff
@@ -17,10 +17,12 @@ dependencies:
...
   - openff-units >=0.2
+  - openff-models
+  - openff-amber-ff-ports
   - openff-utilities >=0.1.5
-  - openff-interchange-base >=0.3.3
-  - openff-nagl >=0.2.2
-  - openff-nagl-models >=0.0.2
+  # openff-interchange-base >=0.3.3
+  # openff-nagl >=0.2.2
+  # openff-nagl-models >=0.0.2
```
j-wags, line 25: Am I correct in thinking that all the changes to this file would be reverted before merging?

j-wags:
The cost/benefit of complicating  the testing infrastructure to sidestep the unavailability of AT on python 3.11 here doesn't seem worth it. Once AT is available for py3.11 I'll be happy to enable this testing and update the package recipe.

mattwthompson:
The complexities here are because of how the toolkit is packaged, not because of AmberTools.

This will at least assess whether or not the test suite passes with 3.11

mattwthompson:
Thanks to unspecified magic happening upstream, this is actually working without much complexity!

j-wags (approved):
Approved pending tests passing! (Excluding RTD)

PR #1637, opened 2023-06-10
Convert ParameterAttribute based on units

### LINKED ISSUE #1633: Substituting units for "id" attribute
**Describe the bug**

"id" is another parameter attribute that gets interpreted as a unit (extension of issue https://github.com/openforcefield/openff-toolkit/issues/1493)

**To Reproduce**

```
In [1]: from openff.toolkit import ForceField

In [2]: ff = ForceField("openff-2.1.0.offxml")

In [3]: handler = ff.get_parameter_handler("LibraryCharges")

In [4]: handler.parameters[2].id
Out[4]: 1 <Unit('kelvin')>

In [5]: import openff.toolkit

In [6]: openff.toolkit.__version__
Out[6]: '0.13.1'
```

**Output**

 - Operating system

**Additional context**

lilyminium:
Fixes #1633 

This is a potential long-term fix for accidentally reading stuff like "K+" as kelvin. Unfortunately it is clumsier than expected, because VirtualSiteType.outOfPlaneAngle and VirtualSiteType.inPlaneAngle unexpectedly have `_unit=None`, despite being defined as below. So this is less than ideal but I thought I'd put it up anyway for discussion :)

```
        outOfPlaneAngle = ParameterAttribute(unit=unit.degree)
        inPlaneAngle = ParameterAttribute(unit=unit.degree)
```

### DIFF
--- a/openff/toolkit/typing/engines/smirnoff/forcefield.py
+++ b/openff/toolkit/typing/engines/smirnoff/forcefield.py
@@ -951,7 +951,8 @@ class ForceField:
                 # `unit=unit.degrees`
                 if infotype._ELEMENT_NAME == "VirtualSite":
                     ignore_keys = [
-                        x for x in ignore_keys
+                        x
+                        for x in ignore_keys
                         if x not in ["outOfPlaneAngle", "inPlaneAngle"]
                     ]
 
--- a/openff/toolkit/typing/engines/smirnoff/forcefield.py
+++ b/openff/toolkit/typing/engines/smirnoff/forcefield.py
@@ -930,7 +930,6 @@ class ForceField:
             # Get the parameter types serialization that is not passed to the ParameterHandler constructor.
             ph_class = self._get_parameter_handler_class(parameter_name)
 
-            # special case virtual sites
             try:
                 infotype = ph_class._INFOTYPE
                 parameter_list_tagname = infotype._ELEMENT_NAME
--- a/openff/toolkit/typing/engines/smirnoff/forcefield.py
+++ b/openff/toolkit/typing/engines/smirnoff/forcefield.py
@@ -904,12 +904,6 @@ class ForceField:
         if "Date" in smirnoff_data["SMIRNOFF"]:
             self._add_date(smirnoff_data["SMIRNOFF"]["Date"])
 
-        # Go through the whole SMIRNOFF data structure, trying to convert all strings to Quantity
-        smirnoff_data = convert_all_strings_to_quantity(
-            smirnoff_data,
-            ignore_keys=["smirks", "name"],
-        )
-
         # Go through the subsections, delegating each to the proper ParameterHandler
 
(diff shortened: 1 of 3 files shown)


openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -935,14 +929,39 @@ def _load_smirnoff_data(self, smirnoff_data: dict, allow_cosmetic_attributes=Fal
...
                 parameter_list_dict = section_dict.pop(parameter_list_tagname, {})
+                attributes = infotype._get_parameter_attributes()
+                ignore_keys = [
+                    attribute_name
+                    for attribute_name, attribute in attributes.items()
+                    if attribute._unit is None
+                ]
+                # for some reason the angle parameters in VirtualSite
+                # do not have `_unit`, despite being defined with
+                # `unit=unit.degrees`
```
mattwthompson, line 950: I doubt this detail is intentional - could be updated there vs. handling as a special case here?

mattwthompson (commented):
I don't grok the full scope of the code with all of the different variables passed through `convert_all_strings_to_quantity`, but the approach of ignoring everything without a `_unit` attribute defined seems like it should be an elegant fix.

I'm not sure about all of the cases of this being defined and not being defined, particularly with plugins. Its [default](https://github.com/openforcefield/openff-toolkit/blob/a53a7e1b29d6ec134114da517b17d19b34908602/openff/toolkit/typing/engines/smirnoff/parameters.py#L338-L344) is `None`, so it might be possible for a plugin developer to define an attribute that should have units, but forget to define it, and then that attribute would be ignored during parsing? Then again, that's arguably an undefined case and something else might break before that. I haven't gone any farther down this line of thinking. (I think the plugins we're responsible for define their `unit` attributes, and if they don't, I'd argue they should.)

Sorry for the partial review - hitting send now nonetheless.

openff/toolkit/typing/engines/smirnoff/forcefield.py
lilyminium: Sorry, that wasn't very clear on my part -- I don't know why this is happening. I've raised #1641 for discussion.

lilyminium:
`so it might be possible for a plugin developer to define an attribute that should have units, but forget to define it, and then that attribute would be ignored during parsing?`

Yes I think that's a potential issue. I *think* it would be ok to enforce that it should be defined with units, or at least a converter function, but breaking existing code could be an undesirable side effect here.

j-wags (commented):
This is promising as a general solution but the vsite situation is really strange and I'd prefer to go with the simpler solution immediately. If we can get to the bottom of this vsite thing I'd be happy to consider this as a solution in the future though.

j-wags:
I think this was superseded by #1640 - Please feel free to reopen if I'm mistaken!

Issue #1635, opened 2023-06-09
Not being able to load the 2.1.0 forcefield file


ijpulidos:
**Describe the bug**
I noticed that the latest release (0.13.1) is having errors with loading the `2.1.0` forcefield release. The same happens with `2.1.0.rc-1`.

Even though the files exist in the directory

```bash
❯ ls /home/user/miniconda3/envs/openff-ffs-test/lib/python3.10/site-packages/openforcefields/offxml
opc-1.0.0.offxml         openff-1.2.0.offxml          openff-2.1.0.offxml                    openff_unconstrained-1.2.0.offxml          openff_unconstrained-2.1.0.offxml
opc.offxml               openff-1.2.1.offxml          openff-2.1.0-rc.1.offxml               openff_unconstrained-1.2.1.offxml          openff_unconstrained-2.1.0-rc.1.offxml
openff-1.0.0.offxml      openff-1.3.0.offxml          openff_unconstrained-1.0.0.offxml      openff_unconstrained-1.3.0.offxml          README.md
openff-1.0.0-RC1.offxml  openff-1.3.1-alpha.1.offxml  openff_unconstrained-1.0.0-RC1.offxml  openff_unconstrained-1.3.1-alpha.1.offxml  tip3p-1.0.0.offxml
openff-1.0.0-RC2.offxml  openff-1.3.1.offxml          openff_unconstrained-1.0.0-RC2.offxml  openff_unconstrained-1.3.1.offxml          tip3p_fb-1.0.0.offxml
openff-1.0.1.offxml      openff-2.0.0.offxml          openff_unconstrained-1.0.1.offxml      openff_unconstrained-2.0.0.offxml          tip3p_fb.offxml
openff-1.1.0.offxml      openff-2.0.0-rc.1.offxml     openff_unconstrained-1.1.0.offxml      openff_unconstrained-2.0.0-rc.1.offxml     tip3p.offxml
openff-1.1.1.offxml      openff-2.0.0-rc.2.offxml     openff_unconstrained-1.1.1.offxml      openff_unconstrained-2.0.0-rc.2.offxml
```

**To Reproduce**

```python
In [1]: from openff.toolkit import ForceField

In [2]: force_field = ForceField('openff-2.1.0.offxml')
```

**Output**

```python
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[2], line 1
----> 1 force_field = ForceField('openff-2.1.0.offxml')

[...]

PR #1640, opened 2023-06-13
Fix force field parsing with Pint 0.22 or of OpenFF 2.1.0

mattwthompson:
#1635 #1633


mattwthompson:
#1637 is probably a better long-term solution, though larger in the scope of changes. The added tests are (currently) identical.

j-wags (approved):
Thanks @mattwthompson. I agree that #1637 is a better long term solution but when I add up the workaround for the vsite mystery and the potential that interpreting `attribute._unit` has to affect plugins, I think this PR is the way to go in the immediate tem. 

Please update the releasenotes before you merge!

PR #1646, opened 2023-06-16
Standardize typing and 0.13.2 releasenotes

j-wags:
- [x] Fix mypy failures following pint 0.22 and openff-units 0.2.1 releases
- [x] Try to make `Molecule.from_qcschema` example in docstring less flaky by moving to a molecular formula with fully defined stereo in all entries (cc #1643).
- [x] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/main/docs/releasehistory.rst) for 0.13.2 release

### DIFF
--- a/devtools/conda-envs/beta_rc_env.yaml
+++ b/devtools/conda-envs/beta_rc_env.yaml
@@ -18,7 +18,7 @@ dependencies:
   - openmm >=7.6
   - openff-forcefields >=2023.05.1
   - smirnoff99Frosst
-  - openff-units >=0.2
+  - openff-units =0.2.0
   - openff-utilities >=0.1.5
   - openff-interchange-base >=0.2
   - openff-nagl >=0.2.2
--- a/devtools/conda-envs/openeye-examples.yaml
+++ b/devtools/conda-envs/openeye-examples.yaml
@@ -17,7 +17,7 @@ dependencies:
   - openff-forcefields >=2023.05.1
   - smirnoff99Frosst
   - openff-amber-ff-ports >=0.0.3
-  - openff-units >=0.2
+  - openff-units =0.2.0
   - openff-utilities >=0.1.5
   - openff-interchange-base >=0.3.3
   - openff-nagl >=0.2.2
--- a/devtools/conda-envs/openeye.yaml
+++ b/devtools/conda-envs/openeye.yaml
@@ -16,7 +16,7 @@ dependencies:
   - openmm >=7.6
   - openff-forcefields >=2023.05.1
   - smirnoff99Frosst
-  - openff-units >=0.2
+  - openff-units =0.2.0
   - openff-utilities >=0.1.5
   - openff-interchange-base >=0.3.3
   - openff-nagl >=0.2.2
--- a/devtools/conda-envs/rdkit-examples.yaml
+++ b/devtools/conda-envs/rdkit-examples.yaml
@@ -16,7 +16,7 @@ dependencies:
   - openff-forcefields >=2023.05.1
   - smirnoff99Frosst
   - openff-amber-ff-ports >=0.0.3
-  - openff-units >=0.2
+  - openff-units =0.2.0
   - openff-utilities >=0.1.5
   - openff-interchange-base >=0.3.3
   - openff-nagl >=0.2.2
--- a/devtools/conda-envs/rdkit.yaml
+++ b/devtools/conda-envs/rdkit.yaml
@@ -15,7 +15,7 @@ dependencies:
   - openmm >=7.6
   - openff-forcefields >=2023.05.1
   - smirnoff99Frosst
-  - openff-units >=0.2
+  - openff-units =0.2.0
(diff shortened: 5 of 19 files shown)


j-wags:
Hm, some sort of issue with mypy and openff-units 0.2.1. Looks like these failures happened in nightly CI as well. I'll dig into this. Any tips for where to look first, @mattwthompson?

j-wags:
My best guess right now is something to do with https://github.com/openforcefield/openff-units/pull/56 - I'm guessing that I need to import stuff from openff-units differently or change how I'm using them in type annotations.

j-wags:
Alright - I'm getting these to go away. It looks like type annotations don't like paths. So instead of

```python
from openff.units import unit
foo: unit.Quantity = ...
```

I should do

```python
from openff.units import unit, Quantity
foo: Quantity = ...
```

mattwthompson:
Pint 0.21 and onward introduced serious typing regressions that won't, in my estimation, be resolved anytime soon. I started working on stubs today but it's not trivial and I haven't finished it up.

mattwthompson:
Falling back to an older version skips all of the typing checks, which is not actually a solution but should turn things green

j-wags:
I'm having some success with just removing paths from type annotations (`unit.Quantity` --> `Quantity`) and I think that propagating this pattern will get mypy to stop complaining even with new Pint and openff-units. This is a fine solution for me - given multiple ways to do things I certainly appreciate some tool having a black-style opinion on one way being "right" - though I'm gonna plan to discuss this Monday since I don't want to be cutting a release this late in the day anyway.

mattwthompson:
I don't think that's a bad solution, but in my view the real issue is in the 0.2.1 build. Requiring downstream packages to not use the `unit.Quantity` pattern is a non-starter and that behavior will be fixed in a future release one way or another

j-wags:
I've tried - again - to figure out what's going on with the flaky `from_qcschema` examples, and have - again - failed.

The proximal error is:

```
...
openff/toolkit/utils/toolkit_registry.py::toolkit.utils.toolkit_registry.ToolkitRegistry.resolve PASSED [100%]

_______ [doctest] toolkit.topology.molecule.FrozenMolecule.from_qcschema _______
4629         molecule : openff.toolkit.topology.Molecule
4630             An OpenFF molecule instance.
4631 
4632         Examples
4633         --------
4634         Get Molecule from a QCArchive molecule record:
4635 
4636         >>> from qcportal import FractalClient
4637         >>> client = FractalClient()
4638         >>> offmol = Molecule.from_qcschema(
UNEXPECTED EXCEPTION: KeyError("The record must contain the hydrogen mapped smiles to be safely made from the archive. It is not present in either 'attributes' or 'extras' on the provided `qca_record`")
  File "/home/runner/micromamba/envs/openff-toolkit-test/lib/python3.9/doctest.py", line 1334, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest toolkit.topology.molecule.FrozenMolecule.from_qcschema[2]>", line 1, in <module>
  File "/home/runner/micromamba/envs/openff-toolkit-test/lib/python3.9/site-packages/openff/utilities/utilities.py", line 80, in wrapper
    return function(*args, **kwargs)
  File "/home/runner/work/openff-toolkit/openff-toolkit/openff/toolkit/topology/molecule.py", line 4714, in from_qcschema
    raise KeyError(
KeyError: "The record must contain the hydrogen mapped smiles to be safely made from the archive. It is not present in either 'attributes' or 'extras' on the provided `qca_record`"
/home/runner/work/openff-toolkit/openff-toolkit/openff/toolkit/topology/molecule.py:4638: UnexpectedException
=============================== warnings summary ===============================
openff/toolkit/topology/molecule.py::toolkit.topology.molecule.FrozenMolecule.from_qcschema
...
```

I'm unable to reproduce this locally. To try and reproduce the issue I played around with the `limit` keyword (which seems to default to its maximum value of 2000) and `skip` keyword (which continues returning 2000 molecules even when raised to values >10,000, but doesn't return anything at values >1,000,000, so it's clearly doing something). 

For future work my code is 

```python
from qcportal import FractalClient
client = FractalClient()
#dataset = client.query_molecules(molecular_formula="C16H20N3O5")
dataset = client.query_molecules(molecular_formula="C7H12N2O4",
                                 #skip=17500,
                                 limit=1)
print(len(dataset))
from matplotlib import pyplot
import numpy 
loadable = numpy.zeros((len(dataset), 1))
for idx, entry in enumerate(dataset):
    try:
        Molecule.from_qcschema(entry)
        loadable[idx] = 1
    except Exception as e:
        pass
pyplot.plot(loadable)
```

My current hypotheses are either:
* After some duration of inactivity, the entries on QCArchive revert to some sort of "cached" state, and the first accesses don't pull down their metadata
* There are some entries for this molecular formula that really don't have metadata, and some of those randomly get pulled down as element[0] of the records (this seems really improbable, given how frequently this fails on CI, and how I've been unable to reproduce it locally after tens of attempts)

Issue #1647, opened 2023-06-18
Validating the conversion to GROMACS


8marmar8:
Dear developers,

I was trying to use the Jupiter Notebook from the folder https://github.com/openforcefield/openff-toolkit/blob/main/examples/using_smirnoff_in_amber_or_gromacs/export_with_interchange.ipynb, in order to parametrize a molecule that I want to simulate in Gromacs. I obtained the top and gro files, but during the last step: Validating the conversion to GROMACS and LAMMPS files, in particular:

```
if find_executable("gmx"):
    pprint(get_gromacs_energies(interchange).energies)
I received an error about the energies:
---------------------------------------------------------------------------
UnsupportedExportError                    Traceback (most recent call last)
Cell In[113], line 2
      1 if find_executable("gmx"):
----> 2     pprint(get_gromacs_energies(interchange).energies)

File ~/anaconda3/envs/openff-toolkit/lib/python3.10/site-packages/openff/interchange/drivers/gromacs.py:86, in get_gromacs_energies(interchange, mdp, round_positions, detailed)
     57 def get_gromacs_energies(
     58     interchange: Interchange,
     59     mdp: str = "auto",
     60     round_positions: int = 8,
     61     detailed: bool = False,
     62 ) -> EnergyReport:
     63     """
     64     Given an OpenFF Interchange object, return single-point energies as computed by GROMACS.
     65 
   (...)
     83 
     84     """
     85     return _process(
---> 86         _get_gromacs_energies(
     87             interchange=interchange,
     88             mdp=mdp,
     89             round_positions=round_positions,
     90         ),
     91         detailed=detailed,
     92     )

File ~/anaconda3/envs/openff-toolkit/lib/python3.10/site-packages/openff/interchange/drivers/gromacs.py:108, in _get_gromacs_energies(interchange, mdp, round_positions)
    106     mdconfig = MDConfig.from_interchange(interchange)
    107     mdp_file = "tmp.mdp"
--> 108     mdconfig.write_mdp_file(mdp_file)
    109 else:
    110     mdp_file = _get_mdp_file(mdp)

File ~/anaconda3/envs/openff-toolkit/lib/python3.10/site-packages/openff/interchange/components/mdconfig.py:162, in MDConfig.write_mdp_file(self, mdp_file)
    160     mdp.write(f"rcoulomb = {coul_cutoff}\n")
    161 else:
--> 162     raise UnsupportedExportError(
[...]

PR #1648, opened 2023-06-18
Refresh examples

mattwthompson:
Would prevent #1647 from being first detected by a user

- [ ] Investigate whether or not Interchange should throw a more specific error
- [x] Install GROMACS in examples environments
- [x] Use `Topology.from_pdb` instead of unnecessarily using `Topology.from_openmm`


mattwthompson:
There's a definitive claim made in `examples/SMIRNOFF_simulation/run_simulation.ipynb` that I'm not sure is currently true, but I'm not pushing to update it here:

> Note on partial charges:</b> The full 1.0.0 release will implement support for the definition of semiempirical partial charge treatment directly into the SMIRNOFF force field file (for details, see https://openforcefield.github.io/standards/standards/smirnoff/#partial-charge-and-electrostatics-models). Moreover, it will be possible to import charges directly from

j-wags (approved):
Thanks! I'll update the releasenotes and merge.

PR #1649, opened 2023-06-19
Make tests private(-ish)

mattwthompson:
We agreed to do this a while back but I don't think we actually did it.

- [ ] Tag issue being addressed
- [ ] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/main/openff/toolkit/tests)
- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/main/docs), if applicable
- [x] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [ ] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/main/docs/releasehistory.rst)

PR #1658, opened 2023-06-28
Ensure `pytest` finds `conftest.py`

mattwthompson:
After #1649, the nightly cron job isn't finding `conftest.py` which sets up the `--runslow` option.

https://github.com/openforcefield/openff-toolkit/actions/runs/5395786436

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -148,8 +148,6 @@ jobs:
             PYTEST_ARGS+=" --runslow"
           fi
 
-          PYTEST_ARGS+=" --runslow"
-
           python -m pytest $PYTEST_ARGS $COV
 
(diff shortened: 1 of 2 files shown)


mattwthompson:
False alarm, maybe. I had an ARM version of OpenEye Toolkits installed, which triggered a different marker to skip some tests.

mattwthompson:
Looking like this is behaving as expected after all - the slow tests only run in the OpenEye jobs

https://github.com/openforcefield/openff-toolkit/actions/runs/5402801575/jobs/9814645494

mattwthompson:
The tests, fast and slow, are all passing as hoped. The only missing bit is that CodeCov doesn't work - hopefully @getsentry's acquisition changes this at some point.

Issue #1660, opened 2023-06-29
`Molecule.add_hierarchy_scheme` allows registering hierarchy schemes with iterator names that conflict with existing attributes


mattwthompson:
In #1601 it was suggested that on iterators like `Topology.residues`:

> If a `HierarchyIterator` tries to give itself a name like "atoms", which would conflict with `Topology.atoms`, then an error should be raised

However this behavior currently exists on `Molecule`. It could be caught when creating a `Topology`, but that wouldn't cover all cases.

```python
In [1]: from openff.toolkit.topology.molecule import *

In [2]: molecule = Molecule.from_smiles("CCO")

In [3]: molecule.add_hierarchy_scheme(uniqueness_criteria=("magic",), iterator_name="atoms")
Out[3]: HierarchyScheme with uniqueness_criteria '('magic',)', iterator_name 'atoms', and 1 elements

In [4]: molecule.atoms
Out[4]:
[Atom(name=, atomic number=6),
 Atom(name=, atomic number=6),
 Atom(name=, atomic number=8),
 Atom(name=, atomic number=1),
 Atom(name=, atomic number=1),
 Atom(name=, atomic number=1),
 Atom(name=, atomic number=1),
 Atom(name=, atomic number=1),
 Atom(name=, atomic number=1)]
```

I don't think this should be allowed, and should probably be caught in `add_hierarchy_scheme` and not the `__getattr__` magic.

Issue #1659, opened 2023-06-29
`HierarchyElement` does not implement equality checks


mattwthompson:
**Describe the bug**

I expected two residue `HierarchyElements` representing the same data to be treated as equal in Python.

**To Reproduce**

```python
In [1]: from openff.toolkit.utils import get_data_file_path

In [2]: from openff.toolkit import Molecule

In [3]: protein_path = get_data_file_path("proteins/ace-ala-nh2.pdb")

In [4]: Molecule.from_polymer_pdb(protein_path).residues[0] == Molecule.from_polymer_pdb(protein_path).residues[0]
Out[4]: False
```

Needless to say, I expected the first residue to be equal to itself.

This is based on a498ae0f29625d63dd6d85eb75e930ab272a4239, though the behavior is probably unchanged for some period of time.

j-wags:
I think that supporting a `==` for `HierarchyElements` would be really problematic, and instead think it should just fall back to the `is` operator (or maybe even raise an error saying it's not implemented). Otherwise we get into user-footgun-land pretty quickly:

* Are two chemically-identical alanines in chemically different molecules equal?
* Are two chemically-identical alanines from different parts of the same molecule equal?
* Are two chemically-identical alanines from the same part of two chemically-identical molecules equal?
* Are two alanines from molecules that define the "residue" `HierarchyScheme" differently equal?
* Are two alanines with different stereochemistry equal?

I think these are all cases where at least 10% of users would take the minority opinion and end up confused no matter what we pick or how well we document the choice. So I think that sticking with the `is` operator for `HierarchyElement` identity comparison (so, nothing is equal except for the same in-memory object) is the way to go.

mattwthompson:
I'm not sure the `is` operator will work here:

```python
In [2]: Molecule.from_polymer_pdb(protein_path).residues[0] is Molecule.from_polymer_pdb(pr
   ...: otein_path).residues[0]
Out[2]: False

In [3]: id(Molecule.from_polymer_pdb(protein_path).residues[0])
Out[3]: 6123700128

In [4]: id(Molecule.from_polymer_pdb(protein_path).residues[0])
Out[4]: 4580692592

j-wags:
Ah - To clarify, I think the simple solution is that all of the hypotheticals I listed above should return `False`. The only thing that should return `True` is `protein.residues[0] == protein.residues[0]`.

PR #1661, opened 2023-06-29
Add hierarchy scheme iterators to `Topology`

### LINKED ISSUE #1601: Add hierarchy iterators as attributes of `Topology`

```
ipdb> topology.molecule(0).residues
[HierarchyElement ('1', 1, '', 'JJN') of iterator 'residues' containing 5 atom(s)]
ipdb> topology.residues
*** AttributeError: 'Topology' object has no attribute 'residues'
ipdb> topology.hierarchy_iterator("residues")
<generator object Topology.hierarchy_iterator at 0x1b2d89f90>
```

Make iterators behave on `Topology` the same as they do on `Molecule`.

Going through `Topology.hierarchy_iterators("residues")` is okay but not ideal.
**Additional context**

@j-wags:

> [The intent is to have this added eventually,] but I didn’t have time to finish and test that before the 0.11 release. So now offmol.residues works but offtop.residues doesn’t yet.

mattwthompson:
Resolves #1601 

Possibly blocked by #1660, would be convenient for a path forward with #1659 to also be decided.

  - [x] Error if any molecule is misses requested iterator
  - [x] Order hierarchy elements by molecule, _then_ whatever sorting is defined by the scheme itself
  - [x] Make no effort to sort between molecules
  - [ ] Error out if a name conflict with existing attributes is found


openff/toolkit/topology/topology.py
```diff
@@ -2385,3 +2386,25 @@ def hierarchy_iterator(
             if hasattr(molecule, iter_name):
                 for item in getattr(molecule, iter_name):
                     yield item
+
+    def __getattr__(self, name: str) -> List["HierarchyElement"]:
+        """If a requested attribute is not found, check the hierarchy schemes"""
+        # Avoid attempting to process dunder methods as hierarchy scheme iterator names
+        if name.startswith("__"):
+            raise AttributeError
+
```
mattwthompson, line 2395: This is not my proudest work but I haven't found an issue with it yet and some other workarounds I tinkered with were more complex and less functional. This flared up in tests when `Topology.__add__` or something similar called `deepcopy()` on a topology, which for one reason or another checks to see if `Topology.__deepcopy__` exists (it does not). AFAICT the implementation here is the same with `name.startswith("_")`.

The best argument one could make for this is pointing out that `__getattr__` is only called as a last resort when an attribute it not found on an object, and we don't want any dunder methods to be processed as hierarchy scheme iterator names anyway.

openff/toolkit/topology/topology.py
```diff
@@ -2385,3 +2386,25 @@ def hierarchy_iterator(
...
+        try:
+            return [
+                element
+                for molecule in self.molecules
+                for element in getattr(molecule, name)
+            ]
+        except AttributeError as error:
+            raise AttributeError(
+                f"'{self.__class__.__name__}' object has no attribute {name!r}"
+            ) from error
```
j-wags, line 2410: (not blocking) I think the nested error here is a pretty good idea. It may be good to use this chance to make the error message more informative instead of mirroring the Python default AttributeError. Maybe something like "Before raising this AttributeError, the Topology class checked whether the Molecules it contains have a HierarchyScheme with a matching name, however at least one Molecule in this Topology does not."

openff/toolkit/topology/topology.py
```diff
@@ -2385,3 +2386,25 @@ def hierarchy_iterator(
...
+            raise HierarchyIteratorNameConflictError(
+                f"Name {name} is already defined as an attribute/method of `Topology`."
+            )
+
+        try:
+            return [
+                element
+                for molecule in self.molecules
+                for element in getattr(molecule, name)
+            ]
```
j-wags, line 2406: (blocking) This runs the risk of forwarding the entire Molecule API to Topology, giving us some weird stuff that we didn't intend for iteration like `Topology.partial_charges`. This should make it so only HierarchySchemes get forwarded to the Topology API. 

```suggestion
            return [
                element
                for molecule in self.molecules
                for element in molecule.hierarchy_schemes[name].hierarchy_elements)
            ]
```

j-wags (changes requested):
Thanks for taking a stab at this - I hadn't appreciated how complex this would end up being. Left a few suggestions but overall this is on a good track.

openff/toolkit/topology/topology.py
```diff
@@ -2385,3 +2386,25 @@ def hierarchy_iterator(
...
+    def __getattr__(self, name: str) -> List["HierarchyElement"]:
+        """If a requested attribute is not found, check the hierarchy schemes"""
+        # Avoid attempting to process dunder methods as hierarchy scheme iterator names
+        if name.startswith("__"):
+            raise AttributeError
+
+        if name in dir(self):
+            raise HierarchyIteratorNameConflictError(
+                f"Name {name} is already defined as an attribute/method of `Topology`."
+            )
```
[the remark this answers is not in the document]
mattwthompson: This was meant to be something of a double safeguard even though it can't be reached in the current implementation. When there is a clash, I'm not sure which behavior from `Topology` the user would get.

j-wags (approved):
Excellent! Thanks @mattwthompson.

Issue #1668, opened 2023-07-11
Check energies reported in using_smirnoff_in_amber_or_gromacs /export_with_interchange.ipynb


pbuslaev:
I tried to reproduce the example [notebook for using SMIRNOFF with GROMACS](https://github.com/openforcefield/openff-toolkit/blob/main/examples/using_smirnoff_in_amber_or_gromacs/export_with_interchange.ipynb). I used **`openff-toolkit`** version `0.14.0` and I used the structure file provided in the tutorial directory. The energies that I got are

| Software | Bond | Angle | Torsion | Electrostatics | vdW | RBTorsion |
| ---- | ---- | ---- | ---- | ---- | ---- | ---- | 
| **OpenMM**	| 0.233879 | 7.141818 | 25.580587 |	-6.840315 | 9.658657 | NaN |
| **GROMACS** | 0.233865 | 7.141832 | 25.580585 | -6.858640 | 9.667050 | 0.0 |

The reported in the [notebook](https://github.com/openforcefield/openff-toolkit/blob/main/examples/using_smirnoff_in_amber_or_gromacs/export_with_interchange.ipynb) values are:

| Software | Bond | Angle | Torsion | Electrostatics | vdW | RBTorsion |
| ---- | ---- | ---- | ---- | ---- | ---- | ---- | 
| **OpenMM** | 0.233879 | 7.141818 | 25.580587 | -15.215644 | 9.658657 | NaN |
| **Amber** | 0.233886 | 7.141670 | 25.580558 | -15.216371 | 9.727382 | NaN |
| **GROMACS** | 0.233869 | 7.141835 | 25.580585 | -15.235316 | 9.667050 | 0.0 |

The difference is only in the Electrostatics term. So I tried to check the assigned charges, by running `antechamber` on structures of cyclohexane and ethanol from [`1_cyclohexane_1_ethanol.pdb`](https://github.com/openforcefield/openff-toolkit/blob/main/examples/using_smirnoff_in_amber_or_gromacs/1_cyclohexane_1_ethanol.pdb) separately. I compared those charges with charges assigned by **`openff-toolkit`**, by applying 
```python
molecule.assign_partial_charges(
    "am1bcc",
    use_conformers=molecule.conformers
)
```
[...]

PR #1671, opened 2023-07-18
Render examples without an OpenEye license

mattwthompson:
All of the notebooks tracked in the repo were run with access to OpenEye, which most of our users don't also have. This leads to some reasonable confusion (#1668). For context, the centralized examples are currently rendered without OpenEye.

This also uncovered a couple of cobwebs, like some deprecation warnings that are no longer relevant, with some of these notebooks not run for a couple of years.

This would also be a good time to switch notebooks over to Sage 2.1.0 - I'm just opening this as-is to keep it a bit smaller.


j-wags (approved):
LGTM - Thanks @mattwthompson!

PR #1683, opened 2023-08-03
Do not require OpenMM for tests

mattwthompson:
Some users do not wish to download the CUDA toolkit to do things that don't require it. Much of the toolkit's API has been updated to accommodate this. The testing layer still has OpenMM, baked in, however. It makes sense that you'd want to run tests with OpenMM, but importing anything from `openff.toolkit._tests`, such as `create_ethanol` requires OpenMM.

This PR skips tests when OpenMM is required but not installed and also tucks away from heavy OpenMM functions into a separate submodule.

I also noticed a pattern of
* Do some toolkit stuff, like prepare molecules and force fields with slightly funky charge methods
* Create an OpenMM System
* Inspect the OpenMM System's charges

which does not need OpenMM to be installed and wastes a small amount of execution time calling `Interchange.to_openmm`. I didn't fix that here, but I'd like to in the future. I expect this would knock a several seconds off of test runtime.


mattwthompson:
I no longer _need_ this for my original purpose, so I'm going to strip out a few things. But there are some improvements I'd still like to get across the finish line, so I'm not dumping it. It'll just not get eyes in a week or two, I think.

PR #1687, opened 2023-08-04
Fix docs builds

mattwthompson:
Starting grasping at straws

https://pip.pypa.io/en/stable/news/#v23-2

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -5,7 +5,7 @@ channels:
     - conda-forge
     - defaults
 dependencies:
-    - pip
+    - pip ==23.2
     # readthedocs dependencies
     - myst-nb
     - myst-parser>=0.14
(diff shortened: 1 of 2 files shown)


mattwthompson:
That ... worked?

mattwthompson:
If I'm not missing anything, I think this is good to go.

Pinning to an old version of `pip` is not good, but our docs not being build it not good-er.

j-wags:
I redid the "stable" pin assignment in the 0.14.2 release around the same time as you opened this PR - I've reverted the changes to kick CI so we can tease apart whether this was the effective fix.

j-wags:
Looks like the RTD build ran with no changes. Thanks for hopping on this so quickly, but I think it was doing the new release and stable-tag-setting correctly that got RTD back.

PR #1698, opened 2023-08-16
Publicize `_toolkit_registry_manager`

### LINKED ISSUE #493: GLOBAL_TOOLKIT_REGISTRY does not behave as anticipated
_Reported by @SimonBoothroyd_

**Describe the bug**
Changing `GLOBAL_TOOLKIT_REGISTRY` should allow the user to change the the cheminformatics toolkit that's used on the backend. However, this doesn't seem to work

**To Reproduce**
```
from openforcefield.utils.toolkits import ToolkitRegistry, RDKitToolkitWrapper, GLOBAL_TOOLKIT_REGISTRY
mol.generate_conformers()
print(mol.conformers[0][0])
```
> [ 0.81511319 -0.53833634  0.49281269] A

These are the coordinates of the first atom in the first conformer when OpenEye generates it
```
mol.generate_conformers(toolkit_registry=RDKitToolkitWrapper())
print(mol.conformers[0][0])
```
> [-0.74552317  0.04144446  0.01170607] A

These are the coordinates of the first atom of the first conformer when RDKit makes the conformer
```
GLOBAL_TOOLKIT_REGISTRY = ToolkitRegistry(toolkit_precedence=[RDKitToolkitWrapper])
#import openforcefield.topology.molecule
print(GLOBAL_TOOLKIT_REGISTRY.registered_toolkits)
```
> [<openforcefield.utils.toolkits.RDKitToolkitWrapper object at 0x118bc65f8>]

I am able to override the imported `GLOBAL_TOOLKIT_REGISTRY` to only have RDKit
```
mol.generate_conformers()
print(mol.conformers[0][0])
```
> [ 0.81511319 -0.53833634  0.49281269] A

But this function still generates the OpenEye conformer

**Additional information**

The GLOBAL_TOOLKIT_REGISTRY can be imported from a number of modules -- for example:
* `openforcefield.topology.molecule.GLOBAL_TOOLKIT_REGISTRY`
* `openforcefield.utils.GLOBAL_TOOLKIT_REGISTRY`
* `openforcefield.topology.molecule.openforcefield.utils.toolkits.GLOBAL_TOOLKIT_REGISTRY`

In my testing, changing any one of these doesn't appear to affect the others. 

I'm concerned that what's really happening here is that a new GLOBAL_TOOLKIT_REGISTRY is being imported for each module. I haven't worked with this pattern of globally-shared objects before, so I'm not sure how to proceed. 

@jaimergp suggested using the context manager approach, where we could use something like `with using_toolkit(RDKit): ...` to define blocks where certain toolkits or registries are used. This seems much smoother than tweaking a "GLOBAL" variable every time we want different defaults. I'm not sure how we'll choose to solve this issue, but this may be a convenient feature to add/test if this issue initiates a larger refactor.

mattwthompson:
Resolves #493 

- [x] Does not break API


openff/toolkit/utils/toolkit_registry.py
```diff
@@ -403,3 +403,6 @@ def _toolkit_registry_manager(toolkit_registry: Union[ToolkitRegistry, ToolkitWr
             GLOBAL_TOOLKIT_REGISTRY.deregister_toolkit(toolkit)
         for toolkit in original_toolkits:
             GLOBAL_TOOLKIT_REGISTRY.register_toolkit(toolkit)
+
+
+_toolkit_registry_manager = toolkit_registry_manager
```
mattwthompson, line 408: This seems like a simpler implementation than having a public context manager call a private context manager to do (literally) the same thing.

j-wags: Agreed!

j-wags:
Gonna update docstrings and check that they render, then I'll approve and merge!

openff/toolkit/_tests/test_toolkits.py
```diff
@@ -4529,6 +4529,50 @@ def test_call_raise_first_error(self):
...
+
+        with toolkit_registry_manager(openeye_registry):
+            assert "OpenEye" in str(GLOBAL_TOOLKIT_REGISTRY)
+            assert "RDKit" not in str(GLOBAL_TOOLKIT_REGISTRY)
+
+            create_ethanol().assign_partial_charges(
+                partial_charge_method="am1bccelf10",
+            )
+
+    def test_no_openeye_registry(self):
```
mattwthompson, line 4549: The best way to test this would probably involve a method that RDKit provides but OpenEye doesn't, but I couldn't think of any. I think the checks here should be enough to cover the bases.

mattwthompson:
Much appreciate the typo fixes 🛠️

j-wags (approved):
Wunderbar! Thanks @mattwthompson.

Issue #1706, opened 2023-09-01
Do not recommend `conda install`


mattwthompson:
**Is your feature request related to a problem? Please describe.**

Many points on the toolkit recommend `conda` instead of `mamba`, which is both a questionable recommendation and not in line with our [ecosystem recommendation](https://docs.openforcefield.org/en/latest/install.html).

**Describe the solution you'd like**

`%s/conda install /mamba install /g`
`%s/conda env /mamba env /g`

**Describe alternatives you've considered**

Dealing with users who still use `conda`

**Additional context**

PR #1707, opened 2023-09-01
Treat Mamba as preferred package manager in embedded documentation

mattwthompson:
#1706 

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -80,7 +80,7 @@ mamba activate <my_new_env>
 ```
 
 Installing into a new environment avoids forcing mamba to satisfy the dependencies of both the toolkit and all existing packages in that environment.
-Taking the approach that virtual environments are generally disposable, even ephemeral, minimizes the chances for hard-to-diagnose dependency issues.
+Taking the approach that conda/mamba environments are generally disposable, even ephemeral, minimizes the chances for hard-to-diagnose dependency issues.
 
 ## My mamba/conda installation of the toolkit STILL doesn't appear to work.
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -70,19 +70,19 @@ As of August 2022, some upstreams (at least AmberTools, possibly more) are not b
 
 (Keywords `osx-arm64`, M1 Mac, M2 Mac)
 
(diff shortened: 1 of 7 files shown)


docs/users/developing.md
```diff
@@ -335,9 +335,10 @@ These checks will use the most recent versions of each linter.
 To ensure that changes follow these standards, you can install and run these tools locally:
 
 ```shell
-conda install black isort -c conda-forge
+mamba install black isort flake8 -c conda-forge
```
j-wags, line 338: (not blocking) Do we also need flake8-bugbear?

j-wags (approved):
Looks great - One proposed change I feel somewhat strongly about. Once that's handled, feel free to merge!

FAQ.md
```diff
@@ -70,19 +70,19 @@ As of August 2022, some upstreams (at least AmberTools, possibly more) are not b
...
-conda create -n <my_new_env> -c conda-forge openff-toolkit
-conda activate <my_new_env>
+mamba create -n <my_new_env> -c conda-forge openff-toolkit
+mamba activate <my_new_env>
 ```
 
-Installing into a new environment avoids forcing conda to satisfy the dependencies of both the toolkit and all existing packages in that environment.
-Taking the approach that conda environments are generally disposable, even ephemeral, minimizes the chances for hard-to-diagnose dependency issues.
+Installing into a new environment avoids forcing mamba to satisfy the dependencies of both the toolkit and all existing packages in that environment.
+Taking the approach that virtual environments are generally disposable, even ephemeral, minimizes the chances for hard-to-diagnose dependency issues.
```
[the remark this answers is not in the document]
mattwthompson: "virtual environment" is the umbrella term for all solutions in this space, of which conda environments are subsets - I don't think it's synonymous with `venv`, `virtualenv`, `pyenv`, or any similar tools

docs/users/developing.md
mattwthompson: Strictly speaking, plugins should be installed here, but I don't think it's too important and I'd rather keep the pre-commit config as the sole source of truth. With the toolkit's config, there shouldn't be many cases in which the base package passes but a plugin throws an error

PR #1715, opened 2023-09-05
Run tests in random order

mattwthompson:


mattwthompson:
Huh, well that doesn't reproduce the error I expected but a different one.

Issue #1727, opened 2023-09-22
Negatively charged cysteines in the main chain are not supported.


pbuslaev:
**Describe the bug**
I tried to to generate topology from pdb with deprotonated, but not covalently connected to other cysteine, system. Such cysteines are often observed in protein systems, especially if they are close to ions. Toolkit gave me an error, saying that some bonds or atoms in the input could not be identified.

**To Reproduce**
I encountered the with `openff-toolkit=0.14.3` (clean installation into empty conda environment).

```python
from openff.toolkit import Molecule, Topology, ForceField
protein = Topology.from_pdb("./cyx_test.pdb")
```

**Output**
This is the error message I got:

```
---------------------------------------------------------------------------
UnassignedChemistryInPDBError             Traceback (most recent call last)
Cell In[10], line 1
----> 1 protein = Topology.from_pdb("/home/pbuslaev/projects/gromacs_fep_pipeline/scripts/test_runs/tmp_6/cyx_test.pdb")
      2 protein_intrcg = Interchange.from_smirnoff(force_field=ff14sb, topology=protein)

File /tmp_mnt/filer1/unix/pbuslaev/conda_env/openff_dev/lib/python3.10/site-packages/openff/utilities/utilities.py:80, in requires_package.<locals>.inner_decorator.<locals>.wrapper(*args, **kwargs)
     77 except Exception as e:
     78     raise e
---> 80 return function(*args, **kwargs)

File /tmp_mnt/filer1/unix/pbuslaev/conda_env/openff_dev/lib/python3.10/site-packages/openff/toolkit/topology/topology.py:1758, in Topology.from_pdb(cls, file_path, unique_molecules, toolkit_registry, _custom_substructures, _additional_substructures)
   1752 substructure_dictionary["ADDITIONAL_SUBSTRUCTURE_OVERLAP"] = {}
   1754 coords_angstrom = np.array(
   1755     [[*vec3.value_in_unit(openmm_unit.angstrom)] for vec3 in pdb.getPositions()]
   1756 )
-> 1758 topology = toolkit_registry.call(
   1759     "_polymer_openmm_pdbfile_to_offtop",
   1760     cls,
   1761     pdb,
   1762     substructure_dictionary,
   1763     coords_angstrom,
   1764     _custom_substructures,
   1765 )
   1767 for off_atom, atom in zip([*topology.atoms], pdb.topology.atoms()):
   1768     off_atom.metadata["residue_name"] = atom.residue.name

[...]

PR #1728, opened 2023-09-22
Update aa_residues_substructures_explicit_bond_orders_with_caps_expli…

pbuslaev:
This patch should solve #1727


j-wags (changes requested):
Thanks for starting on the fix for this! 

Patching that file will only solve the problem temporarily, since we periodically regenerate it from the chemical components dictionary using [this script](https://github.com/openforcefield/openff-toolkit/blob/c84bf62769b46e2f7f0e10b8ebb8fa2afe1ba575/utilities/make_substructure_dict/_cif_to_substructure_dict.py). Also, I'm seeing [CYS residues in our current dictionary with a negative sulfur](https://github.com/openforcefield/openff-toolkit/blob/88ce0b337e45d9ecee0cc012034b88d6fb561020/openff/toolkit/data/proteins/aa_residues_substructures_explicit_bond_orders_with_caps_explicit_connectivity.json#L1751). So either there's something wrong with a existing substructure in this file that's causing it not to match the deprotonated CYS in the input, or the mainchain substructure is fully missing. 

It'd be great if we could solve the problem by substituting an existing substructure in the [`_patch_known_problems` function](https://github.com/openforcefield/openff-toolkit/blob/88ce0b337e45d9ecee0cc012034b88d6fb561020/openff/toolkit/data/proteins/aa_residues_substructures_explicit_bond_orders_with_caps_explicit_connectivity.json#L1751), but barring that, just adding a new working substructure in the [`_add_common_substructures` function](https://github.com/openforcefield/openff-toolkit/blob/c84bf62769b46e2f7f0e10b8ebb8fa2afe1ba575/utilities/make_substructure_dict/_cif_to_substructure_dict.py#L683) would be fine.

pbuslaev:
I think that the changes I added should lead to generation of the correct `json` file. I have also updated the `json` file and fixed a typo in rdkit_wrapper docstring

j-wags:
Wow - Thanks for diving into our crazy substructure code! The resulting patterns look good visually but I tried loading the PDB file from your original post and it failed with:

```
E           openff.toolkit.utils.exceptions.UnassignedChemistryInPDBError: Some bonds or atoms in the input could not be identified.
E           
E           Hint: The following residues were assigned names that do not match the residue name in the input, or could not be assigned residue names at all. This may indicate that atoms are missing from the input or some other error. The OpenFF Toolkit requires all atoms, including hydrogens, to be explicit in the input to avoid ambiguities in protonation state or bond order:
E               Input residue  :CYS#0002 contains atoms matching substructures {'No match', 'PEPTIDE_BOND'}
E           
E           Error: The following 5 atoms exist in the input but could not be assigned chemical information from the substructure library:
E               Atom     9 (HA) in residue  :CYS#0002
E               Atom    10 (CB) in residue  :CYS#0002
E               Atom    11 (HB2) in residue  :CYS#0002
E               Atom    12 (HB3) in residue  :CYS#0002
E               Atom    13 (SG) in residue  :CYS#0002

openff/toolkit/utils/rdkit_wrapper.py:818: UnassignedChemistryInPDBError

```

I tried shaking a few permutations of things like residue names, but to no avail. Are you able to load the original PDB successfully? 

In either case, I have an example file and test we can hammer on, but I don't have write access to your fork to send it over. Would you be OK if I merged this into a branch so OpenEye CI stops complaining, and then gave you write access?

pbuslaev:
> Wow - Thanks for diving into our crazy substructure code! The resulting patterns look good visually but I tried loading the PDB file from your original post and it failed with:
> 
> ```
> E           openff.toolkit.utils.exceptions.UnassignedChemistryInPDBError: Some bonds or atoms in the input could not be identified.
> E           
> E           Hint: The following residues were assigned names that do not match the residue name in the input, or could not be assigned residue names at all. This may indicate that atoms are missing from the input or some other error. The OpenFF Toolkit requires all atoms, including hydrogens, to be explicit in the input to avoid ambiguities in protonation state or bond order:
> E               Input residue  :CYS#0002 contains atoms matching substructures {'No match', 'PEPTIDE_BOND'}
> E           
> E           Error: The following 5 atoms exist in the input but could not be assigned chemical information from the substructure library:
> E               Atom     9 (HA) in residue  :CYS#0002
> E               Atom    10 (CB) in residue  :CYS#0002
> E               Atom    11 (HB2) in residue  :CYS#0002
> E               Atom    12 (HB3) in residue  :CYS#0002
> E               Atom    13 (SG) in residue  :CYS#0002
> 
> openff/toolkit/utils/rdkit_wrapper.py:818: UnassignedChemistryInPDBError
> ```
> 
> I tried shaking a few permutations of things like residue names, but to no avail. Are you able to load the original PDB successfully?

This is strange. With updated `json` substructure file everything works perfectly for me. I do see the exact same SMARTS pattern as I added manually in the generated substructure file, so I am a bit confused why you are getting the error.
 
> In either case, I have an example file and test we can hammer on, but I don't have write access to your fork to send it over. Would you be OK if I merged this into a branch so OpenEye CI stops complaining, and then gave you write access?

Yes, this is fine with me.

PR #1754, opened 2023-10-23
Pin compatible Interchange

mattwthompson:
I'm reverting a few changes that broke the examples, but it may take a couple days to propagate out. Until then, I'm going to stop the bleeding and prevent interim frustration from the (examples) CI failing for unrelated reasons.

### DIFF
--- a/.github/workflows/beta_rc.yaml
+++ b/.github/workflows/beta_rc.yaml
@@ -4,12 +4,6 @@ on:
   schedule:
     - cron: "0 0 * * 0"
   workflow_dispatch:
-  push:
-    branches:
-      - "main"
-  pull_request:
-    branches:
-      - "main"
 
 defaults:
   run:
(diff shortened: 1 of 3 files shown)


mattwthompson:
The only change relates to the examples automation, so I'm merging before everything else passes (which it should).

I will revert this pin when I can, either this week or next

mattwthompson:
I'm also reverting some dev-only code I forgot to take out in #1753

j-wags (approved):
Excellent - I see this is still a draft but I trust your judgement on when this is good to merge. No need for releasenotes or feedstock changes IMO. Excited to see green CI again!

mattwthompson:
Indeed, just wanted to wait for CI to catch up before hitting the Cape Canaveral button

mattwthompson:
Thanks for the review!

mattwthompson:
The CI failures are somewhat reliable, but reliable due to GROMACS builds being incompatible with the options provided by runners

https://github.com/conda-forge/gromacs-feedstock/issues/23

PR #1755, opened 2023-10-31
Test on Python 3.12 in (non-examples) CI

mattwthompson:


mattwthompson:
Everything seems to be working fine here. The only change I want to revert is using the OpenMM RC channel. It looks like that release will be 2-3 weeks out, so I'll leave this until then.

mattwthompson:
A quick update after OpenMM 8.1:

It looks like the vanilla CI is working fine, top to bottom. I guess all of our upstreams are built now.

Examples are failing for two separate reasons:
* MDTraj does not have macOS builds on Python 3.12 (but it does have them on Linux). This is a known issue that will probably get fixed in the next release, unknown timeline.
* The RDKit-only examples are pinned to use AmberTools 22 since 23 brings in ParmEd 4, which breaks some of the examples for reasons we haven't gone into yet.

Each of these will get fixed eventually (I don't see either as urgent, and am happy to see somebody else jump the line). I'm also fine with merging this with 3.12 in the CI, but not examples workflows.

mattwthompson:
I'm just not doing examples for now

Issue #1464, opened 2022-11-14
Molecule.enumerate_tautomers() can return empty list


jchodera:
**Describe the bug**
`Molecule.enumerate_tautomers()` can return an empty list---rather than the single tautomer or original molecule---if tautomer enumeration fails.

**To Reproduce**
```python
>>> print(openff.toolkit.__version__)
0.11.3
>>> from openff.toolkit.topology import Molecule
>>> offmol = Molecule.from_smiles('[H]C1(C(C(C(C(C1([H])[H])([H])[H])([H])[C@]([H])(C([H])([H])[H])N([H])[H])([H])[H])([H])[H])[H]')
>>> offmol.enumerate_tautomers()
[]
```

**Computing environment (please complete the following information):**
 - Operating system: `linux64`
 - Output of running `conda list`
 <details>
 # packages in environment at /lila/home/chodera/miniconda/envs/perses-dev:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
alabaster                 0.7.12                     py_0    conda-forge
alsa-lib                  1.2.3.2              h166bdaf_0    conda-forge
amberlite                 22.0                     pypi_0    pypi
ambertools                22.0            py310h206695f_3    conda-forge
amberutils                21.0                     pypi_0    pypi
anyio                     3.6.2              pyhd8ed1ab_0    conda-forge
argon2-cffi               21.3.0             pyhd8ed1ab_0    conda-forge
argon2-cffi-bindings      21.2.0                   pypi_0    pypi
arpack                    3.7.0                hdefa2d7_2    conda-forge
arsenic                   0.2.1           py310hff52083_0    conda-forge
asttokens                 2.1.0              pyhd8ed1ab_0    conda-forge
astunparse                1.6.3              pyhd8ed1ab_0    conda-forge
attrs                     22.1.0             pyh71513ae_1    conda-forge
autograd                  1.5                pyhd8ed1ab_0    conda-forge
babel                     2.11.0             pyhd8ed1ab_0    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
backports                 1.1                pyhd3eb1b0_0  
backports.functools_lru_cache 1.6.4              pyhd8ed1ab_0    conda-forge
beautifulsoup4            4.11.1             pyha770c72_0    conda-forge
blas                      1.1                    openblas    conda-forge
[...]

PR #1779, opened 2023-11-28
Overhaul protomer enumeration

mattwthompson:
Suggestion #<space>1 in https://github.com/openforcefield/openff-toolkit/issues/1464#issuecomment-1328660311

  - [x] Test a molecule which is itself not a strong protomer is **not** returned
  - [x] Test a molecule which is a sensible protomer is returned
  - [x] Test default `max_states` finds as many states as it can
  - [x] Test too-large `max_states` just returns whatever OpenEye finds
  - [x] Test that `0 < max_states < len(openeye_protomers)` returns `max_states` number of molecules
  - [x] Test some large molecule returns a bunch of protomers
  - [x] Document that input molecule cannot be counted on as returned


openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -826,8 +828,8 @@ def enumerate_protomers(
 
         Returns
         -------
-        molecules: List[openff.toolkit.topology.Molecule],
-            A list of the protomers of the input molecules not including the input.
+        molecules: list[openff.toolkit.Molecule],
+            A list of the protomers of the input molecules including the input.
```
Yoshanuikabundi, line 832:
```suggestion
            A list of the protomers of the input molecules, including the input molecule itself unless it is pruned by the ``max_states`` argument.
```

openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -840,13 +842,14 @@ def enumerate_protomers(
```
Yoshanuikabundi, line 842: I wish GitHub would let you leave suggestions anywhere.

In line 835/837:

```diff
- # add one as the input is included
- options.SetMaxCount(max_states + 1)
+ options.SetMaxCount(max_states)
```

The original behavior was to ask for n+1 protomers, and then throw one away to end up with n. Since we now return the original, we just want to ask for n. Incidentally this means the original behavior was subtly wrong, as if OpenEye did not return the input molecule we would give n+1 molecules back. It also meant there was no way to ask for all protomers (`SetMaxCount(0)`).

openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -811,8 +811,10 @@ def _assign_aromaticity_and_stereo_from_3d(self, offmol):
         return offmol_w_stereo_and_aro
 
     def enumerate_protomers(
-        self, molecule: "Molecule", max_states: int = 10
-    ) -> List["Molecule"]:
+        self,
+        molecule: "Molecule",
+        max_states: int = 10,
```
Yoshanuikabundi, line 816: I think with this change max_states should be 0 by default, or at least a number much bigger than 10.

```suggestion
        max_states: int = 0,
```

I bet if we leave it at 10, this issue comes up again and again from people who don't know about max_states (like me until 10 minutes ago). OE seems to be able to put thousands of protomers together in only a few seconds:

```python
>>> %%timeit
... 
... molecule = Molecule.from_smiles("C(C(=O)O)(C(=O)O)C(C(=O)O)(C(=O)O)C(C(=O)O)(C(=O)O)C(C(=O)O)(C(=O)O)C(C(=O)O)(C(=O)O)")
... 
... 
... options = oequacpac.OEFormalChargeOptions()
... options.SetMaxCount(0)
... 
... molecules = []
... 
... oemol = molecule.to_openeye()
... for protomer in oequacpac.OEEnumerateFormalCharges(oemol, options):
...     molecules.append(
...         Molecule.from_openeye(
...             protomer,
...             allow_undefined_stereo=True,
...         )
...     )
2.45 s ± 47.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
>>> len(molecules)
6400
```

docs/releasehistory.md
```diff
@@ -12,6 +12,8 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ### Behavior changes
 
+- [PR #17XX](https://github.com/openforcefield/openff-toolkit/pull/17XX): `Molecule.enumerate_protomers` now includes the input molecule in the returned list.
```
Yoshanuikabundi, line 15:
```suggestion
- [PR #1779](https://github.com/openforcefield/openff-toolkit/pull/1779): `Molecule.enumerate_protomers` now includes the input molecule in the returned list.
```

openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -811,8 +811,10 @@ def _assign_aromaticity_and_stereo_from_3d(self, offmol):
...
+    ) -> list["Molecule"]:
         """
         Enumerate the formal charges of a molecule to generate different protomoers.
```
Yoshanuikabundi, line 819:
```suggestion
        Enumerate the formal charges of a molecule to generate different protomers.
```

Yoshanuikabundi (changes requested):
This is great! I think this is a slightly more intuitive behavior.

Blocking: Works great for molecules with a small number of protomers, but I found that most of the tetracarboxylic acids I tried did not return the original molecule (including the only [apparently real](https://en.wikipedia.org/wiki/Ethylenetetracarboxylic_acid) tetracarboxylic acid I could find in a quick search - SMILES `"C(C(=O)O)(C(=O)O)=C(C(=O)O)(C(=O)O)"`).

```python
>>> molecule = Molecule.from_smiles("C(C(=O)O)(C(=O)O)=C(C(=O)O)(C(=O)O)")
>>> protomers = molecule.enumerate_protomers()
>>> molecule in protomers
False
```

I tracked this down to being because we by default only ask for the first 11 protomers:

```python
>>> protomers2 = molecule.enumerate_protomers(max_states=999)
>>> molecule in protomers2
True
```

I think this means that we can't claim comprehensively in the docstring that the input is or isn't included. I can think of a few ways to solve this:

1. Ask for all protomers by default (this assumes OpenEye doesn't have some built in limit)
2. Inject the original molecule into the returned protomers if it's missing
3. Don't claim to always return the input molecule

I think 3 is the best way forward, 1 may be worth doing (11 is certainly too low a default), and 2 is probably a waste of time.

Apart from that, I just found a few typos in the original docstrings and noticed that `SetMaxCount` no longer needs a +1 - those changes are in suggestions below.

Non-blocking: We may also want a test with more protomers than `max_states`.

Yoshanuikabundi:
The notebook I was messing around in: [protomers.zip](https://github.com/openforcefield/openff-toolkit/files/13726981/protomers.zip)

openff/toolkit/utils/openeye_wrapper.py
j-wags: I agree with this - The max states should be set to 0 (unlimited) by default. I'm a bit surprised that limiting this is an option at all (if we hit the max, which ones get returned? Users are guaranteed to be surprised) but I'd rather keep this PR at "behavior change" and not "API break", so just changing the default here should be good.  Please also record this behavior change in the changelog.

mattwthompson:
Could you elaborate on why 2 is a waste of time? It's what I intuited the objective to be here. Is the issue that it might return a different number of states if the input molecule happens to be in what was returned by the underlying toolkit?

Yoshanuikabundi:
I guess the issue is that if the input molecule isn't an important protomer, then when the user asks for the top n protomers and we give them those n plus the original molecule that might be a surprise (just like if they ask for the top n and we give them the top n minus the original). We need to decide what the desired behaviour is I guess.

j-wags:
Talked a bit at my check-in with @mattwthompson about this. To move this forward, let's do the following:
* Set the default max_states to 0
* Not manually re-add the input to the list of outputs
* Change the docstring to say that, if max_states is limited (set to something other than 0) the input molecule may not be returned

openff/toolkit/_tests/molecule/test_state_enumeration.py
```diff
@@ -0,0 +1,340 @@
...
+    def test_tetracarboxylic_acid(self):
+        acid = Molecule.from_smiles("C(C(=O)O)(C(=O)O)=C(C(=O)O)(C(=O)O)")
+
+        protomers = acid.enumerate_protomers()
+
+        assert len(protomers) == 16
+        assert acid in protomers
+
+    @requires_openeye
+    def test_bad_state_not_necessarily_in_output(self):
```
j-wags, line 119: 👍 This is a clever test

openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -812,43 +812,46 @@ def _assign_aromaticity_and_stereo_from_3d(self, offmol):
...
-        molecule: openff.toolkit.topology.Molecule
+        molecule
             The molecule whose state we should enumerate
 
-        max_states: int optional, default=10,
-            The maximum number of protomer states to be returned.
+        max_states
+            The maximum number of protomer states to be returned. If 0, the default,
+            attempt to return all protomers. If set to a non-zero number, the input molecule
+            is not guaranteed to be included in the returned list.
```
j-wags, line 830: Even if set to 0, the input isn't guaranteed to be returned (Ex. `NH2-`)
```suggestion
        max_states
            The maximum number of protomer states to be returned. If 0, the default,
            attempt to return all protomers. 
```

openff/toolkit/utils/openeye_wrapper.py
```diff
@@ -812,43 +812,46 @@ def _assign_aromaticity_and_stereo_from_3d(self, offmol):
...
 
     def enumerate_protomers(
-        self, molecule: "Molecule", max_states: int = 10
+        self,
+        molecule: "Molecule",
+        max_states: int = 0,
     ) -> list["Molecule"]:
         """
-        Enumerate the formal charges of a molecule to generate different protomoers.
+        Enumerate the formal charges of a molecule to generate different protomers.
```
j-wags, line 820:
```suggestion
        Enumerate the formal charges of a molecule to generate different protomers.
        Note that, in cases where the input molecule has an uncommon protonation state 
        (for example ``[NH2-]``), the input molecule may not be included in the output. 
```

docs/releasehistory.md
```diff
@@ -13,6 +13,10 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ### Behavior changes
 
+- [PR #1779](https://github.com/openforcefield/openff-toolkit/pull/17XX): Overhauls `Molecule.enumerate_protomers`:
```
j-wags, line 16:
```suggestion
- [PR #1779](https://github.com/openforcefield/openff-toolkit/pull/1779): Overhauls `Molecule.enumerate_protomers`:
```

j-wags (approved):
Excellent - Thanks @mattwthompson!

PR #1784, opened 2023-12-06
Improve `Topology` compatibility with `_SimpleMolecule`

mattwthompson:

### DIFF
--- a/openff/toolkit/topology/topology.py
+++ b/openff/toolkit/topology/topology.py
@@ -80,6 +80,9 @@ if TYPE_CHECKING:
     from openff.toolkit.utils import ToolkitRegistry, ToolkitWrapper
 
 TKR: TypeAlias = Union["ToolkitRegistry", "ToolkitWrapper"]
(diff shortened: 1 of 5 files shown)


openff/toolkit/topology/topology.py
```diff
@@ -1204,9 +1210,22 @@ def _invalidate_cached_properties(self):
...
+
+        self._aromaticity_model = other.aromaticity_model
+
+        for pair, distance in other.constrained_atom_pairs.items():
+            self.add_constraint(pair, distance)
+
+        self.box_vectors = other.box_vectors
+
+        for other_molecule in other.molecules:
+            self._add_molecule_keep_cache(other_molecule.__class__(other_molecule))  # type: ignore[call-arg]
```
j-wags, line 1226: 👍 I broadly agree with this approach for this PR. In the long run we'll want to think through a more robust way to enable serialization of Topologies with SimpleMolecules, but in the short run this is the best way forward.

j-wags:
Unless it's really important to getting things working, I'd prefer for this PR to not be the place where we introduce a hard dep on pydantic.

PR #1786, opened 2023-12-07
Jupyter widget docs

Yoshanuikabundi:
As [requested](https://github.com/openforcefield/openff-toolkit/pull/1732#issuecomment-1766753143) by @mattwthompson, I have bolstered the developers guide with information about how embedding notebooks in Sphinx documentation works, including the tricks needed to embed widgets.

This PR also embeds widget state in the molecule cookbook, which apparently hasn't had working NGLView embeds since 0.13 or so, and updates the SciPy intersphinx URL to silence a warning.


docs/users/developing.md
```diff
@@ -388,6 +388,46 @@ $ open htmlcov/index.html
...
+```
+
+Note that some features that will render correctly in Sphinx rely on extensions that are not available to `jupyterlab-myst`, and so may still appear incorrect even with the extension.
+
+[MyST-NB]: https://myst-nb.readthedocs.io/
+[`jupyterlab-myst`]: https://mystmd.org/guide/quickstart-jupyter-lab-myst
+
+#### Widgets
+
+Jupyter Notebooks can produce interactive [widgets] as output, and MyST-NB supports rendering them in HTML output. Any given widget includes both Python code and JavaScript code, and Jupyter defines ways for them to speak to each other which allows Python code to be run interactively via the widget in the notebook.
```
mattwthompson, line 419: I don't buy the logical structure of this paragraph - the second sentence introduces Python <-> JavaScript communication, but I don't see how that enables Python <-> Jupyter communication "which allows Python code to be run interactively via the widget in the notebook" as I thought that works with vanilla Python + Jupyter. Maybe I mis-understanding the contents of a widget or where the JavaScript lies?

docs/users/developing.md
```diff
@@ -388,6 +388,46 @@ $ open htmlcov/index.html
...
+
+Secondly, widget initialization happens in Python, so any state shared between Python and JavaScript must be included in the notebook itself before Sphinx can render it. Any time these notebooks are re-executed, the internal IDs of the widgets may change, causing the widget to forget its original state and be missing from the rendered documentation. To fix this, it's essential to ask Jupyter to **save widget state**. In Jupyter Notebook 7+ and Jupyter Lab, this option appears under `Settings -> Save Widget State Automatically`, which if ticked will save widget state when the notebook is saved. Note that this menu item will not appear if the open notebook has no widgets. In earlier versions of Notebook, the `Widgets -> Save Widget State` menu item must be selected any time the notebook is re-executed and saved.
```
mattwthompson, line 423: Not blocking: is there a way to do this with `jupyter` CLI? Same question for VS code. Both questions are because I'm lazy

docs/users/developing.md
```diff
@@ -388,6 +388,46 @@ $ open htmlcov/index.html
...
+
+Widget state can also be saved programmatically by executing the notebook with [NBConvert], though since NBConvert does not include a JavaScript runtime the widget must be able to initialize itself from stored state. This is not possible, for example, in NGLView before version 3.0.6.
```
mattwthompson, line 425: Well, there's the answer

mattwthompson (changes requested):
Perfect, this is exactly what I was after. I have a couple of non-blocking questions for my own curiosity.

The only thing I would like to see changed is removing the "The value of the smallest subnormal for ..." warnings from the saved notebook file. I'm not sure how they slipped past the filter - the regex looks like it matches and I don't see any change that should have caused this.

j-wags:
@mattwthompson - I'd be happy to have you as primary reviewer on this on (to review+merge as you see fit.) Any objection?

mattwthompson:
Nope, I'll take it

docs/users/developing.md
Yoshanuikabundi: CLI is addressed in the next paragraph as you noted, but as for Code - I have no idea, sorry!

docs/users/developing.md
Yoshanuikabundi: Ha!

docs/users/developing.md
Yoshanuikabundi: I'm confused about your confusion :/ I'm not really sure what distinction you're drawing between Jupyter and JavaScript. I think by Jupyter you mean, like, how its rendered in the notebook - but as far as I know Jupyter just says "OK put some JS and HTML here" and all the rendering happens in JS. The JS just sorta renders directly to the browser, it doesn't really communicate with Jupyter at all beyond being told "here's where you should put stuff". Python can do something similar by writing HTML output I think, which is rendered richly by Jupyter, but Widgets specifically go further and incorporate JS to provide interactivity, and in Widgets specifically its JS that does the rendering.

Here, I'll describe at a lower level in case one of us is misunderstanding something. I've contributed to NGLView on both the Python and JS side, so I think I have a working understanding, but it still feels very much like black magic:

Widgets are usually created in Python by subclassing `ipywidgets.Widget`, which provides Python and JS APIs to the `jupyterlab-widgets` and `widgetsnbextension` Jupyter extensions that are distributed with `ipywidgets`. The resulting `Widget` instance is synchronized by the extension with a widget model class (`WidgetModel`), written in JS. So if you change one of its special "traitlet" attributes in Python, the extension updates the same attribute in JS, and vice versa (assuming the traitlet has sync enabled). That's all a high level API over a message sending thingy that some widgets also use, but the general idea is that the widget and the model have the same data and you can define callbacks in either language to react to changes in that data. And then the model is represented in the notebook by a view (`WidgetView`), also written in JS, which does the rendering.

So rendering and interactivity is all written in JS, and that updates the shared data, and then Python is free to react to that (possibly changing the state). The result is that you can generally write responses to interactivity in either language, as long as you have both runtimes... running. And then `ipywidgets` includes some widgets that compose the others so it *looks* like you can basically write everything in Python... until you render everything to HTML and half your widgets break and you run off on a surprisingly successful wild goose chase to figure all this out.

Details: https://ipywidgets.readthedocs.io/en/stable/examples/Widget%20Low%20Level.html

I think my original paragraph accurately summarizes this, but it would be great to know how to make this clearer or where I've made a mistake!

mattwthompson: I'm not sure where my confusion was, but this all makes enough sense to me that it's well worth including as-is. Thanks for the explainer!

mattwthompson (approved):
Please remove the warnings in one of the cell outputs - otherwise I think this is good to go

PR #1787, opened 2023-12-11
Improve Toolkit Wrapper and Registry API docs

Yoshanuikabundi:
Updates some out-of-date docs tooling and fixes warnings so that the next time things break, we get a red check mark! Also adds `NAGLToolkitWrapper` to the docs and to the `toolkits` module, and breaks the Toolkit API reference out of the Utilities section so it can be discovered more easily.


Yoshanuikabundi:
I think remaining CI failures are fixed in #1786

docs/api/toolkits.md
```diff
@@ -0,0 +1,112 @@
...
+The toolkit wrappers provide a simple uniform API for accessing minimal functionality of cheminformatics toolkits.
+
+These toolkit wrappers are generally used through a {class}`ToolkitRegistry`, which can be constructed with a desired precedence of toolkits:
+
+```python
+>>> from openff.toolkit.utils.toolkits import ToolkitRegistry, OpenEyeToolkitWrapper, RDKitToolkitWrapper, AmberToolsToolkitWrapper
+>>> toolkit_registry = ToolkitRegistry()
+>>> toolkit_precedence = [OpenEyeToolkitWrapper, RDKitToolkitWrapper, AmberToolsToolkitWrapper]
+>>> [ toolkit_registry.register_toolkit(toolkit) for toolkit in toolkit_precedence if toolkit.is_available() ]
+[None, None, None]
```
mattwthompson, line 12: Why not

```python
toolkit_registry = ToolkitRegistry(toolkit_precedence = [OpenEyeToolkitWrapper, RDKitToolkitWrapper, AmberToolsToolkitWrapper])
```

This block implies to me that users are responsible for registering their own toolkits. I don't know if this is the case, and I assume/hope that's automatically handled when you make a `ToolkitRegistry` in one go.

I might be useful to append

```
>>> toolkit_registry
<ToolkitRegistry containing OpenEye Toolkit, The RDKit, AmberTools>
```
just to demonstrate what's in the thing

docs/api/toolkits.md
```diff
@@ -0,0 +1,112 @@
...
+>>> from openff.toolkit.utils.toolkits import GLOBAL_TOOLKIT_REGISTRY as toolkit_registry
+>>> from openff.toolkit import Molecule
+>>> molecule = Molecule.from_smiles('Cc1ccccc1')
+>>> smiles = toolkit_registry.call('to_smiles', molecule)
+
+```
+
+The order of toolkits, as specified in `toolkit_precedence` above, determines the order in which
+the called method is resolved, i.e. if the toolkit with highest precedence has a `to_smiles`
+method, that is the toolkit that will be called. If the toolkit with highest precedence does not
```
mattwthompson, line 28: Text wrapping here is inconsistent

docs/api/toolkits.md
```diff
@@ -0,0 +1,112 @@
...
+have such a method, it is attempted with other toolkits until one is found. By default, if a toolkit with an appropriately-named method raises an exception of any type, then iteration over the registered toolkits stops and that exception is raised. To continue iteration if specific exceptions are encountered, customize this behavior using the optional `raise_exception_types` keyword argument to `ToolkitRegistry.call`. If no registered
+toolkits have the method, a ValueError is raised, containing a message listing the registered toolkits and exceptions (if any) that were ignored.
```
mattwthompson, line 30: I had a very long comment written that got eaten by my browser ... I'll try to summarize it.

I suggest this be restructure a little bit, drawing on what is already here. I'd like to see something that communicates, mostly in this order

* How a registry selects wrappers to try
* What happens when the wrapper runs a method without error
* What happens when the (first attempted) wrapper throws an error
* How to raise specific errors but skip others
* What happens when no wrappers have a matching method name
* Example showing default behavior of something working
* Example showing default behavior of a wrapper throwing an error
* Example showing how `raise_exception_types` can be used to skip one type of error but throw another one

This is fairly long, but I think this paragraph currently tries to convey much more information that can fit into one section of text. I find this behavior of registries and wrappers pretty confusing [^1]- my suggestion here is not a criticism of the prose as-is but my attempt to project backwards in time what I wish existed in the docs three years ago.

> To continue iteration if specific exceptions are encountered, customize this behavior using the optional `raise_exception_types` keyword argument to `ToolkitRegistry.call`.

If I understand the point of this Markdown file, this detail needs to be expanded on (even if the source code already exists as a source of truth). I had to look at the docstrings to remember how I would use this argument to do that.

[^1]: For example, `raise_exception_types` is meant to be of type `list[Exception]` but its default value is `None`. In that case, it's transformed into `list[Exception]`, and in English that translates to something like "raise only these specific exception types, but oh by the way if you don't pick any I will assume you meant 'all of them'."

docs/api/toolkits.md
```diff
@@ -0,0 +1,112 @@
...
+>>> print(len(toolkit_registry.registered_toolkits))
+3
+>>> toolkit_registry.register_toolkit(RDKitToolkitWrapper)
+>>> print(len(toolkit_registry.registered_toolkits))
+4
+
+```
+
+For example, differences in `to_smiles` functionality between OpenEye toolkits and The RDKit can
+be explored by selecting which toolkit(s) are and are not registered.
```
mattwthompson, line 75: If a user want to do this, I'd suggest they use the context manager, not this approach. I see this as a useful example for demonstrating the mechanics of the registry, but could we tack on at the end a sort of "by the way, here's another way to do the same thing" section? I think it would look like

```python
>>> # imports
>>> molecule = Molecule.from_smiles('Cc1ccccc1')
>>> with toolkit_registry_manager(ToolkitRegistry([OpenEyeToolkitWrapper()]):
...     print(molecule.to_smiles())
[H]c1c(c(c(c(c1[H])[H])C([H])([H])[H])[H])[H]
>>> with toolkit_registry_manager(ToolkitRegistry([RDKitToolkitWrapper()]):
...     print(molecule.to_smiles())
[H][c]1[c]([H])[c]([H])[c]([C]([H])([H])[H])[c]([H])[c]1[H]
```

I may be biased here - I am lazy and I like being able to get this in one go with the context manager.

mattwthompson (commented):
This is awesome work, the sort of stuff I wish was written years ago but didn't have the motivation to do myself.

I haven't read everything yet - I am happy to back later to finish a review @j-wags

j-wags:
> I haven't read everything yet - I am happy to back later to finish a review @j-wags

Yes @mattwthompson - Happy to have you take this one on. Please go ahead and review+merge when approved. Thanks!

openff/toolkit/utils/nagl_wrapper.py
```diff
@@ -17,6 +17,12 @@
 
 
 class NAGLToolkitWrapper(ToolkitWrapper):
+    """NAGL toolkit wrapper for applying partial charges with a GCN model.
```
mattwthompson, line 20: I'm marginally concerned that the prose uses "NAGL" and "OpenFF NAGL" synonymously. I don't think now is the time to advocate for changing the name of the class ... but it'd be nice to consistently use "OpenFF NAGL" in text where practical

mattwthompson (commented):
I'm pretty much rubber-stamping the tooling changes and trusting they look okay, but I can also go through particular parts of the rendered site if that would be useful.

The changes to the content of the docs will take a few iterations to finish, I predict, so if you want the tooling changes in sooner, you could split those out into a separate PR and get those in quickly. (That's a bit of work, I recognize, so I'm not pushing for it.)

Yoshanuikabundi:
> This is awesome work, the sort of stuff I wish was written years ago but didn't have the motivation to do myself.

I would love to take credit, but `toolkits.md` is all cut and pasted from `utilities.rst` - it's existing documentation that I've just moved to be a bit more discoverable. I can see that it needs some editing, but should we leave that for a future PR?

mattwthompson:
Okay, so then the changes in this PR are

1. Moving that file around as you describe
2. Lots of (waves hands, ghostly noises appear out of the background, smoke machine revs up, lighting rains down from the sky) changes to make Sphinx/readthedocs happy
3. Adding a docstring into the NAGL wrapper

Even though it's not part of what you originally wanted, I would like to see a couple of changes to that `toolkits.md` file. What's the lowest-friction way of moving this forward? I don't want to hang up the other changes on the basis of wanting to refresh existing docs, and it might be worth splitting this out into 2-3 separate PRs even though that incurs some cost.

Yoshanuikabundi:
OK I've broken this PR up into:

1. #1795, which should be simple to review
2. #1796, which has a hellish change list but is mostly just converting ReST links to MarkDown links
3. This PR, which now only breaks the Toolkit Wrapper docs into their own page (to preserve existing review comments)

I also rebased this PR on #1795 to avoid future merge conflicts, so #1795 needs to be merged before this one (because it's gonna be a while before I get to refreshing these docs). Accordingly, I've marked this as a draft. #1796 is now independent.

Fortunately I seem to have actually had my head on straight when I was making commits and this was actually very doable!

docs/api/toolkits.md
Yoshanuikabundi: Interesting - Is there a good reason I'm not familiar with not to just use the `toolkit_registry` argument?

Yoshanuikabundi:
Hey @mattwthompson, finally got around to this, I've implemented everything you suggested I think but have also moved stuff around to be less overwhelming at first glance. A lot more detailed info is in docstrings now. I also couldn't think of a way to skip only one type of error, so the example for that *throws* only one kind of error and skips the rest. That might have been what you meant anyway now that I think about it!

docs/api/toolkits.md
mattwthompson: For one-liners it's just style

For larger sequences of logic (imagine setting up conformers and later calling out to some cornucopia of `openmmforcefields` and NAGL), it's still a question of style but less ambiguous


```
with toolkit_registry_manager(no_openeye_registry):
    # do thing
    # do other thing
    # oh no there's more
    # oh god it doesn't stop
    # but wait there's more
```

and I'd rather not wire that through on each line

mattwthompson (approved):
I think I got lost in the sauce of my own over-reviewing months back. I took a read through everything (in the rendered docs) that changed and liked it all.

PR #1782, opened 2023-11-30
Assorted typing improvements

mattwthompson:
While working on updating the state enumerators, I noticed a lot of `molecule.py` had some missing annotations, historic shims, or quirks that could be updated. The changes here are  as follows:

* Use built-in types were possible. Previously i.e. `from typing import Dict` was needed to support Python <3.9 (although `from __future__ import annotations` was another option). I didn't change all of them, but I was able to drop some of the `typing` imports changing the common types.
* Add `warn_unused_ignores = True` into the mypy configuration. It does just what it describes - throws a warning (from the type checker only) when a `# type: ignore` is not needed. I like this option.
* Add many missing annotations to `molecule.py`.

This started by trying to sneak some changes into those PRs, but I realized quickly that wasn't the place (common code across several different changesets) and it'd be better to just consolidate everything in one go.

- [ ] Tag issue being addressed
- [ ] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/main/openff/toolkit/_tests)
- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/main/docs), if applicable
- [x] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [ ] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/main/docs/releasehistory.rst)

PR #1788, opened 2023-12-14
Remove obsolete comments to `OrderedDict`

mattwthompson:
One detail that came up in #1782 is that `OrderedDict` is still made reference to in a couple locations. I removed most of them in  #1053 but some comments were missed.


j-wags (approved):
LGTM, thanks! No releasenotes needed on this one.

PR #1796, opened 2023-12-20
Fix all sphinx warnings

Yoshanuikabundi:
This PR fixes all existing Sphinx warnings, in part by updating to Sphinx 6. It also sets RTD to fail a docs build on a warning. I'm hoping this will stop a recurring issue where incorrect ReStructuredText gets into releases without anyone noticing (because ReST is impossible to write correctly without a detailed dialogue between the parser and whitespace)

One thing this needed is to convert the remaining RST in the changelog to MarkDown. This uncovered a bunch of broken links and also renders better as there no longer needs to be a distinction between old releases and new releases.


docs/users/developing.md
```diff
@@ -414,4 +414,4 @@ The CI matrix is currently as follows:
 +-----------------------+------------+-----------+-------------+------------+-----------+-------------+
 | Python 3.12 and newer | Pending official releases and upstream support                              | 
 +-----------------------+------------+-----------+-------------+------------+-----------+-------------+
```
mattwthompson, line 416: You are welcome to update this table so that 3.12 looks like 3.11 and this line applies to 3.13+, but that's not needed to get this merged

mattwthompson (approved):
I read or skimmed ever file in the changeset, making an effort to sample some links and actually look at rendered webpages along the way. I found only one thing that stands out as needing fixing - no need for a re-review!

The end of the release history page looks funky:
image

PR #1797, opened 2023-12-20
Support `io.StringIO` in `Topology.from_pdb()`

Yoshanuikabundi:
In `Topology.from_pdb`, we check if the file is a file-like object by calling `isinstance(file, io.TextIOWrapper)`. This works on file objects created with `open()` in text mode, but not with `io.StringIO` objects, which are very useful when you want to work entirely in memory. The abstract base class `io.TextIOBase` supplies the necessary API for OpenMM to be able to read a PDB file, and works with both:

```python
>>> import io
>>> file = io.StringIO()
>>> isinstance(file, io.TextIOWrapper)
False
>>> isinstance(file, io.TextIOBase)
True
>>> file = open("README.md", 'r')
>>> isinstance(file, io.TextIOWrapper)
True
>>> isinstance(file, io.TextIOBase)
True
```


mattwthompson (approved):

openff/toolkit/_tests/test_topology.py
```diff
@@ -780,6 +781,10 @@ def test_from_pdb_input_types(self):
         with open(protein_path) as f:
             Topology.from_pdb(f)
 
+        pdb_string = pathlib.Path(protein_path).read_text()
+        with StringIO(pdb_string) as f:
+            Topology.from_pdb(f)
+
```
mattwthompson, line 787: Would a trick like you did here also work? I stole it for use elsewhere so I'm curious if there is nuance outside of "if it works, it works"

https://github.com/openforcefield/openff-toolkit/blob/d0e63b9dbb540bf8a9a6d9aa48921f0c01d257b4/openff/toolkit/topology/topology.py#L2035-L2039

PR #1801, opened 2023-12-24
Passing custom SQM parameters to Antechamber using the `-ek` flag.

### LINKED ISSUE #1800: Allow passing the `-ek` flag to Antechamber (passing custom SQM keywords)

This is related to our work digging down into some of the issues related to: https://github.com/orgs/openforcefield/discussions/34
It has become increasingly clear to us that at least a portion of the reproducibility issues faced are down to unstable SQM options being passed by antechamber by default.

Antechamber has a `-ek` flag for these circumstances which allows users to manually specify the SQM options desired.

The proposal here is to add an option to feed an optional string for this `-ek` flag.

Currently we are validating things for https://github.com/orgs/openforcefield/discussions/34 using a custom branch with that feature, but we are limited by not having an OpenFF release with it. E.g. we won't be able to run full benchmark checks over alchemiscale.

Optionally allow passing the `-ek` flag as a user supplied string.

1. Fixing this upstream would be ideal, but we are too data poor to make a proper recommendation for now.
2. We could implement our own wrappers around antechamber, althought avoiding duplication of effort would be preferred.

**Additional context**

https://github.com/orgs/openforcefield/discussions/34

IAlibay:
Fixes #1800 

Note: I opened up this PR as an example of what we would need for #1800, please feel free to supersede or take over as needed.


j-wags (changes requested):
I don't like that this adds `**kwargs` to a public API point, since it 1) will mask typos like `use_cofnomers` and 2) opens up an unboundedly large surface in the public API.

Could you replace uses of `**kwargs` with one or more specific kwargs prepended with underscore to indicate that it's private and temporary?

IAlibay:
@j-wags :/ maybe worth a call? From the look of things right now, this might not be so temporary - as in this really seems to be the way we pin down local producibility.

j-wags:
Sure. Happy to have a call - Let's coordinate on Slack.

My intent in saying "temporary" is that we are very likely to adopt your recommendations at the conclusion of this study. So the keyword argument in this method call would be temporary because the antechamber flags you settle on would be hard-coded into the OpenFF Toolkit.

IAlibay:
>  because the antechamber flags you settle on would be hard-coded into the OpenFF Toolkit.

😅 This is probably a good thing for a call then - you may find that my recommendation is very nuanced with lots of "in some situations" 🙈

PR #1814, opened 2024-01-29
Ensure v1 `ValidationError` is checked in test

mattwthompson:
The current code blindly does `from Pydantic import ValidationError`, which later causes problems when attempting to later catch the error which is always from v1.

This works locally with each of v1 and v2 installed


j-wags (approved):
Thanks @mattwthompson!

PR #1815, opened 2024-01-29
Re-lint for Black 24

mattwthompson:
Black decided to change things; the linting action does no version control (pre-commit.ci does) so these fixes need to be introduced ASAP


j-wags (approved):
Good catch! I'd wondered what changed.

PR #1825, opened 2024-02-15
Skip some tests if unpackaged files not present, make some files path-y-er

mattwthompson:
The tests currently make some unsafe assumptions about files present relative to the Python package. For example, this does not necessarily point to the root of the repository, since many files are not packaged:

https://github.com/openforcefield/openff-toolkit/blob/5ffcb58714c93143d8c00841c00e73f46a6c296f/openff/toolkit/_tests/test_links.py#L7-L9

From this, peeking at the `README.md` file (or looking for an `examples/` directory) can introduce some surprises:

```python
In [1]: import pathlib, os

In [2]: from openff.toolkit._tests.test_links import __file__

In [3]: os.system(f"head {pathlib.Path(__file__).parents[3] / 'README.md'}")
# autoflake

[![Build Status](https://github.com/PyCQA/autoflake/actions/workflows/main.yaml/badge.svg?branch=main)](https://github.com/PyCQA/autoflake/actions/workflows/main.yaml)

## Introduction

_autoflake_ removes unused imports and unused variables from Python code. It
makes use of [pyflakes](https://pypi.org/pypi/pyflakes) to do this.

By default, autoflake only removes unused imports for modules that are part of
Out[3]: 0
```

This change simply has fixtures return empty lists when it looks like tests are being collected from a packaged path (as distinct from a local path, where the full contents of the repository may be present).

### DIFF
--- a/openff/toolkit/_tests/test_examples.py
+++ b/openff/toolkit/_tests/test_examples.py
@@ -2,7 +2,6 @@
 Test that the examples in the repo run without errors.
 """
 
-import os
 import pathlib
 import re
 import subprocess
@@ -11,9 +10,8 @@ import textwrap
 
 import pytest
 
+from openff.toolkit._tests.utils import _get_readme_path
 from openff.toolkit.utils import RDKIT_AVAILABLE, get_data_file_path, temporary_cd
-
-ROOT_DIR_PATH = pathlib.Path(__file__).joinpath("../../../../").resolve()
 
 
 def run_script_file(file_path):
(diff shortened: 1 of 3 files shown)


openff/toolkit/_tests/test_examples.py
```diff
@@ -38,55 +36,70 @@ def run_script_str(script_str):
...
+    else:
+        examples_dir_path = pathlib.Path(_get_readme_path().parent, "examples")
 
-    # Examples that require RDKit
-    rdkit_examples = {
-        examples_dir_path.joinpath("conformer_energies/conformer_energies.py"),
-    }
+        # Examples that require RDKit
+        rdkit_examples = {
+            examples_dir_path / "conformer_energies/conformer_energies.py"
```
j-wags, line 71: Wow, trippy. I didn't realize that Path objects overrode the `/` operator. Made my day.

j-wags (approved):
Thanks @mattwthompson!

Issue #1820, opened 2024-02-02
Allow `Topology.set_positions(None)`


mattwthompson:
**Is your feature request related to a problem? Please describe.**

I'm dealing with a use case in which I want to modify a topology's positions but reset them to what they were before. The getter and setter introduced in #1368 makes this clean and first-class except for the (common, I'd argue) case in which there aren't originally positions at all. 

Here is an example of how this fails and what it takes to work around it:

```python
import numpy
from openff.toolkit import Molecule, Quantity, unit
from openff.toolkit.utils.exceptions import IncompatibleUnitError


def temporarily_modify(topology, new_positions):
    """Modify a topology's positions, but then reset them."""
    original_positions = topology.get_positions()
    topology.set_positions(new_positions)
    print(f"Modified positions mean: {numpy.mean(topology.get_positions().m)}")
    topology.set_positions(original_positions)
    print(f"Reset positions mean: {numpy.mean(topology.get_positions().m)}")


topology = Molecule.from_smiles("CCO").to_topology()

ones = Quantity(
    numpy.ones(shape=(topology.n_atoms, 3)),
    unit.nanometer,
)

try:
    temporarily_modify(topology, 3 * ones)
except IncompatibleUnitError as error:
    # Fails because the original positions (None) can't be set
    print("Can't reset positions: " + str(error))
    # Prints
    # Can't reset positions: array should be an OpenFF Quantity with dimensions of length

# Can work around this by only calling it on a topology with positions
molecule = Molecule.from_smiles("CCO")
molecule._conformers = [0 * ones]
topology_with_positions = molecule.to_topology()

# Works; prints 4.0 and 0.0
temporarily_modify(topology_with_positions, 4.0 * ones)

# or can work around by special-casing downstream, which is not desirable


def workaround(topology, new_positions):
    """Modify a topology's positions, but then reset them."""
    original_positions = topology.get_positions()
    topology.set_positions(new_positions)
    print(f"Modified positions mean: {numpy.mean(topology.get_positions().m)}")
    if original_positions is None:
        for molecule in topology.molecules:
            molecule._conformers = None
        print(f"Reset positions mean: {topology.get_positions()}")
    else:
        topology.set_positions(original_positions)
        print(f"Reset positions mean: {numpy.mean(topology.get_positions().m)}")


topology = Molecule.from_smiles("CCO").to_topology()

# Works; prints 5.0 and 0.0
workaround(topology, 5.0 * ones)
```
[...]

PR #1827, opened 2024-02-22
Add `Topology.clear_positions`

mattwthompson:
This adds `Topology.clear_positions()` which does what the name implies, instead of allowing `Topology.set_positions(None)` as I originally advocated for in #1820. Today I think it's better to just have a separate method that does this one thing - it handles what I remember my use case to be - instead of allowing `set_positions` to take different types.

I hold this preference weakly (I've already changed my mine once on my own) and can easily be talked out of it


j-wags (approved):

PR #1836, opened 2024-03-13
Use new-ish NumPy random

mattwthompson:
Context https://docs.astral.sh/ruff/rules/numpy-legacy-random/#numpy-legacy-random-npy002


j-wags:
After looking at it a bit, I think the point of this (admittedly strange) test IS to run with different random inputs each time. So I don't think this change is needed, the (weird) test is (marginally) better without it, and the code is more readable in the old form. 

But I am still bothered by how weird this test is. So an alternative to closing this PR would be making the test not-weird by hardcoding a random-ish conformer.

j-wags (approved):
Thanks. It took me a few minutes to find where numpy authoritatively said "don't use `numpy.random.random` any more" but if I navigate to its [API docs](https://numpy.org/doc/stable/reference/random/generated/numpy.random.random.html) and follow the breadcrumbs they do indeed say it. The Ruff page says stuff about performance, reproducibility, and statistical properties, but what they really need to say is "numpy says to change your existing code involving `numpy.random.random` to a more complex thing involving a creating a new type of object, even though your current code is working just fine and not so much as emitting a warning". 

I'll update the releasenotes and merge!

mattwthompson:
Yeah it's a bit confusing, just trusting the "rules" here. It may never get removed but following their (admittedly weak) suggestion prevents this from being a confusing deprecation warning in the future

PR #1837, opened 2024-03-14
Decouple vdW and virtual site parameters

mattwthompson:
The current structure of `VirtualSiteHandler.VirtualSiteType` couples together electrostatics, geometry, and (12-6) vdW interactions. This forces unnecessary copying of hundreds of lines of code if anybody wishes to build off of this class using non-standard functional forms.

This change refactors `VirtualSiteType` to enable this downstream use case without changing the public API or behavior.

In the future it may be worth moving the charge increment logic from `_BasicVirtualSiteType` to `VirtualSiteType`; this change neither does that nor prevents it from happening in the future.

Credit to @jthorton who came up with this idea and implemented it; I just plucked it over here since it doesn't make sense for downstream packages to re-implement toolkit offerings

- [x] De-couple non-bonded and virtual site parameters


j-wags:
Will this require coordination with an Interchange release/break downstream stuff, or can it be released at will?

mattwthompson:
I don't think this would affect any Interchange code since the behavior of virtual sites with LJ or no vdW interactions should be unchanged, just like force field parsing.

Let's see (some NAGL stuff will probably fail for unrelated reasons): https://github.com/openforcefield/status/pull/71/checks

j-wags:
Awesome - Tests are all green. I'll see if I can get this in for the upcoming toolkit release!

mattwthompson:
Great, if there's reason to hold off, 0.16.1 could work as well

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -3428,149 +3429,193 @@ def check_handler_compatibility(self, other_handler: "GBSAHandler"):
...
             )
 
-            if not cls._supports_match(type_, match, is_in_plane):
-                raise SMIRNOFFSpecError(
-                    f"match='{match}' not supported with type='{type_}'"
-                    + ("" if is_in_plane is None else f" and is_in_plane={is_in_plane}")
-                )
+    @classmethod
+    def _supports_in_plane_angle(cls, type_: _VirtualSiteType) -> bool:
+        return type_ in {"MonovalentLonePair"}
```
j-wags, line 3562: Does this make https://github.com/openforcefield/openff-toolkit/pull/1837/files#diff-fd6b852eaa3bb076240130325c7023d4041de6b0f45bee26c509e61b8a661a23R3619-R3621 redundant?

mattwthompson: Nice find, I think so!

j-wags (approved):

PR #1852, opened 2024-04-02
[WIP] Allow duplicate SMIRKS and have lookups return the LAST one found

### LINKED ISSUE #1363: `ForceField` allows multiple parameters to have the same SMIRKS patterns
**Describe the bug**
The toolkit's force field class allows multiple parameters within the same handler to have the same SMIRKS patterns. While it may seem like a strange use case, the inclusion of TIP3P in Sage makes this an issue anytime somebody wishes to use a different water model alongside it. I'd argue this state should not be reached and somewhere an exception should be raised when attempting to add a parameter with a smirks already found in the handler's list of parameters.

This also causes issues downstream when [we assume](https://github.com/openforcefield/openff-interchange/pull/494) this sort of 1:1 mapping, which I think we should be able to get away with.

**To Reproduce**

Using the file at `openff/toolkit/data/test_forcefields/tip3p.offxml`:

```python3
>>> from openff.toolkit import ForceField
>>> import pprint
>>> ff = ForceField("openff-2.0.0.offxml", "tip3p.offxml")
>>> pprint.pprint(ff['vdW'].parameters[-4:])
[<vdWType with smirks: [#1]-[#8X2H2+0:1]-[#1]  epsilon: 0.1521 kilocalorie / mole  id: n-tip3p-O  sigma: 3.1507 angstrom  >,
 <vdWType with smirks: [#1:1]-[#8X2H2+0]-[#1]  epsilon: 0.0 kilocalorie / mole  id: n-tip3p-H  sigma: 1 angstrom  >,
 <vdWType with smirks: [#1]-[#8X2H2+0:1]-[#1]  epsilon: 0.635968 kilojoules_per_mole  id: n1  sigma: 0.3150752406575124 nanometer  >,
 <vdWType with smirks: [#1:1]-[#8X2H2+0]-[#1]  epsilon: 0 kilojoules_per_mole  id: n2  sigma: 1 nanometer  >]
>>> assert ff['vdW'].parameters[-1].smirks == ff['vdW'].parameters[-3].smirks
>>> assert ff['vdW'].parameters[-2].smirks == ff['vdW'].parameters[-4].smirks
```

**Output**
We ran into a case in which `ForceField("openff-2.0.0.offxml", "tip4p.offxml")` ultimately assigned TIP3P vdW parameters (from Sage) where I expected TIP4P vdW parameters to be assigned. This caused issues since other parameters (i.e. virtual sites) came from TIP4P and the result was an invalid mismatch of water models that unfortunately ran well enough to hide this error for weeks.

Python 3.9, macOS, most recent development head (0f1611fe0b08c42276befeb8cba767caa7cda58b)

j-wags:
- [x] Resolves #1363 
    - If multiple parameters have identical SMIRKS, this will make `__getitem__` calls return the LAST instance in the ParameterList/ParameterHandler (rather than the first)
    - Adds named argument `ParameterHander.add_parameter(..., allow_duplicates=True)`
    - Since the default behavior remains the same, this won't affect amber-ff-porting https://github.com/openforcefield/amber-ff-porting/blob/c6d3412f5c94cd1c3ed1b8b6ba56909fe4706080/ConvertResidueParameters.py#L747

### DIFF
--- a/openff/toolkit/_tests/test_forcefield.py
+++ b/openff/toolkit/_tests/test_forcefield.py
@@ -1711,6 +1711,18 @@ class TestForceField(_ForceFieldFixtures):
         ):
             force_field["vdW"][smirks]
 
+    def test_lookup_duplicate_parameter_type(self, force_field):
+        """Test that if a parameter SMIRKS appears twice, lookups will yield the second copy"""
+        first_bond = force_field["Bonds"][0]
+        force_field["Bonds"].add_parameter({"smirks": first_bond.smirks,
+                                            "k": first_bond.k / 2,
+                                            "length": first_bond.length/2,
+                                            "id": first_bond.id+"2"})
+        result = force_field["Bonds"][first_bond.smirks]
+        assert result.id == force_field["Bonds"][0].id + "2"
+        assert result.k == first_bond.k / 2
+        assert result.length == first_bond.length/2
+
     @pytest.mark.parametrize(
         "to_deregister",
         [
--- a/openff/toolkit/_tests/test_forcefield.py
+++ b/openff/toolkit/_tests/test_forcefield.py
@@ -1724,7 +1724,7 @@ class TestForceField(_ForceFieldFixtures):
             allow_duplicate_smirks=True,
         )
         result = force_field["Bonds"][first_bond.smirks]
-        assert result == forcefield['Bonds'].parameters[first_bond.smirks]
+        assert result == forcefield["Bonds"].parameters[first_bond.smirks]
         assert result.id == force_field["Bonds"][0].id + "2"
         assert result.k == first_bond.k / 2
         assert result.length == first_bond.length / 2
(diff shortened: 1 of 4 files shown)


openff/toolkit/_tests/test_forcefield.py
```diff
@@ -1711,6 +1711,23 @@ def test_lookup_parameter_type(self, force_field):
...
+        force_field["Bonds"].add_parameter(
+            {
+                "smirks": first_bond.smirks,
+                "k": first_bond.k / 2,
+                "length": first_bond.length / 2,
+                "id": first_bond.id + "2",
+            },
+            allow_duplicate_smirks=True,
+        )
+        result = force_field["Bonds"][first_bond.smirks]
```
mattwthompson, line 1726: Worth testing at both `ParameterHandler` and `ParameterList`? i.e.
```suggestion
        result = force_field["Bonds"][first_bond.smirks]
        assert result == forcefield['Bonds'].parameters[first_bond.smirks]
```

j-wags: Good call, thanks!

PR #1856, opened 2024-04-05
Resolve #1096 with an optional argument for <_generate_unique_atom_names()>

PEFrankel:
- [ ] Resolves: [Make atom names more unique #1096](https://github.com/openforcefield/openff-toolkit/pull/1096), with an optional suffix parameter. Avoids possible conflict with analyses that create atom dictionaries without referencing the resname. Also improves readability of simulation files.

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -1089,7 +1089,7 @@ class FrozenMolecule(Serializable):
         """``True`` if the molecule has unique atom names, ``False`` otherwise."""
         return _has_unique_atom_names(self)
 
-    def generate_unique_atom_names(self, suffix:str = None):
+    def generate_unique_atom_names(self, suffix: str = None):
         """
         Generate unique atom names from the element symbol and count.
 
@@ -6128,7 +6128,7 @@ class HierarchyElement:
         """``True`` if the element has unique atom names, ``False`` otherwise."""
         return _has_unique_atom_names(self)
 
-    def generate_unique_atom_names(self, suffix:str = None):
+    def generate_unique_atom_names(self, suffix: str = None):
         """
         Generate unique atom names from the element symbol and count.
 
@@ -6154,7 +6154,9 @@ def _has_unique_atom_names(
     return True
 
 
-def _generate_unique_atom_names(obj: Union[FrozenMolecule, HierarchyElement], suffix:str = "x"):
+def _generate_unique_atom_names(
+    obj: Union[FrozenMolecule, HierarchyElement], suffix: str = "x"
+):
     """
     Generate unique atom names from the element symbol and count.
 
@@ -6165,7 +6167,7 @@ def _generate_unique_atom_names(obj: Union[FrozenMolecule, HierarchyElement], su
     atom names might begin 'C1x', 'H1x', 'O1x', 'C2x', etc.
     """
     from collections import defaultdict
- 
+
     element_counts: DefaultDict[str, int] = defaultdict(int)
     for atom in obj.atoms:
         symbol = atom.symbol
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -1089,17 +1089,20 @@ class FrozenMolecule(Serializable):
         """``True`` if the molecule has unique atom names, ``False`` otherwise."""
         return _has_unique_atom_names(self)
 
-    def generate_unique_atom_names(self):
+    def generate_unique_atom_names(self, suffix:str = None):
         """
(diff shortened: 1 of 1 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -1089,17 +1089,20 @@ def has_unique_atom_names(self) -> bool:
         """``True`` if the molecule has unique atom names, ``False`` otherwise."""
         return _has_unique_atom_names(self)
 
-    def generate_unique_atom_names(self):
+    def generate_unique_atom_names(self, suffix: str = None):
```
mattwthompson, line 1092:
```suggestion
    def generate_unique_atom_names(self, suffix: str = None):
```
I think it'd be simpler to have the default be the current behavior. This also has the benefit of this argument always being a string

openff/toolkit/topology/molecule.py
```diff
@@ -1089,17 +1089,20 @@ def has_unique_atom_names(self) -> bool:
...
+        that element is found in the hierarchy element. The character 'x' is
+        appended to these generated names to reduce the odds that they clash
+        with an atom name or type imported from another source. For example,
+        generated atom names might begin 'C1x', 'H1x', 'O1x', 'C2x', etc.
         """
-        return _generate_unique_atom_names(self)
+        if suffix == None:
+            return _generate_unique_atom_names(self)
+        else:
+            return _generate_unique_atom_names(self, suffix)
```
mattwthompson, line 1105:
```suggestion
        return _generate_unique_atom_names(self, suffix)
```
See aboe

openff/toolkit/topology/molecule.py
```diff
@@ -6125,7 +6128,7 @@ def has_unique_atom_names(self) -> bool:
         """``True`` if the element has unique atom names, ``False`` otherwise."""
         return _has_unique_atom_names(self)
 
-    def generate_unique_atom_names(self):
+    def generate_unique_atom_names(self, suffix: str = None):
```
mattwthompson, line 6131:
```suggestion
    def generate_unique_atom_names(self, suffix: str = "x"):
```
Again, same as above

openff/toolkit/topology/molecule.py
```diff
@@ -6135,7 +6138,10 @@ def generate_unique_atom_names(self):
         with an atom name or type imported from another source. For example,
         generated atom names might begin 'C1x', 'H1x', 'O1x', 'C2x', etc.
         """
-        return _generate_unique_atom_names(self)
+        if suffix == None:
+            return _generate_unique_atom_names(self)
+        else:
+            return _generate_unique_atom_names(self, suffix)
```
mattwthompson, line 6144:
```suggestion
        return _generate_unique_atom_names(self, suffix)
```
And again, same as above

mattwthompson (commented):
If I'm understand the removed comments correctly (I may have written them), this is a good change as it implements an idea we previously put off implementing. I suggest keeping the default behavior something like `suffix="x"` (see comments below) since I think people wouldn't be calling this method at all if they don't want unique atom names.

Could you add a couple of tests for this behavior? I'm thinking along the lines of
* Having a non-default suffix and making sure it's passed through to the resulting atom names
* Making sure `suffix=None` raises an error (injecting my idea here)
* What happens with `suffix=""`? Maybe this should be an error, I don't know
* Anything else relevant to your use case I may be forgetting


For my own reference, this method is called a few times in Interchange: https://github.com/search?q=org%3Aopenforcefield+generate_unique_atom_names+repo%3Aopenforcefield%2Fopenff-interchange&type=code

mattwthompson:
P.S. don't worry about
* failing tests on `OpenEye=true` runs - these will always fail since we don't expose the OpenEye license on forks, and since these changes probably don't hit RDKit/OpenEye behavior they shouldn't be necessary
* the bot reformatting code - it's only for style consistency. You can [run it yourself](https://docs.openforcefield.org/projects/toolkit/en/stable/users/developing.html#pre-commit) or let it auto-fix stuff (you might need to `git pull origin HEAD` to avoid merge conflicts, where `origin` is the remote name of your fork)

openff/toolkit/topology/molecule.py
PEFrankel: The default is set in _generate_unique_atom_names to "x", but I guess consistency in the calls will avoid a slip.

j-wags:
Apologies for the big delay on this. This morning I'm going to:

* incorporate Matt's suggestions 
* move this PR on to a branch so CI will fully run
* add @PEFrankel to the developers group so he can modify the branch.

PR #1859, opened 2024-04-12
Update NAGL pins in test environments

lilyminium:
Would it be possible to update the NAGL pins to stay current with new 0.3.x releases?
I dropped Python 3.9 in 0.3.8, whereas the toolkit still tests on it, so I've set the lower bound to 0.3.7 here.


mattwthompson:
FWIW, Interchange is now 3.10+ so any new toolkit release is subject to that constraint (implicitly or explicitly)

j-wags (approved):

PR #1862, opened 2024-04-17
[WIP] Revert seemingly innocuous change that causes some OpenEye methods to fail

j-wags:
- [x] Resolves https://github.com/openforcefield/openff-qcsubmit/issues/278 by reverting [this change](https://github.com/openforcefield/openff-toolkit/pull/1798/files#diff-e2e1869da0d73e84cb7e83cc8c464ef6bfe93f5d1b075de6ed3a21e212853a70R1613)

### DIFF
--- a/openff/toolkit/utils/openeye_wrapper.py
+++ b/openff/toolkit/utils/openeye_wrapper.py
@@ -1675,7 +1675,7 @@ class OpenEyeToolkitWrapper(ToolkitWrapper):
         assert None not in oemol_bonds
 
         # Retain conformations, if present
-        if molecule._conformers is not None:
+        if molecule.n_conformers != 0:
             oemol.DeleteConfs()
             for conf in molecule._conformers:
                 # OE needs a 1 x (3*n_Atoms) double array as input


mattwthompson (commented):
Frankly, I'm baffled as to why this would be causing issues. But if it fixes it, it fixes it.

I get that length-checking might be safer than attribute access, but this attribute access felt innocuous to me when writing it and in fact happens in `Molecule.n_conformers` itself. What am I missing?

https://github.com/openforcefield/openff-toolkit/blob/0.16.0/openff/toolkit/topology/molecule.py#L3288-L3296

mattwthompson:
This seems to fix the failing test in QCSubmit, and I don't want to stand in the way of a rapid patch, but

1. I don't think we ever got a minimal reproduction (at least I'm not sure how to do so)
1. There might be other code paths in which I made this sort of error

j-wags:
Right, my plan A right now is to come up with a test for this. This has taken long enough to find that I'd like to keep our future selves from getting dinged by it.

j-wags:
I've managed to fix this entirely within QCSubmit by changing its calls to the private `offmol._conformers` attribute. So this PR is superseded by https://github.com/openforcefield/openff-qcsubmit/pull/277

Issue #1865, opened 2024-04-25
`antechamber -L` reports a different version than AmberTools package


mattwthompson:
The toolkit stores the header of `antechamber -L` as the AmberTools version for provenance and reporting, but in odd versions these two dffier:

```
$ antechamber -L | head -n2 && micromamba list ambert

Welcome to antechamber 22.0: molecular input file processor.
List of packages in environment: "/Users/mattthompson/micromamba/envs/qcsubmit-test-basic"

  Name        Version  Build            Channel
─────────────────────────────────────────────────────
  ambertools  23.3     py311hd5bda53_6  conda-forge
```

We have a different way of grabbing this which is meant to more directly query AmberTools: https://github.com/openforcefield/openff-utilities/blob/ee9f969c252d02548c08537fb34cb713923eef70/openff/utilities/provenance.py#L39

There's a good chance I wrote this TODO: https://github.com/openforcefield/openff-toolkit/blob/f79f48bef13a997619c4391ca6c800806238237a/openff/toolkit/utils/ambertools_wrapper.py#L72-L75

mattwthompson:
#1866 / 0.16.1 (probably)

PR #1866, opened 2024-04-25
Safer AmberTools version check

mattwthompson:
#1865

### DIFF
--- a/openff/toolkit/utils/ambertools_wrapper.py
+++ b/openff/toolkit/utils/ambertools_wrapper.py
@@ -11,6 +11,7 @@ from shutil import which
 from typing import TYPE_CHECKING, Optional
 
 import numpy as np
+from openff.utilities.provenance import get_ambertools_version
 
 from openff.toolkit import Quantity, unit
 from openff.toolkit.utils import base_wrapper, rdkit_wrapper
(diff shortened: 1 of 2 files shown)


openff/toolkit/utils/ambertools_wrapper.py
```diff
@@ -69,10 +70,7 @@ def __init__(self):
                 f"available. {self._toolkit_installation_instructions}"
             )
 
-        # TODO: More reliable way to extract AmberTools version
-        out = subprocess.check_output(["antechamber", "-L"])
-        ambertools_version = out.decode("utf-8").split("\n")[1].split()[3].strip(":")
-        self._toolkit_version = ambertools_version
+        self._toolkit_version = get_ambertools_version()
```
j-wags, line 73: For reference, source code is here: https://github.com/openforcefield/openff-utilities/blob/ee9f969c252d02548c08537fb34cb713923eef70/openff/utilities/provenance.py#L39-L42

j-wags (approved):
Thanks @mattwthompson - I'll update releasenotes and merge.

pbuslaev:
Maybe a bit late, but the current solution does not work well in combination with slurm. The use case is as following:

I want to assign charges with `ambertools` to a molecule and I am submitting a script that is doing that to the external cluster with [sbatch](https://slurm.schedmd.com/sbatch.html). By default (check --export documentation for sbatch) local environment variables (including PATH) will also be exported to cluster. And the problem occurs, when conda path is different at local and server node (this can be quite a frequent scenario when running on the cloud). In this case, `get_ambertools_version`, which is [looking for `conda` path](https://github.com/openforcefield/openff-utilities/blob/90d49544c6f2c131d7c3edbaef68ade9883d7861/openff/utilities/provenance.py#L39) will fail. While there are ways around it on the user side, it took me quite a while to detect the issue. I wonder if there is a way on the openff side to solve it? For instance, [`_get_conda_list_package_versions()`](https://github.com/openforcefield/openff-utilities/blob/90d49544c6f2c131d7c3edbaef68ade9883d7861/openff/utilities/provenance.py#L7) can in addition check the output of `conda info --base` or smth like this?

mattwthompson:
> conda path is different at local and server node

Doesn't this by definition mean different environments are being used?

I'm open to stuffing something through to `_get_conda_list_package_versions` but this is the sort of thing that can have side effects

PR #1873, opened 2024-05-02
Drop py311 from conda tests

j-wags:
OpenMMForceFields requires AmberTools <23, but AmberTools <=22 doesn't support python 3.11

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -104,46 +104,26 @@ jobs:
 
           cd openff-toolkit
 
-      #- name: Test the package
-      #  run: |
-      #    python -m pip install utilities/test_plugins
+      - name: Test the package
+        run: |
+          python -m pip install utilities/test_plugins
 
-      #    pwd
-      #    ls
+          pwd
+          ls
 
-      #    if [[ "$OPENEYE" == true ]]; then
-      #      python -c "import openff.toolkit; print(openff.toolkit.__file__)"
-      #      python -c "import openeye; print(openeye.oechem.OEChemIsLicensed())"
-      #    fi
+          if [[ "$OPENEYE" == true ]]; then
+            python -c "import openff.toolkit; print(openff.toolkit.__file__)"
+            python -c "import openeye; print(openeye.oechem.OEChemIsLicensed())"
+          fi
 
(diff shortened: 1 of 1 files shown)


mattwthompson:
Oh, I see, it's because of this trick: https://github.com/openforcefield/openff-toolkit/blob/f282282c6c5fb14df4c479cb7c689984bf3952f0/devtools/conda-envs/conda.yaml#L5-L6

I think this should be changed until `openmmforcefields` is a base dependency of the toolkit

j-wags:
> I think this should be changed until openmmforcefields is a base dependency of the toolkit

I'd like to interpret "this should be changed..." as "the changes proposed in this PR should remain until...", but I'm not sure that's what you meant. 

These conda tests are the "whole kit and kaboodle", including external examples which require OMMFFs. I don't currently plan for OMMFFs to become a dependency of `openff-toolkit` or `openff-toolkit-base`.

j-wags:
And the immediate issue is that we have our [conda tests](https://github.com/openforcefield/openff-toolkit/actions/runs/8916569010/job/24488207845#step:4:44) running with py3.10 and 3.11, the latter of which will be failing until the deployment issue is resolved.

mattwthompson:
I have no idea how my hands threw an "until" in there, "unless" fits better (I don't think it should become a dependency!). Sorry for the confusion

> These conda tests are the "whole kit and kaboodle"

Right, it brings in everything needed for base functionality, examples, and maybe some other peripheral stuff. But this workflow doesn't run examples right now, so bringing in examples-only dependencies is bloat (at the level of what's installed) at the benefit of simplicity (only declare one "package")

So here I'd advocate for either

1. Swap out `openff-toolkit-examples` for `openff-toolkit` and whatever other packages are needed for the tests that currently run (my preference) or
2. Update this workflow to also run examples nightly from the latest conda package


(There's also the 1a option of merging this as-is to get stuff green for tonight's run, and then following it up with a change later on that brings back 3.11 and installs without `openmmforcefields`.)

j-wags:
I'm manually kicking off conda tests [here](https://github.com/openforcefield/openff-toolkit/actions/workflows/conda.yml)

PR #1874, opened 2024-05-08
Lazy load `networkx` in topology modules

mattwthompson:
I was profiling some other code and found that a ~100 ms import of `networkx` happens in the toolkit whether or not it's actually used. I remember lazy-loading this in the past, but that's reflected in the current code.

This way of profiling here shows roughly a 50-100 ms speedup depending on how one squints their eyes

```shell
$ git checkout upstream/main
$ hyperfine 'python -c "from openff.toolkit import Molecule"' --prepare 'python -c "from openff.toolkit import Molecule"'
  Time (mean ± σ):      1.119 s ±  0.047 s    [User: 0.909 s, System: 0.142 s]
  Range (min … max):    1.085 s …  1.241 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

$ git checkout lazy-load-networkx
Previous HEAD position was ef2f514f Drop py311 from conda tests (#1873)
Switched to branch 'lazy-load-networkx'
$ hyperfine 'python -c "from openff.toolkit import Molecule"' --prepare 'python -c "from openff.toolkit import Molecule"'
  Time (mean ± σ):      1.050 s ±  0.048 s    [User: 0.864 s, System: 0.121 s]
  Range (min … max):    1.007 s …  1.143 s    10 runs
```

With this patch, here's a rough picture of what takes so long to run this import. There are some other obvious candidates for lazy-loading, but I'm not going for them now:

image

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -25,6 +25,7 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 ### New features
 
 - [PR #1866](https://github.com/openforcefield/openff-toolkit/pull/1866): Implements a safer AmberTools version check.
+- [PR #1874](https://github.com/openforcefield/openff-toolkit/pull/1874): Improves some loading times by lazy-loading `networkx`.
 
 ### Improved documentation and warnings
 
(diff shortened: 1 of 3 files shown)


j-wags (approved):
Thanks @mattwthompson  - Please update the releasenotes before you merge :-)

PR #1881, opened 2024-05-29
Resolve #1096 with an optional argument for <_generate_unique_atom_names()>

j-wags:
Continued from @PEFrankel 's original PR at  https://github.com/openforcefield/openff-toolkit/pull/1856, which was on a fork so some linting actions were sad and OpenEye CI wasn't running.

- [ ] Resolves: [Make atom names more unique #1096](https://github.com/openforcefield/openff-toolkit/pull/1096), with an optional suffix parameter. Avoids possible conflict with analyses that create atom dictionaries without referencing the resname. Also improves readability of simulation files.


openff/toolkit/topology/molecule.py
```diff
@@ -6162,9 +6164,4 @@ def _generate_unique_atom_names(obj: Union[FrozenMolecule, HierarchyElement]):
     for atom in obj.atoms:
         symbol = atom.symbol
         element_counts[symbol] += 1
-        # TODO: It may be worth exposing this as a user option, i.e. to avoid multiple ligands
-        # parameterized with OpenFF clashing because they have atom names like O1x, H3x, etc.
-        # i.e. an optional argument could enable a user to `generate_unique_atom_names(blah="y")
-        # to have one ligand be O1y, etc.
-        # https://github.com/openforcefield/openff-toolkit/pull/1096#pullrequestreview-767227391
-        atom.name = symbol + str(element_counts[symbol]) + "x"
+        atom.name = symbol + str(element_counts[symbol]) + suffix
```
mattwthompson, line 6167: It looks like there are two defined `generate_unique_atom_names` functions? (Messy merge?) This is unhappy because one of them is trying to stringify `None`

```python
>>> '' + None
TypeError: can only concatenate str (not "NoneType") to str

Comment (j-wags) [reply openff/toolkit/topology/molecule.py:6167]: I'm somewhat confident (~70%) there should be two definitions of `generate_unique_atom_names` (one on FrozenMolecule and another on HierarchyElement). So I don't think the root cause here is a merge conflict, rather it's a genuine code issue (TM) caused by me merging in some code suggestions from the previous PR. 

I'm contacting @PEFrankel to see if he wants to carry this over the finish line. Otherwise I'll clean up my mess here :-P

Comment (mattwthompson) [reply openff/toolkit/topology/molecule.py:6167]: Okay yeah that would make sense - I may have glossed over that in the diff

Comment (mattwthompson) [reply openff/toolkit/topology/molecule.py:6167]: Still needs to be a way to prevent this concatenation, `+ suffix if suffix else ''` is probably lazy enough to work

Comment (PEFrankel) [reply openff/toolkit/topology/molecule.py:6167]: Yes, I've been using "else" flags in the calls since I planned on adding tests there, but this works as well.

**Review by j-wags — COMMENTED** [review]
@PEFrankel the remaining steps for this are:

* Fix the code so the existing tests pass
* Update the docstring of the methods that this PR changes to list the new `suffix` argument (you'll need to add a `Parameters` section to their docstrings, see other methods in `molecule.py` for a template)
* Add tests (Note: Our test organization is inconsistent; I'd recommend adding the new tests above or below [this one](https://github.com/openforcefield/openff-toolkit/blob/main/openff/toolkit/_tests/test_molecule.py#L907). This test loops over all of the molecules in minidrugbank, your tests don't need to do this. But do feel free to add a minimal multi-residue polymer that's relevant to your work to this repo to use in the new tests)
    * Add a test ensuring that the default value of `suffix` is "x" 
    * Add a test ensuring that non-default string values of `suffix` behave sanely
    * Add a test ensuring that `suffix=""` behaves sanely
    * (optional) Add tests for any other weird edge cases you've encountered in your work (maybe two-letter element names, or cases where there are more than 10 of one element so the atom name has to become one character longer)
* Add this PR as a "new feature" in the changelog in the hidden [REAL releasenotes file](https://github.com/openforcefield/openff-toolkit/blob/main/docs/releasehistory.md) (I've apparently kept this link broken for a long time to deter first time contributors 🤦🤦🤦 )

File: openff/toolkit/topology/molecule.py [hunk 2]
```diff
@@ -1089,17 +1089,17 @@ def has_unique_atom_names(self) -> bool:
         """``True`` if the molecule has unique atom names, ``False`` otherwise."""
         return _has_unique_atom_names(self)
 
-    def generate_unique_atom_names(self):
+    def generate_unique_atom_names(self, suffix: str = None):
```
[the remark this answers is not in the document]
PEFrankel: I define the default string only once to make any future changes of the function/default easier, but changing it is fine if you'd prefer.

docs/releasehistory.md
```diff
@@ -26,6 +26,7 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 - [PR #1866](https://github.com/openforcefield/openff-toolkit/pull/1866): Implements a safer AmberTools version check.
 - [PR #1874](https://github.com/openforcefield/openff-toolkit/pull/1874): Improves some loading times by lazy-loading `networkx`.
+- [PR #1881](https://github.com/openforcefield/openff-toolkit/pull/1881): Adds an optional suffix parameter to `_generate_unique_atom_names`.
```
j-wags, line 29: We don't have to tell users about changes to the private API (stuff prepended with `_`) but we do need to mention changes to the _public_ API. 
```suggestion
- [PR #1881](https://github.com/openforcefield/openff-toolkit/pull/1881): Adds an optional suffix parameter to `generate_unique_atom_names`.
```

openff/toolkit/_tests/test_molecule.py
```diff
@@ -924,7 +924,39 @@ def test_unique_atom_names(self, molecule):
         assert (
             len(set([atom.name for atom in molecule.atoms])) == molecule.n_atoms
         ) == molecule.has_unique_atom_names
-        assert all("x" in a.name for a in molecule.atoms)
+        assert all(
+            a.name.endswith("x") for a in molecule.atoms
+        )  # fails for xenon, and nothing here checks the x is in the correct place
+        # generate_unique_atom_names tests
```
j-wags, line 930: Cleaning up some comments

```suggestion
        )  
```

j-wags (commented):
Looks good! There are two more changes I've listed here (and feel free to do any other cleanups you want). Also, since I technically opened this new PR, I can't approve it. So please leave an approving comment yourself and merge when you're ready!

Issue #1888, opened 2024-06-10
`unit` argument to `ParameterAttribute` not sanitized


mattwthompson:
**Describe the bug**

I'm working on a isolating something from a more involved example, but for now I can demonstrate that this argument is not checked for proper types

https://github.com/openforcefield/openff-toolkit/blob/fdf383261717a1ca9fe72586e2463a425c04e0f9/openff/toolkit/typing/engines/smirnoff/parameters.py#L342-L350

which later causes this clause to bubble up in confusing ways - below because `self._unit` is allowed to be a string, it bubbles up a confusing error that nanometers aren't compatible with nanometers.

https://github.com/openforcefield/openff-toolkit/blob/fdf383261717a1ca9fe72586e2463a425c04e0f9/openff/toolkit/typing/engines/smirnoff/parameters.py#L416-L420

It's possible I'm missing something with how stuff is passed through these classes, but it's hard for me to figure out from the code itself.

**To Reproduce**

Passing a string to the `unit` argument is bad, but for now just consider the possibly that something could get botched and let it through like that or that a user would give it a try.

```python
In [1]: from openff.toolkit.typing.engines.smirnoff.parameters import (
   ...:     ParameterAttribute,
   ...:     unit,
   ...:     Quantity,
   ...: )

In [2]: parameter_attribute = ParameterAttribute(
   ...:     default=Quantity("1 nanometer"),
   ...:     unit="nanometer",
   ...: )

In [3]: parameter_attribute._name = "_foo"  # Not relevant to this reproduction

In [4]: parameter_attribute._validate_units(value=Quantity("2 nanometer"))
/Users/mattthompson/micromamba/envs/new-models/lib/python3.12/site-packages/openff/toolkit/typing/engines/smirnoff/parameters.py:585: SyntaxWarning: invalid escape sequence '\ '
  """The attribute of a parameter with an unspecified number of terms, where
/Users/mattthompson/micromamba/envs/new-models/lib/python3.12/site-packages/openff/toolkit/typing/engines/smirnoff/parameters.py:1614: SyntaxWarning: invalid escape sequence '\ '
  """
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File ~/micromamba/envs/new-models/lib/python3.12/site-packages/openff/toolkit/typing/engines/smirnoff/parameters.py:412, in ParameterAttribute._validate_units(self, value)
    411 try:
--> 412     if not self._unit.is_compatible_with(value.units):
[...]

PR #1890, opened 2024-06-13
De-stringify `ParameterAttribute.unit`

mattwthompson:
#1888

### DIFF
--- a/openff/toolkit/typing/engines/smirnoff/parameters.py
+++ b/openff/toolkit/typing/engines/smirnoff/parameters.py
@@ -342,10 +342,14 @@ class ParameterAttribute:
     def __init__(
         self,
         default: Any = UNDEFINED,
-        unit: Optional[Unit] = None,
+        unit: Unit | str | None = None,
         converter: Optional[Callable] = None,
         docstring: str = "",
     ):
+        if isinstance(unit, str):
+            # be careful with module & variable names
+            unit = Unit(unit)
+
         self.default = default
         self._unit = unit
         self._converter = converter
--- a/utilities/test_plugins/custom_plugins/handler_plugins.py
+++ b/utilities/test_plugins/custom_plugins/handler_plugins.py
@@ -31,10 +31,9 @@ class BuckinghamHandler(_NonbondedHandler):
                                unit="nanometer**-1",
                                )
 
-        # Define as None, since that's supported
         c = ParameterAttribute(
             default=None,
-            unit=None,
+            unit=unit.kilojoule_per_mole * unit.nanometer ** 6,
         )
 
     _TAGNAME = "TestBuckingham"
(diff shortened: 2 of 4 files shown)


openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -342,10 +342,14 @@ class ParameterAttribute:
...
         self,
         default: Any = UNDEFINED,
-        unit: Optional[Unit] = None,
+        unit: Unit | str | None = None,
         converter: Optional[Callable] = None,
         docstring: str = "",
     ):
+        if isinstance(unit, str):
+            # be careful with module & variable names
+            unit = Unit(unit)
```
mattwthompson, line 351: Another option is simply
```suggestion
        if isinstance(unit, str):
            raise ValueError(f"Input argument {unit} must be `Unit` or `None`")
```

Either fixes my use case

utilities/test_plugins/custom_plugins/handler_plugins.py
```diff
@@ -11,3 +12,29 @@ class WrongSubclass(list):
 
 class CustomIOHandler(ParameterIOHandler):
     _FORMAT = "JSON"
+
+class BuckinghamHandler(_NonbondedHandler):
+    """A custom parameter handler for buckingham interactions."""
+
+    class BuckinghamType(ParameterType):
```
j-wags, line 19: (blocking) Could these be given names like `TestBuckinghamHandler` and `TestBuckinghamType` to remove all chance that they come into conflict with a real plugin in the future?

openff/toolkit/typing/engines/smirnoff/parameters.py
j-wags: (not blocking) I'm in favor of the current fix, but out of curiosity, did you ever figure out what was going on with the trying-to-treat-a-string-as-a-quantity error arising even when a quantity was explicitly given?

j-wags (approved):
Thanks @mattwthompson - Approved pending one blocking comment and updating the releasenotes. I'm happy to hold off on 0.16.1 until the end of your workday if you can merge this before then.

mattwthompson:
@j-wags the issue I showed you yesterday was a combination of something funky with my local system and the virtual site and atomistic parameters _both_ having a/b/c/ parameters, I think the issue was stemming from a different file altogether https://github.com/openforcefield/openff-toolkit/issues/1888#issuecomment-2166400789

utilities/test_plugins/custom_plugins/handler_plugins.py
mattwthompson: Good idea - I'm rolling with `FOO` as a prefix, though, just in case it's collected as a test by `pytest`

mattwthompson:
I think this is good to go, but out of an abundance of caution (and since it's not blocking my work) I'm inclined to merge this _after_ you tag 0.16.1. Feel free to hit the green button if you disagree

PR #1891, opened 2024-06-13
Lint utilities/ subdirectory

mattwthompson:
When working on #1890 I noticed that `utilities/` was not including in the linting checks - this PR does that (and only that) instead of doing it all there

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 ci:
   autoupdate_schedule: "quarterly"
-files: ^openff|(^examples/((?!deprecated).)*$)|^docs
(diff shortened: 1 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 ci:
   autoupdate_schedule: "quarterly"
-files: ^openff|(^examples/((?!deprecated).)*$)|^docs
+files: ^openff|(^examples/((?!deprecated).)*$)|^docs|(^utilities/((?!deprecated).)*$)
```
mattwthompson, line 3: This (and a similar change below) are pretty much the only non-automated changes here

j-wags (approved):
Thanks, no releasenotes needed since this isn't user-facing.

PR #1898, opened 2024-06-20
Raise error when encountering empty InChi string

mattwthompson:
This is the surface-level fix for #1897 

- [ ] Raise same error from `to_inchikey`?
- [x] Tag issue being addressed
- [x] Add [tests](https://github.com/openforcefield/openff-toolkit/tree/main/openff/toolkit/_tests)
- [ ] Update docstrings/[documentation](https://github.com/openforcefield/openff-toolkit/tree/main/docs), if applicable
- [x] [Lint](https://open-forcefield-toolkit.readthedocs.io/en/latest/developing.html#style-guide) codebase
- [ ] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/main/docs/releasehistory.md)

Issue #1897, opened 2024-06-20
to_inchi() returns empty string on failure


lilyminium:
**Describe the bug**

`"[H:5][S:3]#[N+:2][S:1][H:4]"` is a molecule that cannot be turned into an InChI due to what looks like an internal software error in the inchi code itself. However, the toolkit returns an empty string upon this failure instead of raising an error, which can be problematic when mass-converting a bunch of molecules while being unaware that this can happen.

**To Reproduce**

```python
from openff.toolkit import Molecule
from openff.toolkit.utils import OpenEyeToolkitWrapper, RDKitToolkitWrapper

Molecule.from_mapped_smiles("[H:5][S:3]#[N+:2][S:1][H:4]").to_inchi(toolkit_registry=OpenEyeToolkitWrapper())
Molecule.from_mapped_smiles("[H:5][S:3]#[N+:2][S:1][H:4]").to_inchi(toolkit_registry=RDKitToolkitWrapper())
```

**Output**

Both toolkits yield warnings along the lines of `Warning: No description(-9995)`, but the OpenFF toolkit still returns an empty string. Could we please raise an error if there's an empty string, just to really warn the user?

**Computing environment (please complete the following information):**
 - Operating system
 - Output of running `conda list`

**Additional context**

mattwthompson:
> Could we please raise an error if there's an empty string, just to really warn the user?

I've done this in #1898 - do we care about the same thing for InChi keys as well?

lilyminium:
Thanks for the quick fix @mattwthompson!!

I haven't checked the behaviour of inchi key but as a user I'd like to see an error there as well for similar reasons.

PR #1904, opened 2024-07-09
Also raise error on empty InChi key

mattwthompson:
This is a brief follow-on to #1898 and should finish out #1897

IIUC, the key is just a hash of the InChi representation so it can be tested with the same setup. A counter-example would be a case in which either the InChi or its key can be empty, but not the other. Based on this, I didn't make a new error, but that could be done.

### DIFF
--- a/openff/toolkit/_tests/test_toolkits.py
+++ b/openff/toolkit/_tests/test_toolkits.py
@@ -843,15 +843,18 @@ class TestOpenEyeToolkitWrapper:
         with pytest.raises(InChIParseError, match="ksbfksfksfksbfks"):
             Molecule.from_inchi(inchi, toolkit_registry=toolkit)
 
-    def test_empty_inchi(self):
+    @pytest.mark.parametrize("method", ["to_inchi", "to_inchikey"])
+    def test_empty_inchi(self, method):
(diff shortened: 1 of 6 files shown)


openff/toolkit/_tests/test_toolkits.py
```diff
@@ -843,15 +843,18 @@ def test_from_bad_inchi(self):
...
+            match="failed to generate" + (".*key" if method == "to_inchikey" else ""),
         ):
-            Molecule.from_mapped_smiles("[H:5][S:3]#[N+:2][S:1][H:4]").to_inchi(
-                toolkit_registry=OpenEyeToolkitWrapper()
-            )
+            # call either .to_inchi or .to_inchikey
+            getattr(
+                Molecule.from_mapped_smiles("[H:5][S:3]#[N+:2][S:1][H:4]"),
+                method,
+            )(toolkit_registry=OpenEyeToolkitWrapper())
```
j-wags, line 857: faaaancy

j-wags (approved):
Looks great, thanks @mattwthompson. Thanks for catching the missing releasenotes, I suggested one change, but you can merge without it if you disagree.

PR #1912, opened 2024-07-26
[DNM] Test CI on osx-ar64

mikemhenry:
RE: #1911 I wanted to see what sort of stuff crops up in CI, I think enough of the ecosystem on conda-forge has osx-arm64 builds, but it also might not work at all.

Issue #1911, opened 2024-07-26
Partial charge methods fail in clean environment


slochower:
**Describe the bug**
In a new, clean OpenFF environment on an x86-64 Mac (14.4.1), partial charge methods fail. This is either from `antechamber` or possibly `sqm` and probably related to issues that I've seen @mattwthompson talking about with the Amber folks.

For this test script:

```python
from openff.toolkit.utils import get_data_file_path
from openff.toolkit.topology import Molecule, Topology
from openff.toolkit.typing.engines.smirnoff import ForceField

# This prepared PDB file from the toolkit's test suite is a box of solvents
pdb_path = get_data_file_path(
    "systems/packmol_boxes/propane_methane_butanol_0.2_0.3_0.5.pdb"
)
molecules = [Molecule.from_smiles(smi) for smi in ["CCC", "C", "CCCCO"]]

# The OpenFF Toolkit can directly read PDB files!
topology = Topology.from_pdb(pdb_path, unique_molecules=molecules)

# Construct the Interchange with the OpenFF "Sage" force field
interchange = ForceField("openff-2.0.0.offxml").create_interchange(topology)
```

I'm seeing:

```
 ToolkitWrapper around The RDKit version 2024.03.5 <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : partial_charge_method 'am1bcc' is not available from RDKitToolkitWrapper. Available charge methods are {'mmff94': {}, 'gasteiger': {}}
 ToolkitWrapper around AmberTools version 23.6 <class 'subprocess.CalledProcessError'> : Command '['antechamber', '-i', 'molecule.sdf', '-fi', 'sdf', '-o', 'charged.mol2', '-fo', 'mol2', '-pf', 'yes', '-dr', 'n', '-c', 'bcc', '-nc', '0.0']' returned non-zero exit status 1.
 ToolkitWrapper around Built-in Toolkit version None <class 'openff.toolkit.utils.exceptions.ChargeMethodUnavailableError'> : Partial charge method "am1bcc"" is not supported by the Built-in toolkit. Available charge methods are {'zeros': {'rec_confs': 0, 'min_confs': 0, 'max_confs': 0}, 'formal_charge': {'rec_confs': 0, 'min_confs': 0, 'max_confs': 0}}
```

The test cases also fail.

**To Reproduce**
1. `conda create --name openff-clean`
2. `mamba install -c conda-forge openff-toolkit`
[...]

PR #1913, opened 2024-07-26
[DNM] Test osx arm64

mikemhenry:
See #1912

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-12]
+        os: [ubuntu-latest, macos-12, macos-14]
         python-version: ["3.10", "3.11", "3.12"]
         rdkit: [true, false]
         openeye: [true, false]
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-12, macos-14]
+        os: [macos-14]
         python-version: ["3.10", "3.11", "3.12"]
         rdkit: [true, false]
         openeye: [true, false]


mikemhenry:
I can add back in the other operating systems if we need to compare output, but I figured as a first pass we can see what sort of errors crop up.

mikemhenry:
@j-wags Things are green, but lots of skips, know I know there are some tests for openeye and others for rdkit, but are any tests skiped via a heuristic like "antechaber failed so we will skip tests that use it" since that would give us misleading picture of what is going on. `= 2302 passed, 9840 skipped, 7 xfailed, 2 xpassed, 282 warnings in 170.03s (0:02:50) =`

j-wags:
> are any tests skiped via a heuristic like "antechaber failed so we will skip tests that use it"

Not to the best of my knowledge. If RDK+AT are installed, the tests for those backends should be unconditionally run.

Thanks for your iteration on this. I'm out of the office today but can follow up on anything else Monday!

mattwthompson:
FWIW these fall under the umbrella of our ecosystem tests https://github.com/openforcefield/status/actions/runs/10136579213/job/28025563565

mikemhenry:
So is there a test case missing that #1911 is exposing OR is it "we can't reproduce these errors on our test infrastructure?"

mikemhenry:
Also thanks Matt, I didn't think to look there to see if you are testing on osx-arm64!

mattwthompson:
> So is there a test case missing ... OR is it "we can't reproduce these errors on our test infrastructure?"

I strongly suspect it's the latter since reproducing it in isolation seems to be difficult.

FWIW, `osx-arm64` tests have not been flaky in the months we've had them, at least not with respect to charge assignment. In principle, we want to wrap all of that investigation into a new unit tests, but that doesn't seem to be so simple for the steps needed to isolate what was reported there

.github/workflows/CI.yml
```diff
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-12]
+        os: [macos-14]
```
mattwthompson, line 30:
```suggestion
        os: [ubuntu-latest, macos-12, macos-latest]
```
This change is strictly unnecessary but I think it'd be useful to have here instead of downstream

mikemhenry: Ah I had intended to just test osx-arm64 -- but would you like to get osx-arm64 added into your CI matrix?

mattwthompson: up to @j-wags in this project but it makes sense to test on all three IMHO

j-wags:
Sorry to be off in orbit for a year here, and thanks for pitching in @mikemhenry. This has been superseded by https://github.com/openforcefield/openff-toolkit/pull/1973 (which does indeed just do what this was proposing 🤦 )

Issue #1917, opened 2024-08-01
`Topology.identical_molecule_groups` is slow and order-dependent with some `_SimpleMoleucles`


mattwthompson:
**Describe the bug**

Repeated calls to `_SimpleAtom.molecule_atom_index` result in `Topology.identical_molecule_groups` being quite slow for some systems, at least when
* Molecules are represented, at least in part, by `_SimpleMolecule`s
* There is a single large molecule and many copies of a smaller molecule (i.e. a protein with waters)
* The large molecule is listed first

On its face (i.e. just reading the reproduction without context) this sounds like a wild use case, but it's only the basis by which `Interchange.from_openmm` is able to process externally-prepared systems (at least now [^1]). OpenMM infrastructure in particularly strong with protein force fields in ways that OpenFF's is not yet, so this is a common state for users to end up at.

**To Reproduce**

```python
import time

from openff.toolkit.topology._mm_molecule import _SimpleMolecule
from openff.toolkit import Molecule, Topology


# wget -O protein.pdb \
# https://raw.githubusercontent.com/openforcefield/protein-ligand-benchmark/a0b67b93c4f12f63a8bb417a1f6e4df4dc10c6c5/data/shp2/01_protein/crd/protein.pdb
protein = _SimpleMolecule.from_molecule(Topology.from_pdb("protein.pdb").molecule(0))
water = _SimpleMolecule.from_molecule(Molecule.from_smiles("O"))

n = 100

for order in ("before", "after", "middle"):
    match order:
        case "before":
            topology = Topology.from_molecules([protein] + n * [water])
        case "after":
            topology = Topology.from_molecules(n * [water] + [protein])
        case "middle":
            topology = Topology.from_molecules(
                int(n / 2) * [water] + [protein] + int(n / 2) * [water]
            )

    start = time.time()

    topology.identical_molecule_groups

    print(order, round(time.time() - start, 3))

    del topology
```

**Output**

```shell
$ sudo py-spy record -o profile.svg -- python protein-check.py                                          8:06:11  ☁  fix-1022 ☂
py-spy> Sampling process 100 times a second. Press Control-C to exit.

LICENSE: Could not open license file "oe_license.txt" in local directory
LICENSE: N.B. OE_LICENSE environment variable is not set
LICENSE: N.B. OE_DIR environment variable is not set
LICENSE: No product keys!
LICENSE: No product keys!
LICENSE: No product keys!
LICENSE: No product keys!
[...]

PR #1918, opened 2024-08-01
Cache indexing of atoms in molecules

mattwthompson:
One possible solution for #1917


j-wags (approved):
greatest PR of all time?

j-wags:
Yeah, I looked into the same thing, but there doesn't seem to be any way in the public API to modify a SimpleMolecule. If this had been a topology index, then there would have been some danger from situations where people add a molecule from one topology into another. But since it's just a molecule property it should be safe.

PR #1919, opened 2024-08-01
Update some inherited method signatures flagged by mypy 1.11

j-wags:
- [x] Update some method signatures that don't agree with inherited base class


mattwthompson (approved):
LGTM and meshes with the documented changes in `mypy`, happy it doesn't seem to have any side effects

IMO it's worth re-visiting if these classes are the right way to handle these data since some stuff has changed in the ecosystem over the years. Worth re-visiting ... someday that's not now

j-wags:
Agree, ParameterList always quacked more like a dict than a list. But that's for another day.

PR #1923, opened 2024-08-05
Fix indexing while adding molecules to topologies

### LINKED ISSUE #1922: `Topology.add_molecule` returns 1-indexed indices, toolkit elsewhere uses 0-index
**Describe the bug**

`Topology.add_molecule` returns an index, implied to be the index of the returned molecule. This value is 1-indexed, which is IMHO not in line with user expectations of the toolkit and Python ecosystem broadly. This quirk is [not documented](https://docs.openforcefield.org/projects/toolkit/en/stable/api/generated/openff.toolkit.topology.Topology.html#openff.toolkit.topology.Topology.add_molecule) and I suspect completely accidental. I don't believe anybody actually uses this behavior, but if it's there it should be correct (or documented as a quirk).

The cause is ultimately [here](https://github.com/openforcefield/openff-toolkit/blob/0.16.2/openff/toolkit/topology/topology.py#L2399-L2401) - the 0-index of the last molecule in a topology is the number of molecules _minus one_ - all code paths I checked ended up here by one means or another.

**To Reproduce**

```python
>>> from openff.toolkit import *
>>> topology = Topology()
>>> added_molecule_index = topology.add_molecule(Molecule.from_smiles("O"))
>>> topology.molecule(added_molecule_index)
  File "/Users/mattthompson/software/openff-toolkit/openff/toolkit/topology/topology.py", line 717, in molecule
    return self._molecules[index]
IndexError: list index out of range
>>> added_molecule_index
1
```

 - Operating system
 - Output of running `conda list`: recent environment, using 0.16.2

mattwthompson:
Fixes #1922

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -16,6 +16,8 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ### New features
 
+- [PR #1923](https://github.com/openforcefield/openff-toolkit/pull/1923): `Topology.add_molecule` and `Topology.add_molecules` now return indices starting at 0. Previously, these indices started at 1, which does not match other indexing in the toolkit.
+
 ### Improved documentation and warnings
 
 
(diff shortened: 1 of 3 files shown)


j-wags (approved):
Thanks @mattwthompson!

PR #1925, opened 2024-08-07
Add question to FAQ about sending a system to HPC

Yoshanuikabundi:
This PR adds a question to the FAQ describing how to transfer a system prepared from OpenFF tools to HPC.

[Rendered](https://openff-docs--1925.org.readthedocs.build/projects/toolkit/en/1925/faq.html#how-can-i-transfer-my-prepared-system-to-hpc-resources-for-simulation)


mattwthompson:
Would you consider having this live in the Interchange docs and have the FAQ entry point there?

My thinking is that
* this is an important topic (I've heard it multiple times from pharma partners)
* it's important to have it in this FAQ since it's a common landing point
* it's something that would evolve over time
* we could have short instructions on how different engines might need special treatment
* ... okay, I admit this is something that should already be in Interchange's docs

Yoshanuikabundi:
Yeah I was actually thinking the Interchange docs would be a good place for this to live while I was writing it. 

TBH the whole FAQ is getting unwieldy - it kind of represents every undocumented Toolkit question from the last three or four years. It'd be good to spend a week or two at some point using it to design documentation for the Toolkit that covers the common stumbling blocks in one narrative sweep, and then simplify the questions and have the answers just point to the parts of the docs that address the issue. 

The goal would be for users who read the introductory documentation to have already had most of the questions answered, and then the FAQ can just pick up the pieces, rather than it being the de facto documentation page for the entire toolkit.

I'd actually really like to get to a place where we can move it to the top-level docs, and have it link to all the underlying projects as appropriate.

j-wags:
Agreed. The two good paths forward I see here are 

1. immediately migrate the Toolkit FAQ to an `openff-docs`-level page (with this addition)
2. Put this addition in the interchange docs initially, with the eventual goal of doing item 1 further down the line

But now that you've written this nice prose, I see how interchange-y the process and concept is, so I'm hesitant to take a PR that puts it here.

mattwthompson:
I'm a big +1 to this FAQ moving to the `openff-docs` level (on any timeline)

j-wags (approved):
Just talked with @Yoshanuikabundi - Instead of keeping track of two things (this FAQ addition AND the FAQ migration), let's handle this serially. So this PR first, and then migrate (and possibly refactor) the FAQ.

PR #1941, opened 2024-10-01
Fix CI

mattwthompson:


j-wags:
Mysteriously, I'd noticed that Toolkit CI wasn't actually failing (though I'd totally expected it to be). But patching never hurts since we'll need to UNpin the micromamba action in all our repos on some future date.

PR #1951, opened 2024-10-24
Remove unused fixture(s)

mattwthompson:
I found this [nice plugin](https://pypi.org/project/pytest-deadfixtures/) while debugging some unforced errors in another project. Fortunately it looks like the toolkit isn't hanging onto too much dead code in this respect

```shell
$ git checkout 0.16.5 && pytest --dead-fixtures                                                 8:30:21  ☁  0.16.5 ☀
HEAD is now at f8110b56 Fix merge (#1943)

Hey there, I believe the following fixture(s) are not being used:
Fixture name: create_circular_handler_dependencies, location: openff/toolkit/_tests/test_forcefield.py:1262
Fixture name: alkethoh_forcefield, location: openff/toolkit/_tests/test_forcefield.py:771

1 error in 8.75s
```
`alkethoh_forcefield` is actually used in a few places, it seems like


j-wags (approved):
Good catch, thanks. It looks like this had come in with [a test for raising an error](https://github.com/openforcefield/openff-toolkit/pull/433/files#diff-97717c1ca13fbf0c1aea9fbcae5bd27c94c30cd560ab9d8b0f2d3395140adcecR717) when ParameterHandler dependencies created an un-solvable situation, and it looks like that went away in version 0.11, when Interchange started being used as a backend while the fixture was left behind. I don't see Interchange importing this fixture so I think it's safe to remove.

PR #1959, opened 2024-11-06
Defer/fix some minor import issues

mattwthompson:
I'm hunting down some issues with slow import times. Some are likely due to OpenMM sneaking in where it doesn't need to be (might be fixed upstream https://github.com/openforcefield/openff-units/pull/97#issuecomment-2459903434) but I found a few other minor improvements when digging.

### DIFF
--- a/openff/toolkit/topology/topology.py
+++ b/openff/toolkit/topology/topology.py
@@ -31,7 +31,6 @@ from typing import (
 
 import numpy as np
 from numpy.typing import NDArray
-from openff.units import ensure_quantity
 from typing_extensions import TypeAlias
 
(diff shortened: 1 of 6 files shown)


openff/toolkit/topology/topology.py
```diff
@@ -31,7 +31,6 @@
 
 import numpy as np
 from numpy.typing import NDArray
-from openff.units import ensure_quantity
```
mattwthompson, line 34: This method is in `openff.units.openmm`, which means importing it results in an `import openmm` call whenever anything in this module is imported. Obviously it's going to be in memory for many use cases, but it's unnecessary to import it here

PR #1969, opened 2024-11-26
Update linters

mattwthompson:
Development currently gets a questionable negative in linting because `.github/workflows/lint.yml` always points to the most recent release of all linters, so when one updates with changes everything is broken.

`pre-commit` allows for slower cadences of updating these, but syncing them requires manual updates every time something changes


j-wags (commented):
Hm, I think we should switch from auto-updated linter versions to manually-updated ones. Linting is a nice-to-have but in the event that auto-updates cause a conflict with merging a functional PR (as they just did), I'd rather not have the auto-updates. I know this means we'll periodically need to manually update the linter versions (which we'll say we'll do and then forget) but that's preferable to having linting be a sporadic point of CI failure (since we already have plenty of those).

Following this reasoning, what would you think of removing `lint.yml` entirely and having the pre-commit action run the show? It looks like they're doing the same things currently, and using the latter lets us manually control the linter versions in one place.

mattwthompson:
> I think we should switch from auto-updated linter versions to manually-updated ones

I am -1 on this (in a vacuum). 

> what would you think of removing lint.yml entirely

I am +1 on this

I think you grok the conflict in how these two things work at the moment, but to put it to paper for posterity and passers-by:

A very intentional part of the design of `pre-commit` is that individual hooks point to (static) specific versions/revisions. This ensures that code changes and tooling updates happen in completely isolated contexts and don't cross streams. Note that [here](https://github.com/openforcefield/openff-toolkit/blob/a1d2bee331844aac437a3c545c4bda8a4e857212/.pre-commit-config.yaml#L6) the version is not the most recent but is what most recently passed with the rest of the codebase. If a linter has any behavior change, it's _only_ going to show up in the autoupdate PRs, not in development PRs. By comparison, `lint.yaml` just grabs whatever was most recently release, so any (impactful) behavior change in one of these tools immediately causes confusion

j-wags:
> If a linter has any behavior change, it's only going to show up in the autoupdate PRs, not in development PRs

Ahh, ok. Thanks for the explanation. Then I agree with removing the `lint.yml` workflow and leaving the pre-commit one with autoupdate.

j-wags (approved):
Thanks @mattwthompson!

PR #1996, opened 2025-01-14
Add `ForceField.combine`

### LINKED ISSUE #1948: Combine force fields in-memory

I frequently wish to combine force fields, like Sage with a different water model. This works fine when loading from disk: 

```python
sage_with_opc = ForceField("openff-2.2.1.offxml", "opc.offxml")
```

However, if I'm in the middle of a workflow and have already loaded them up into memory, there's no straightforward way to combine `ForceFields`.

It'd be nice to be able to do something like
```python
sage = SmallMoleculeForceFieldProvider.get("openff")
opc = WaterModelProvider.get("opc")
sage_with_opc = sage.combine(opc)
```

Parameter ordering, parameter section compatibility, etc. should follow the existing behavior.

I can handle this with an extra trip to disk, but this is slow and I'd argue should not be necessary.

```python
sage.to_file("file1.offxml")
opc.to_file("file2.offxml")
sage_with_opc = ForceField("file1.offxml", "file2.offxml")
```

**Additional context**

Combining force fields has behavior consequences which are potentially significant and often result in worse performance. Combining force fields, however, is [part of the specification](https://openforcefield.github.io/standards/standards/smirnoff/#multiple-smirnoff-representations-can-be-processed-in-sequence) and currently supported in de-serialization, so I don't think it's reasonable to claim this behavior out to be unsupported.

mattwthompson:
Resolves #1948


mattwthompson:
Failures appear to be identical to upstream (https://github.com/openforcefield/openff-toolkit/issues/1997)

mattwthompson:
Those type-checker failures are surprising given that that `topology.py` isn't updated here - can have a look tomorrow if you wish

openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -1137,6 +1140,29 @@ def to_file(
...
+        import copy
+
+        combined = copy.deepcopy(self)
+
+        combined._load_smirnoff_data(
+            smirnoff_data=other._to_smirnoff_data(
+                discard_cosmetic_attributes=False,
+            ),
+            allow_cosmetic_attributes=allow_cosmetic_attributes,
+        )
```
j-wags, line 1162: (blocking) After some thought, I don't think that `allow_cosmetic_attributes` should be exposed on this method. My thinking is that, if a user wants to do a nonstandard/dangerous thing, they should need to opt-in with something like `allow_cosmetic_attributes=True`, but once they've opted in we should get out of their way. It seems like, by the time the user is in a situation where they want to use `ForceField.combine` with a FF object with cosmetic attributes, they've already opted in during the construction of those FF objects. So I think that this named argument should be dropped and have it be hard-wired to `True` in the `_load_smirnoff_data` call below below. 

```suggestion
    def combine(
        self,
        other: "ForceField",
    ) -> "ForceField":
        """
        Combine this `ForceField` with another `ForceField`, returning a new `ForceField`.

        The same rules as `ForceField.__init__` are followed.
        """
        import copy

        combined = copy.deepcopy(self)

        combined._load_smirnoff_data(
            smirnoff_data=other._to_smirnoff_data(
                discard_cosmetic_attributes=False,
            ),
            allow_cosmetic_attributes=True,
        )
```

j-wags (changes requested):
One blocking change to make with the method design, and tests to correspondingly update for it. Otherwise this looks good, feel free to also update the releasenotes before I review again, but expect approval once the suggested changes are made.

j-wags (approved):
Awesome, thanks @mattwthompson!

Issue #1997, opened 2025-01-16
Scheduled CI failing


mattwthompson:
https://github.com/openforcefield/openff-toolkit/actions/runs/12799622748/job/35686014018

mattwthompson:
Note that I opened this _before_ #1822 was merged; the linked log above points to a few issues that stand on their own if not magically fixed

There are some new failures _after_ #1822 which are confusing on the surface; nothing obviously changed with the plugins we use but there is some cross-interaction complaining:

```
============================= test session starts ==============================
platform linux -- Python 3.10.16, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/runner/work/openff-toolkit/openff-toolkit
configfile: pyproject.toml
plugins: anyio-4.8.0, rerunfailures-15.0, xdist-3.6.1, nbval-0.11.0, timeout-2.3.1, cov-6.0.0
created: 2/2 workers
2 workers [0 items]



------------------------------ coverage ------------------------------
---------------------- coverage: failed workers ----------------------
The following workers failed to return coverage data, ensure that pytest-cov is installed on these workers.
gw0
gw1
Coverage XML written to file coverage.xml

============================ no tests ran in 1.19s =============================
```

https://github.com/openforcefield/openff-toolkit/actions/runs/12940806133/job/36095749431

Nothing too suspicious in the recent changes of the plugins we use:


https://pytest-cov.readthedocs.io/en/latest/changelog.html#id1

https://pytest-xdist.readthedocs.io/en/latest/changelog.html#pytest-xdist-3-6-1-2024-04-28

https://github.com/pytest-dev/pytest-randomly/blob/main/CHANGELOG.rst

j-wags:
The main things I found in #2004 are:
* I can't reproduce the failure locally, even though all my versions of python/pytest/etc seem to match up and I ran using the exact same keywords. When I run locally the tests do actually run. 
* Doing `pytest ... -m "slow"`(instead of `-m "slow or not slow"`) leads to a new and possibly informative failure in CI:

```
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/home/runner/micromamba/envs/openff-toolkit-test/lib/python3.10/site-packages/_pytest/main.py", line 283, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
[...]

PR #2004, opened 2025-01-29
Skip slow CI in nightly runs until failures can be debugged

j-wags:
- [x] Works around #1997 by disabling slow tests, but leaves it open for future resolution


j-wags:
Thanks for the input. It would take me a while to tease apart #1970 into the minimal changes to revert to advance this, so my current thinking is to temporarily disable the slow tests so that nightlies regain some of their utility, and add a high-priority ticket to the backlog to debug and re-enable them.

j-wags:
#2017 is a better solution

PR #2016, opened 2025-02-07
Type-checking maintenance updates

### LINKED ISSUE #2014: Mypy 1.15.0 changes
Mypy 1.15.0, released two days ago (between finishing and merging #2007, unfortunately), includes some performance optimizations but also change that break our CI.

mattwthompson:
Resolves #2014 

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -22,12 +22,11 @@ dependencies:
   - openff-interchange-base =0.4
   - openff-nagl-base >=0.4.0
   - openff-nagl-models >=0.3.0
+  - typing_extensions
     # Toolkit-specific
   - ambertools >=22
     # rdkit 2024.03.6 and 2024.09.1 packages fail when run natively on osx-arm64 macs
     # https://github.com/rdkit/rdkit/issues/7583
-    # and a bad stub (causing mypy failure only)
-    # https://github.com/rdkit/rdkit/issues/7221
   - rdkit !=2024.03.6,!=2024.09.1
 
   - openeye-toolkits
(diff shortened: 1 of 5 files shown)


devtools/conda-envs/test_env.yaml
```diff
@@ -22,12 +22,11 @@ dependencies:
   - openff-interchange-base =0.4
   - openff-nagl-base >=0.4.0
   - openff-nagl-models >=0.3.0
+  - typing_extensions
```
mattwthompson, line 25: This is a runtime requirement, not a "only when we run the type-checker" requirement. It's brought in by several other packages so it's not likely to be missed at runtime, but I'm listing it here according to the philosophy of listing all (non-optional) runtime requirements in this sections

pyproject.toml
```diff
@@ -76,6 +76,7 @@ exclude_lines = [
 ]
 
 [tool.mypy]
+python_version=3.12
```
mattwthompson, line 79: If I'm reading things correctly, this setting allows for Mypy to "use" a particular version of Python no matter what is actually installed in the environment. I think this is a better setting to use than having the conditional live in the CI workflow because this (should!) ensure that all developers get a similar Mypy experience. I've been bit by something like this before

devtools/conda-envs/test_env.yaml
```diff
@@ -50,13 +49,3 @@ dependencies:
   - nbval
   # No idea why this is necessary, see https://github.com/openforcefield/openff-toolkit/pull/1821
   - nomkl
-  - mypy
```
j-wags, line 53: (not blocking) just double checking that this removal is intended

j-wags (approved):
Thanks @mattwthompson!

devtools/conda-envs/test_env.yaml
mattwthompson: Yep - it's not run when this environment file is used

Issue #2008, opened 2025-02-03
`Topology.identical_molecule_groups` scales poorly with many large molecules


mattwthompson:
**Describe the bug**

`Topology.identical_molecule_groups` scales super-linearly with molecule size when multiple large components are present in a system. This makes parametrizing large polymer systems unworkably slow. I believe this is a root cause of https://github.com/openforcefield/openff-interchange/issues/1156

**To Reproduce**

This is a simple and imperfect reproduction, but it shows that when multiple copies of a large molecule are present, runtime explodes when that molecule is a few hundred or thousand heavy atoms, which is not necessarily a large system in materials science.

```python
import time
import numpy
from matplotlib import pyplot
from openff.toolkit import Molecule, Topology
import pathlib


def time_isomorphism(
    num_heavy_atoms: int,
    num_molecule_copies: int) -> float:
    molecule = Molecule.from_smiles("C" * num_heavy_atoms)

    file_name = f"topologies/{num_heavy_atoms}-{num_molecule_copies}.json"

    if pathlib.Path(file_name).exists():
        topology = Topology.from_json(open(file_name).read())
    else:
        topology = Topology.from_molecules(num_molecule_copies * [molecule])

        open(file_name, "w").write(topology.to_json())
        
    start_time = time.perf_counter()
    topology.identical_molecule_groups

    return_time = time.perf_counter() - start_time

    print(f"n={num_heavy_atoms=}, m={num_molecule_copies}, time={round(return_time, 10)}")

    del topology

    return time.perf_counter() - start_time

fig, (ax1, ax2) = pyplot.subplots(2, 1)

for m in numpy.logspace(0, 2.0, num=3, dtype=int):
    n = numpy.unique(numpy.logspace(0, 4.0, num=41, dtype=int))
    times = numpy.array([time_isomorphism(_n, m) for _n in n])
    ax1.plot(n, times, 'o-', label=m)
    ax2.plot(n, times/m, 'o-', label=m)
    print(f"finished {m=}")

ax1.legend(loc=0, title='num copies of each molecule')
ax1.set_xlabel("Alkane length $C_n$")
ax1.set_ylabel("total runtime (seconds)")
ax1.set_xscale("log")
ax1.set_yscale("log")

ax2.set_xscale("log")
ax2.set_ylabel("runtime per copy (seconds)")

fig.show()
```
**Output**

[...]

PR #2025, opened 2025-02-23
Speed up identical molecule detection

j-wags:
- [x] Addresses #2008 
- [x] Document change in Molecule.ordered_connection_table_hash

Using the excellent reproducing code snippet from #2008 (though restricting mol size to 10^3 instead of 10^4) we see about a 50x speedup

Before
Screenshot 2025-02-23 at 1 32 11 PM

After 
Screenshot 2025-02-23 at 1 31 53 PM


mattwthompson:
I'd like to volunteer be the reviewer once you think it's ready for feedback

hannaomi:
Hi both,
before_commit
after_commit

Many thanks for your patience whilst I collected the rest of the data. As you can see I did not see a significant difference in the create interchange run time at higher unique molecule numbers, despite initially seeing a time saving with smaller topologies. Perhaps there is a significant difference in the way I am preparing my topologies vs. the reproducing code?

@j-wags  happy to talk this over in our meeting on Thursday if that would work for you?

j-wags:
As an update - @hannaomi and I met after the last message and she handed off a reproducing example of the whole workflow to me. @mattwthompson took a look at this example and opened several issues for performance improvement with the `polymer-performance` label. Those issues are getting triaged for our upcoming sprints.

mattwthompson:
Is there a downside of moving forward with this change now?

j-wags:
I'd put this together quickly as a proof of concept solution to this use case, without considering its effect on anything else. I'll give this a once-over today and will open for review shortly if it seems safe.

openff/toolkit/topology/molecule.py
```diff
@@ -1971,23 +1981,7 @@ def from_smiles(
...
-        for bond1, bond2 in zip(self.bonds, other.bonds):
-            if (
-                (bond1.atom1_index != bond2.atom1_index)
-                or (bond1.atom2_index != bond2.atom2_index)
-                or (bond1.is_aromatic != bond2.is_aromatic)
-                or (bond1.stereochemistry != bond2.stereochemistry)
-            ):
-                return False
-        return True
+        return self.ordered_connection_table_hash() == other.ordered_connection_table_hash()
```
Yoshanuikabundi, line 1984:
```suggestion
        self_id = (
            tuple((atom.atomic_number, atom.formal_charge.magnitude, atom.stereochemistry) for atom in self.atoms),
            tuple((bond.bond_order, bond.stereochemistry, bond.atom1_index, bond.atom2_index) for bond in self.bonds),
        )
        other_id = (
            tuple((atom.atomic_number, atom.formal_charge.magnitude, atom.stereochemistry) for atom in other.atoms),
            tuple((bond.bond_order, bond.stereochemistry, bond.atom1_index, bond.atom2_index) for bond in other.bonds),
        )
        return self_id == other_id
                for atom1, atom2 in zip(self.atoms, other.atoms):
```

openff/toolkit/topology/molecule.py
```diff
@@ -1255,13 +1255,23 @@ def __hash__(self):
...
+        compared with hashes generated using different versions.
+        """
         if self._ordered_connection_table_hash is not None:
             return self._ordered_connection_table_hash
 
+        # Pre-assign molecule atom indices in O(N) time to avoid use of List.index to get index of each one
+        # in O(N^2) time
+        for index, atom in enumerate(self.atoms):
+            atom._molecule_atom_index = index
+
```
Yoshanuikabundi, line 1271: This could be moved to `_add_atom()`

Yoshanuikabundi (changes requested):
This speedup is great! 

I think the approach of checking for equality by comparing Python builtin `hash()`es of the relevant information is very dangerous. As written, this PR doesn't actually do the same thing as the original code because of the possibility of a hash collision. Python's `hash()` builtin doesn't use a cryptographic hash function and is quite vulnerable to collisons - The canonical example is `hash(-2) == hash(-1)` (eg, see <https://stackoverflow.com/a/68048420>). A hash in, say, a dict is only used to narrow down the search, the keys are still directly compared - the actual datastructure is like `dict[int, list[tuple[KeyType, ValueType]]`, with the list hopefully being length 1. The builtin hash function also salts string inputs, so hashes are not consistent from one Python process to the next even in the same environment (try running `python -c 'print(hash("hi"))'` a few times in a row, and see <https://docs.python.org/3/reference/datamodel.html#object.__hash__>) - this might cause a serious issue if a `Molecule` gets pickled. 

I think the chances of a hash collision are worth taking seriously - partly because the hash, on my machine, is only 64 bits long, much shorter than the string that would be fed into it, but more because it would be such an infuriating and embarrassing bug to run into in the unlikely case that it came up.

I've done some testing and it seems that the vast majority of the speedup comes from looping over the atoms and assigning their `_molecule_atom_index` attribute. Once that's done, the difference between comparing hashes or just comparing the big `id` string that gets hashed is basically nil. I've also found that comparing tuples of the attributes you want to compare is much faster than iterating over the actual attributes; in other words,

```py
    def _is_exactly_the_same_as(self, other):
        self_id = (
            tuple((atom.atomic_number, atom.formal_charge.magnitude, atom.stereochemistry) for atom in self.atoms),
            tuple((bond.bond_order, bond.stereochemistry, bond.atom1_index, bond.atom2_index) for bond in self.bonds),
        )
        other_id = (
            tuple((atom.atomic_number, atom.formal_charge.magnitude, atom.stereochemistry) for atom in other.atoms),
            tuple((bond.bond_order, bond.stereochemistry, bond.atom1_index, bond.atom2_index) for bond in other.bonds),
        )
        return self_id == other_id
```

Is about 5x faster (in this case where the molecules are large and the same) than

```py
    def _is_exactly_the_same_as(self, other):
        for atom1, atom2 in zip(self.atoms, other.atoms):
            if (
                (atom1.atomic_number != atom2.atomic_number)
                or (atom1.formal_charge != atom2.formal_charge)
                or (atom1.is_aromatic != atom2.is_aromatic)
                or (atom1.stereochemistry != atom2.stereochemistry)
            ):
                return False
        for bond1, bond2 in zip(self.bonds, other.bonds):
            if (
                (bond1.atom1_index != bond2.atom1_index)
                or (bond1.atom2_index != bond2.atom2_index)
                or (bond1.is_aromatic != bond2.is_aromatic)
                or (bond1.stereochemistry != bond2.stereochemistry)
            ):
                return False
        return True
```

In fact, that first implementation if `_is_exactly_the_same_as()` is just as fast as computing and comparing the hashes if the molecule atom indices are already assigned.

So I'd propose just assigning `_molecule_atom_indices` in `Molecule._add_atom()` (the value is already computed in the return statement!), and then changing `_is_exactly_the_same_as` to construct and compare big tuples. This would mean that there would not need to be any change in the `ordered_connection_table_hash()` function, though I would suggest we still update its docstring to say that hashes cannot be used across different Python processes.

If that approach is no good, I think this is at least blocked by needing a collision-resistant hash function like `hashlib.sha3_224`. Incidentally, this would not involve any salting, so hashes could be compared across Python processes (within a given version of the Toolkit) - however it would mean changing the return type from `int` to `bytes`. It feels like overkill, but since it didn't slow the benchmark down at all on my machine, I think it's worth it.

openff/toolkit/topology/molecule.py
j-wags: That's a complicated enough change (will become wrong if we ever allow more than just appending atoms, then this would need to be handled as an invalidatable cached property) that I'm gonna leave this as-is until a significant runtime benefit is demonstrated.

openff/toolkit/topology/molecule.py
```diff
@@ -1255,17 +1256,27 @@ def __hash__(self):
         return hash(self.to_smiles())
 
     def ordered_connection_table_hash(self) -> int:
-        """Compute an ordered hash of the atoms and bonds in the molecule"""
+        """
+        Compute an ordered hash of the atoms and bonds in the molecule.
+        This hash method is intended for comparison of Molecule objects at
+        runtime, and hashes from one version of the software should not be
+        compared with hashes generated using different versions.
+        """
```
Yoshanuikabundi, line 1264:
```suggestion
        """
        Compute an ordered hash of the atoms and bonds in the molecule.
        
        This hash method is intended for comparison of Molecule objects at
        runtime, and hashes from one version of the software should not be
        compared with hashes generated using different versions.
        """
```
Sorry for the pedantry.

openff/toolkit/topology/molecule.py
```diff
@@ -1971,23 +1982,15 @@ def from_smiles(
...
-        for bond1, bond2 in zip(self.bonds, other.bonds):
-            if (
-                (bond1.atom1_index != bond2.atom1_index)
-                or (bond1.atom2_index != bond2.atom2_index)
-                or (bond1.is_aromatic != bond2.is_aromatic)
-                or (bond1.stereochemistry != bond2.stereochemistry)
-            ):
-                return False
-        return True
+        self_id = (
```
Yoshanuikabundi, line 1985:
```suggestion
        # Pre-assign molecule atom indices in O(N) time to avoid use of List.index to get index of each one
        # in O(N^2) time
        for index, atom in enumerate(self.atoms):
            atom._molecule_atom_index = index
        for index, atom in enumerate(other.atoms):
            atom._molecule_atom_index = index

        self_id = (
```

Sorry for burying this point earlier - My finding was that the majority of the speedup from using the hash function comes from pre-assigning molecule atom indices, and not from the actual hash comparison. To get the speedup, they need to be somewhere - I think when I suggested the tuple comparison it was in the context of moving them to `_add_atom()`. So if that's not happening, this needs to be here.

Yoshanuikabundi (approved):
Just one more change to make sure we're still getting the main optimization and this is good to go!

openff/toolkit/topology/molecule.py
j-wags: Ah, gotcha. Thanks for the clarification!

hannaomi:
Hi all,

I have reran my reproducing example with this new commit - much faster performance parameterizing 50 unique polymer chains in one topology.
1879c22

Many thanks!

j-wags:
Fantastic, thanks for the feedback! There's lots to unpack in the shape of that plot, but a 40x speedup for the worst case is good news. This PR is now in the OpenFF Toolkit 0.16.9 release, so you shouldn't need to install from the branch/commit any more :-)

Issue #1977, opened 2024-12-09
Cannot create InChi of large molecule


mattwthompson:
**Describe the bug**

The toolkit fails to generate InChi for molecules with more than 1024 atoms.

**To Reproduce**

In this example, C340 passes but C341 errors. This is where the number of atoms crosses 1024.

```python
from openff.toolkit import ForceField, Molecule
from openff.toolkit.utils.exceptions import EmptyInChiError


for n in [340, 341]:
    molecule = Molecule.from_smiles(n * "C")

    try:
        molecule.to_inchi()
    except EmptyInChiError as error:
        print(f"failed with {n=}")
        print(f"{error=}")
        continue

    print(f"passed with {n=}")

    # double-check that the SMILES conversion doesn't crash, and is probably correct
    assert n == len(molecule.to_smiles(explicit_hydrogens=False))
```

**Output**

With OpenEye installed and licensed:
```
passed with n=340
Warning: OECreateInChI: InChI only supports molecules with between 1 and 1023 atoms! (note: large molecule support is experimental)
failed with n=341
error=EmptyInChiError('OEChem failed to generate an InChI for the molecule.')
```

Without the license accessible:
```
passed with n=340
[14:27:42] ERROR: Too many atoms [did you forget 'LargeMolecules' switch?]

failed with n=341
error=EmptyInChiError('RDKit failed to generate an InChI for the molecule.')
```

**Computing environment (please complete the following information):**
 - Operating system macOS 15.1.1
 - Output of running `conda list`: https://gist.github.com/mattwthompson/39e450a1cb21e088debde9fc1756b677

**Additional context**

InChi versions prior to 1.05 (2017) were limited to 1024 atoms. OpenEye says they use version 1.06, so their error message is outdated in saying that InChi only supports molecules up to 1023 atoms and, as far as the release notes indices, that large molecule support is experimental.

I can't find what version of InChi is used by RDKit, but its error message indicates it uses 1.05+ and does support it with a flag being passed through.

https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemFunctions/OEMolToInChI.html?highlight=inchi#OEChem::OEMolToInChI

https://github.com/IUPAC-InChI/InChI/blob/v1.07.1/INCHI-1-DOC/CHANGELOG.md#v105-release-2017-02-04

mattwthompson:
RDKit does support this, though

PR #2040, opened 2025-03-27
Support large molecule InChi with RDKit

mattwthompson:
Partial fix for #1977 since OEChem insists on not allowing the user to pass the argument through to InChI's tool

Arguably the `/LargeMolecules` option should be opt-in, but that's an API addition with no obvious upside.

This would change, but not break, how NAGL uses InChI to key its lookup table (https://github.com/openforcefield/openff-nagl/pull/165/). Currently, it makes an attempt to gracefully handle `EmptyInChiError`; when RDKit is that case shouldn't be needed any more.

This still appears to be relatively fast for a large molecule, not accounting for the time it takes to instantiate `RDKitToolkitWrapper`, delegate out, etc.:

```python
In [6]: molecule = Molecule.from_smiles(1000 * "C")

In [7]: %timeit molecule.to_inchi(toolkit_registry=RDKitToolkitWrapper())
39.7 ms ± 1.54 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Simple demo:

```python
from openff.toolkit.utils.rdkit_wrapper import RDKitToolkitWrapper, EmptyInChiError

try:
    Molecule.from_smiles(342 * "C").to_inchi(toolkit_registry=RDKitToolkitWrapper())
except EmptyInChiError as error:
    print(error)
else:
    print("InChI fine ... ")

try:
    Molecule.from_smiles(342 * "C").to_inchikey(toolkit_registry=RDKitToolkitWrapper())
except EmptyInChiError as error:
    print(error)
else:
    print("InChI key fine ... ")
```
Out:
```console
$ git checkout upstream/main && python large_molecule_inchi.py
HEAD is now at 4cc57064 re-ignore external examples in conda tests (#2038)
[13:51:05] ERROR: Too many atoms [did you forget 'LargeMolecules' switch?]

RDKit failed to generate an InChI for the molecule.
[13:51:05] Invalid InChI prefix in generating InChI Key
RDKit failed to generate an InChI key for the molecule.
(openff-toolkit-test) [openff-toolkit]
$ git checkout rdkit-large-molecule-inchi && python large_molecule_inchi.py
Previous HEAD position was 4cc57064 re-ignore external examples in conda tests (#2038)
Switched to branch 'rdkit-large-molecule-inchi'
InChI fine ...
InChI key fine ...
```

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -19,7 +19,7 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 ### New features
 
 - [PR #1996](https://github.com/openforcefield/openff-toolkit/pull/1996): Adds `ForceField.combine`.
-- [PR #20XX](https://github.com/openforcefield/openff-toolkit/pull/20XX): Adds support for InChI with large molecules when using RDKit.
+- [PR #2040](https://github.com/openforcefield/openff-toolkit/pull/2040): Adds support for InChI with large molecules when using RDKit.
 
 ### Improved documentation and warnings
 
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -19,6 +19,7 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 ### New features
 
 - [PR #1996](https://github.com/openforcefield/openff-toolkit/pull/1996): Adds `ForceField.combine`.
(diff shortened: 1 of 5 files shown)


lilyminium (approved):
LGTM, both the PR iteself and its effect on NAGL -- thank you Matt!

PR #2054, opened 2025-05-22
Drop use of `importlib_metadata`

mattwthompson:
`importlib-metadata` is a third-party package which backports features of `importlib.metadata` (which is in the standard library) to older Python versions. We were using it as a bridge while it was "provisional" but that ship has long passed.

This change is likely 3.10+ based on my reading of the docs, but I haven't pinned it down completely: https://docs.python.org/3/library/importlib.metadata.html

This was *not* declared as a dependency in test environments - it probably should have been - but is a requirement in packaging. This change would allow that to be dropped from the conda recipe.

### DIFF
--- a/openff/toolkit/typing/engines/smirnoff/forcefield.py
+++ b/openff/toolkit/typing/engines/smirnoff/forcefield.py
@@ -13,6 +13,7 @@ Parameter assignment tools for the SMIRNOFF (SMIRKS Native Open Force Field) for
 import logging
 import os
 import pathlib
+from importlib.metadata import entry_points
 from typing import IO, TYPE_CHECKING, Any, Optional, Union
 
 from packaging.version import Version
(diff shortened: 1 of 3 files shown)


openff/toolkit/typing/engines/smirnoff/forcefield.py
```diff
@@ -13,6 +13,7 @@
 import logging
 import os
 import pathlib
+from importlib.metadata import entry_points
```
mattwthompson, line 16: This is arguably a style change but I will defend it on a few points

* I question whether or not it makes sense to have standard library imports guarded
* The code paths are pretty hot with respect to each module, i.e. almost all functionality that requires importing from these files will end up calling one of these code paths
* It's a pretty quick import, not counting anything else that's already brought in by other imports

```shell
$ time python -c ""                                                                        16:11:52  ☁  stdlib-entry-points ☂ ✭
python -c ""  0.01s user 0.01s system 51% cpu 0.049 total
(bespokefit-test) [openff-toolkit]
$ time python -c "from importlib.metadata import entry_points"                             16:12:22  ☁  stdlib-entry-points ☂ ✭
python -c "from importlib.metadata import entry_points"  0.05s user 0.02s system 67% cpu 0.105 total

PR #2081, opened 2025-07-09
Serialization/conversion performance improvements

mattwthompson:

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -154,7 +154,7 @@ jobs:
           # TODO: Flip back to schedule condition before merge
           PYTEST_ARGS+=" --runslow"
 
-          python -m pytest -x --durations=20 $PYTEST_ARGS $COV openff/toolkit/_tests
+          python -m pytest --durations=20 $PYTEST_ARGS $COV openff/toolkit/_tests
 
       - name: Run code snippets in docs
         if: ${{ matrix.rdkit == true && matrix.openeye == true }}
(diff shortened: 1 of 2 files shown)


mattwthompson:
Bummer, the behavior seems fine:

```python
In [3]: hash(ForceField("openff-2.0.0.offxml")['vdW'])
Out[3]: -7474900899989015509

In [4]: hash(ForceField("openff-2.1.0.offxml")['vdW'])
Out[4]: -7474900899989015509

In [5]: hash(ForceField("openff-2.2.1.offxml")['vdW'])
Out[5]: -2512421923988562758
```

and the raw performance isn't terrible:
```
In [6]: %%timeit
   ...: hash(ForceField("openff-2.2.1.offxml")['vdW'])
   ...:
   ...:
80.5 ms ± 1.44 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

but something somewhere is causing huge slowdowns, or this just doesn't scale well to large systems. That's what's up next.

openff/toolkit/utils/utils.py
```diff
@@ -163,10 +165,10 @@ def quantity_to_string(input_quantity: Quantity) -> str:
         The serialized quantity
 
     """
-    unitless_value: float | int | NDArray | list = input_quantity.m_as(input_quantity.units)
+    unitless_value: float | int | NDArray | list = input_quantity.m
```
mattwthompson, line 168: I don't think there was a reason to convert this quantity into the units it already has

openff/toolkit/typing/engines/smirnoff/parameters.py
```diff
@@ -1086,37 +1086,38 @@ def to_dict(
     def __getattr__(self, item):
         """Take care of mapping indexed attributes to their respective list elements."""
 
-        # Try matching the case where there are two indices
-        # this indicates a index_mapped parameter
-        attr_name, index, key = self._split_attribute_index_mapping(item)
+        if not item.isalpha():
```
mattwthompson, line 1089: This check is not free, but it improves the current design by not checking for indexed or indexed + mapped attributes if an attribute name is alphabetic

PR #2085, opened 2025-07-11
Actually format with `ruff`

mattwthompson:
When I switched this repo over to `ruff`, I botched an important detail. It is split into two parts, a linter (think Flake8) and a formatter (think Black). At the time, I neglected to add the _formatter_ to automation.

I tried to do this with minimal disruptions, but there just isn't a straightforward way to do that in such a large codebase. Most (estimate 70%+) of the changes are re-wrapping some things that Black was eager to unwrap in the past. This is evidenced by how the changes should be identical but the codebase is reduced by ~3000 LOC. (I'm fine with this, line lengths of ~80 are too narrow to me on modern displays.) This can be tinkered with by changing the line length config, but I left that in place.


j-wags:
Oh, some of these conflicts are substantial-ish. Would it be easy to reopen this PR from a fresh rerendering from main or is that a lot of manual work?

j-wags (approved):

Issue #2084, opened 2025-07-10
Atom stereochemistry setter does no input validation


mattwthompson:
**Describe the bug**

I think one would have to try to run into this problem, but at very least the docstring is incorrect 

https://github.com/openforcefield/openff-toolkit/blame/e71757033cd6a81a1e188d62e8b5d80186c61934/openff/toolkit/topology/molecule.py#L446-L460

**To Reproduce**

```python
>>> from openff.toolkit import Molecule
>>> water = Molecule.from_smiles("O")
>>> water.atom(2).stereochemistry
>>> water.atom(2).stereochemistry = "The Mythical California Zephyr"
>>> water.atom(2).stereochemistry
'The Mythical California Zephyr'
```

j-wags:
Agree, if the docstring says this can only have certain values, then we should make sure that's in line with the behavior

j-wags:
I made a relevant comment in https://github.com/openforcefield/openff-toolkit/pull/2086:

> I'm weakly against validation [in molecule.py], since we don't know which custom cheminformatics toolkitwrapper users might have, and there are valid stereo definitions beyond R and S. If we really want to validate, we could have the cheminformatics wrappers raise an exception if a stereochemistry they can't handle comes through.

PR #2000, opened 2025-01-23
Improve some typing

Yoshanuikabundi:
This PR includes many changes needed for OpenFF Pablo to type check, and 1 for it to run correctly. Pablo development so far has and will continue to test and type check against this branch (or `main` once it is merged) until it gets into a release.

- [x] improves typing as needed by OpenFF Pablo by adding and correcting many type annotations
- [x] Handle an exception previously raised when `Molecule.to_rdkit()` is run on a molecule with residue numbers that cannot be converted to `int` (such as hexadecimal residue numbers used by OpenMM in large PDB files)
- [x] Makes `Molecule.is_isomorphic_with()` so that its keyword arguments are actually keyword arguments. This is not a breaking change as the `**kwargs` argument is still present, it's just ignored - we should deprecate and remove this ASAP 
- [x] Update [changelog](https://github.com/openforcefield/openff-toolkit/blob/main/docs/releasehistory.md)

Sorry for the behemoth PR!

PR #2086, opened 2025-07-14
Update `Atom.stereochemistry` docstring

mattwthompson:
Interacts with #2084 but does **not** close it


j-wags (changes requested):
The actual commonly used values are `R` and `S` - we default to CIP stereo instead of local (`CW` and `CCW`).

I'm weakly against validation here, since we don't know which custom cheminformatics toolkitwrapper users might have, and there are valid stereo definitions beyond R and S. If we really want to validate, we could have the cheminformatics wrappers raise an exception if a stereochemistry they can't handle comes through.

We could make an issue for this, but I'd icebox it, so I'm not sure it's worth it. I'd recommend closing this PR as #2000 makes the correct changes

PR #2096, opened 2025-08-19
Allow NAGL model paths to be `pathlib.Path | str`

mattwthompson:
The `partial_charge_method` is meant to be a string, but some helper functions from `openff.nagl_models` return `list[pathlib.Path]`. This change sanitizes inputs in order to support both

### DIFF
--- a/openff/toolkit/utils/nagl_wrapper.py
+++ b/openff/toolkit/utils/nagl_wrapper.py
@@ -61,7 +61,7 @@ class NAGLToolkitWrapper(ToolkitWrapper):
     def assign_partial_charges(
         self,
         molecule: "Molecule",
-        partial_charge_method: str,
+        partial_charge_method: pathlib.Path | str,
         use_conformers: Optional[list["Quantity"]] = None,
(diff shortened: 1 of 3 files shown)


openff/toolkit/utils/nagl_wrapper.py
```diff
@@ -61,7 +61,7 @@ def is_available(cls) -> bool:
     def assign_partial_charges(
         self,
         molecule: "Molecule",
-        partial_charge_method: str,
+        partial_charge_method: pathlib.Path | str,
```
j-wags, line 64: I'm uncomfortable with this change because the ToolkitRegistry magic relies on ToolkitWrappers having the same API, and this pushes `NAGLToolkitWrapper` out of sync with `OpenEyeToolkitWrapper`, `RDKitToolkitWrapper`, and `BuiltInToolkitWrapper`. I can't name a specific thing that would go wrong by adding this but it invites some ugly edge cases down the road. 

And in practical terms, I think we can contain all the needed changes for our current set of issues to openff-nagl-models for faster turnaround.

PR #2097, opened 2025-08-19
Declare minimum Python version

mattwthompson:
The metadata of the Python project did not declare what Python version(s) are supported, at least the lower bound. This adds a one-liner into `pyproject.toml` to do just that. This will make Python-centric tooling and packaging more useful since this is where this information is supposed to live.

Some tooling is aware of this and, as a consequence, made some automated changes. I see mostly
* `Optional[Foo]` becomes `Foo | None`
  * https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/
* `Union[Foo, Bar]` becomes `Foo | Bar`
* `isinstance(mylist, (list, tuple))` becomes `isinstance(mylist, list | tuple)`
  * https://docs.astral.sh/ruff/rules/non-pep604-isinstance/
* Some imports from `typing_extensions` are in the standard library from the minimum supported Python
  * https://docs.astral.sh/ruff/rules/deprecated-import/

I don't love all of these changes, but they're automatic and consistent.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ ci:
 files: ^openff|(^examples/((?!deprecated).)*$)|^docs|(^utilities/((?!deprecated).)*$)
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.12.8
+  rev: v0.12.11
   hooks:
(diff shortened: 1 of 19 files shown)


j-wags (approved):
LGTM, thanks @mattwthompson!

PR #2101, opened 2025-08-25
Update conformer energies example

### LINKED ISSUE #2100: Update conformer energies example
https://github.com/openforcefield/openff-toolkit/tree/main/examples/conformer_energies

- says the toolkit produces deterministic charge
- uses OpenMM instead of Interchange.minimize

While it would be tempting to switch this over to use NAGL, don't do that unless openff-2.3.0 is out

mattwthompson:
Closes #2100

- [x] Use 2.2.1
- [x] Use `Interchange.minimize`
- [x] Do not use OpenMM units registry
- [x] Try to sync details between notebook and script

### DIFF
--- a/examples/conformer_energies/conformer_energies.ipynb
+++ b/examples/conformer_energies/conformer_energies.ipynb
@@ -21,24 +21,9 @@
    "metadata": {
     "tags": []
    },
-   "outputs": [
-    {
-     "data": {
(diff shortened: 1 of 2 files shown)


j-wags (approved):
One not-blocking comment, other than that this is good to go!

PR #2114, opened 2025-10-14
Speedup for molecule isomorphism checking

### LINKED ISSUE #2035: `Topology.identical_molecule_groups` is too slow
**Describe the bug**

Subgraph isormorphism is central to using the toolkit on multi-molecule systems but it is slow, especially for large and complicated systems.

**To Reproduce**

[topology.json.zip](https://github.com/user-attachments/files/19322417/topology.json.zip)

With a topology of fairly modest size and refactoring to use a Rust re-implementation of `networkx` (#2033), this takes about 10 minutes:

```python
In [1]: from openff.toolkit import Topology

In [2]: topology = Topology.from_json(open("topology.json").read())

In [3]: %timeit -o -r1 -n1 topology.identical_molecule_groups
11min 2s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
Out[3]: <TimeitResult : 11min 2s ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)>

In [4]: !open .

In [5]: topology.n_atoms, topology.n_molecules
Out[5]: (4119, 10)
```

Using the current `main` branch, it takes at least twice that (24 minutes, still running):

Image

**Output**

 - Operating system

**Additional context**

#1143 #1734 #353 #2008 https://github.com/openforcefield/openff-interchange/issues/1156 etc.

j-wags:
- [x] Fix #2035
    - This makes the reproducing case run on my machine in ~0.5 sec, compared to 11 min using the old stack 

### DIFF
--- a/.github/workflows/installer.yml
+++ b/.github/workflows/installer.yml
@@ -69,7 +69,7 @@ jobs:
         ls build
 
     - name: Upload installer as artifact
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v5
       with:
         name: ${{ matrix.os }}_py${{ matrix.python-version }}.sh
         path: toolkit-installer-constructor/build/openff-toolkit*/openff-toolkit*.sh
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -16,6 +16,9 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 - [PR #2114](https://github.com/openforcefield/openff-toolkit/pull/2114): Fixes [Issue #2035](https://github.com/openforcefield/openff-toolkit/issues/2035) (and touches on several others) by improving the runtime of Molecule.are_isomorphic/is_isomorphic_with, which can greatly improve interchange creation and export runtime.  
 - [PR #2114](https://github.com/openforcefield/openff-toolkit/pull/2114): May change the specific mapping of symmetric atoms returned by `Molecule.are_isomorphic` (to a different but chemically equivalent mapping) 
 
+### Bugfixes
+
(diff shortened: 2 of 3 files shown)


openff/toolkit/topology/molecule.py
```diff
@@ -2151,6 +2151,12 @@ def to_networkx(data: FrozenMolecule | nx.Graph) -> nx.Graph:
                 return data.to_networkx()
 
             elif isinstance(data, nx.Graph):
+                for node in data:
+                    h_counter = -1
+                    for neighbor in data.neighbors(node):
+                        if data.nodes[neighbor]['atomic_number'] == 1:
+                            data.nodes[neighbor]['atomic_number'] = h_counter
+                            h_counter -= 1
```
mattwthompson: Does this depend on the order of the `data.neighbors(node)` iterator (in problematic ways)? I will try to think of an antagonistic test case tomorrow, if possible

mattwthompson:
So far I'm seeing a net improvement in some `Interchange.to_x` export times. I have **not** looked at anything beforehand (molecule loading, topology preparation, parameter assignment, partial charge assignment, etc.) nor is this a completely comprehensive analysis.[^2]

The two environments I used are my Interchange development environment (with this branch installed) and a second with only released versions of our software[^1]

|         | OpenFE JACS set (vacuum) | Mixed solvent system | Ligand in water | Tim's polymers   |
|---------|-----------------|----------------------|-----------------|------------------|
| GROMACS | No change       | Marginally worse     | Slightly better | Huge improvement |
| OpenMM  | No change       | Marginally worse     | Somewhat better | No change        |


[^1]: `micromamba create --name released openff-toolkit seaborn polars tqdm matplotlib`
[^2]: I have receipts for everything elsewhere if people want lots of logs files

j-wags:
A quick update here - I'm offline for the rest of this week and half of the next, but my plan is to come back around the 27th and try to prove to myself that this is a safe approach and put in tests to raise the alarm if a later change violates some assumption that this needs (no divalent H, no H isotopes, no negative atomic numbers, etc). If folks want to help out by thinking of edge cases/regressions, that could help me get through the QA process faster.

PR #2137, opened 2025-12-16
Fix examples CI

### LINKED ISSUE #2129: CI is failing
Examples and Conda Latest CIs are intermittently failing: https://github.com/openforcefield/openff-toolkit/actions/runs/20016453230/job/57395051837

The [failure](https://github.com/openforcefield/openff-toolkit/actions/runs/20009573573/job/57376950231#step:12:261) appears to be in charge assignment in the Toolkit Showcase example, and only appears in non-openeye builds (ie, using AmberToolsToolkitWrapper for charge assignment)

mattwthompson:
- [ ] Fixes #2129

### DIFF
--- a/devtools/conda-envs/rdkit-examples.yaml
+++ b/devtools/conda-envs/rdkit-examples.yaml
@@ -17,6 +17,7 @@ dependencies:
   - openff-units
   - openff-utilities >=0.1.5
   - openff-interchange-base =0.4
+  # No idea why this is necessary, see https://github.com/openforcefield/openff-toolkit/pull/1821
   - nomkl
   - openff-nagl-base >=0.4.0
   - openff-nagl-models >=0.3.0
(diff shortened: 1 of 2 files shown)


devtools/conda-envs/rdkit-examples.yaml
```diff
@@ -17,6 +17,7 @@ dependencies:
   - openff-units
   - openff-utilities >=0.1.5
   - openff-interchange-base =0.4
+  - nomkl
```
mattwthompson: See for context: https://github.com/openforcefield/openff-toolkit/blob/73d62fecd60751d4ee6799e3a23ea3544eab903e/devtools/conda-envs/test_env.yaml#L48-L49

j-wags: Wow, that issue mirrors what I did this time, all the way down to not reproducing the issue in a docker container.

Issue #2150, opened 2026-02-16
Vsite serialization is failing


j-wags:
https://github.com/openforcefield/openff-toolkit/actions/runs/22046462342/job/63782661923

Tests are failing, both nightly and deployment:

```pytb
_______________ TestVirtualSiteHandler.test_serialize_roundtrip ________________
[gw1] linux -- Python 3.13.12 /home/runner/micromamba/envs/openff-toolkit-test-rdkit/bin/python
openff/toolkit/_tests/test_parameters.py:2165: in test_serialize_roundtrip
    roundtripped_force_field = ForceField(offxml_string)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^
openff/toolkit/typing/engines/smirnoff/forcefield.py:336: in __init__
    self.parse_sources(sources, allow_cosmetic_attributes=allow_cosmetic_attributes)
openff/toolkit/typing/engines/smirnoff/forcefield.py:791: in parse_sources
    self._load_smirnoff_data(smirnoff_data, allow_cosmetic_attributes=allow_cosmetic_attributes)
openff/toolkit/typing/engines/smirnoff/forcefield.py:900: in _load_smirnoff_data
    smirnoff_data = convert_all_strings_to_quantity(
openff/toolkit/utils/utils.py:261: in convert_all_strings_to_quantity
    smirnoff_data[key] = convert_all_strings_to_quantity(
openff/toolkit/utils/utils.py:261: in convert_all_strings_to_quantity
    smirnoff_data[key] = convert_all_strings_to_quantity(
openff/toolkit/utils/utils.py:261: in convert_all_strings_to_quantity
    smirnoff_data[key] = convert_all_strings_to_quantity(
openff/toolkit/utils/utils.py:269: in convert_all_strings_to_quantity
    smirnoff_data[index] = convert_all_strings_to_quantity(
openff/toolkit/utils/utils.py:261: in convert_all_strings_to_quantity
    smirnoff_data[key] = convert_all_strings_to_quantity(
openff/toolkit/utils/utils.py:280: in convert_all_strings_to_quantity
    obj_to_return = object_to_quantity(smirnoff_data)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../micromamba/envs/openff-toolkit-test-rdkit/lib/python3.13/functools.py:934: in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
[...]

PR #2151, opened 2026-02-17
Pin xmltodict <=1.0.2 in CI

j-wags:
- [x] Bandaids (but doesn't close) #2150


mattwthompson (approved):
LGTM

If there was more or problematic history with this package a lower constraint might be useful, but I don't think that's the case

Issue #2082, opened 2025-07-10
Remove all instances of "This API is experimental and subject to change"


j-wags:

jameseastwood:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/OpenFF-6747890d1f1c2b0017d11bd0/issues/gh/openforcefield/openff-toolkit/2082?planning-poker) with Zenhub @chrisiacovella @j-wags @lilyminium @mattwthompson

PR #2160, opened 2026-03-06
Improve `Molecule.remap` docstring

mattwthompson:
When working with `Molecule.remap` today, I came across a few behaviors which weren't documented and I had to discover through trial-and-error.

* This method returns a new object; I thought it modified in-place
* Conformers are updated - I don't need to shuffle them myself
* Same with partial charges
* This API doesn't seem particularly experimental, or at least not more unstable than I'd expect from a 0.x package (#2082)

I can move this into an issue but it's minor and, I hope, self-explanatory.

### DIFF
--- a/openff/toolkit/topology/molecule.py
+++ b/openff/toolkit/topology/molecule.py
@@ -4854,7 +4854,7 @@ class FrozenMolecule(Serializable):
             )
 
     def remap(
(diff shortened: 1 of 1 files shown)


j-wags (approved):
Thanks @mattwthompson!

Issue #2179, opened 2026-05-13
Treat AmberTools as an optional dependency


mattwthompson:
**Is your feature request related to a problem? Please describe.**

AmberTools is no longer used to assign AM1-BCC partial charges in mainline OpenFF force fields. This is expected to be the case in the indefinite future.

We should treat AmberTools as an optional dependency and update packaging to reflect this.

**Describe the solution you'd like**

Don't assume in source code that AmberTools is installed, also don't install it by default in packaging.

**Describe alternatives you've considered**

* Continue to use old versions of other upstream packages (Boost and NetCDF come to mind) because of compiled incompatibilities
* Continue to maintain AmberTools package

**Additional context**

Related #2178

j-wags:
We've removed AmberTools as a requirement from openff-toolkit packaging, but the next step on this would be to make a new slot in the test matrix that exercises the case of no AmberTools AND no OpenEye.

PR #2193, opened 2026-06-05
Test Python 3.14

mikemhenry:
Just testing python 3.14

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.12", "3.13"]
+        python-version: ["3.12", "3.13", "3.14"]
         rdkit: [true, false]
         openeye: [true, false]
         nagl: [true, false]
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -25,15 +25,11 @@ jobs:
         os: [ubuntu-latest, macos-latest]
         python-version: ["3.12", "3.13", "3.14"]
         rdkit: [true, false]
-        openeye: [true, false]
+        openeye: [false]
         nagl: [true, false]
         exclude:
           - rdkit: false
             openeye: false
-          - openeye: true
-            python-version: "3.13"
-          - openeye: true
-            python-version: "3.14"
           - rdkit: false
             nagl: true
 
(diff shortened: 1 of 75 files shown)


mattwthompson:
We re-built the toolkit and Interchange, so I'm re-kicking CI now

mikemhenry:
Weird, some http issue since it looks like it can't find these packages?
```

      The following packages are incompatible
      ├─ ambertools >=22 * does not exist (perhaps a typo or a missing channel);
      └─ numpy <2.3 * does not exist (perhaps a typo or a missing channel).
```

mattwthompson:
For better or worse you've come across two outstanding issues in this infrastructure

* AmberTools isn't built for 3.14, I don't think. I'm trying to make it optional (#2179) but that's a non-trivial set of changes. For now can simply drop it from the testing environment(s) and let ~30 tests pass
* The NumPy constraint is just outdated, there was a broken ParmEd build but that has since been fixed

mattwthompson:
Those Pydantic errors seem to stem from NAGL whose testing we're also in the middle of revamping. This implies to me there's value in just some basic tests of the Python code on 3.14 (https://github.com/openforcefield/openff-nagl/issues/236) even if most of the dependencies aren't updated yet

I want to drop v1 compatibility altogether (https://github.com/openforcefield/openff-nagl/issues/262) but it won't be in a release for a while and also might not be the core issue

mikemhenry:
@mattwthompson It looks like the issue now is we need `from __future__ import annotations` in a few spots since the behavior of type annotations changes in python 3.14 https://docs.python.org/3/reference/compound_stmts.html#annotations

I'm going to take a stab at adding them mostly just to get the tests to run and not worry too much about doing it the right way, I'm mostly just trying to find all the sharp edges around python 3.14 support

mattwthompson:
That's ... gonna take some patches to smooth over.

Have you come across a tool that checks and/or fixes that automagically for us?

mikemhenry:
I think there are some tools that can do it in a fancy way but this is easy:

`lint.isort.required-imports = ["from __future__ import annotations"]`

j-wags:
@mikemhenry Feel free to make a branch instead of a fork since otherwise being unable to access the OpenEye license secret will hamstring the CI runs.

mikemhenry:
@j-wags Ha, I just noticed that issue, I decided to save some carbon and just turn off the openeye stuff for now

mikemhenry:
`qcportal` & friends looks like it will be a problem, but it looks like I might need to start in `openff-nagl` to see where the 3.14 blockers exist in the direct openff ecosystem.

mikemhenry:
Ah and that will hit the dgl blocker...

mattwthompson:
DGL is needed for training but not inference (and the toolkit only needs inference functionality)

PR #2202, opened 2026-06-15
Fall back through toolkits for InChI conversion methods

### LINKED ISSUE #2172: `Molecule.to_inchi` does not fall back through wrapped toolkits as expected
**Describe the bug**

When calling `Molecule.to_inchi()` without special arguments or toolkit registry management, I expect it to iterated through wrapped toolkits until it finds one that provide the method (without a failure) and return the result. Instead, I observe that it does not fall back to using the RDKit after OpenEye's `OEMolToInChI` fails.

**To Reproduce**
This can be reproduced with any molecule larger than 1024 atoms; RDKit includes InChI's "large molecule" support but OpenEye does not, on the sketchy evaluation that it's experimental. That alone is its own issue, the problem for the OpenFF user is the fallback behavior in the OpenFF Toolkit.

```python

from openff.toolkit.utils.toolkits import (
    toolkit_registry_manager,
    RDKitToolkitWrapper,
    OpenEyeToolkitWrapper,
    GLOBAL_TOOLKIT_REGISTRY,
)


for tag, backend in zip(
    [
        "RDKit",
        "OpenEye",
        "Global",
    ],
    [
        RDKitToolkitWrapper(),
        OpenEyeToolkitWrapper(),
        GLOBAL_TOOLKIT_REGISTRY,
    ],
):
    print(f"Testing with {tag} toolkit registry...")

    with toolkit_registry_manager(backend):
        try:
            inchi = Molecule.from_smiles(342 * "C").to_inchi()

            # just make sure it isn't accidentally an empty string or something
            assert len(inchi) > 10

            print("\t\t... Success! made an InChI!")
        except Exception as e:
            print("\t\t... Failure! did NOT make an InChI!")
            print(type(e), str(e))

    print()
```

**Output**

```
$ python repro.py
Testing with RDKit toolkit registry...
		... Success! made an InChI!

Testing with OpenEye toolkit registry...
Warning: OECreateInChI: InChI only supports molecules with between 1 and 1023 atoms! (note: large molecule support is experimental)
		... Failure! did NOT make an InChI!
<class 'openff.toolkit.utils.exceptions.EmptyInChiError'> OEChem failed to generate an InChI for the molecule.

Testing with Global toolkit registry...
Warning: OECreateInChI: InChI only supports molecules with between 1 and 1023 atoms! (note: large molecule support is experimental)
		... Failure! did NOT make an InChI!
<class 'openff.toolkit.utils.exceptions.EmptyInChiError'> OEChem failed to generate an InChI for the molecule.
```

 - Operating system

**Additional context**

The case of `OEMolToInChI` / `OEMolToSTDInChI` returning an empty string is actually handled; I land [here](https://github.com/openforcefield/openff-toolkit/blob/b56ac244b9ddb0e03e68cd24e54bc6a21c78bd15/openff/toolkit/utils/openeye_wrapper.py#L1811-L1812) in that case

gaoflow:
## Description

`Molecule.to_inchi`, `Molecule.to_inchikey`, and `Molecule.from_inchi` call `ToolkitRegistry.call(...)` without `raise_exception_types=[]`, so they get the default behavior of re-raising the *first* toolkit's exception instead of continuing to the next registered toolkit. As a result the documented toolkit precedence/fallback never happens for InChI conversion: per #2172, OpenEye refuses InChI generation for molecules with more than 1024 atoms (which RDKit supports), but the registry propagates OpenEye's error rather than falling back to RDKit.

## Fix

Pass `raise_exception_types=[]` at these three call sites, matching the convention already used throughout the rest of the `Molecule` API (e.g. `assign_partial_charges`, and as the test suite documents: "the Molecule API passes raise_exception_types=[] to ToolkitRegistry.call"). A higher-precedence toolkit that raises is now skipped in favor of the next one; if every registered toolkit fails, `ToolkitRegistry.call` still raises a single aggregate `ValueError` describing each failure.

## Testing

Added `TestToolkitRegistry::test_inchi_methods_fall_back_through_toolkits`, which registers a stub wrapper whose InChI methods always raise ahead of `RDKitToolkitWrapper` and asserts that `to_inchi`, `to_inchikey`, and `from_inchi` skip the failing wrapper and succeed via RDKit. The test fails on `main` (the first toolkit's `ValueError` propagates) and passes with this change; it runs offline with only RDKit (no OpenEye needed). All existing InChI tests continue to pass.

Closes #2172.

---
Prepared with AI assistance under my direction; I reviewed and tested the change.

### DIFF
--- a/docs/releasehistory.md
+++ b/docs/releasehistory.md
@@ -14,6 +14,8 @@ Releases follow the `major.minor.micro` scheme recommended by [PEP440](https://w
 
 ### Bugfixes
 
+- [Issue #2172](https://github.com/openforcefield/openff-toolkit/issues/2172): `Molecule.to_inchi`, `Molecule.to_inchikey`, and `Molecule.from_inchi` now fall back to the next registered toolkit when a higher-precedence toolkit raises, instead of aborting on the first toolkit's error. For example, InChI generation for molecules with more than 1024 atoms now falls back to RDKit when OpenEye declines to process them.
+
 ### New features
 
 ### Improved documentation and warnings
--- a/openff/toolkit/_tests/test_toolkits.py
+++ b/openff/toolkit/_tests/test_toolkits.py
@@ -4333,6 +4333,43 @@ class TestToolkitRegistry:
                 raise_exception_types=[],
             )
 
+    @requires_rdkit
+    def test_inchi_methods_fall_back_through_toolkits(self):
+        """InChI conversion methods should skip a higher-precedence toolkit that
+        raises and fall back to the next one (issue #2172). For example, OpenEye
+        refuses InChI generation for >1024-atom molecules while RDKit supports
+        it, so ``to_inchi`` must fall back to RDKit rather than propagating the
+        first toolkit's error."""
+
+        class _FailingInChIWrapper(RDKitToolkitWrapper):
+            """A wrapper that advertises the InChI methods but always fails."""
+
+            def to_inchi(self, *args, **kwargs):
+                raise ValueError("simulated to_inchi failure")
+
+            def to_inchikey(self, *args, **kwargs):
+                raise ValueError("simulated to_inchikey failure")
+
+            def from_inchi(self, *args, **kwargs):
+                raise ValueError("simulated from_inchi failure")
+
+        registry = ToolkitRegistry(toolkit_precedence=[])
+        registry.register_toolkit(_FailingInChIWrapper())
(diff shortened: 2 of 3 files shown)


j-wags:
Thanks @gaoflow - This looks promising. However because it is substantially AI generated, and because reviewing this PR will take about as much effort as making the changes myself, I'm just going to keep this at the same level of priority as the underlying issue. So I will review this when the underlying issue is actioned in our normal task planning.

Issue #2178, opened 2026-05-13
Treat RDKit as a required dependency


mattwthompson:
**Is your feature request related to a problem? Please describe.**

With the major force field releases using NAGL charge models - AshGC now, hopefully future iterations to come - which require RDKit and incoming PDB loading infrastructure also explicitly requiring RDKit, I don't think it makes sense to continue to treat it as an optional dependency. (Historical releases are unaffected by definition.)

**Describe the solution you'd like**

Assume RDKit is always installed, and also make sure it's always installed

**Describe alternatives you've considered**

Continue to do [complex operations](https://github.com/openforcefield/openff-toolkit/blob/6e2fc3520f1dce0f4f79e5674446b348ceadeba8/.github/workflows/CI.yml#L57-L63) to force RDKit removal and treat it as "optional" in most code

**Additional context**

RDKit and AmberTools ought to be decoupled whereas in the past they were heavily tied to each other.

j-wags:
I think this is a good idea. Let's move forward with this, and not close this issue until the change is both made in CI and in packaging dependencies.

mattwthompson:
This is partially in #2215 and partially in #2214

mattwthompson:
Also see https://github.com/orgs/openforcefield/discussions/58

PR #2203, opened 2026-06-18
Update examples dependencies

### LINKED ISSUE #2204: Use Sage 2.3.0 in examples
Examples currently use Sage 2.2.x or older force fields. This implies they are the recommended starting point for new users, which is not the cases.

Examples should be updated to use the new flagship force field, Sage 2.3.0.

mattwthompson:
Fixes #2204

Tag #2178 #2179

- [x] Do not install `openmmforcefields`, which is not used in examples
- [x] Handle `rdkit` as a required dependency
- [x] Handle `openff-nagl` as a required dependency
- [ ] Handle `ambertools` as an optional dependency
---


j-wags (changes requested):
This is complex to review as it touches several test env yamls but they expect a state of `CI.yml` from midway through the PRs commit history. I'd prefer to review this once it's in its final form.

PR #2214, opened 2026-07-08
Update examples to use Sage 2.3.0, so make rdkit+nagl required and drop OE examples builds

### LINKED ISSUE #2204: Use Sage 2.3.0 in examples

mattwthompson:
Closes #2204
- [x] Use Sage 2.3.0 in examples
- [x] Treat NAGL as required in examples
- [x] Treat RDKit as required in examples
- [x] Do not install AmberTools in examples
- [x] Do not install OpenMMForceFields in examples
- [x] Update other dependencies to minimum versions
- [x] Re-run notebooks


devtools/conda-envs/examples.yaml
```diff
@@ -12,17 +12,16 @@ dependencies:
...
-  - openff-forcefields >=2023.11
+  - openff-forcefields >=2025
   - openff-amber-ff-ports >=0.0.3
   - openff-units
   - openff-utilities >=0.1.5
   - openff-interchange-base >=0.5
-  - openff-nagl-base >=0.4.0
-  - openff-nagl-models >=0.3.0
+  - openff-nagl-base >=0.5.3
+  - openff-nagl-models >=2025
```
mattwthompson, line 21: These do not need to be here - they are not imported in code - and ought to be implicitly available if the `openff-forcefields` constraint is accurate

j-wags: I don't understand. How would NAGL otherwise be pulled in? The openff-forcefields package does not list them as a dependency.

mattwthompson: I was wrong

mattwthompson:
7289d37947d59548db52745802039e0492a19a33 rerenders with AmberTools 26 (build string `cuda_None_nompi_py313h47fe9ef_100`) installed. I ran this command, as I did for all notebooks in this PR

```
$ jupyter nbconvert --to notebook --execute --inplace examples/using_smirnoff_in_amber_or_gromacs/export_with_interchange.ipynb
```

mattwthompson:
cb687bb25decb900f47844c5488173fc222b389d adds a check for OpenEye being installed and available in the appropriate workflow runs

See it in action [here](https://github.com/openforcefield/openff-toolkit/actions/runs/29517926513/job/87687523218?pr=2214)

.github/workflows/examples.yml
```diff
@@ -18,23 +18,15 @@ defaults:
...
+    name: ${{ matrix.os }}, Python ${{ matrix.python-version }}, OpenEye=${{ matrix.openeye }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
         python-version: [ "3.12"]
-        rdkit: [true, false]
+        rdkit: [true]
         openeye: [true, false]
```
j-wags: Forensic note to self: Matt and I reviewed this synchronously and I advocated for the following: There are now NO examples that require openeye, since we assume that rdkit+nagl will ALWAYS be present, and those will be reached for as a fallback even in openeye-having builds. So we gain nothing from an openeye-having examples build and are removing it. Openeye-only functionality (eg am1bcc charge assignment tests) are still included in the main CI, and will fail if OE is not present because they target the OpenEyeToolkitWrapper specifically.

mattwthompson:
And we changed our minds about this - b9d0557dc77c7147fd8082e3900706a725d38ca8 drops OpenEye from this infrastructure altogether. (It remains in the normal full CI testing!)

j-wags (approved):
Thanks @mattwthompson - This was a very productive synchronous review, the forensic note attached to this explains that this PR is quite a bit larger than the title currently indicates so I'll update that. This is good to merge once the tests pass and the releasenotes are updated.

PR #2215, opened 2026-07-08
Require RDKit in CI

### LINKED ISSUE #2178: Treat RDKit as a required dependency

mattwthompson:
Closes #2178


.github/workflows/CI.yml
```diff
@@ -130,7 +84,7 @@ jobs:
         # of available builds.
         # See ex https://github.com/rdkit/rdkit/issues/7221
         # and https://github.com/rdkit/rdkit/issues/7583
```
j-wags: (not blocking) Given that mypy tests now pass on RDKit-containing builds, is this outdated?

.github/workflows/CI.yml
```diff
@@ -18,22 +18,16 @@ defaults:
 jobs:
   test:
     if: (github.event_name == 'schedule' && github.repository == 'openforcefield/openff-toolkit') || (github.event_name != 'schedule')
-    name: Test on ${{ matrix.os }}, Python ${{ matrix.python-version }}, RDKit=${{ matrix.rdkit }}, OpenEye=${{ matrix.openeye }}, NAGL=${{ matrix.nagl }}
```
j-wags, line 21: (blocking) NAGL is still optional. Making RDKit required is not equivalent to making NAGL required, and users should still get functional+tested builds if NAGL is not present.

devtools/conda-envs/test_env_no_openeye.yaml
```diff
@@ -38,3 +38,12 @@ dependencies:
...
   - nglview
+  - pip:
+    - mypy ==2
+    - types-setuptools
+    - types-toml
+    - types-PyYAML
+    - types-networkx
+    - types-xmltodict
+    - types-cachetools
+    - mongo-types
```
j-wags, line 49: (not blocking) These don't do anything if mypy doesn't run in rdkit=true builds. But since CI passes in OE+RDK builds, maybe we keep this here and make the mypy step run unconditionally?

j-wags (changes requested):

j-wags (approved):
Thanks @mattwthompson!

PR #2216, opened 2026-07-10
Manage virtual environments and tasks with Pixi

### LINKED ISSUE #2178: Treat RDKit as a required dependency

### LINKED ISSUE #2179: Treat AmberTools as an optional dependency

### LINKED ISSUE #2184: Revamp testing environments

While working on #2180 #2183 and other recent changes iit is clear to me that we need to re-think our testing infrastructure, possibly from the ground-up.

I'm working off of the following assumptions:
1. Basic toolkit usage requires
  a. OpenFF NAGL[^1]
  b. the RDKit[^2]
1. Basic toolkit usage does NOT require
  a. OpenEye toolkits (OEChem, OEOmega, etc.)
  b. AmberTools
1. AmberTools and the RDKit are not particularly coupled at install time
1. Charge assignment hierarchy should still follow the SMIRNOFF specification

Revamp testing infrastructure with the following requirements in mind
* OpenFF Nagl or RDKit can be assured to be installed and removing them is an advanced use case
* Tests using any force field older than 2.3.0 requires AmberTools OR OpenEye
* Some tests may continue to require OpenEye

I think this could be accomplished with three CI environments:
* One "base" with only required dependencies
* One environment with OpenEye, and nothing else, added into the base
* One environment with AmberTools, and nothing else, added into the base

We could continue to have multiple examples environments, but to me the tradeoffs here don't make as much sense; our examples ought to taste basic usage and our CI setup should catch most breaks.

Lots and lots of import guards, complexity in automation, and fragility with i.e. forcibly removing required dependencies.

**Additional context**

This idea loses an "everything bagel" environment, but this one tends to be the hardest one to maintain.

[^1]: For example, using Sage 2.3.0 or more recent force fields
[^2]: Above but also PDB loading

mattwthompson:
Closes #2178
Closes #2184
Closes #2179


j-wags:
Thanks. This is good to have as a starting point for a future switchover, but I'm not interested in moving CI from conda to pixi right now.

PR #2230, opened 2026-09-02
More gently fail if AmberTools provenance fails

mattwthompson:
See https://github.com/openforcefield/openff-bespokefit/issues/438#issuecomment-5320725045 for parallel context, though this is a standalone change

### DIFF
--- a/openff/toolkit/utils/ambertools_wrapper.py
+++ b/openff/toolkit/utils/ambertools_wrapper.py
@@ -74,9 +74,12 @@ class AmberToolsToolkitWrapper(base_wrapper.ToolkitWrapper):
 
     @functools.cached_property
     def _toolkit_version(self):
-        from openff.utilities.provenance import get_ambertools_version
+        try:
+           from openff.utilities.provenance import get_ambertools_version
 
-        return get_ambertools_version()
+           return get_ambertools_version()
+        except Exception:
+            return 'Unknown'
 
     @staticmethod
     def is_available() -> bool:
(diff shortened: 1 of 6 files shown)


j-wags (commented):
I think this approach is good - The remaining thing to do is to add tests similar to https://github.com/openforcefield/openff-bespokefit/pull/440 so we know that this behaves correctly under a variety of scenarios.

openff/toolkit/utils/ambertools_wrapper.py
```diff
@@ -74,9 +74,18 @@ def __init__(self):
...
     @functools.cached_property
     def _toolkit_version(self):
-        from openff.utilities.provenance import get_ambertools_version
-
-        return get_ambertools_version()
+        # See https://github.com/openforcefield/openff-bespokefit/pull/440
+        try:
+            from openff.utilities.provenance import get_ambertools_version
+
+            ambertools_version = get_ambertools_version()
```
j-wags, line 81: (not blocking) `_toolkit_version`  is implied to have a return type of `str` (since its return value is used directly as a value in a return type of `dict[str, str]` [here](https://github.com/openforcefield/openff-toolkit/blob/a93553259b490f89e7a9f71c4ff019e34d526ae5/openff/toolkit/utils/toolkit_registry.py#L145)). But `get_ambertools_version` will sometimes return `None`. It could be good to guard against that here, and return `"Unknown"` if `None` comes out of the `get_ambertools_version` call.

j-wags (approved):
LGTM, one nitpicky suggestion that may save us some headache later.

openff/toolkit/utils/ambertools_wrapper.py
mattwthompson: Nice find!

PR #2231, opened 2026-09-04
Stop OpenEyeToolkitWrapper from erroring on planar trivalent nitrogens if they have trivalent carbon neighbors

### LINKED ISSUE #626: from_openeye doesn't percieve nitrogen stereochemistry, even if coordinates are available
**Describe the bug**

The following code snippet creates on OEMol from a SMILES with undefined stereo around a nitrogen, generates conformers of it (which do indeed show the nitrogen being pyrimidal), and then raises an undefined stereo error if we try to run `Molecule.from_openeye`. This molecule SHOULD have defined stereo, since it has coordinates. The code snippet below even deletes all but one conformer of the molecule, so that there's no ambiguity over which one to take the stereochemistry from, but it still fails. 

```
from openeye import oechem
# create a new molecule
oemol = oechem.OEMol()
# convert the SMILES string into a molecule
oechem.OESmilesToMol(oemol, "CN(CC)c1ccccn1")
from openeye import oeomega
omega = oeomega.OEOmega()
status = omega(oemol)
print(len([i for i in oemol.GetConfs()]))
# Delete all but one conformer
for idx, conf in enumerate(oemol.GetConfs()):    
    if idx ==  0:
        continue
    oemol.DeleteConf(conf)
print(len([i for i in oemol.GetConfs()]))
offmol = Molecule.from_openeye(oemol)
print(offmol.n_conformers)
```

> 11
> 1
> 
> ---------------------------------------------------------------------------
> UndefinedStereochemistryError             Traceback (most recent call last)
> 

**To Reproduce**

**Output**

OFFTK 0.7.0 with OpenEye (but it also reproduces on 0.6.0)

### LINKED ISSUE #725: Undefined stereo error depends on the toolkit.
**Describe the bug**
Undefined stereo errors seem to depend on the backend toolkit used to create the molecule see the example below I think this is mainly due to nitrogen stereochemistry being picked up or not.

This also causes an error in QCSubmit when enumerating stereochemistry, in the workflow we have a component which enumerates all undefined stereochemistry and checks that the molecule is valid before adding it to a dataset by doing a round trip and looking for undefined stereo errors. In the case of this molecule, the nitrogen stereo chem is enumerated and 2 isomers are made but no new molecules are returned from the method as the toolkit makes sure to not return the same molecule and as nitrogen stereo chem is no longer checked in the isomorphism check the toolkit thinks the two new molecules are the same as the input and they are removed. Then when we do the round trip test on the input molecule it fails as openeye identifies missing stereochemistry meaning any input molecules with only undefined nitrogen stereochemistry are accidentally removed when building datasets.

**To Reproduce**
```python
mol = Molecule.from_smiles("[H]c1c(c(c(c(c1C#N)[H])[H])N(C([H])([H])c2c(c(c(c(c2[H])Br)OS(=O)(=O)N([H])[H])[H])[H])N3C(=NN=C3[H])[H])[H]", allow_undefined_stereo=False, toolkit_registry=OpenEyeToolkitWrapper())
```
> Problematic atoms are:
Atom atomic num: 7, name: , idx: 11, aromatic: False, chiral: True with bonds:
bond order: 1, chiral: False to atom atomic num: 6, name: , idx: 3, aromatic: True, chiral: False
bond order: 1, chiral: False to atom atomic num: 6, name: , idx: 12, aromatic: False, chiral: False
bond order: 1, chiral: False to atom atomic num: 7, name: , idx: 32, aromatic: False, chiral: False

```python
mol = Molecule.from_smiles("[H]c1c(c(c(c(c1C#N)[H])[H])N(C([H])([H])c2c(c(c(c(c2[H])Br)OS(=O)(=O)N([H])[H])[H])[H])N3C(=NN=C3[H])[H])[H]", allow_undefined_stereo=False, toolkit_registry=RDKitToolkitWrapper())
```
> Molecule with name '' and SMILES '[H]c1c(c(c(c(c1C#N)[H])[H])N(C([H])([H])c2c(c(c(c(c2[H])Br)OS(=O)(=O)N([H])[H])[H])[H])N3C(=NN=C3[H])[H])[H]'

### LINKED ISSUE #1547: Saturated Planner nitrogen
**Describe the bug**
I had a ligand 
[ligand_12t.sdf.zip](https://github.com/openforcefield/openff-toolkit/files/10731178/ligand_12t.sdf.zip)
, which contains a planner saturated nitrogen. If I pass this molecule into openff-tookit, it will complain this planner nitrogen being chiral and has no stereochemistry.

**To Reproduce**
```
from openeye import oechem
from openff.toolkit.topology import Molecule as OpenFFMol
ifs = oechem.oemolistream()
if ifs.open('ligand_12t.sdf'):
	for mol in ifs.GetOEGraphMols():
		openffmol = OpenFFMol.from_openeye(mol)
```

**Output**
```
  File "/Users/zwu/mambaforge/envs/BSSOrion/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3442, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-8-a7df804a7ba1>", line 3, in <module>
    openffmol = OpenFFMol.from_openeye(mol)
  File "/Users/zwu/mambaforge/envs/BSSOrion/lib/python3.8/site-packages/openff/toolkit/utils/base_wrapper.py", line 52, in wrapped_function
    value = func(*args, **kwargs)
  File "/Users/zwu/mambaforge/envs/BSSOrion/lib/python3.8/site-packages/openff/toolkit/topology/molecule.py", line 4265, in from_openeye
    molecule = toolkit.from_openeye(
  File "/Users/zwu/mambaforge/envs/BSSOrion/lib/python3.8/site-packages/openff/toolkit/utils/openeye_wrapper.py", line 1159, in from_openeye
openff.toolkit.utils.exceptions.UndefinedStereochemistryError: Unable to make OFFMol from OEMol: OEMol has unspecified stereochemistry. oemol.GetTitle(): ligand_12t
Problematic atoms are:
Atom atomic num: 7, name: , idx: 27, aromatic: False, chiral: True with bonds:
bond order: 1, chiral: False to atom atomic num: 6, name: , idx: 18, aromatic: False, chiral: False
bond order: 1, chiral: False to atom atomic num: 6, name: , idx: 16, aromatic: False, chiral: False
bond order: 1, chiral: False to atom atomic num: 6, name: , idx: 7, aromatic: True, chiral: False
```

 - Operating system: 

**Additional context**
The cause of this is that the nitrogen is characterised as chiral as it is saturated. However, it has no stereochemistry as it is planner.
I think the solution could be to change 

https://github.com/openforcefield/openff-toolkit/blob/main/openff/toolkit/utils/openeye_wrapper.py#L1124

To 

```
carbonOnly = True
oechem.OEPerceiveChiral(mol, carbonOnly)
```

Thus, only Chiral carbon will be recognised.

j-wags:
- [x] Makes `OpenEyeToolkitWrapper.from_openeye` not raise an undefined-stereo error/warning for planar Ns if they have at least one trivalent carbon neighbor (This isn't a perfect solution but does cover a large fraction of the annoying over-policing of nitrogen stereo by OETKW)
- [x] Closes #725 
- [x] Closes #626 
- [x] Closes #1547 


j-wags:
Fingers crossed that the docs build issues are transient - I'll try a rebuild after the long weekend.

Yoshanuikabundi (approved):
I've found one class of false negatives introduced by this change, but am unsure how much of a dealbreaker this is. Otherwise, LGTM! The DrugBank entries removed from the test files is very pleasing. The tests look right to me and pass.

# New false negative

*(by false negative, I mean an inarguably chiral molecule that is allowed to be constructed with ambiguous stereochemistry without the `allow_undefined_stereo` switch turned on)*

If it's important that all chiral molecules be detected, then I believe this change introduces a failure mode in the form of bridgehead amines in conjugated systems such as [Troger's base](https://en.wikipedia.org/wiki/Tr%C3%B6ger%27s_base). This is a bridgehead, which makes it pyramidal, but is still bonded to an sp2 carbon, so I believe it would be detected as chiral before this change but slip by after this change. As this change is advertised as reducing overpolicing, it might be surprising or undesirable that it introduces a false negative.

You can see this by adding 
```py
            # Troger's base (bridgehead conjugated tertiary amine)
            ("c1(ccc3c(c1)C[N@]4c2ccc(cc2C[N@]3C4)C)C", ["S", "R"], None),
            ("c1(ccc3c(c1)CN4c2ccc(cc2CN3C4)C)C", [None, None], "unspecified stereochemistry"),
```
to the `test_smiles_nitrogen_chirality_skip` parametrization and changing `[expected_stereo]` to `expected_stereo` in the test body. The first new line succeeds both before and after the change, and the second succeeds only before the change.

My only proffered alternative (that I'm not sure is in scope or rises to the level of a suggestion) is that we could go further. There's an argument that basically all tertiary amines are achiral, as the planar ones are planar and the pyramidal ones interconvert rapidly. This seems to be what's suggested by the "spec for undefined chemistry" issue's table (#146, line with SMILES `N(Cl)(F)C`).

We could follow [RDKit's example](https://github.com/rdkit/rdkit/blob/20331b5101183089580840759c3ddf11a73efe31/Code/GraphMol/FindStereo.cpp#L95-L108) and specify that a chiral tertiary nitrogen must either be in a 3-membered ring or be a bridgehead.

